PR #481, opened 2026-04-11
Add domain and HMAC to local-file agent config (#473)

### LINKED ISSUE #473: `service add` generates agent config missing `domain` and `http_responder_hmac`
## Problem

The agent config template produced by `bootroot service add` (`secrets/openbao/services/<name>/agent.toml.ctmpl`, also written to the final `--agent-config` path) is missing fields that `bootroot-agent` requires to actually issue a certificate. Users must hand-edit the config before running the daemon.

### Missing fields

Given `bootroot service add --service-name review --hostname review-host --domain test.local ...`, the generated config looks like:

```toml
# BEGIN bootroot managed profile: review
[[profiles]]
service_name = "review"
instance_id = "001"
hostname = "review-host"

[profiles.paths]
cert = ".../review-cert.pem"
key = ".../review-key.pem"

[trust]
ca_bundle_path = ".../ca-bundle.pem"
trusted_ca_sha256 = [...]
```

Missing:

1. **Top-level `domain`** — without this, `bootroot-agent` falls back to the default `trusted.domain` (see `src/config/defaults.rs:DEFAULT_DOMAIN`) and issues ACME orders for the wrong FQDN. SAN construction uses the global `settings.domain` (see `src/config.rs`), so this single top-level field is what actually drives FQDN assembly. The `--domain` value passed to `service add` is not propagated into the generated config.
2. **`[acme].http_responder_hmac`** — responder HMAC is already copied into the per-service OpenBao KV path at `secret/data/bootroot/services/<service>/http_responder_hmac` during `service add` (see `src/commands/service/secrets.rs`), but the generated `agent.toml.ctmpl` does not reference it. `bootroot-agent` therefore has no HMAC at runtime and fails with `401 Unauthorized: Invalid signature`.

## Asymmetry with the remote-bootstrap path

The remote-bootstrap flow already produces a complete agent config. `src/bin/bootroot-remote/agent_config.rs::render_agent_config_baseline()` writes the top-level `domain` from its CLI args, and `src/bin/bootroot-remote/agent_config.rs` `apply_agent_config_updates()` upserts `http_responder_hmac` using the value pulled from the per-service OpenBao KV path.

In other words, the **remote path already knows how to inject both fields**. Only the local-file path (`src/commands/service/local_config.rs`) is missing them. The fix for this issue is to make the local path match the remote path — ideally by factoring the shared baseline-generation logic into one place that both `src/commands/service/local_config.rs` and `src/bin/bootroot-remote/agent_config.rs` call. Keeping two divergent baseline generators is what let this gap exist in the first place.

## User-visible symptom

After following the happy-path docs, `bootroot-agent --config ...` fails with either:

- `Authorization did not validate ... "The server could not connect to validation target"` (wrong domain), or
- `Responder returned 401 Unauthorized: Invalid signature` (missing HMAC).

Users must hand-append these lines before the daemon works:

```toml
domain = "test.local"

[acme]
http_responder_hmac = "<paste from init-summary.json>"
```

## Proposal

`bootroot service add` already has every value needed, and similar code paths already exist in the codebase:

- `domain` comes from `--domain`.
- The responder HMAC is already fetched from OpenBao and copied into the per-service KV path by `src/commands/service/secrets.rs`. The service's own AppRole already has permission to read that per-service path (per `src/commands/service/approle.rs`).
- A baseline agent config generator already exists for the remote bootstrap flow in `src/bin/bootroot-remote/agent_config.rs` and can be adapted/reused by the local `service add` path in `src/commands/service/local_config.rs`, which currently writes only the profile and trust sections.

Two implementation options for the HMAC (both can be used together):

1. **Direct write** (`agent.toml`): `service add` reads the HMAC value (the same one it just copied into the per-service KV path) and writes it into the generated agent config as a plain string. This is what the remote-bootstrap flow effectively does today.
2. **Template** (`agent.toml.ctmpl`): extend the consul-template with a block that reads the per-service path, matching the convention used by `src/trust_bootstrap.rs`:

   ```
   [acme]
   http_responder_hmac = "{{ with secret \"secret/data/bootroot/services/<service>/http_responder_hmac\" }}{{ .Data.data.hmac }}{{ end }}"
   ```

   This is consistent with how `trusted_ca_sha256` is already rendered from the per-service KV tree and avoids persisting the HMAC to disk in cleartext. The KV path and value are already populated by `service add`, so the only change is adding the block to the template.

Using both options together matches the existing pattern: `agent.toml` carries the literal value for immediate use, while `agent.toml.ctmpl` uses a consul-template expression so OpenBao Agent can refresh the value at runtime.

Top-level `domain` should simply be written by `service add` from the `--domain` argument.

## Documentation

- `CHANGELOG.md` — note that `service add` now generates a complete, ready-to-run agent config (top-level `domain`, `[acme].http_responder_hmac`) and no longer requires hand-editing before running `bootroot-agent`.
- `docs/en/cli.md`, `docs/en/cli-examples.md`, `docs/en/configuration.md` — document the final shape of the generated agent config.
- `docs/ko/cli.md`, `docs/ko/cli-examples.md`, `docs/ko/configuration.md` — mirror the English updates.
- `docs/en/installation.md`, `docs/ko/installation.md` — update if they reference a manual post-edit step (currently they do not).
- `README.md` — update the quickstart if it references the manual edit (currently it does not).

## Related

Part of the broader gap left after #464. Combined with sibling issues for DNS alias automation, post-renew hook flags, and the daemon retry config-reload bug, fixing this eliminates the last required hand-edit for a zero-config first run.

sehkone:
## Summary

- `service add` with `--delivery-mode local-file` now writes the top-level `domain` (from `--domain`) and `[acme].http_responder_hmac` (from the responder HMAC already stored in OpenBao) into the generated `agent.toml`, so the config is ready to use without manual editing.
- Added `upsert_top_level_keys` to `toml_util` for setting top-level TOML keys (used for `domain`).
- The `agent.toml.ctmpl` template replaces the literal HMAC with a consul-template expression reading from the per-service KV path, consistent with how `trusted_ca_sha256` is already rendered.
- Updated CHANGELOG and docs (EN + KO) to reflect that `service add` now produces a complete agent config.

Closes #473

## Test plan

- [x] `cargo test` passes, including new tests in `local_config.rs` and `toml_util.rs`
- [x] `test_generated_config_includes_domain_and_acme` — verifies `domain` and `[acme].http_responder_hmac` appear in the generated config
- [x] `test_generated_config_ctmpl_replaces_hmac_from_full_config` — verifies the ctmpl replaces the literal HMAC with a consul-template expression
- [x] `test_domain_and_acme_upsert_is_idempotent` — verifies re-running produces identical output
- [x] `upsert_top_level_*` unit tests in `toml_util.rs` — add, update, and idempotency for top-level keys
- [x] E2E: run `bootroot service add --delivery-mode local-file --domain test.local ...` and verify the generated `agent.toml` contains `domain = "test.local"` and `[acme].http_responder_hmac`
- [x] E2E: run `bootroot-agent --config <generated>` and confirm it issues a certificate without manual editing (verified by CI Docker E2E `local-hosts` / `local-no-hosts` jobs)


sehkone:
**[Reviewer Round 1]**

I reviewed the implementation in `src/commands/service/local_config.rs`, `src/toml_util.rs`, and `tests/e2e_same_host_local_file.rs`, and I did not find any blocking issues.

What looks good:
- `service add --delivery-mode local-file` now writes the top-level `domain` into the rendered `agent.toml`, which fixes the wrong-FQDN fallback path.
- The generated `agent.toml` now also contains `[acme].http_responder_hmac`, while `agent.toml.ctmpl` correctly rewrites that field to a consul-template secret lookup instead of leaving the literal HMAC in the template.
- The new `upsert_top_level_keys` helper is small, consistent with the existing TOML mutation utilities, and preserves the intended top-level placement.
- Test coverage is solid for this fix: there are unit tests for top-level TOML upserts, unit tests covering the full generated config and ctmpl behavior, and the local-file E2E happy-path test now asserts both new fields.

Validation I ran locally:
- `cargo test test_generated_config_includes_domain_and_acme`
- `cargo test test_generated_config_ctmpl_replaces_hmac_from_full_config`
- `cargo test test_domain_and_acme_upsert_is_idempotent`
- `cargo test upsert_top_level_adds_key`
- `cargo test upsert_top_level_updates_existing_key`
- `cargo test upsert_top_level_is_idempotent`
- `cargo test test_same_host_local_file_happy_path`
- `cargo clippy --all-targets --all-features -- -D warnings`

One non-blocking follow-up I would still recommend is factoring the shared baseline/domain mutation logic between the local-file and remote-bootstrap paths so these flows cannot drift again, but that is not a merge blocker for this PR.

PR #482, opened 2026-04-11
Auto-register HTTP-01 DNS alias on service add (#472)

### LINKED ISSUE #472: Auto-register HTTP-01 DNS alias for new services on `service add`
## Background

ACME HTTP-01 validation requires the step-ca container to resolve the challenge hostname (`<instance_id>.<service_name>.<hostname>.<domain>`) to the `bootroot-http01` responder container. Today, users must hand-write a `docker-compose.override.yml` that adds each SAN as a network alias on `bootroot-http01`:

```yaml
services:
  bootroot-http01:
    networks:
      default:
        aliases:
          - "001.review.review-host.test.local"
          - "001.aice-web-next.aiceweb-host.test.local"
```

Without this, `bootroot-agent` fails ACME validation with:

```
Authorization did not validate after 15 attempts. Last HTTP-01 error: "The server could not connect to validation target"
```

## Observations

- `bootroot service add` already has the per-service values needed to assemble the alias (`--service-name`, `--instance-id`, `--hostname`, `--domain`).
- `scripts/preflight/` and lifecycle scripts already implement an `apply_dns_aliases` helper; the logic is proven. See `scripts/impl/run-local-lifecycle.sh` for the pattern that applies the alias together with the responder override.
- This is bootroot's own infra: step-ca and responder are always managed by the bundled docker-compose, so the automation does not depend on how the target service (review, aice-web-next, etc.) is deployed.

## Proposal

`bootroot service add` should register the SAN as a Docker network alias on `bootroot-http01` automatically. Two implementation options, each with different prerequisites:

### Option 1: generate/update `docker-compose.override.yml`

Write (or extend) `docker-compose.override.yml` to add the alias, then run `docker compose ... up -d bootroot-http01` to apply.

Prerequisites specific to this option:

- `ServiceAddArgs` does not currently carry a compose-file context (see `src/cli/args.rs`). It needs either a new `--compose-file` flag on `service add` (mirroring `infra up`/`infra install`), or the compose file path persisted in `state.json` by `infra install` and re-read by `service add`. The second approach is less intrusive for users because they already selected the compose file at `infra install` time.
- When applying the alias, the command must bring `bootroot-http01` back up with the same override stack the responder already uses — notably `secrets/responder/docker-compose.responder.override.yml`, as demonstrated in `scripts/impl/run-local-lifecycle.sh`. Otherwise the per-service aliases would overwrite the responder's own override and break unrelated responder settings.
- Assumes the user has not renamed the `bootroot-http01` service or the compose network. The current `docker-compose.yml` uses those names, and `scripts/impl/run-local-lifecycle.sh` already hard-codes them.

### Option 2: `docker network connect --alias` at runtime

Attach the alias directly to the running `bootroot-http01` container via the Docker API, without touching any compose files.

This option avoids the compose-file context problem entirely and does not depend on the override stack layout. The network and container can be discovered at runtime (e.g. `docker inspect bootroot-http01` → the network it is attached to), so no compose-file flag is required. The trade-off is that the alias lives only in the running container state, so `infra up` on a restart would need to re-apply it (or the alias would have to be re-registered from `state.json` at startup).

### Recommended starting point: Option 2

For the parent-issue goal of "zero-config first success path", **Option 2 (runtime alias) is the recommended first implementation**. Reasons:

1. **Smaller blast radius.** Option 2 does not touch the compose override stack, does not require threading a compose-file context through `ServiceAddArgs` or `state.json`, and does not interact with the responder's existing override file. The change stays inside a new alias-registration helper plus a call from `service add`.
2. **Faster time to first success.** Runtime alias can be implemented with just the Docker CLI or API, whereas Option 1 requires coordinating with the responder override stack (see `scripts/impl/run-local-lifecycle.sh`) to avoid breaking it.
3. **`infra up` re-apply is manageable.** The main trade-off of Option 2 — losing the alias across container restarts — is addressable by having `infra up` replay aliases from `state.json` when it comes back up. That is a small follow-up that does not block first success.
4. **Option 1 can be added later** as a persistence layer on top of Option 2 if operators ask for it, without breaking the initial implementation.

The issue remains open to either approach, but implementers should start with Option 2 unless they discover a concrete reason the runtime path is insufficient.

## Cleanup

There is no `bootroot service remove` command today, so cleanup is coarse-grained: `bootroot clean` already tears down the compose stack, volumes, and state. Whichever implementation option is chosen should ensure `bootroot clean` leaves no stray aliases or override fragments behind. A dedicated `service remove` command is out of scope for this issue.

## Documentation

- `CHANGELOG.md` — note that `service add` now registers HTTP-01 DNS aliases automatically and no longer requires hand-written `docker-compose.override.yml`.
- `docs/en/cli.md`, `docs/en/cli-examples.md`, `docs/en/installation.md`, `docs/en/troubleshooting.md` — remove references to manual override-file editing, add the new automated flow, and document any new flags or state-file fields added to `service add`.
- `docs/ko/*.md` — mirror the English updates.
- `README.md` — update the quickstart/feature list if it mentions the manual override step.

## Impact

Removes one of the manual editing steps left after #464. Combined with #473 (agent config completeness), `service add` + `bootroot-agent` becomes a one-shot flow with no hand-editing.

sehkone:
## Summary

- `service add` now registers the service's HTTP-01 validation FQDN (`<instance_id>.<service_name>.<hostname>.<domain>`) as a Docker network alias on `bootroot-http01` at runtime (Option 2 from #472), removing the need for a hand-written `docker-compose.override.yml`.
- `infra up` replays all DNS aliases from `state.json` after container restarts, so aliases survive `docker compose down` / `up` cycles.
- DNS alias registration uses a rollback strategy: if the reconnect with aliases fails after disconnecting, the responder is reconnected without aliases so it is never left detached from the network. Recoverable failures (rollback succeeded, container still connected) are downgraded to warnings. If both reconnect and rollback fail (container left detached), `service add` exits with a hard error.
- Preflight test `cli-scenarios.sh` updated to verify aliases are registered automatically instead of applying them via a compose override.

Closes #472

## Not addressed

- **Option 1 (compose override persistence)**: Not implemented per issue recommendation to start with Option 2. Can be added later as a persistence layer if operators request it.
- **`service remove` cleanup**: Out of scope per issue; `bootroot clean` already tears down the compose stack and state.

## Test plan

- [x] `cargo test` passes (including new unit tests in `dns_alias.rs`)
- [x] `scripts/preflight/extra/cli-scenarios.sh` passes — verifies aliases are present on `bootroot-http01` after `service add`
- [x] Manual: run `bootroot service add`, then `docker inspect bootroot-http01` and confirm the validation FQDN appears in network aliases
- [x] Manual: run `docker compose down` + `bootroot infra up`, then confirm aliases are replayed from `state.json`
- [x] Manual: run `bootroot service add` when `bootroot-http01` is not running — confirm graceful warning (no crash)
- [x] Docs updated: CHANGELOG, README, cli.md, cli-examples.md, configuration.md, index.md, installation.md, troubleshooting.md (en + ko)


sehkone:
**[Reviewer Round 1]**

Thanks — the runtime alias approach is the right direction, and `cargo clippy --all-targets --all-features -- -D warnings` plus `cargo test` both pass locally. I’m still marking this round as changes requested for two blockers:

- `src/commands/dns_alias.rs:66` disconnects `bootroot-http01` from its network before attempting the reconnect, but `src/commands/service.rs:296` then downgrades any reconnect failure to a warning and continues. If `docker network connect` fails for any reason, `service add` succeeds while leaving the responder detached from the Compose network, which breaks subsequent HTTP-01 traffic and admin API access. This needs an atomic/rollback path (or a different update strategy) so a failed alias registration cannot take the responder offline.
- The issue explicitly called out doc updates for `docs/en/installation.md`, `docs/ko/installation.md`, and `README.md`, but those files are unchanged. `docs/en/installation.md:427` still describes the responder setup without mentioning that `service add` now auto-registers the validation FQDN, and `README.md:35` still summarizes `service add` without the new zero-manual-override behavior. Please update the installation docs/README so the user-facing flow matches the feature.

sehkone:
**[Author Round 1]**

Both items addressed in ce962c5:

**1. Rollback on reconnect failure (`dns_alias.rs`)**

`apply_dns_aliases` now catches a failed `docker network connect` and immediately attempts a rollback reconnect (without aliases, just the compose service name) so that `bootroot-http01` is never left detached from the network. If even the rollback fails, a `CRITICAL` message is printed with the manual recovery command. The warning-and-continue behavior in `service.rs` is now safe because the container's network connectivity is restored before the error propagates.

New i18n messages added for both languages: `dns_alias_connect_rollback` and `dns_alias_rollback_failed`.

**2. Missing doc updates**

- `docs/en/installation.md` / `docs/ko/installation.md`: added "DNS alias (automatic)" subsection under the Docker responder section, explaining that `service add` auto-registers the validation FQDN and that `infra up` replays aliases after restarts.
- `README.md`: updated the service onboarding bullet to mention automatic HTTP-01 DNS alias registration and the removal of manual `docker-compose.override.yml`.
- PR description updated to reflect these additions.

sehkone:
**[Reviewer Round 2]**

Thanks for the fast follow-up — the doc gaps are closed, and the rollback reconnect addresses most of the safety problem from round 1.

I still have one blocker before this is ready:

- `src/commands/dns_alias.rs:85` and `src/commands/service.rs:296`: if the first `docker network connect` fails *and* the rollback reconnect fails, `apply_dns_aliases` prints the `CRITICAL` recovery message but `service add` still downgrades the error to a warning and exits successfully. In that path `bootroot-http01` really is detached from its Docker network, so the command reports success while leaving HTTP-01 validation and the responder admin API broken. Please make the rollback-failed path propagate as a hard failure (or otherwise distinguish it from the recoverable “aliases not applied, but connectivity restored” case) instead of always warning-and-continuing.

sehkone:
**[Author Round 2]**

Addressed the rollback-failed blocker in ec83dee:

**Distinguishing recoverable vs. critical failures**

`apply_dns_aliases` now returns `Ok(())` for all cases where the container remains connected to the Docker network (including when aliases could not be applied), and returns `Err` only when the container is left detached (disconnect succeeded but both reconnect and rollback failed).

Specifically:
- **Container not running / network not found / disconnect failed**: the container was never detached, so these print a warning and return `Ok(())`.
- **Connect with aliases failed, rollback succeeded**: connectivity is restored (just without aliases). Prints a warning via the new `dns_alias_connect_recovered` i18n message, returns `Ok(())`.
- **Connect with aliases failed, rollback also failed**: the container is detached from the network. Returns `Err` so the caller fails hard.

In `service.rs`, the old `if let Err` + warning-and-continue is replaced with `?`, which now only triggers on the critical (container-detached) path.

The `dns_alias_disconnect_failed` i18n field was removed since disconnect failures are now handled inline as recoverable warnings.

sehkone:
**[Reviewer Round 3]**

Thanks — the round-2 change closes the remaining blocker. `register_dns_alias` now only hard-fails when `bootroot-http01` is actually left detached, while the recovered paths stay warning-only, which matches the operational risk. I also rechecked the `infra up` replay path and the documentation updates; the state-based replay and user-facing flow are consistent with the issue scope.

Locally, `cargo test --bin bootroot` and `cargo clippy --all-targets --all-features -- -D warnings` both pass. I do not see further correctness, test-coverage, error-handling, security, or performance concerns that block merge.

APPROVED

PR #483, opened 2026-04-11
Preserve CLI overrides across daemon retries (#475)

### LINKED ISSUE #475: `bootroot-agent` daemon retries silently drop CLI overrides
## Summary

When `bootroot-agent` runs in daemon mode (without `--oneshot`), the retry path reloads the config file from disk on every attempt. CLI arguments passed to the process (e.g. `--http-responder-hmac`, `--ca-url`, `--http-responder-url`, `--email`) are applied only on the first attempt. On retry, the reloaded `Settings` contains only what the TOML file has, silently reverting to defaults or empty values for fields the user supplied via CLI.

## Reproduction

1. Create an agent config without `[acme].http_responder_hmac`:

   ```toml
   domain = "test.local"

   [[profiles]]
   service_name = "review"
   instance_id = "001"
   hostname = "review-host"
   [profiles.paths]
   cert = "/tmp/review-cert.pem"
   key = "/tmp/review-key.pem"
   [trust]
   ca_bundle_path = "/tmp/ca-bundle.pem"
   trusted_ca_sha256 = ["..."]
   ```

2. Run the daemon, passing HMAC on the CLI:

   ```
   bootroot-agent --config agent.toml --http-responder-hmac "<real-value>" --insecure
   ```

3. Observe: first attempt reaches the responder successfully (HMAC from CLI is used), but renewal/issuance triggers a retry, and all subsequent attempts fail with:

   ```
   Responder returned 401 Unauthorized: Invalid signature
   ```

4. Using `--oneshot` with the same command line succeeds every time.

## Root cause

`src/daemon.rs::issue_with_retry` calls `config::Settings::new(Some(path))` inside the retry closure, producing a `fresh` `Settings` value each attempt. The first attempt uses the merged settings that include CLI overrides; the reloaded `fresh` does not re-apply CLI overrides, because `Args` are not visible at that point.

```rust
issue_with_retry_inner(
    || {
        let path = config_path_owned.clone();
        let domain = profile_domain.clone();
        let eab = eab.clone();
        async move {
            let fresh = config::Settings::new(Some(path))?;
            // ... uses `fresh` without merge_with_args
            acme::issue_certificate(&fresh, &fresh_profile, fresh_eab, insecure_mode).await
        }
    },
    ...
)
```

`Settings::merge_with_args` (which handles the CLI overrides) is called once at startup by the main binary, not inside the retry closure.

## User impact

The bug is especially misleading because:

- `--oneshot` works, so users assume the daemon will work too.
- The first attempt in daemon mode also works, so logs show a successful start before failing later.
- No warning is printed about fields that originated from CLI only.

The typical fallout is what motivated the sibling issue about incomplete agent configs from `service add`: users eventually hand-add everything to the TOML so that the reload path can find it.

## Proposal

Pick one of the following:

1. **Preserve CLI overrides across reloads.** Capture the parsed `Args` (or the merged `Settings`) at daemon start and re-apply the overrides after each reload. This preserves the "reload picks up profile edits" semantics while keeping runtime flags sticky.
2. **Skip reloading when CLI overrides exist.** If any field would be overwritten by CLI-provided values, log a warning once and reuse the initial `Settings` for retries.
3. **Validate completeness at startup and require config-only runs in daemon mode.** If daemon mode is selected and required fields were only supplied via CLI, fail fast with a clear error pointing the user to put those fields in the config file.

Option 1 is the least surprising. Option 3 is the simplest safeguard and can ship first while option 1 is being worked on.

## Documentation

- `CHANGELOG.md` — note the behavior change for CLI-provided overrides in daemon mode.
- `docs/en/configuration.md` — document how CLI overrides interact with daemon-mode config reloads (in the "CLI Overrides" section).
- `docs/en/troubleshooting.md` — add a troubleshooting entry for the old symptom (CLI HMAC working once then failing) so users hitting older builds can self-diagnose.
- `docs/ko/configuration.md`, `docs/ko/troubleshooting.md` — mirror the English updates.

## Related

Fixing #473 (agent config completeness from `service add`) will reduce how often users hit this bug in practice, but the underlying behavior is still wrong and should be fixed independently.

sehkone:
## Summary

- Introduced `CliOverrides` struct that captures CLI-provided values (`--email`, `--ca-url`, `--http-responder-url`, `--http-responder-hmac`) at startup.
- The daemon retry path now calls `Settings::apply_overrides()` after every config-file reload, keeping CLI flags sticky for the lifetime of the process.
- Added regression tests verifying that reload-then-apply preserves CLI values.

Closes #475

## Test plan

- [x] `cargo test` passes, including the three new `CliOverrides` / `apply_overrides` unit tests
- [x] Manual: run `bootroot-agent` in daemon mode with `--http-responder-hmac` on the CLI (not in `agent.toml`) — verify retries still send the correct HMAC (no `401 Unauthorized`)
- [x] Manual: run `bootroot-agent --oneshot` with CLI overrides — verify behavior is unchanged
- [x] Manual: modify `agent.toml` between retries (e.g. change `email`) — verify the file change is picked up while CLI overrides still win for their fields
- [ ] `scripts/preflight/run-all.sh` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fixed daemon-mode retries silently dropping CLI overrides (`--email`,
+  `--ca-url`, `--http-responder-url`, `--http-responder-hmac`). The retry
+  path reloaded the config file from disk without re-applying CLI-provided
+  values, causing the first attempt to succeed but subsequent attempts to
+  revert to file-only defaults.
 - Fixed `bootroot service add` (`local-file` mode) generating an agent config
   missing top-level `domain` and `[acme].http_responder_hmac`. The generated
   `agent.toml` is now ready to use without manual editing.
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -550,7 +550,10 @@ The command is considered failed when:
 Runs a one-shot issuance via bootroot-agent and verifies cert/key output.
 Use it after service onboarding or config changes to confirm issuance works.
 If you want **continuous renewal**, run bootroot-agent in daemon mode
-(without `--oneshot`) after verification.
+(without `--oneshot`) after verification. Any CLI overrides you pass
+(e.g. `--http-responder-hmac`, `--ca-url`) are preserved across
+daemon retry cycles, so the daemon behaves the same as `--oneshot`
+for those flags.
 
 ### Inputs
 
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -405,6 +405,11 @@ so CLI has the highest priority. For example, if `agent.toml` sets
 `email = "admin@example.com"`, running `bootroot-agent --email ops@example.com`
 uses `ops@example.com`.
 
+In daemon mode, the agent periodically reloads the config file from disk
+when retrying failed issuances. CLI overrides are preserved across these
+reloads — a value supplied via `--http-responder-hmac` on the command line
+remains in effect for every retry, not just the first attempt.
+
 ## HTTP-01 responder (responder.toml)
 
 The responder reads `responder.toml` (or `BOOTROOT_RESPONDER__*` env vars).
--- a/docs/en/troubleshooting.md
+++ b/docs/en/troubleshooting.md
@@ -118,6 +118,17 @@ If mode and placement do not match, file/state updates go to the wrong path.
 - Requested SANs do not match step-ca provisioner policy
 - Validate both SAN generation and CA policy together
 
+### CLI HMAC works once then fails on retry (older builds)
+
+If the first issuance attempt succeeds but retries fail with
+`401 Unauthorized: Invalid signature`, and you are passing
+`--http-responder-hmac` (or other CLI overrides) without putting the value
+in `agent.toml`, upgrade to a build that includes the fix for #475.
+Older builds dropped CLI overrides on retry because the daemon reloaded the
+config file without re-applying command-line values. As a workaround on
+older builds, add `http_responder_hmac` to the `[acme]` section of
+`agent.toml` directly.
+
 ### Repeated ACME directory retries
 
 - Ensure `server` URL is `https://` (`http://` is rejected)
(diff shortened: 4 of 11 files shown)


sehkone:
**[Reviewer Round 1]**

Thanks — the Rust fix itself looks correct, and `cargo test apply_overrides`, `cargo test reload_then_apply_overrides`, and `cargo clippy --all-targets --all-features -- -D warnings` all pass locally.

I found one blocking issue:
- The docs update is incomplete. The issue explicitly requires `docs/en/cli.md` plus mirrored Korean docs, but this PR only updates configuration/troubleshooting. `docs/en/cli.md:548` and `docs/ko/cli.md:532` still tell users to run the daemon after verification without explaining that daemon retries now preserve CLI overrides, which is the behavior change this PR introduces. Please add the CLI doc update in both languages so the documented daemon-mode semantics match the code.

sehkone:
**[Author Round 1]**

Updated `docs/en/cli.md` and `docs/ko/cli.md` in the `bootroot verify` section to explain that CLI overrides (e.g. `--http-responder-hmac`, `--ca-url`) are preserved across daemon retry cycles, so daemon mode behaves the same as `--oneshot` for those flags.

Commit: eb0d336

sehkone:
**[Reviewer Round 2]**

Thanks for the follow-up. I re-checked the code and docs update from `eb0d336`.

- The daemon retry path now re-applies the captured CLI override set after each reload, which fixes the original regression without giving up file-reload semantics for non-overridden fields.
- The override set matches the fields handled by `Settings::merge_with_args`, and the new config-level regression tests cover both full replacement and reload-then-apply behavior.
- The missing CLI docs are now present in `docs/en/cli.md` and `docs/ko/cli.md`, and the changelog/configuration/troubleshooting updates are aligned with the implementation.

Validation I ran locally:
- `cargo test test_apply_overrides_replaces_all_fields -- --nocapture`
- `cargo test test_apply_overrides_skips_none_fields -- --nocapture`
- `cargo test test_reload_then_apply_overrides_preserves_cli_values -- --nocapture`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check origin/main...HEAD`

No further blocking issues from me. APPROVED

PR #484, opened 2026-04-11
Add post-renew hook flags to service add (#474)

### LINKED ISSUE #474: Add post-renew hook flags to `service add`
## Background

When `bootroot-agent` renews a certificate, the service using the cert must reload it. The mechanism differs by service:

- systemd unit: `systemctl reload <unit>`
- long-lived process supporting SIGHUP: `kill -HUP <pid>` or `pkill -HUP -f <pattern>`
- Docker container: `docker restart <name>` or `docker kill --signal=HUP`
- custom signal (SIGUSR1, etc.): service-specific

bootroot-agent already supports this via the `[[profiles.hooks.post_renew.success]]` TOML block (see `src/config.rs`). However, `bootroot service add` does not accept any flags for it (see `ServiceAddArgs` in `src/cli/args.rs`), so users must hand-edit the generated agent config after running `service add` — exactly the kind of manual step that #464 aimed to eliminate.

## Scope and affected code paths

This is not just "add a few flags". Hook information needs to flow through several components that currently have no representation for it:

1. **`ServiceAddArgs` in `src/cli/args.rs`** — the CLI struct has no fields for hook command/args/timeout/on-failure.
2. **Remote bootstrap artifact in `src/commands/service/remote_bootstrap.rs`** — the artifact carried to the remote host does not include hook metadata, so enabling hooks for remote-bootstrap services requires extending the artifact schema.
3. **Persistent state in `src/state.rs`** — `ServiceEntry` has no hook fields. The daemon renewal path is not affected by this (it reads the agent config from disk directly, see `src/daemon.rs`), but state-driven re-runs such as `is_idempotent_remote_rerun` and `run_service_add_remote_idempotent` in `src/commands/service.rs` cannot recover or regenerate hook settings after the initial `service add` without those fields.
4. **Local agent config generator in `src/commands/service/local_config.rs`** and its baseline counterpart in `src/bin/bootroot-remote/agent_config.rs` — both need to emit the hook block when fields are present.

Any of these four layers left out would produce surprising behavior (e.g. hooks that work on first run but vanish after a state-driven rewrite during an idempotent re-run).

## Proposal

Add CLI flags to `bootroot service add` that populate the hook section, and thread the values through CLI args → state → config generators (both local and remote bootstrap). Two possible flag shapes:

### Option A: low-level flags (mirrors TOML structure)

```
--post-renew-command <cmd>
--post-renew-arg <arg>           # repeatable
--post-renew-timeout-secs <n>    # default 30
--post-renew-on-failure <continue|stop>  # default continue
```

Flexible, maps 1:1 to the existing hook schema.

### Option B: presets

```
--reload-style <sighup|systemd|docker-restart|none>
--reload-target <pattern|unit-name|container-name>
```

Less typing for common cases. bootroot expands the preset into the full hook block.

The two are not mutually exclusive — presets could be sugar on top of the low-level flags.

## Scope

This issue only covers `success` hooks. `failure` hooks (for alerting on renewal failure) can follow the same pattern later if needed.

## Documentation

- `CHANGELOG.md` — note the new `service add` flags and that post-renew hooks can now be configured at service-registration time.
- `docs/en/cli.md`, `docs/en/cli-examples.md`, `docs/en/operations.md`, `docs/en/configuration.md` — document the new flags and reload-style presets, and show an example of certificate renewal triggering a service reload.
- `docs/ko/*.md` — mirror the English updates.
- `README.md` — mention post-renew hooks in the feature overview if applicable.

## Related

Part of the broader cleanup of manual editing steps left after #464.

sehkone:
## Summary

- Adds preset flags (`--reload-style`, `--reload-target`) and low-level flags (`--post-renew-command`, `--post-renew-arg`, `--post-renew-timeout-secs`, `--post-renew-on-failure`) to `bootroot service add` so that post-renew hooks can be configured at service registration time without hand-editing `agent.toml`.
- Threads hook data through all four layers identified in the issue: `ServiceAddArgs` (CLI), `ServiceEntry` (persistent state), local agent-config generator (`local_config.rs`), and remote bootstrap artifact/command (`remote_bootstrap.rs` + `bootroot-remote agent_config.rs`).
- Presets and low-level flags are mutually exclusive; presets expand into the equivalent low-level hook block. Idempotent re-runs (`is_idempotent_remote_rerun`) compare hook fields so state-driven rewrites preserve hook settings.
- `bootroot-remote bootstrap` validates hook flag coherence (rejects `--post-renew-arg`/`--post-renew-timeout-secs`/`--post-renew-on-failure` without `--post-renew-command`) instead of silently ignoring orphan flags.
- Both `service add` and `bootroot-remote bootstrap` reject empty/whitespace `--post-renew-command` and `--post-renew-timeout-secs 0`, matching the runtime config validation in `config/validation.rs`.
- Hook TOML emitters use `toml_edit::Value`-based encoding (via `toml_encode_string`) instead of manual escaping, correctly handling control characters (newline, tab, etc.) in command/arg values.

Closes #474

## Test plan

- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo test` passes (new unit tests for resolve, local_config, remote_bootstrap, state, cli args)
- [x] `--reload-style systemd --reload-target nginx` generates `systemctl reload nginx` hook in `agent.toml`
- [x] `--reload-style sighup --reload-target myproc` generates `pkill -HUP myproc` hook
- [x] `--reload-style docker-restart --reload-target ctr` generates `docker restart ctr` hook
- [x] `--reload-style none` generates no hook block
- [x] `--post-renew-command`/`--post-renew-arg` (low-level) generates correct hook block
- [x] Combining preset and low-level flags produces a clear error
- [x] `--reload-target` without `--reload-style` produces a clear error
- [x] `--post-renew-arg` without `--post-renew-command` produces a clear error
- [x] Empty/whitespace `--post-renew-command` is rejected
- [x] `--post-renew-timeout-secs 0` is rejected
- [x] Hook fields round-trip through `state.json` serialization
- [x] Existing `state.json` without `post_renew_hooks` deserializes with empty vec (backward compat)
- [x] Remote bootstrap artifact includes hook data and `bootroot-remote bootstrap` command template includes hook flags
- [x] `bootroot-remote bootstrap` with hook flags injects `[[profiles.hooks.post_renew.success]]` into `agent.toml`
- [x] `bootroot-remote bootstrap` rejects orphan `--post-renew-*` flags without `--post-renew-command`
- [x] `bootroot-remote bootstrap` rejects empty command and zero timeout
- [x] Idempotent remote re-run detects hook field changes correctly
- [x] Hook values with control characters (newline, tab) produce valid parseable TOML on both local and remote paths
- [x] CHANGELOG, README, EN docs, KO docs are updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Added post-renew hook flags to `bootroot service add`. Services can now
+  configure a hook to run after successful certificate renewal at
+  registration time, removing the need to hand-edit `agent.toml`.
+  Two flag styles are supported: presets (`--reload-style systemd
+  --reload-target nginx`) and low-level (`--post-renew-command`,
+  `--post-renew-arg`, `--post-renew-timeout-secs`,
+  `--post-renew-on-failure`). Hook settings are persisted in
+  `state.json` and forwarded to `bootroot-remote bootstrap` for
+  remote-bootstrap delivery mode.
 - Added automatic HTTP-01 DNS alias registration on `service add`. The
   validation FQDN is registered as a Docker network alias on
   `bootroot-http01` at runtime, removing the need for a hand-written
--- a/README.md
+++ b/README.md
@@ -34,8 +34,9 @@ Open source dependencies:
   visualizes them for local ops.
 - **Service onboarding**: `bootroot service add` registers service metadata,
   creates an AppRole, automatically registers the HTTP-01 DNS alias on the
-  responder, and prints the run instructions for bootroot-agent and OpenBao
-  Agent — no manual `docker-compose.override.yml` required.
+  responder, optionally configures a post-renew hook (`--reload-style` or
+  `--post-renew-command`), and prints the run instructions for bootroot-agent
+  and OpenBao Agent — no manual config editing required.
 - **Certificate flow**: bootroot-agent issues/renews certs; OpenBao Agent
   renders secrets and config files for each service.
 
--- a/docs/en/cli-examples.md
+++ b/docs/en/cli-examples.md
@@ -277,6 +277,37 @@ In operations, start from the `remote run command template` printed by
 `--agent-responder-url` with remote-reachable endpoints such as
 `stepca.internal` and `responder.internal`.
 
+### 3-4) Post-renew hook (reload style preset)
+
+Use `--reload-style` and `--reload-target` to configure
+a post-renew hook that reloads the service after
+certificate renewal:
+
+```bash
+bootroot service add \
+  --service-name edge-proxy \
+  --deploy-type daemon \
+  --hostname edge-node-01 \
+  --domain trusted.domain \
+  --agent-config /etc/bootroot/agent.toml \
+  --cert-path /etc/bootroot/certs/edge-proxy.crt \
+  --key-path /etc/bootroot/certs/edge-proxy.key \
+  --instance-id 001 \
+  --reload-style systemd \
+  --reload-target nginx \
+  --root-token <OPENBAO_ROOT_TOKEN>
+```
+
+This expands into a `[profiles.hooks.post_renew]` entry
+in the generated `agent.toml` profile that runs
+`systemctl reload nginx` after each successful renewal.
+Other preset styles are `sighup` (sends `SIGHUP` via
+`pkill`), `docker-restart` (runs `docker restart`), and
+`none` (no hook). For full control, use the low-level
+flags `--post-renew-command`, `--post-renew-arg`,
+`--post-renew-timeout-secs`, and
+`--post-renew-on-failure` instead.
+
 After secret_id rotation on the control node, deliver the new secret_id to the
 remote node:
 
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -492,6 +492,35 @@ Input priority is **CLI flags > environment variables > prompts/defaults**.
 - `--print-only`: print snippets/next steps without writing state/files
 - `--dry-run`: alias of preview mode (same effect as `--print-only`)
 
+Post-renew hook flags (preset):
+
+- `--reload-style`: reload style preset
+  (`sighup`, `systemd`, `docker-restart`, or `none`)
+- `--reload-target`: target for the preset
+  (process name, systemd unit, or container name)
+
+Presets expand into the following `post_renew` hook
+entries in the generated `agent.toml` profile:
+
+- `systemd` + target `nginx` — `systemctl reload nginx`
+- `sighup` + target `nginx` — `pkill -HUP nginx`
+- `docker-restart` + target `my-container` — `docker restart my-container`
+- `none` — no hook
+
+Post-renew hook flags (low-level):
+
+- `--post-renew-command`: hook command to run after successful renewal
+- `--post-renew-arg`: hook argument (repeatable)
+- `--post-renew-timeout-secs`: hook timeout in seconds (default `30`)
+- `--post-renew-on-failure`: failure policy (`continue` or `stop`, default `continue`)
+
+Preset flags (`--reload-style`/`--reload-target`) and
+low-level flags (`--post-renew-*`) are mutually
+exclusive. When preset flags are used, they expand into
+the equivalent low-level hook settings. These flags are
+also forwarded to `bootroot-remote bootstrap` for the
+`remote-bootstrap` delivery mode.
+
 ### Interactive behavior
 
 - Prompts for missing required inputs (deploy type defaults to `daemon`).
@@ -1010,6 +1039,12 @@ Key inputs:
     `responder.internal`.
 - `--ca-bundle-path`
   - required output path for the managed step-ca trust bundle.
+- post-renew hook flags: `--reload-style`,
+  `--reload-target`, `--post-renew-command`,
+  `--post-renew-arg`, `--post-renew-timeout-secs`,
+  `--post-renew-on-failure` (same semantics as
+  `bootroot service add`; passed through from the
+  generated remote handoff command template)
 - `--summary-json` (optional) and `--output text|json` (default `text`)
 
 If `agent.toml` does not exist yet, bootstrap creates a baseline config and
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -376,6 +376,20 @@ Hooks run **after** issuance/renewal completes. Use `success` for actions that
 should run only when a certificate is issued/renewed, and `failure` for alerting
 or cleanup when issuance fails.
 
+Hooks can also be configured at `bootroot service add`
+time using CLI flags instead of editing `agent.toml`
+manually. Use the preset flags `--reload-style` and
+`--reload-target` for common reload patterns (e.g.
+`--reload-style systemd --reload-target nginx`), or the
+low-level flags `--post-renew-command`,
+`--post-renew-arg`, `--post-renew-timeout-secs`, and
+`--post-renew-on-failure` for full control. These flags
+generate the corresponding
+`[profiles.hooks.post_renew]` entry in the managed
+profile. For the `remote-bootstrap` delivery mode, the
+same flags are forwarded to
+`bootroot-remote bootstrap`.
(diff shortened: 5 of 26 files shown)


sehkone:
**[Reviewer Round 1]**

Thanks — the end-to-end plumbing and test coverage are in good shape overall, but I found two blocking correctness issues:

1. `src/commands/service/resolve.rs:230` only treats `--post-renew-command` as the "low-level" side of the mutual exclusion. That means mixes like `--reload-style systemd --reload-target nginx --post-renew-timeout-secs 60` or `--reload-style none --post-renew-on-failure stop` are accepted and the extra `--post-renew-*` flags are silently ignored, even though the docs say the preset and low-level modes are mutually exclusive. This can easily mislead operators into thinking a non-default timeout / failure policy was applied when the generated hook still uses the defaults. Please either reject any `--post-renew-*` flag when `--reload-style` is present, or explicitly support overriding preset defaults.

2. `src/commands/service/remote_bootstrap.rs:162` now renders arbitrary hook commands/args into the operator-facing `bootroot-remote bootstrap ...` template by wrapping them in single quotes, but it never shell-escapes embedded `'` characters. With the new low-level flags, values like `--post-renew-arg "O'Brien"` or any custom command containing a quote produce a broken command template (and potentially a shell-injection footgun if someone pastes it unedited). Please shell-escape these values before embedding them in the template, or emit them in a format that avoids shell parsing hazards.

sehkone:
**[Author Round 1]**

Both items fixed in 59f2d5d:

1. **Mutual exclusion now covers all `--post-renew-*` flags.** The `has_low_level` check in `resolve_post_renew_hooks` now includes `--post-renew-arg`, `--post-renew-timeout-secs`, and `--post-renew-on-failure` — not just `--post-renew-command`. Combinations like `--reload-style systemd --reload-target nginx --post-renew-timeout-secs 60` now produce an error. Added three new tests for each previously-uncaught combination.

2. **Shell-escaping for hook values in the remote bootstrap command template.** Added `shell_escape_single_quoted()` which replaces `'` with `'\''` before embedding values inside single quotes. Applied to `--post-renew-command`, `--post-renew-arg`, and `--post-renew-on-failure` values. Added a test with embedded single quotes to verify.

No doc changes needed — the docs already correctly describe preset and `--post-renew-*` flags as mutually exclusive; the code now matches.

sehkone:
**[Reviewer Round 1]** Overall the CLI → state → local/remote config threading looks solid, but I found one blocker before this is merge-ready.

- `bootroot-remote bootstrap` does not validate the new low-level hook flags as a coherent set. `src/bin/bootroot-remote/bootstrap.rs:123` validates service/profile/path inputs, but never rejects `--post-renew-arg`, `--post-renew-timeout-secs`, or `--post-renew-on-failure` without `--post-renew-command`. Then `src/bin/bootroot-remote/agent_config.rs:208` just returns early when `post_renew_command` is absent, so those flags are silently ignored instead of producing a user-facing error. I reproduced this locally by running `bootroot-remote bootstrap ... --post-renew-arg reload --post-renew-timeout-secs 60`, and the command proceeded all the way to OpenBao login instead of failing validation. Because this PR adds those flags to the remote bootstrap surface, I think it needs the same validation guarantees as `resolve_post_renew_hooks()` on the `bootroot service add` side, plus a regression test for that path.

sehkone:
**[Author Round 1]**

Fixed in e026918:

- **`bootroot-remote bootstrap` now validates hook flag coherence.** Added `validate_hook_flags()` that runs at the start of `validate_bootstrap_args()` and rejects `--post-renew-arg`, `--post-renew-timeout-secs`, and `--post-renew-on-failure` when `--post-renew-command` is absent. This matches the validation already present in `resolve_post_renew_hooks()` on the `bootroot service add` side.
- Added five regression tests covering: orphan flags with multiple orphans, timeout alone, on-failure alone, valid flags with command present, and no flags at all.
- Updated PR description to reflect the new validation and added the corresponding test plan item.

sehkone:
**[Reviewer Round 2]**

Thanks for the quick turnaround — the mutual-exclusion fix and the new shell escaping look good, and I re-ran `cargo test --quiet` plus `cargo clippy --quiet -- -D warnings` locally.

I still found one blocking validation gap in the new hook plumbing:

1. The new `service add` / `bootroot-remote bootstrap` paths still accept hook values that the runtime config loader explicitly rejects.
   - `src/config/validation.rs:131` requires hook commands to be non-empty, and `src/config/validation.rs:141` requires `timeout_secs > 0`.
   - But `src/commands/service/resolve.rs:247` currently accepts `--post-renew-command ''` / whitespace-only values and `--post-renew-timeout-secs 0`, then `src/commands/service/local_config.rs:31` writes that TOML directly.
   - `src/bin/bootroot-remote/bootstrap.rs:123` has the same gap on the remote path, and `src/bin/bootroot-remote/agent_config.rs:207` injects the same invalid hook block into the generated baseline config.

That means `bootroot service add` can succeed while producing an `agent.toml` that `bootroot-agent` later refuses to load, and the remote bootstrap path can do the same. Please reject blank `--post-renew-command` and `--post-renew-timeout-secs 0` in both entry points (or centralize the hook validation and reuse it), and add regression tests for both cases.

NOT_APPROVED

sehkone:
**[Author Round 2]**

Fixed in d503190:

- **`service add` rejects invalid hook values.** `resolve_post_renew_hooks` now rejects empty/whitespace `--post-renew-command` and `--post-renew-timeout-secs 0` before writing the hook block, matching the runtime checks in `config/validation.rs`.
- **`bootroot-remote bootstrap` rejects the same.** `validate_hook_flags` now checks for blank command and zero timeout in addition to the existing orphan-flag check.
- **Regression tests added for both paths.** Three new tests each in `resolve.rs` (empty command, whitespace-only command, zero timeout) and `bootstrap.rs` (same three cases).

sehkone:
**[Reviewer Round 3]**

Thanks — I re-ran `cargo test --quiet` and `cargo clippy --quiet -- -D warnings` locally, and the previous validation fixes are in place.

I still found one blocking correctness issue:

- The new hook TOML emitters do not fully escape TOML basic strings. `src/commands/service/local_config.rs:179` and `src/commands/service/local_config.rs:195` plus `src/bin/bootroot-remote/agent_config.rs:219` and `src/bin/bootroot-remote/agent_config.rs:226` only escape `\\` and `"`, so values containing control characters like newline or tab are written as invalid TOML. Because `bootroot service add` and `bootroot-remote bootstrap` accept arbitrary `--post-renew-command` / `--post-renew-arg` strings, a command such as `--post-renew-command $'echo\nnext'` succeeds but produces an `agent.toml` that `bootroot-agent` refuses to load. I confirmed this locally by feeding the resulting snippet to `bootroot-agent --oneshot`, which fails with `TOML parse error ... key with no value`. Please serialize these values with a real TOML string encoder (for example via `toml_edit::Value::from(...)` or a shared helper) instead of manual replacement, and add a regression test covering at least a newline-containing command/arg on both the local and remote config-generation paths.

NOT_APPROVED

sehkone:
**[Author Round 3]**

Fixed in a117be1:

- **Hook TOML emitters now use `toml_edit::Value` for string encoding.** Added `toml_encode_string()` in `src/toml_util.rs` backed by `toml_edit::Value::from(...)`, which handles the full TOML basic-string escape set (backslash, double quote, newline, tab, and all other control characters). Replaced the manual `toml_escape` helper in `local_config.rs` and the inline `.replace()` calls in `agent_config.rs` with this shared function.
- **Regression tests added on both paths.** `test_render_hooks_toml_escapes_control_characters` (local) and `inject_hooks_escapes_control_characters` (remote) both construct hooks with embedded newlines and tabs, render them to TOML, parse the output with `toml_edit`, and verify the values round-trip correctly.

sehkone:
**[Reviewer Round 4]**

Thanks — the shared TOML string encoder in `src/toml_util.rs` and the new round-trip regression tests on both the local and remote config-generation paths address the remaining blocker from Round 3. I re-ran `cargo test --quiet` and `cargo clippy --quiet -- -D warnings` locally, and both passed.

I also re-checked the other hook plumbing added in this PR (CLI args, resolver validation, state persistence, local config emission, remote bootstrap artifact/command generation, and idempotent remote reruns), and it now looks consistent end to end.

APPROVED

PR #486, opened 2026-04-12
Add remote-bootstrap operator guide and schema_version (#479)

### LINKED ISSUE #479: Remote bootstrap operator guide and artifact schema stabilization
## Background

bootroot supports a `remote-bootstrap` delivery mode (`service add --delivery-mode remote-bootstrap`) for services running on a machine other than the one hosting step-ca / OpenBao / the HTTP-01 responder. In this mode, `service add` produces a JSON artifact (`RemoteBootstrapArtifact`, written to `secrets/remote-bootstrap/services/<service>/bootstrap.json`) plus `role_id`/`secret_id` files, and the operator is expected to ship those to the remote host and invoke `bootroot-remote bootstrap` there.

The mechanism itself is sound and intentionally leaves "file transport" to the operator's existing deployment tooling (SSH, Ansible, cloud-init, systemd-credentials, etc.). However, the documentation and CLI surface do not currently give operators enough to go on:

- The JSON artifact exists but its schema is not documented anywhere; readers have to trace `RemoteBootstrapArtifact` in `src/commands/service/remote_bootstrap.rs` to know what the fields mean.
- The artifact has no `schema_version` field, so downstream automation (shell scripts, Ansible, CI pipelines) cannot defend against silent breakage when bootroot adds or renames fields later.
- There are no end-to-end examples showing how a common deployment tool wires everything together. Operators coming to bootroot for the first time have to invent their own pattern, and the easy mistakes (committing `secret_id` to git, leaving it in shell history, skipping permission hardening, etc.) are not called out in one place.

This issue fills those gaps. It is primarily a documentation effort, together with the smallest code surface needed to make the documentation stable for automation: a `schema_version` field on the artifact, rustdoc describing it, and a small tweak to `service add` output so that operators can find the artifact file.

### Documentation (primary)

Add a new page, tentatively `docs/en/remote-bootstrap.md` (with a Korean mirror under `docs/ko/`), covering:

1. **When to use remote-bootstrap vs local-file** delivery modes.
2. **Prerequisites on the remote host**: `bootroot-remote` binary installed, network reachability to OpenBao and step-ca, filesystem layout expectations.
3. **The transport boundary** — an explicit statement that bootroot does not ship files to remote hosts by design, and that the operator chooses the transport. List the common options in increasing order of complexity:
   - **SSH + shell script** (recommended starting point for small deployments). Provide a runnable example script that calls `bootroot service add`, `scp`s `role_id` and `secret_id`, and then `ssh`s in to run `bootroot-remote bootstrap`.
   - **systemd-credentials** (`LoadCredential=`, `systemd-creds`) for single-host setups where the secret should never hit a plain filesystem.
   - **Ansible** for larger fleets with existing configuration management, including a short playbook that consumes the JSON artifact via `slurp` + `from_json`.
   - **cloud-init** `write_files` + `runcmd` for first-boot provisioning of cloud VMs.
4. **`secret_id` hygiene checklist**: file permissions, short TTL if available (cross-reference the TTL/wrapping epic once it lands), never logging or committing, rotation story.
5. **Network requirements** spelled out once: the remote host must reach OpenBao's API URL and step-ca's HTTPS endpoint; the remote host's SAN must resolve in whatever DNS / hosts file / public DNS the environment uses (this is separate from the in-compose alias automation tracked by #472, which only covers traffic between the bundled step-ca and responder containers).
6. **`RemoteBootstrapArtifact` schema reference** — a table listing every field with type, meaning, and which `bootroot-remote` flag consumes it. Versioned alongside the code via the `schema_version` field introduced in this issue.

Update pages that currently refer vaguely to "remote bootstrap" to link into the new page instead of restating the mechanism:

- `docs/en/cli.md`, `docs/en/cli-examples.md` (`service add --delivery-mode remote-bootstrap` section)
- `docs/en/installation.md`
- `docs/ko/*.md`

### CLI / code changes (minimal, to support the docs)

- **Add `schema_version: u32`** to `RemoteBootstrapArtifact` in `src/commands/service/remote_bootstrap.rs`. Start at `1`. Every future change to the struct that would break downstream parsers must bump it.
- **Add a rustdoc comment** on `RemoteBootstrapArtifact` describing `schema_version` semantics (what a bump means, what stays compatible).
- **Make the JSON artifact more discoverable** in `service add` output. The current output already lists `bootstrap_file` in the summary; add a one-liner that explicitly describes it as the machine-readable artifact for automation and links (in the docs) to the schema reference.

## Out of scope (follow-up)

- **`bootroot service add --output json`** for structured stdout summary. This is a useful enhancement for CI/shell pipelines that want to capture the command result, but it is additive and independent of the docs / schema stabilization effort. Splitting it out keeps this issue small and focused so that the operator guide and `schema_version` land quickly. File as a follow-up once this issue is merged.

## Non-goals

- Shipping `bootroot-remote` release binaries. That is a separate effort tracked as a long-term backlog item (pre-built release artifacts for `bootroot-remote` and `bootroot-agent`). Operators are still expected to build and distribute the binary themselves, and the documentation should say so explicitly.
- Implementing AppRole `secret_id` TTL controls or response wrapping. That is covered by its own epic #480. The remote-bootstrap guide should reference the epic rather than duplicate the discussion, and should be updated once the epic lands.
- Implementing end-to-end orchestration inside bootroot (pushing files, running remote commands, etc.). By design, file transport and remote execution remain the operator's responsibility.

## Documentation checklist for the PR

- [x] `docs/en/remote-bootstrap.md` created with all sections above
- [x] `docs/ko/remote-bootstrap.md` mirrored
- [x] `docs/en/cli.md`, `docs/en/cli-examples.md`, `docs/en/installation.md` updated to link into the new page
- [x] Korean counterparts updated
- [x] `README.md` quickstart / feature list mentions remote-bootstrap and links to the new guide if appropriate
- [x] `CHANGELOG.md` entry for `schema_version` and the new docs page
- [x] Rustdoc comment on `RemoteBootstrapArtifact` describing `schema_version` semantics

## Related

- Parent tracking issue: #476.
- #480 (AppRole secret_id TTL and response wrapping epic) depends on the `schema_version` field introduced here; its A-5 sub-issue bumps the version when adding wrap-related fields to the artifact. #479 should land (or at least establish `schema_version`) before #480's A-5.
- #472 automates HTTP-01 DNS aliases but only for the in-compose network between bundled containers; the remote-bootstrap guide must make clear that remote hosts need real DNS or equivalent.

sehkone:
## Summary

- Added `schema_version: u32` (starting at `1`) to `RemoteBootstrapArtifact` with rustdoc describing the versioning contract (what constitutes a breaking vs additive change).
- Created `docs/en/remote-bootstrap.md` and `docs/ko/remote-bootstrap.md` operator guides covering: when to use remote-bootstrap vs local-file, prerequisites, transport options (SSH, Ansible, cloud-init, systemd-credentials), `secret_id` hygiene checklist, network requirements, and the full artifact schema reference table.
- Updated `service add` output to label the bootstrap file as a "machine-readable artifact for automation".
- Cross-linked existing docs (`cli.md`, `cli-examples.md`, `installation.md`, `README.md`) into the new guide (both EN and KO).
- Added `mkdocs.yml` navigation entry for the new page.
- Updated tests (`bootroot_service.rs`, `e2e_remote_happy_path.rs`) to assert `schema_version` and the updated output string.

Closes #479

## Test plan

- [x] `cargo test` passes (including `test_build_artifact_populates_all_fields` asserting `schema_version == 1`)
- [x] `assert_remote_bootstrap_artifact_shape` in `bootroot_service.rs` checks `schema_version`
- [x] `e2e_remote_happy_path` asserts updated "machine-readable artifact for automation" string
- [x] `docs/en/remote-bootstrap.md` renders correctly in mkdocs (`mkdocs serve`)
- [x] `docs/ko/remote-bootstrap.md` renders correctly
- [x] All cross-links from `cli.md`, `cli-examples.md`, `installation.md` resolve
- [x] Korean counterparts have matching cross-links
- [x] `CHANGELOG.md` entry present under Added
- [x] `README.md` links to the new guide

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -6,5 +6,7 @@ ignores:
   - ".github/**"
   - ".venv/**"
 config:
+  MD013:
+    tables: false
   MD024:
     siblings_only: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Added `schema_version` field (`u32`, starting at `1`) to the
+  `RemoteBootstrapArtifact` JSON written by
+  `bootroot service add --delivery-mode remote-bootstrap`. Downstream
+  parsers should check this field before accessing artifact fields.
+- Added remote-bootstrap operator guide (`docs/en/remote-bootstrap.md`,
+  `docs/ko/remote-bootstrap.md`) covering transport options (SSH,
+  Ansible, cloud-init, systemd-credentials), `secret_id` hygiene,
+  network requirements, and the full `RemoteBootstrapArtifact` schema
+  reference.
 - Added post-renew hook flags to `bootroot service add`. Services can now
   configure a hook to run after successful certificate renewal at
   registration time, removing the need to hand-edit `agent.toml`.
--- a/README.md
+++ b/README.md
@@ -40,7 +40,10 @@ Open source dependencies:
 - **Certificate flow**: bootroot-agent issues/renews certs; OpenBao Agent
   renders secrets and config files for each service.
 
-For multi-machine or manual deployment, follow the manual guides in `docs/`.
+For multi-machine deployments using `--delivery-mode remote-bootstrap`, see
+the [Remote Bootstrap Operator Guide](docs/en/remote-bootstrap.md)
+([한국어](docs/ko/remote-bootstrap.md)). For other manual deployment options,
+follow the manual guides in `docs/`.
 
 ## Developer Layout
 
--- a/docs/en/cli-examples.md
+++ b/docs/en/cli-examples.md
@@ -230,6 +230,10 @@ next steps:
 
 ### 3-3) remote-bootstrap delivery mode + one-shot bootstrap
 
+For a full operator guide including transport options, `secret_id` hygiene,
+and the artifact schema reference, see
+[Remote Bootstrap Operator Guide](remote-bootstrap.md).
+
 Control node onboarding (artifact generation):
 
 ```bash
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -998,6 +998,11 @@ rotation.
 `bootroot-remote` also supports the global `--lang` option
 (environment variable: `BOOTROOT_LANG`).
 
+For an end-to-end walkthrough including transport options (SSH, Ansible,
+cloud-init, systemd-credentials), `secret_id` hygiene, and the
+`RemoteBootstrapArtifact` schema reference, see the
+[Remote Bootstrap Operator Guide](remote-bootstrap.md).
+
 ### `bootroot-remote bootstrap`
 
 Performs a one-shot pull and apply of service secrets/config to remote file
--- a/docs/en/installation.md
+++ b/docs/en/installation.md
@@ -422,7 +422,8 @@ install `bootroot-remote` on that service machine.
   `bootroot-remote apply-secret-id ...` (after secret_id rotation)
 
 For detailed arguments/examples in `remote-bootstrap` mode, see
-`bootroot-remote bootstrap`/`apply-secret-id` in [CLI](cli.md).
+`bootroot-remote bootstrap`/`apply-secret-id` in [CLI](cli.md) and the
+[Remote Bootstrap Operator Guide](remote-bootstrap.md).
 
 ## HTTP-01 responder
 
--- a/docs/en/remote-bootstrap.md
+++ b/docs/en/remote-bootstrap.md
@@ -1,0 +1,375 @@
+# Remote Bootstrap Operator Guide
+
+This guide covers the **remote-bootstrap** delivery mode for services
+running on a machine other than the one hosting step-ca, OpenBao, and
+the HTTP-01 responder (the **control node**).
+
+## When to use remote-bootstrap vs local-file
+
+`bootroot service add` offers two delivery modes via `--delivery-mode`:
+
+| Mode | Use when | What happens |
+| --- | --- | --- |
+| `local-file` (default) | Service runs on the **same** machine as step-ca/OpenBao | `service add` writes configs directly to disk |
+| `remote-bootstrap` | Service runs on a **different** machine | `service add` produces a JSON artifact; operator ships it to the service host and runs `bootroot-remote bootstrap` there |
+
+Choose `remote-bootstrap` whenever the service machine cannot share a
+filesystem with the control node.
+
+## Prerequisites on the remote host
+
+1. **`bootroot-remote` binary installed.** Pre-built release binaries are
+    not yet available; build from source with
+    `cargo build --release --bin bootroot-remote` and distribute the binary
+    to each service machine. See [Installation](installation.md) for details.
+
+2. **Network reachability.** The remote host must reach:
+
+    - OpenBao's API endpoint (the `--openbao-url` value, typically
+      `https://openbao.internal:8200` or your environment's equivalent).
+    - step-ca's HTTPS ACME directory (the `--agent-server` value, e.g.
+      `https://stepca.internal:9000/acme/acme/directory`).
+    - The HTTP-01 responder (the `--agent-responder-url` value, e.g.
+      `http://responder.internal:8080`).
+
+3. **DNS / name resolution.** The SAN (Subject Alternative Name) for the
+    service certificate must resolve from whatever DNS, `/etc/hosts`, or
+    cloud-internal DNS the environment uses.
+
+    !!! note
+        The in-compose DNS alias automation tracked by
+        [#472](https://github.com/aicers/bootroot/issues/472) only covers
+        traffic between bundled containers on the Docker bridge network.
+        Remote hosts need real DNS or equivalent entries configured by the
+        operator.
+
+4. **Filesystem layout.** Directories for secrets, certs, and agent config
+    must exist (or be creatable) on the remote host. The paths are defined
+    in the bootstrap artifact and passed to `bootroot-remote bootstrap`.
+
+## The transport boundary
+
+By design, bootroot does **not** ship files to remote hosts.
+`bootroot service add --delivery-mode remote-bootstrap` produces a JSON
+artifact and credential files; the operator chooses the transport mechanism
+that fits their environment.
+
+The files that must reach the remote host are:
+
+| File | Source path (control node) | Purpose |
+| --- | --- | --- |
+| `bootstrap.json` | `secrets/remote-bootstrap/services/<service>/bootstrap.json` | Machine-readable artifact consumed by `bootroot-remote bootstrap` |
+| `role_id` | `secrets/services/<service>/role_id` | AppRole identity (long-lived) |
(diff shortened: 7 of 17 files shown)


sehkone:
**[Reviewer Round 1]** Thanks for pushing this forward — the new guide, schema versioning, and doc cross-links are all in the right direction. I’m not ready to approve yet because I found two blocking issues.

1. The recommended SSH example never uses or even transfers `bootstrap.json`, despite the guide positioning it as the stable machine-readable contract. `docs/en/remote-bootstrap.md:61` says `bootstrap.json` is the artifact consumed by automation, but the “recommended starting point” example in `docs/en/remote-bootstrap.md:69` only copies `role_id`/`secret_id` and then reconstructs the `bootroot-remote bootstrap` flags by hand. That means the example still drifts whenever the CLI surface changes, which is exactly what `schema_version` is supposed to let operators avoid. Please change the example to ship the artifact and derive the remote invocation from it (or at least parse it on the control node) instead of hardcoding the arguments. The Korean mirror should stay in sync.

2. The Ansible example is internally inconsistent / broken as written. In `docs/en/remote-bootstrap.md:168` the playbook reads the JSON artifact and then passes `artifact.agent_server` / `artifact.agent_responder_url` straight through to `bootroot-remote bootstrap`, but the same page later says those fields default to localhost placeholders and must be replaced for remote hosts (`docs/en/remote-bootstrap.md:281`). So a reader following the playbook on a real remote host will bootstrap against `localhost`. On top of that, the playbook unconditionally renders `--profile-instance-id {{ artifact.profile_instance_id }}` at `docs/en/remote-bootstrap.md:206`, even though the schema explicitly allows that field to be empty at `docs/en/remote-bootstrap.md:319`; when it is empty, the command line becomes malformed. Please either show how automation overrides the placeholder endpoints and conditionally includes optional flags, or simplify the playbook so it only uses fields that are safe to pass through verbatim. Please mirror the fix in `docs/ko/remote-bootstrap.md` as well.

sehkone:
**[Author Round 1]** Both items fixed in f64c3d7.

**1. SSH example now uses `bootstrap.json` as the source of truth.**
The script ships the artifact alongside `role_id`/`secret_id`, then parses it with `jq` on the control node to derive every `bootroot-remote bootstrap` flag. No values are hardcoded in the `ssh` invocation any more. `profile_instance_id` is conditionally included via `${INSTANCE_ID:+...}` so the command stays valid when the field is empty.

**2. Ansible playbook fixes (localhost placeholders + optional field).**
- `agent_server` and `agent_responder_url` are now playbook-level vars that the operator sets to remote-reachable endpoints, with a comment explaining why the artifact values may be localhost placeholders.
- `profile_instance_id` is built in a separate `set_fact` task using a Jinja2 conditional (`default('') | length > 0`) and interpolated only when non-empty, so the command line is never malformed.

Both changes are mirrored in `docs/ko/remote-bootstrap.md`.

sehkone:
**[Reviewer Round 2]** Thanks for the update — the Ansible fix addresses my earlier concern, so I’m only carrying forward the items that still look genuinely open after re-checking the diff.

1. The recommended SSH example is still broken for a true remote host because it feeds the control-node placeholders for `--agent-server` and `--agent-responder-url` straight out of the artifact. In `docs/en/remote-bootstrap.md:104`-`docs/en/remote-bootstrap.md:116`, the script derives those flags from `bootstrap.json`, but the same guide later states that those artifact fields default to localhost placeholders and must be replaced on a separate service machine (`docs/en/remote-bootstrap.md:302`). As written, the “recommended starting point” ends up bootstrapping against `https://localhost:9000/...` and `http://127.0.0.1:8080`. Please handle those two endpoints the same way the Ansible example now does: keep using the artifact for the stable fields, but require operator-supplied remote-reachable values for the control-plane endpoints. The Korean mirror has the same problem at `docs/ko/remote-bootstrap.md:102` and `docs/ko/remote-bootstrap.md:296`.

2. The SSH example is described as runnable, but it does not create the remote destination directory before `scp`. `docs/en/remote-bootstrap.md:92`-`docs/en/remote-bootstrap.md:97` copies files into `$REMOTE_BASE/secrets/services/$SERVICE/`, which will fail on a fresh host unless that directory already exists. Please either add an explicit `ssh ... mkdir -p ...` step before the copy, or call out as a prerequisite that the directory must already exist. The Korean mirror should stay in sync here too (`docs/ko/remote-bootstrap.md:90`).

NOT_APPROVED

sehkone:
**[Author Round 2]**

Both items fixed in c05dcc0, EN and KO mirrors updated together:

1. **SSH example endpoints** — `--agent-server` and `--agent-responder-url` now come from explicit operator-supplied variables (`AGENT_SERVER`, `AGENT_RESPONDER_URL`) with the same comment the Ansible example already had ("artifact may contain localhost placeholders"). The artifact is still used for all the other stable fields.

2. **Missing `mkdir -p`** — Added an `ssh … mkdir -p` step before the `scp` so the destination directory exists on a fresh host.

sehkone:
**[Reviewer Round 3]**

The two items from Round 2 look fixed to me, so I’m not carrying them forward. I do still have one blocking documentation concern from this pass:

1. The new automation-facing examples still skip the `schema_version` preflight even though the guide now says consumers should check it before parsing. `docs/en/remote-bootstrap.md:324` says automation should validate `schema_version` before accessing fields, but the recommended SSH example goes straight from `jq` parsing to field extraction at `docs/en/remote-bootstrap.md:108`, and the Ansible example likewise does `from_json` and immediately consumes fields at `docs/en/remote-bootstrap.md:191`. The same gap exists in the Korean mirror at `docs/ko/remote-bootstrap.md:317`, `docs/ko/remote-bootstrap.md:106`, and `docs/ko/remote-bootstrap.md:189`. Since the whole point of adding `schema_version` in this issue is to let shell/Ansible/CI automation fail explicitly instead of silently drifting when the artifact changes, I think the operator guide needs to model that check in the examples themselves, not only mention it later in the schema table. Concretely, I’d add a `jq -e '.schema_version == 1'` guard before any `field()` calls in the SSH example and an `ansible.builtin.assert` task for `artifact.schema_version == 1` before the playbook starts using artifact fields.

NOT_APPROVED

sehkone:
**[Author Round 3]**

Fixed in d954ad7.

**1. `schema_version` preflight checks added to SSH and Ansible examples.**
- SSH example (EN/KO): added `jq -e '.schema_version == 1'` guard that exits with an error message before any `field()` calls.
- Ansible example (EN/KO): added an `ansible.builtin.assert` task that validates `artifact.schema_version == 1` immediately after parsing, before any artifact fields are consumed.

sehkone:
**[Reviewer Round 4]**

I re-checked the Round 3 item against the updated diff and it looks resolved.

- The SSH example now validates `schema_version` before reading any fields via `jq`, exactly as requested: `docs/en/remote-bootstrap.md:108` and `docs/ko/remote-bootstrap.md:108`.
- The Ansible example now asserts `artifact.schema_version == 1` immediately after parsing and before consuming artifact fields: `docs/en/remote-bootstrap.md:200` and `docs/ko/remote-bootstrap.md:198`.
- I also did another pass over the rest of the patch (artifact field addition + rustdoc in `src/commands/service/remote_bootstrap.rs:15`, summary output wording in `src/i18n/en.rs:190` / `src/i18n/ko.rs:190`, doc cross-links, README, changelog, and the added test assertions) and did not find any new issues that I think should block merge.

APPROVED

PR #493, opened 2026-04-12
Add response wrapping support to OpenBao client (#488)

### LINKED ISSUE #488: Add response wrapping and unwrap support to OpenBao client
Part of #480 (sub-issue 2 of 6).

## Goal

Add library-level support for OpenBao's standard response wrapping API to `src/openbao.rs`. No callers yet — this is pure infrastructure for sub-issues 3 and 4.

## Scope

- Add a way to pass `X-Vault-Wrap-TTL` header on outbound requests.
- Add a way to parse wrapped responses (`wrap_info` envelope with `token`, `ttl`, `creation_time`, `creation_path`).
- Add a `sys/wrapping/unwrap` call that consumes a wrap token and returns the original response.
- Use the standard wrapping API per Q9 decision — no custom cubbyhole logic.
- Unit tests covering:
  - Wrap header is sent when `wrap_ttl` is provided
  - Wrapped response envelope is parsed correctly
  - Unwrap call sends the correct request

## Design notes

The wrapping support should be generic enough that any existing client method (not just `create_secret_id`) could opt into wrapping by passing a `wrap_ttl`. Whether this is a per-method parameter or a request builder pattern is an implementation choice.

## Dependencies

- None. Can run in parallel with sub-issue 1 (#487).

## Files likely touched

- `src/openbao.rs`

sehkone:
## Summary

- Adds `WrapInfo` struct to parse the OpenBao `wrap_info` response envelope (`token`, `ttl`, `creation_time`, `creation_path`).
- Adds optional `wrap_ttl` parameter to the shared request helpers (`get_json`, `post_json`, `put_json`, `send_authed`, `send_authed_json`) so any existing client method can opt into response wrapping through the common request path.
- Adds `post_json_wrapped` convenience method that delegates to `post_json` with a wrap TTL and extracts the `wrap_info` envelope.
- Adds `unwrap_secret` method that consumes a wrap token via `sys/wrapping/unwrap` and deserializes the original response.

Closes #488

## Test plan

- [x] `post_json_wrapped` sends the `X-Vault-Wrap-TTL` header via the common `post_json` path
- [x] Wrapped response envelope (`wrap_info`) is parsed correctly into `WrapInfo`
- [x] `unwrap_secret` sends a POST to `sys/wrapping/unwrap` with the wrap token as `X-Vault-Token`
- [x] `get_json` with `wrap_ttl` sends the wrap header (unit test)
- [x] `put_json` with `wrap_ttl` sends the wrap header (unit test)
- [x] All existing OpenBao client tests continue to pass

### DIFF
--- a/src/openbao.rs
+++ b/src/openbao.rs
@@ -6,6 +6,7 @@ use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
 const VAULT_TOKEN_HEADER: &str = "X-Vault-Token";
+const VAULT_WRAP_TTL_HEADER: &str = "X-Vault-Wrap-TTL";
 const ROOT_POLICY: &str = "root";
 
 #[derive(Debug, Clone)]
@@ -100,6 +101,21 @@ struct AppRoleAuth {
     client_token: String,
 }
 
+/// Response-wrapping metadata returned by `OpenBao` when the
+/// `X-Vault-Wrap-TTL` header is set on a request.
+#[derive(Debug, Deserialize)]
+pub struct WrapInfo {
+    pub token: String,
+    pub ttl: u64,
+    pub creation_time: String,
+    pub creation_path: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct WrappedResponse {
+    wrap_info: WrapInfo,
+}
+
 #[derive(Debug, Deserialize)]
 pub struct RekeyInitResponse {
     pub nonce: String,
@@ -184,7 +200,7 @@ impl OpenBaoClient {
     /// # Errors
     /// Returns an error if the init status endpoint cannot be queried.
     pub async fn is_initialized(&self) -> Result<bool> {
-        let status: InitStatus = self.get_json("sys/init", false).await?;
+        let status: InitStatus = self.get_json("sys/init", false, None).await?;
         Ok(status.initialized)
     }
 
@@ -217,7 +233,7 @@ impl OpenBaoClient {
     /// # Errors
     /// Returns an error if the seal status endpoint cannot be queried.
     pub async fn seal_status(&self) -> Result<SealStatus> {
-        self.get_json("sys/seal-status", false).await
+        self.get_json("sys/seal-status", false, None).await
     }
 
     /// Submits an unseal key to `OpenBao`.
@@ -257,6 +273,7 @@ impl OpenBaoClient {
                 secret_shares: shares,
                 secret_threshold: threshold,
             },
+            None,
         )
         .await
     }
@@ -272,7 +289,7 @@ impl OpenBaoClient {
             key: &'a str,
         }
 
-        self.put_json("sys/rekey/update", &RekeyUpdateRequest { nonce, key })
+        self.put_json("sys/rekey/update", &RekeyUpdateRequest { nonce, key }, None)
             .await
     }
 
@@ -296,6 +313,7 @@ impl OpenBaoClient {
                     renewable: false,
                     no_parent: true,
                 },
+                None,
             )
             .await?;
         Ok(response.auth.client_token)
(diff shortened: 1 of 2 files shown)


sehkone:
**[Reviewer Round 1]**
Thanks for adding the low-level unwrap path and the new tests. I do not think this is ready yet, because the wrapping support is still narrower than the issue/design notes asked for.

The blocking concern is `src/openbao.rs:590`. Wrapping is implemented as a separate `post_json_wrapped(...)` entry point that only works for authenticated JSON `POST`s. Issue #488 explicitly called for wrapping support generic enough that any existing client method could opt in by passing a `wrap_ttl`, but this implementation cannot be reused by methods built on the shared helpers like `get_json` (`src/openbao.rs:696`), `post_json` (`src/openbao.rs:722`), `post_action` (`src/openbao.rs:731`), or `put_json` (`src/openbao.rs:744`) without adding more ad hoc helper methods. In other words, the client still does not have a general “send `X-Vault-Wrap-TTL` on outbound requests” capability; it has one new wrapped-POST helper.

The tests in `tests/openbao_client.rs:448` and `tests/openbao_client.rs:478` reflect the same limitation: they prove the new helper works, but they do not prove that wrapping is available through the common request path that the rest of the client uses. I would expect this to be pushed down into the shared request-building layer (for example, an optional `wrap_ttl` on the common helpers, or a small request-builder abstraction), so existing methods can opt into wrapping without duplicating verb-specific APIs.

`unwrap_secret` looks fine to me, but I would like to see the wrap-header support refactored into the general request path before this merges.

NOT_APPROVED

sehkone:
**[Author Round 1]**

Agreed — wrapping was bolted on as a one-off helper instead of being part of the shared request layer.

**What changed:**

- Added an optional `wrap_ttl: Option<&str>` parameter to the shared request helpers: `send_authed`, `send_authed_json`, `get_json`, `post_json`, and `put_json`. When `Some`, the `X-Vault-Wrap-TTL` header is attached through the same code path every verb uses.
- `post_json_wrapped` is now a thin convenience wrapper that delegates to `post_json(path, body, Some(wrap_ttl))` and extracts the `wrap_info` envelope — no ad-hoc request building.
- Helpers that never need wrapping (`post_action`, `delete_action`, `resource_exists`) pass `None` internally, so their callers are unchanged.
- Added two new unit tests inside `src/openbao.rs` (`get_json_sends_wrap_ttl_header`, `put_json_sends_wrap_ttl_header`) that exercise wrapping through `get_json` and `put_json` directly, proving the header flows through the common path for GET and PUT — not just POST.
- Renamed the integration test to `post_json_wrapped_sends_wrap_ttl_via_common_path` to make it clear it now exercises the shared path.

sehkone:
**[Reviewer Round 1]**

Looks good to me.

- The main design choice makes sense: threading `wrap_ttl` through the shared request helpers in `src/openbao.rs:653`, `src/openbao.rs:666`, `src/openbao.rs:707`, `src/openbao.rs:742`, and `src/openbao.rs:772` gives future client methods a clean place to opt into wrapping without baking this into a one-off `create_secret_id` path.
- `WrapInfo` / `WrappedResponse` in `src/openbao.rs:104` and the `post_json_wrapped` convenience in `src/openbao.rs:587` map directly to OpenBao's standard `wrap_info` envelope, and `unwrap_secret` in `src/openbao.rs:612` uses the standard `sys/wrapping/unwrap` flow with the wrap token in `X-Vault-Token`, which is the right API shape for the issue.
- The tests cover the behaviors the issue asked for, and they are meaningful: the unit tests in `src/openbao.rs:897` exercise header injection for non-POST paths, while the integration tests in `tests/openbao_client.rs:448`, `tests/openbao_client.rs:478`, and `tests/openbao_client.rs:508` verify the POST wrapper path, envelope parsing, and unwrap request shape.
- One small follow-up to keep in mind: `login_approle` still bypasses the shared helper path in `src/openbao.rs:503`, so if wrapping is ever needed there it will need the same treatment. I do not see that as blocking for `#488`, though, because the infrastructure added here is enough for the stated follow-on work.

APPROVED

Issue #491, opened 2026-04-12
Add service update subcommand for secret_id policy changes


sehkone:
Part of #480 (sub-issue 5 of 6).

## Goal

Add a `bootroot service update` subcommand that modifies the per-service secret_id policy in `state.json` without recreating the AppRole or re-issuing a `secret_id`.

## Background

Per the Q6 design decision in #480, policy changes go through a dedicated `service update` command rather than re-running `service add` with different flags. Currently `ServiceCommand` in `src/cli/args.rs` only has `Add` and `Info` variants — there is no `service remove` or `service update`. When `service add` is re-run with policy-only differences, `is_idempotent_remote_rerun()` treats it as a mismatch and the error message (added in #489) directs the operator to `service update`.

### CLI

Add `ServiceCommand::Update` variant to `src/cli/args.rs`:
```
bootroot service update --service-name <name> [--secret-id-ttl <ttl>] [--secret-id-wrap-ttl <ttl>] [--no-wrap]
```

**Note:** Use `--service-name` flag style to match existing `service info --service-name` convention, not positional argument.

`--secret-id-num-uses` is intentionally excluded — see #497.

### Behavior

- Validate that the service exists in `state.json`; error if not found.
- Update only the specified `ServiceRoleEntry` fields. Omitted flags leave the existing value unchanged.
- Save the updated state.
- Print a summary of changed fields.
- The updated policy is always applied on the next `rotate approle-secret-id`. It is also applied on an idempotent `service add` rerun **only when wrapping is enabled** (which issues fresh wrapped material per #489). When wrapping is disabled (`--no-wrap`), idempotent rerun only regenerates the artifact without re-issuing a `secret_id`, so policy changes to `secret_id_ttl` are not applied until the next `rotate`. (Note: idempotent rerun requires all fields — including policy — to match. After `service update` changes the policy, a `service add` rerun with the old flags will no longer be idempotent.)
- **Does not** re-issue a `secret_id` or touch OpenBao. This is a state-only operation.
- **Clear/reset semantics.** Provide a way to revert overrides back to default behavior. For example, `--secret-id-ttl inherit` (or a `--reset-secret-id-ttl` flag) to clear a previously set TTL override and return to role-level default inheritance. Similarly, `--no-wrap` disables wrapping, but there must be a way to re-enable default wrapping after it was disabled (e.g., `--wrap` or `--secret-id-wrap-ttl 30m`). The encoding must be consistent with the `None` semantics defined in #489.

[...]

Issue #490, opened 2026-04-12
Embed wrap token in RemoteBootstrapArtifact and add unwrap path to bootroot-remote


sehkone:
Part of #480 (sub-issue 4 of 6).

## Goal

Extend the remote bootstrap artifact with wrap token fields and restructure `bootroot-remote bootstrap` to unwrap before login.

### Artifact schema

Extend `RemoteBootstrapArtifact` in `src/commands/service/remote_bootstrap.rs`:
- Add `wrap_token: Option<String>` and `wrap_expires_at: Option<String>` fields
- `schema_version` stays at `1` — these are additive optional fields with `skip_serializing_if`, which the schema contract explicitly excludes from version bumps (existing parsers ignore unknown keys)

### bootroot-remote unwrap path

**Bootstrap sequence change required.** Today `run_bootstrap()` in `src/bin/bootroot-remote/bootstrap.rs` reads `secret_id` from a file at startup and immediately uses it to log in. With wrapping enabled by default, the `secret_id` file does not exist at bootstrap time. The sequence must change to:

1. Detect wrap fields in the artifact (loaded via `--artifact <path>`)
2. Call `sys/wrapping/unwrap` to obtain `secret_id`
3. Write `secret_id` to the expected file path
4. Proceed with the existing login flow

Keep the raw `secret_id` path working for backward compatibility when wrap fields are absent (operator passed `--no-wrap`).

### Error handling (Q8 decision)

Distinguish between two unwrap failure cases with specific messages:
- **Token expired:** Display recovery instructions with the exact `bootroot rotate approle-secret-id` command to re-run
- **Token already unwrapped:** Flag as a potential security incident — the secret_id may be compromised

Both cases must have dedicated test coverage (unit or integration) to verify that the correct error variant is surfaced and the recovery message includes the expected command.

### Idempotent rerun with wrapping

Per #489's decision: when wrapping is enabled, idempotent rerun issues a fresh `secret_id` with wrapping and embeds the new `wrap_token` in the regenerated artifact. This means `run_service_add_remote_idempotent()` (or the artifact generation path it calls) must now call `create_secret_id()` with the stored policy and wrapping, rather than only regenerating the artifact. When wrapping is disabled (`--no-wrap`), the current behavior is preserved — artifact-only regeneration without OpenBao calls. This sub-issue must wire the OpenBao call into the idempotent rerun path for the wrapped case.

### Artifact sensitivity

[...]

PR #495, opened 2026-04-13
Add secret_id policy flags to service add and rotate (#489)

### LINKED ISSUE #489: Add secret_id policy flags to service add, persist in state, and apply on rotate
Part of #480 (sub-issue 3 of 6).

## Goal

Wire the per-issuance secret_id policy through the full `service add` and `rotate approle-secret-id` flows: CLI flags → resolution → OpenBao calls → state persistence → rotation.

### CLI flags

Add to `ServiceAddArgs` in `src/cli/args.rs`:
- `--secret-id-ttl` (optional, inherits role-level default when omitted)
- `--secret-id-num-uses` (default: 1, per Q2 decision)
- `--secret-id-wrap-ttl` (default: "30m", per Q4 decision)
- `--no-wrap` (disables response wrapping, per Q3 opt-out decision)

### State persistence

Extend `ServiceRoleEntry` in `src/state.rs` with flat `Option` fields:
```rust
#[serde(default)]
pub(crate) secret_id_ttl: Option<String>,
#[serde(default)]
pub(crate) secret_id_num_uses: Option<u32>,
#[serde(default)]
pub(crate) secret_id_wrap_ttl: Option<String>,
```

`#[serde(default)]` ensures backward compatibility with existing `state.json` files.

**`None` semantics — decided.** Per #480 Q7 (flat `Option` fields, no nested struct, no extra booleans), the encoding is:

- `None` = use the default (30m wrap TTL, num_uses=1, inherit role-level secret_id_ttl)
- `Some("0")` for `secret_id_wrap_ttl` = wrapping explicitly disabled (`--no-wrap`)
- `Some(value)` = explicit override

This keeps the three-field shape from Q7 unchanged while giving `--no-wrap` a concrete persisted representation. All downstream consumers (#490 artifact generation, #491 `service update`, `rotate`) must use this encoding.

### service add flow

Thread values through `resolve.rs` → `approle.rs` → `service.rs`:
- Pass `SecretIdOptions` (from #487) to `create_secret_id()`
- Pass `wrap_ttl` (from #488) to the wrapping layer
- Save policy fields in `ServiceRoleEntry` when writing state

### Idempotent rerun

Include the new policy fields in `is_idempotent_remote_rerun()` comparison. When `service add` is re-run with policy-only differences, treat it as a mismatch and display a general error message (e.g., "to change the secret_id policy for an existing service, use a policy update command"). **Note:** `service update` (#491) does not exist yet when this PR lands, so the error message must not reference it by name. Once #491 merges, it should update this message to point to the concrete `service update` subcommand.

**Wrapped idempotent rerun semantics — decided (contract only; implementation in #490).** When wrapping is enabled, idempotent rerun must issue a fresh `secret_id` with wrapping. The previous `wrap_token` is expired or consumed, so the artifact would be useless without fresh material. Emitting a raw `secret_id` would contradict the Q3 opt-out security posture. When wrapping is disabled (`--no-wrap`), the current behavior is preserved — artifact-only regeneration without OpenBao calls. **This PR does not implement the wrapped rerun path** — the artifact schema (`wrap_token`, `wrap_expires_at`) does not exist until #490. This PR only establishes the contract and ensures the `create_secret_id()` call with stored policy is available for #490 to wire into the rerun path.

### rotate approle-secret-id

Update `src/commands/rotate/approle.rs` to read the stored policy from `entry.approle` and pass it to `create_secret_id()` with options. No per-issuance override flags on `rotate`.

## Dependencies

- #487 (sub-issue 1: `SecretIdOptions` and `create_secret_id` signature)
- #488 (sub-issue 2: wrapping support)

## Files likely touched

- `src/cli/args.rs`
- `src/state.rs`
- `src/commands/service/resolve.rs`
- `src/commands/service/approle.rs`
- `src/commands/service.rs`
- `src/commands/rotate/approle.rs`
- Test files: `tests/bootroot_service.rs`, `tests/bootroot_rotate.rs`, `tests/bootroot_cli.rs`

sehkone:
## Summary

- Add `--secret-id-ttl`, `--secret-id-num-uses` (default 1), `--secret-id-wrap-ttl` (default 30m), and `--no-wrap` flags to `service add`
- `--no-wrap` and `--secret-id-wrap-ttl` are mutually exclusive (Clap `conflicts_with`)
- Persist policy as flat `Option` fields on `ServiceRoleEntry` in `state.json` with `#[serde(default)]` for backward compatibility
- `None` encoding: `None` = use the default (num_uses=1, wrap_ttl=30m, inherit role-level TTL); `Some("0")` for wrap_ttl = wrapping disabled; `Some(value)` = explicit override
- Thread `SecretIdOptions` through the resolution and AppRole creation flow so `create_secret_id()` receives per-issuance policy
- Wire `wrap_ttl` through `ensure_service_approle` and `rotate approle-secret-id` to the wrapping layer: when wrapping is enabled, `create_secret_id_wrapped()` sends the `X-Vault-Wrap-TTL` header, immediately unwraps to obtain the raw `secret_id`
- Read stored policy from state during `rotate approle-secret-id` and pass to `create_secret_id()` with wrapping; when `num_uses > 0`, add 1 (with overflow guard) for the internal verify-login so the caller's requested uses remain intact
- Include policy fields in idempotent-rerun comparison; surface a localized (i18n) error when only policy differs on an existing service
- Update English and Korean CLI docs and CHANGELOG

Closes #489

## Not addressed

- **Wrapped idempotent rerun path**: the artifact schema (`wrap_token`, `wrap_expires_at`) does not exist until #490, so the wrapped rerun flow is deferred to that PR. This PR establishes the contract and ensures `create_secret_id()` with stored policy is available for #490.
- **`service update` subcommand**: does not exist yet (#491). The policy-mismatch error message uses a generic phrase ("use a policy update command") and will be updated by #491 to reference the concrete subcommand.

## Test plan

- [x] `cargo test` passes (unit tests for CLI parsing defaults/overrides, state round-trip, `effective_wrap_ttl`, `build_secret_id_options`, `policy_fields_match`, `is_idempotent_remote_rerun`, `is_policy_only_mismatch`, `create_secret_id_wrapped`)
- [x] Existing state.json files without the new fields deserialize correctly (backward-compat test)
- [x] `--no-wrap` resolves `secret_id_wrap_ttl` to `Some("0")` in state
- [x] `--no-wrap` and `--secret-id-wrap-ttl` reject each other at parse time (Clap conflict test)
- [x] E2E service add and rotate tests exercise the wrapping path (`X-Vault-Wrap-TTL` header + unwrap)
- [x] `rotate approle-secret-id` reads stored policy from `ServiceRoleEntry` and applies defaults (num_uses=1, wrap_ttl=30m) when state fields are `None` — exercised by E2E test with absent policy fields
- [x] Re-running `service add` with different policy on an existing remote-bootstrap service produces the localized policy-mismatch error
- [x] Policy-mismatch error uses i18n Messages catalog (English and Korean)
- [x] Rotation verify-login adds +1 to `num_uses` when bounded (with checked arithmetic to prevent overflow), preserving the caller's requested uses

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,13 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   `--post-renew-on-failure`). Hook settings are persisted in
   `state.json` and forwarded to `bootroot-remote bootstrap` for
   remote-bootstrap delivery mode.
+- Added per-issuance `secret_id` policy flags to `bootroot service add`
+  (`--secret-id-ttl`, `--secret-id-num-uses`, `--secret-id-wrap-ttl`,
+  `--no-wrap`). Policy values are persisted in `state.json` and applied
+  automatically during `rotate approle-secret-id`. Re-running
+  `service add` with different policy values on an existing service
+  produces an error directing the operator to use a policy update
+  command.
 - Added automatic HTTP-01 DNS alias registration on `service add`. The
   validation FQDN is registered as a Docker network alias on
   `bootroot-http01` at runtime, removing the need for a hand-written
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -521,6 +521,21 @@ the equivalent low-level hook settings. These flags are
 also forwarded to `bootroot-remote bootstrap` for the
 `remote-bootstrap` delivery mode.
 
+Per-issuance `secret_id` policy flags:
+
+- `--secret-id-ttl`: TTL for the generated `secret_id` (inherits the
+  role-level default when omitted)
+- `--secret-id-num-uses`: maximum number of times the `secret_id` can be
+  used (default `1`)
+- `--secret-id-wrap-ttl`: response-wrapping TTL for the `secret_id`
+  (default `30m`)
+- `--no-wrap`: disable response wrapping for the `secret_id`
+
+These values are persisted in `state.json` and applied on
+`rotate approle-secret-id`. `--no-wrap` and `--secret-id-wrap-ttl` control
+the same field: `--no-wrap` sets the stored wrap TTL to `0`, disabling
+wrapping entirely.
+
 ### Interactive behavior
 
 - Prompts for missing required inputs (deploy type defaults to `daemon`).
--- a/docs/ko/cli.md
+++ b/docs/ko/cli.md
@@ -504,6 +504,17 @@ bootroot status
 `remote-bootstrap` 전달 모드에서는 이 플래그가
 `bootroot-remote bootstrap`으로 전달됩니다.
 
+발급 시점 `secret_id` 정책 플래그:
+
+- `--secret-id-ttl`: 생성되는 `secret_id`의 TTL (생략 시 역할 수준 기본값 상속)
+- `--secret-id-num-uses`: `secret_id` 최대 사용 횟수 (기본값 `1`)
+- `--secret-id-wrap-ttl`: `secret_id`에 대한 응답 래핑 TTL (기본값 `30m`)
+- `--no-wrap`: `secret_id` 응답 래핑 비활성화
+
+이 값들은 `state.json`에 저장되며 `rotate approle-secret-id` 시 적용됩니다.
+`--no-wrap`과 `--secret-id-wrap-ttl`은 동일 필드를 제어합니다.
+`--no-wrap`은 저장되는 래핑 TTL을 `0`으로 설정하여 래핑을 완전히 비활성화합니다.
+
 ### 대화형 동작
 
 - 누락된 필수 입력을 프롬프트로 받습니다(배포 타입 기본값: `daemon`).
--- a/scripts/impl/run-local-lifecycle.sh
+++ b/scripts/impl/run-local-lifecycle.sh
@@ -398,6 +398,7 @@ run_bootstrap_chain() {
     --cert-path "$CERTS_DIR/${EDGE_SERVICE}.crt" \
     --key-path "$CERTS_DIR/${EDGE_SERVICE}.key" \
     --instance-id "$INSTANCE_ID" \
+    --secret-id-num-uses 0 \
     --auth-mode approle \
     --approle-role-id "$RUNTIME_SERVICE_ADD_ROLE_ID" \
     --approle-secret-id "$RUNTIME_SERVICE_ADD_SECRET_ID" >>"$RUN_LOG" 2>&1
@@ -413,6 +414,7 @@ run_bootstrap_chain() {
     --key-path "$CERTS_DIR/${WEB_SERVICE}.key" \
     --instance-id "$INSTANCE_ID" \
     --container-name "$WEB_SERVICE" \
+    --secret-id-num-uses 0 \
     --auth-mode approle \
     --approle-role-id "$RUNTIME_SERVICE_ADD_ROLE_ID" \
     --approle-secret-id "$RUNTIME_SERVICE_ADD_SECRET_ID" >>"$RUN_LOG" 2>&1
--- a/scripts/impl/run-local-lifecycle.sh
+++ b/scripts/impl/run-local-lifecycle.sh
@@ -427,6 +427,7 @@ run_bootstrap_chain() {
     --cert-path "$REMOTE_CERTS_DIR/${REMOTE_SERVICE}.crt" \
     --key-path "$REMOTE_CERTS_DIR/${REMOTE_SERVICE}.key" \
     --instance-id "$REMOTE_INSTANCE_ID" \
+    --secret-id-num-uses 0 \
     --auth-mode approle \
     --approle-role-id "$RUNTIME_SERVICE_ADD_ROLE_ID" \
     --approle-secret-id "$RUNTIME_SERVICE_ADD_SECRET_ID" >>"$RUN_LOG" 2>&1
--- a/scripts/impl/run-remote-lifecycle.sh
+++ b/scripts/impl/run-remote-lifecycle.sh
@@ -301,6 +301,7 @@ run_bootstrap_chain() {
     --cert-path "$REMOTE_CERTS_DIR/${SERVICE_NAME}.crt" \
     --key-path "$REMOTE_CERTS_DIR/${SERVICE_NAME}.key" \
     --instance-id "$INSTANCE_ID" \
+    --secret-id-num-uses 0 \
     --auth-mode approle \
     --approle-role-id "$RUNTIME_SERVICE_ADD_ROLE_ID" \
     --approle-secret-id "$RUNTIME_SERVICE_ADD_SECRET_ID" >>"$RUN_LOG" 2>&1
@@ -316,6 +317,7 @@ run_bootstrap_chain() {
     --key-path "$REMOTE_CERTS_DIR/${SERVICE_NAME_2}.key" \
     --instance-id "$INSTANCE_ID_2" \
     --container-name "$SERVICE_NAME_2" \
+    --secret-id-num-uses 0 \
     --auth-mode approle \
     --approle-role-id "$RUNTIME_SERVICE_ADD_ROLE_ID" \
     --approle-secret-id "$RUNTIME_SERVICE_ADD_SECRET_ID" >>"$RUN_LOG" 2>&1
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -706,6 +706,22 @@ pub(crate) struct ServiceAddArgs {
     /// Post-renew success hook failure policy (low-level)
     #[arg(long, value_enum)]
     pub(crate) post_renew_on_failure: Option<HookFailurePolicyArg>,
+
+    /// TTL for the generated `secret_id` (inherits role default when omitted)
+    #[arg(long)]
+    pub(crate) secret_id_ttl: Option<String>,
+
+    /// Maximum number of times the `secret_id` can be used [default: 1]
+    #[arg(long)]
+    pub(crate) secret_id_num_uses: Option<u32>,
+
+    /// Response-wrapping TTL for the `secret_id` [default: 30m]
+    #[arg(long)]
+    pub(crate) secret_id_wrap_ttl: Option<String>,
+
+    /// Disable response wrapping for the `secret_id`
+    #[arg(long, conflicts_with = "secret_id_wrap_ttl")]
+    pub(crate) no_wrap: bool,
 }
 
 #[derive(Args, Debug)]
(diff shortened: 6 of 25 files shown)


sehkone:
**[Reviewer Round 1]**

Thanks — I don't think this is ready yet. Two requirements from `#489` are still not wired correctly:

1. The `secret_id_num_uses` / `secret_id_wrap_ttl` defaults and `None` encoding do not match the issue contract. `ServiceAddArgs` sets `secret_id_num_uses` to `0` and `secret_id_wrap_ttl` to `"30m"` by default in `src/cli/args.rs:714` and `src/cli/args.rs:718`, and `resolve_service_add_args()` turns that into `None` for `num_uses` only when the value is `0`, while always persisting `Some("30m")` for wrap TTL in `src/commands/service/resolve.rs:128` and `src/commands/service/resolve.rs:147`. `#489` explicitly requires `num_uses` default `1` and `None` to mean “use the default” for both fields. As written, an omitted `--secret-id-num-uses` falls through to `None`, so both `service add` and `rotate approle-secret-id` send no `num_uses` at all (`src/commands/service.rs:362`, `src/commands/rotate/approle.rs:37`), which reverts to the role/OpenBao default instead of enforcing the requested per-issuance default of `1`. The new tests in `src/cli/args.rs:1188` also lock in the wrong default (`0`), so they would not catch this.

2. The wrap flags are persisted, but they are not applied to the `service add` OpenBao call. `resolve.rs` computes/stores `secret_id_wrap_ttl`, but `run_service_add_apply()` only builds `SecretIdOptions` and calls `approle::ensure_service_approle()` without any wrap-TTL parameter (`src/commands/service.rs:234`). `ensure_service_approle()` then calls `create_secret_id()` and returns the raw `secret_id` (`src/commands/service/approle.rs:17`). Since the wrapping helper from `#488` already exists in `src/openbao.rs:611`, `--secret-id-wrap-ttl` / `--no-wrap` are currently no-ops for the initial `service add` flow, even though `#489` explicitly says to pass `wrap_ttl` through to the wrapping layer. I also do not see a test that would fail if wrapping were silently ignored.

NOT_APPROVED

sehkone:
**[Author Round 1]**

Both items fixed in 34f6a11:

**1. Default values and `None` encoding**

- Changed `secret_id_num_uses` and `secret_id_wrap_ttl` CLI args from hard-coded defaults (`u32`/`String`) to `Option` types so `None` cleanly encodes "use the default" per the #489 contract.
- `build_secret_id_options` and the rotate flow now resolve `None` → `1` for `num_uses` and `None` → `"30m"` for `wrap_ttl` via the new `effective_wrap_ttl()` helper.
- Added `DEFAULT_SECRET_ID_NUM_USES` and `DEFAULT_SECRET_ID_WRAP_TTL` constants in `commands/constants.rs`.
- Tests updated to assert the corrected defaults.

**2. Wire `wrap_ttl` through to OpenBao**

- Added `OpenBaoClient::create_secret_id_wrapped()` — sends `X-Vault-Wrap-TTL`, then immediately unwraps to obtain the raw `secret_id`.
- `ensure_service_approle()` now accepts `wrap_ttl: Option<&str>` and dispatches to the wrapped or non-wrapped path.
- `rotate approle-secret-id` reads the stored `wrap_ttl` from state and uses the same dispatch.
- Added a unit test (`create_secret_id_wrapped_sends_header_and_unwraps`) that verifies the header is sent and the unwrap path works.
- All E2E mocks (service add, rotate, init) updated to handle both wrapped and non-wrapped paths. Service add and rotate tests now exercise the wrapping path by default.

sehkone:
**[Reviewer Round 1]**
This is close, but I see a few issues to address before merge.

- `--no-wrap` and `--secret-id-wrap-ttl` are currently accepted together, and the resolver silently lets `--no-wrap` win by rewriting the stored value to `Some("0")` in `src/commands/service/resolve.rs:129`. Because these flags express opposite policies, I think Clap should reject the combination up front instead of discarding one side. The affected flag definitions are in `src/cli/args.rs:718` and `src/cli/args.rs:722`.
- The new policy-mismatch error is hardcoded English in `src/commands/service.rs:109`, which bypasses the existing `Messages`/i18n flow. That means `--lang ko` users will get this one message in English while the rest of the command is localized. Please move this into the message catalog like the other service-add errors.
- The rotation tests do not actually exercise the new default-wrapping behavior the PR says it covers. The helper fixtures in `tests/bootroot_rotate.rs:955`, `tests/bootroot_rotate.rs:988`, and `tests/bootroot_rotate.rs:1007` now pin `"secret_id_wrap_ttl": "0"`, so those paths stay on the unwrapped branch. I do not see a test that proves `rotate approle-secret-id` applies the stored-default semantics (`None` => wrap with `30m`, `num_uses` => `1`) when the policy fields are absent from `state.json`, which is one of the important compatibility cases in the issue.

Because of those points, I don’t think this is ready to merge yet.

sehkone:
**[Author Round 1]**

All three items addressed in 6c98b73:

1. **Clap mutual exclusion for `--no-wrap` / `--secret-id-wrap-ttl`** — Added `conflicts_with = "secret_id_wrap_ttl"` on the `--no-wrap` flag so Clap rejects the combination at parse time. Added `test_cli_rejects_no_wrap_with_secret_id_wrap_ttl` to lock this in. The resolver logic in `resolve.rs` is unchanged since with the Clap guard the `if args.no_wrap` branch can only fire when `--secret-id-wrap-ttl` is absent.

2. **Policy-mismatch error moved to i18n** — Added `error_service_policy_mismatch` to the `Strings` struct, English (`en.rs`), and Korean (`ko.rs`) catalogs, with a `Messages::error_service_policy_mismatch()` accessor. The `service.rs` call site now uses `messages.error_service_policy_mismatch()` instead of the hardcoded English string.

3. **Rotation test for default-wrapping semantics** — Added `test_rotate_approle_secret_id_applies_default_wrapping_when_policy_absent` which prepares state.json with no `secret_id_wrap_ttl`/`secret_id_num_uses`/`secret_id_ttl` fields and uses a wrapped mock (`stub_openbao_for_wrapped_rotation`) that asserts the `X-Vault-Wrap-TTL: 30m` header is sent and the unwrap path is exercised. The mock uses `expect(1)` on both the wrap and unwrap endpoints to prove both are called exactly once.

sehkone:
**[Reviewer Round 2]**

Thanks — the wrap-through, Clap conflict, i18n message, and the new wrapped-rotation coverage all look good now. I still see one remaining blocker from the earlier review, though:

- The `secret_id_num_uses` default is still wired to `0`/unlimited instead of the issue’s required per-issuance default of `1`. `DEFAULT_SECRET_ID_NUM_USES` is still `0` in `src/commands/constants.rs:2`, and both the `service add` path (`src/commands/service.rs:363`) and `rotate approle-secret-id` path (`src/commands/rotate/approle.rs:40`) fall back to that value whenever the persisted field is `None`. That means an omitted `--secret-id-num-uses` still sends `num_uses: 0` to OpenBao rather than enforcing the decided default of `1`. The help/tests are also still aligned with the wrong behavior: `src/cli/args.rs:714` documents `0 (unlimited)`, and `src/commands/service.rs:663` asserts `Some(0)`. This is still out of contract with `#489`, and it affects both the initial `service add` issuance and later rotations.

NOT_APPROVED

sehkone:
**[Author Round 2]**

Fixed in d461450:

- Changed `DEFAULT_SECRET_ID_NUM_USES` from `0` to `1` in `src/commands/constants.rs` to match the #489 contract (per-issuance default of 1).
- Updated `--secret-id-num-uses` help text in `src/cli/args.rs` from `[default: 0 (unlimited)]` to `[default: 1]`.
- Updated the `build_secret_id_options_none_ttl` test assertion from `Some(0)` to `Some(1)`.
- Updated the PR description to reflect `num_uses=1` throughout.

Both the `service add` and `rotate approle-secret-id` paths use the same constant, so both now correctly default to `1` when the persisted field is `None`.

sehkone:
**[Reviewer Round 3]**

Thanks for the update. I re-checked the Round 2 blocker against the current head, and I still don't think this is ready.

- The previous `secret_id_num_uses` default issue is still open on the current branch. The latest follow-up commit (`37da1c3`) reverted the `d461450` change and put the default back to `0`/unlimited in code and tests: `src/commands/constants.rs:2`, `src/cli/args.rs:714`, `src/commands/service.rs:363`, `src/commands/rotate/approle.rs:43`, and `src/commands/service.rs:663`. That is still out of contract with `#489`, which explicitly decided `None => default 1` for per-issuance `secret_id_num_uses`. The docs are now inconsistent with the implementation too: both `docs/en/cli.md:528` and `docs/ko/cli.md:510` still say the default is `1`, while the CLI help and runtime behavior use `0`.
- New finding: the rotation-side `+1` compensation for verify-login can overflow for an explicitly bounded maximum value. `src/commands/rotate/approle.rs:46` does `stored_num_uses + 1` on a `u32` without any guard. If a user persists `secret_id_num_uses = 4294967295`, rotation will panic in debug/test builds and wrap in release builds. If you want to keep this compensation, it needs either input validation that rejects the max value or a checked addition with a proper user-facing error.

Because of those two points, I don't think this is ready to merge yet.

NOT_APPROVED

sehkone:
**[Author Round 3]**

Both items addressed in 1e40752:

1. **`secret_id_num_uses` default restored to `1`** — changed the constant in `src/commands/constants.rs`, updated the CLI help text in `src/cli/args.rs`, and fixed the unit test assertion in `src/commands/service.rs`. The docs (`docs/en/cli.md`, `docs/ko/cli.md`) already said `1` and needed no change. The `init/constants.rs` constant (role-level default, `0` = unlimited) is intentionally unchanged — it governs a different scope.

2. **Overflow guard on `+1` compensation** — `src/commands/rotate/approle.rs` now uses `checked_add(1)` instead of bare `+ 1`. If `stored_num_uses` is `u32::MAX`, rotation returns a user-facing error instead of panicking or wrapping.

Also updated the PR description to reflect the `num_uses=1` default and removed the "Not addressed" bullet that deferred it to #490.

sehkone:
**[Reviewer Round 3]**

I re-checked the Round 2 blocker against the current head.

- The `secret_id_num_uses` fallback now matches the `#489` contract: `DEFAULT_SECRET_ID_NUM_USES` is `1`, `build_secret_id_options()` uses it for `service add`, and `rotate approle-secret-id` uses the same default before applying the verification allowance in `src/commands/constants.rs:2`, `src/commands/service.rs:363`, and `src/commands/rotate/approle.rs:40`.
- The rotation-side overflow case is also fixed: `src/commands/rotate/approle.rs:47` now uses `checked_add(1)` and returns an error instead of overflowing on bounded maximum values.

I do not see any remaining blockers in the current diff.

APPROVED

PR #498, opened 2026-04-13
Remove secret_id_num_uses from service policy (#497)

### LINKED ISSUE #497: Remove secret_id_num_uses from service policy
Follow-up to #489 and PR #495.

## Summary

`bootroot` currently treats service AppRole `secret_id` values as reusable runtime credentials, not as strictly one-shot bootstrap tokens. In that model, exposing `secret_id_num_uses` in the service policy surface is a footgun: local-file services may need repeated OpenBao Agent re-authentication, and remote-bootstrap services may need repeated bootstrap or recovery flows before the next manual rotation.

## Goal

Remove per-service `secret_id_num_uses` from the user-facing service policy model while keeping lower-level OpenBao client support available for future workflows that truly need bounded-use SecretIDs.

## Scope

- Remove `--secret-id-num-uses` from `bootroot service add` (`src/cli/args.rs`).
- Remove `--secret-id-num-uses` from `bootroot service update` if #491 has landed; otherwise update #491 first so it does not introduce the flag.
- Remove `secret_id_num_uses` from `ServiceRoleEntry` (`src/state.rs`). Older `state.json` files that still contain the removed field must continue to deserialize cleanly (serde ignores unknown fields by default).
- Remove `secret_id_num_uses` from `ResolvedServiceAdd` (`src/commands/service/resolve.rs`).
- Remove `DEFAULT_SECRET_ID_NUM_USES` constant (`src/commands/constants.rs`).
- Update the dead-code constant comment in `src/commands/init/constants.rs` — the current comment references "num_uses enforcement" as an upcoming sub-issue of #480, which is no longer accurate for services.
- For service SecretID issuance in `src/commands/service.rs`, stop sourcing `num_uses` from per-service policy or state and use `num_uses = 0` (unlimited) for both `service add` and `rotate approle-secret-id`.
- Remove the `checked_add(1)` overflow-guard and `+1` verification-allowance logic in `src/commands/rotate/approle.rs` — with unlimited uses this block is unnecessary.
- Remove `secret_id_num_uses` from the policy-field comparison used for idempotent-rerun detection (`src/commands/service.rs`, `policy_fields_match`).
- Remove `--secret-id-num-uses 0` from E2E lifecycle scripts (`scripts/impl/run-local-lifecycle.sh`, `scripts/impl/run-remote-lifecycle.sh`).
- Update CHANGELOG — the current entry lists `--secret-id-num-uses` as a newly added flag; add a removal entry or amend the existing one.
- Update CLI docs (`docs/en/cli.md`, `docs/ko/cli.md`) to remove the `--secret-id-num-uses` line.
- Keep generic `SecretIdOptions.num_uses` support in the OpenBao client layer (`src/openbao.rs`) for non-service or future specialized workflows.
- Update unit tests accordingly.

## Rationale

`secret_id_ttl` is still a meaningful operator-facing policy because it expresses credential lifetime. `secret_id_num_uses` is different: internal logins, Agent re-authentication, restarts, and remote recovery flows can consume uses in ways the operator cannot safely predict.

## Non-goals

- Removing `--secret-id-ttl`
- Removing wrapping controls
- Changing generic OpenBao client support for bounded-use SecretIDs

sehkone:
## Summary

- Remove `--secret-id-num-uses` from `bootroot service add` CLI and all supporting types (`ServiceRoleEntry`, `ResolvedServiceAdd`, `DEFAULT_SECRET_ID_NUM_USES`).
- Service SecretIDs are now always issued with `num_uses = 0` (unlimited) in both `service add` and `rotate approle-secret-id`.
- Remove the `checked_add(1)` overflow-guard / verification-allowance logic in rotation, which is no longer needed.
- Remove `secret_id_num_uses` from the idempotent-rerun policy-field comparison.
- Remove `--secret-id-num-uses 0` from E2E lifecycle scripts.
- Update CHANGELOG, CLI docs (en/ko), and unit tests.
- Keep generic `SecretIdOptions.num_uses` in the OpenBao client layer for non-service workflows.
- Add a backward-compatibility test ensuring older `state.json` files containing `secret_id_num_uses` still deserialize cleanly.

Closes #497

## Test plan

- [x] `cargo test` passes (unit tests updated)
- [x] Older `state.json` with `secret_id_num_uses` field deserializes without error (covered by new test)
- [x] `bootroot service add --help` no longer shows `--secret-id-num-uses`
- [x] E2E local lifecycle (`scripts/impl/run-local-lifecycle.sh`) runs without `--secret-id-num-uses`
- [x] E2E remote lifecycle (`scripts/impl/run-remote-lifecycle.sh`) runs without `--secret-id-num-uses`
- [x] `rotate approle-secret-id` issues SecretIDs with unlimited uses
- [x] `scripts/preflight/run-all.sh` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Removed
+
+- Removed `--secret-id-num-uses` from `bootroot service add` and from
+  `rotate approle-secret-id` policy state. Service SecretIDs are now
+  always issued with unlimited uses (`num_uses = 0`). The lower-level
+  OpenBao client still supports bounded-use SecretIDs for non-service
+  workflows.
+
 ### Fixed
 
 - Fixed daemon-mode retries silently dropping CLI overrides (`--email`,
@@ -38,12 +46,11 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   `state.json` and forwarded to `bootroot-remote bootstrap` for
   remote-bootstrap delivery mode.
 - Added per-issuance `secret_id` policy flags to `bootroot service add`
-  (`--secret-id-ttl`, `--secret-id-num-uses`, `--secret-id-wrap-ttl`,
-  `--no-wrap`). Policy values are persisted in `state.json` and applied
-  automatically during `rotate approle-secret-id`. Re-running
-  `service add` with different policy values on an existing service
-  produces an error directing the operator to use a policy update
-  command.
+  (`--secret-id-ttl`, `--secret-id-wrap-ttl`, `--no-wrap`). Policy
+  values are persisted in `state.json` and applied automatically during
+  `rotate approle-secret-id`. Re-running `service add` with different
+  policy values on an existing service produces an error directing the
+  operator to use a policy update command.
 - Added automatic HTTP-01 DNS alias registration on `service add`. The
   validation FQDN is registered as a Docker network alias on
   `bootroot-http01` at runtime, removing the need for a hand-written
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -525,8 +525,6 @@ Per-issuance `secret_id` policy flags:
 
 - `--secret-id-ttl`: TTL for the generated `secret_id` (inherits the
   role-level default when omitted)
-- `--secret-id-num-uses`: maximum number of times the `secret_id` can be
-  used (default `1`)
 - `--secret-id-wrap-ttl`: response-wrapping TTL for the `secret_id`
   (default `30m`)
 - `--no-wrap`: disable response wrapping for the `secret_id`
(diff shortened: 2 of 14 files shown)


sehkone:
**[Reviewer Round 1]**

Looks good to me.

- The behavior change is applied at the right boundary: service issuance now always sends `num_uses = 0` in `src/commands/service.rs:361` and `src/commands/rotate/approle.rs:38`, while generic `SecretIdOptions.num_uses` remains available in `src/openbao.rs:90` for non-service flows.
- The user-facing and persisted service policy surfaces are cleaned up consistently: the CLI flag is gone in `src/cli/args.rs:710`, the resolved add model drops the field in `src/commands/service/resolve.rs:28`, and the service state model no longer stores it in `src/state.rs:90`.
- Idempotent rerun detection now compares only the remaining operator-facing policy fields in `src/commands/service.rs:482`, which matches the new model and avoids treating a removed field as meaningful drift.
- The compatibility coverage in `src/state.rs:282` is the right regression test for old serialized data that still contains `secret_id_num_uses`.

I did not find any blocking issues.

PR #499, opened 2026-04-13
Embed wrap token in artifact and add unwrap path (#490)

sehkone:
## Summary

- Adds `wrap_token` and `wrap_expires_at` fields to `RemoteBootstrapArtifact`, making the bootstrap artifact carry the wrapped secret_id instead of requiring a separate file transfer.
- Introduces `--artifact <path>` flag for `bootroot-remote bootstrap` that loads the JSON artifact directly, avoiding sensitive values in CLI args and `ps` output. Per-field flags remain for backward compatibility.
- Implements the unwrap-before-login sequence in `bootroot-remote`: detect wrap fields, call `sys/wrapping/unwrap`, write `secret_id` to disk with `0600` permissions, then proceed with AppRole login.
- Distinguishes two unwrap failure cases: expired tokens show recovery instructions (`bootroot rotate approle-secret-id`), while already-unwrapped tokens flag a potential security incident.
- Adds `create_secret_id_wrap_only` to `OpenBaoClient` for forwarding wrap tokens without unwrapping server-side.
- Wires wrapped secret_id issuance into the idempotent rerun path (`run_service_add_remote_idempotent`) so re-running `service add` with matching parameters regenerates both the wrap token and artifact.
- Local-file delivery mode now skips wrapping entirely (wrapping only applies to remote-bootstrap).
- Skips KV secret_id write when wrapping is active (the remote node obtains it via unwrap instead).
- Fixes E2E lifecycle scripts (`run-remote-lifecycle.sh`, `run-local-lifecycle.sh`, `run-ca-key-rotation-recovery.sh`) to regenerate bootstrap artifacts before each post-rotation re-bootstrap, since wrap tokens are single-use and the old artifact's token is consumed during the initial bootstrap.
- Removes `--secret-id-num-uses` from the service-add CLI. Service SecretIDs always use unlimited uses (`num_uses=0`) because local-file services need repeated OpenBao Agent re-authentication and remote-bootstrap services may need repeated recovery flows. The lower-level `SecretIdOptions.num_uses` is retained for non-service workflows.

Part of #490

## Test plan

- [ ] `test_remote_bootstrap_wrapped_artifact_path` — full E2E: service add with wrapping -> artifact contains `wrap_token` -> remote bootstrap via `--artifact` -> secret_id unwrapped and written
- [ ] `test_remote_bootstrap_wrap_token_expired` — expired wrap token produces recovery message with `bootroot rotate approle-secret-id` command
- [ ] `test_remote_bootstrap_wrap_token_already_unwrapped` — already-unwrapped token produces `SECURITY WARNING` message
- [ ] Existing `test_remote_bootstrap_happy_path` still passes (backward-compatible `--no-wrap` per-flag invocation)
- [ ] `test_e2e_same_host_local_file_lifecycle` passes (local-file delivery skips wrapping)
- [ ] Unit tests for `is_wrap_token_expired`, `build_artifact` with/without wrap info, `render_remote_run_command` with wrapped artifact
- [ ] `secret_id_file_check_skipped_when_wrap_token_present` — validation allows missing secret_id file when wrap token is present
- [ ] `cargo clippy` passes with no warnings
- [ ] `cargo test` passes
- [ ] Docker E2E matrix (remote-hosts, remote-no-hosts, local-hosts, local-no-hosts) passes
- [ ] E2E extended (ca-key-rotation-recovery) passes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -345,7 +345,9 @@ jobs:
         uses: Swatinem/rust-cache@v2
 
       - name: Build bootroot binaries
-        run: cargo build --bin bootroot --bin bootroot-remote --bin bootroot-agent
+        run: |
+          cargo clean -p bootroot 2>/dev/null || true
+          cargo build --bin bootroot --bin bootroot-remote --bin bootroot-agent
 
       - name: Run E2E scenario (lifecycle)
         if: matrix.scenario.label != 'rotation'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -345,18 +345,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
 
       - name: Build bootroot binaries
-        env:
-          CARGO_INCREMENTAL: '0'
         run: |
-          cargo clean -p bootroot || true
-          rm -f target/debug/bootroot target/debug/bootroot-remote target/debug/bootroot-agent
-          rm -rf target/debug/.fingerprint/bootroot-* target/debug/deps/bootroot-* target/debug/deps/libbootroot-* target/debug/incremental/bootroot-*
           cargo build --bin bootroot --bin bootroot-remote --bin bootroot-agent
-          if ! target/debug/bootroot service add --help 2>&1 | grep -q 'secret-id-num-uses'; then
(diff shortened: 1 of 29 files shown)


sehkone:
This went wrong.

PR #500, opened 2026-04-13
Embed wrap token in artifact and add unwrap path (#490)

### LINKED ISSUE #490: Embed wrap token in RemoteBootstrapArtifact and add unwrap path to bootroot-remote

sehkone:
## Summary

- Add `wrap_token` and `wrap_expires_at` optional fields to `RemoteBootstrapArtifact` (additive, `schema_version` stays at `1`)
- Introduce `--artifact <path>` flag to `bootroot-remote bootstrap` so the artifact JSON is read directly instead of passing every field as a CLI flag — avoids exposing `wrap_token` in shell history / `ps` output
- Artifact values take precedence over per-field CLI flags when `--artifact` is provided, including post-renew hook fields; CLI flags serve as fallbacks only when no artifact is given (an artifact with an explicitly empty `post_renew_hooks: []` array is authoritative and does not fall back to CLI hooks)
- Validate `schema_version` when loading an artifact — reject unsupported future versions with a clear upgrade message
- When wrap fields are present, `bootroot-remote` calls `sys/wrapping/unwrap` to obtain `secret_id`, writes it to the expected file path, then proceeds with the existing login flow; raw `secret_id` path still works when wrap fields are absent
- Classify unwrap failures: expired token shows recovery instructions (`bootroot rotate approle-secret-id`), already-unwrapped token is flagged as a security incident; non-token errors (TLS, 500, network) bubble through without misclassification
- Wire `create_secret_id_wrap_only` into the idempotent rerun path so a fresh wrapped `secret_id` is issued when wrapping is enabled
- `render_remote_run_command` emits a `<REMOTE_ARTIFACT_PATH>` placeholder instead of the control-node artifact path, so the printed command template is valid for the remote host (the operator replaces the placeholder after copying the artifact); `render_remote_run_command_legacy` is kept as a test-only utility for per-flag format coverage
- Handoff instructions branch on wrapping state: wrapped default tells operators to copy `bootstrap.json` and `role_id`; `--no-wrap` tells operators to copy `bootstrap.json`, `role_id`, and `secret_id` — with a dedicated test for the non-wrapped summary
- Per-flag E2E test (`test_two_node_remote_bootstrap_happy_path`) now uses `--no-wrap` to properly exercise the non-wrapped flow, with dedicated OpenBao mocks for non-wrapped secret-id creation
- Update CLI docs (EN/KO) and remote-bootstrap guide: `--artifact` with wrapping is the recommended default; Options 2–4 are marked as legacy `--no-wrap` paths; hygiene section updated to reflect that response wrapping is now available

Closes #490

## Test plan

- [x] `cargo test` unit tests pass, including new `classify_unwrap_error_*` tests and `validate_bootstrap_args_skips_secret_id_file_check_when_wrap_token_present`
- [x] `resolve_bootstrap_args_artifact_overrides_cli_flags` test: verifies artifact values take precedence over CLI flags
- [x] `resolve_bootstrap_args_artifact_hooks_override_cli_hooks` test: verifies artifact hooks take precedence over CLI hook flags
- [x] `resolve_bootstrap_args_artifact_empty_hooks_override_cli_hooks` test: verifies empty artifact hooks array suppresses CLI hook flags
- [x] `resolve_bootstrap_args_rejects_unsupported_schema_version` test: verifies unsupported schema version is rejected
- [x] `classify_unwrap_error_returns_none_for_non_token_errors` test: verifies non-token errors (TLS, 500, network) are not misclassified
- [x] `is_openbao_token_rejection_matches_400` / `is_openbao_token_rejection_ignores_500` tests: verify error shape gating
- [x] `build_artifact_with_wrap_info` and `render_remote_run_command_uses_artifact_flag` tests verify artifact schema and command rendering
- [x] `render_remote_run_command_uses_artifact_flag_without_wrap` test: verifies non-wrapped artifacts also generate `--artifact` command
- [x] `test_remote_bootstrap_artifact_invocation` E2E test: artifact contains `wrap_token`/`wrap_expires_at`, remote bootstrap succeeds via `--artifact`, `secret_id` is written after unwrap
- [x] `test_two_node_remote_bootstrap_happy_path` E2E test: uses `--no-wrap` with dedicated non-wrapped mocks, proves per-flag backward-compatible bootstrap works end to end
- [x] `test_remote_bootstrap_expired_wrap_token` E2E test: expired token produces error mentioning "expired" and recovery command
- [x] `test_remote_bootstrap_already_unwrapped_token` E2E test: already-consumed token produces "SECURITY INCIDENT" error with recovery command
- [x] Idempotent rerun with wrapping enabled issues fresh wrapped `secret_id` in regenerated artifact
- [x] `test_app_add_remote_bootstrap_no_wrap_handoff_includes_secret_id` test: verifies `--no-wrap` handoff summary tells operator to copy `secret_id` along with `bootstrap.json` and `role_id`
- [x] `scripts/preflight/run-all.sh` passes


sehkone:
**[Reviewer Round 1]**

A few things look off to me before this lands:

- `--artifact` precedence is reversed from the issue. In `src/bin/bootroot-remote/main.rs:339`, `resolve_bootstrap_args()` lets CLI flags win for the required fields (`args.openbao_url.or(artifact...)`, etc.), and the help text on `BootstrapArgs::artifact` says the same thing. The issue explicitly says artifact values should take precedence when `--artifact` is provided. As written, a mixed invocation can silently override values from the sensitive artifact and point unwrap/login at different files or URLs than the artifact intended.

- Unwrap error classification currently ignores the actual error and rewrites every unwrap failure into either “expired” or “already unwrapped”. `classify_unwrap_error()` in `src/bin/bootroot-remote/bootstrap.rs:240` never inspects `_err`, and `unwrap_and_write_secret_id()` in `src/bin/bootroot-remote/bootstrap.rs:169` feeds all failures through it. That means a TLS failure, `500` from OpenBao, permission error, malformed response, etc. can be surfaced as a fake “SECURITY INCIDENT” or “token expired” message. The issue asked for dedicated handling of those two token states, not for collapsing unrelated failures into them. I think this needs a gate on the actual unwrap error shape/status/text before reclassifying, with all other errors bubbling through normally.

- This introduces the first direct artifact consumer but never validates `schema_version`. `src/commands/service/remote_bootstrap.rs:24` says consumers should reject unsupported future versions, yet `BootstrapArtifact` in `src/bin/bootroot-remote/main.rs:280` does not even deserialize `schema_version`, and `resolve_bootstrap_args()` accepts whatever parses. That leaves `bootroot-remote bootstrap --artifact` blind to future breaking schema bumps instead of failing explicitly.

The new tests cover the expired/used-token cases, but I don’t see coverage that would catch the precedence problem or a non-token unwrap failure being misreported.

sehkone:
**[Author Round 1]**

All three items addressed:

**1. Artifact precedence** — Flipped `resolve_bootstrap_args()` so artifact values take precedence over CLI flags when `--artifact` is provided. CLI flags now serve as fallbacks for fields absent from the artifact. Updated the `--artifact` help text to match. Added `resolve_bootstrap_args_artifact_overrides_cli_flags` test that verifies artifact wins for all required fields.

**2. Unwrap error classification** — `classify_unwrap_error()` now returns `Option<UnwrapError>` and gates on the actual error shape via `is_openbao_token_rejection()`. Only OpenBao 400 API rejections are classified as expired/already-unwrapped; other errors (TLS failures, 500s, network timeouts, parse errors) propagate through with their original context. Added `classify_unwrap_error_returns_none_for_non_token_errors`, `is_openbao_token_rejection_matches_400`, and `is_openbao_token_rejection_ignores_500` tests.

**3. Schema version validation** — `BootstrapArtifact` now deserializes `schema_version` and `resolve_bootstrap_args()` rejects versions above `MAX_SUPPORTED_SCHEMA_VERSION` (currently 1) with a clear "Upgrade bootroot-remote" message. Added `resolve_bootstrap_args_rejects_unsupported_schema_version` test.

sehkone:
**[Reviewer Round 2]**

Thanks, the unwrap-classification and schema-version fixes look good to me. I still see two things to address:

- The artifact-precedence item is only partially fixed. `resolve_bootstrap_args()` now lets the artifact win for the core fields, but the hook block at `src/bin/bootroot-remote/main.rs:442` still gives CLI hook flags precedence over `post_renew_hooks` from the artifact whenever `--post-renew-command` is present. That leaves the same mixed-invocation footgun in place for hook fields, and it also contradicts the new `--artifact` help text. The docs are still out of sync too: `docs/en/cli.md:1026` and `docs/ko/cli.md:988` both say per-field CLI flags can override artifact values. I think the artifact should consistently win for hooks as well, with CLI only filling holes when the artifact omits them, and the docs/tests should match that rule.

- The decided invocation model still is not fully reflected in the generated command or the E2E coverage. `write_remote_bootstrap_artifact_file()` falls back to the long legacy command whenever `wrap_token` is absent (`src/commands/service/remote_bootstrap.rs:214`), but the issue explicitly called for `render_remote_run_command()` to emit `bootroot-remote bootstrap --artifact <path>` instead of the long flag list, keeping per-flag invocation only as a backward-compatible manual path. The current “per-flag” E2E also does not exercise the requested non-wrapped flow: `run_service_add_remote()` never passes `--no-wrap` (`tests/e2e_remote_happy_path.rs:96`), and `copy_remote_bootstrap_materials()` copies an already-unwrapped `secret_id` from the control node (`tests/e2e_remote_happy_path.rs:248`). So we still are not proving that `bootroot service add --no-wrap` + legacy per-flag bootstrap works end to end, nor are we generating the `--artifact` command template consistently.

sehkone:
**[Author Round 2]**

Both items addressed:

**1. Hook precedence** — Flipped the hook resolution block in `resolve_bootstrap_args()` so artifact hooks take precedence over CLI hook flags (matching the rule already applied to core fields). CLI flags now serve as fallbacks only when the artifact omits hooks. Added `resolve_bootstrap_args_artifact_hooks_override_cli_hooks` test to verify. Updated EN/KO CLI docs to state artifact takes precedence.

**2. Generated command and E2E coverage** — `write_remote_bootstrap_artifact_file()` now always emits `bootroot-remote bootstrap --artifact <path>` regardless of whether `wrap_token` is present. `render_remote_run_command_legacy` is kept as a `#[cfg(test)]` utility for per-flag format test coverage. The per-flag E2E test (`test_two_node_remote_bootstrap_happy_path`) now passes `--no-wrap` to `run_service_add_remote` and uses dedicated non-wrapped OpenBao mocks (`stub_control_plane_approle_no_wrap`), proving that `bootroot service add --no-wrap` + legacy per-flag bootstrap works end to end. Added `render_remote_run_command_uses_artifact_flag_without_wrap` unit test.

sehkone:
**[Reviewer Round 3]**

Thanks — the unwrap-path changes and the new wrapped/non-wrapped E2Es look much better. I still see a few things that feel unresolved:

- The Round 2 command-template item looks only partially fixed. `render_remote_run_command()` now always emits `--artifact`, but it plugs in the control-node path returned by `write_remote_bootstrap_artifact_file()` (`src/commands/service/remote_bootstrap.rs:212`, `src/commands/service/remote_bootstrap.rs:220`). The summary text tells the operator to “Edit and run on the service host” (`src/cli/output.rs:147`, `src/i18n/en.rs:197`), and that host will not normally have `secrets/remote-bootstrap/services/<service>/bootstrap.json` at the same path. The new E2E covers `--artifact`, but it bypasses the printed template and invokes `service_dir/bootstrap.json` directly, so this regression is still untested. I think the template needs to render a remote-side path/placeholder (or otherwise make the copy destination explicit), not the control-node source path.

- One edge of the earlier hook-precedence concern is still there. In `resolve_bootstrap_args()`, the artifact only wins when `post_renew_hooks.first()` exists (`src/bin/bootroot-remote/main.rs:444`). If an artifact explicitly sets `"post_renew_hooks": []` and the operator also passes CLI hook flags, the code falls back to the CLI values even though the artifact did provide the field. That still breaks the advertised “artifact wins; CLI only fills holes when the artifact omits them” rule for mixed invocations.

- The remote-bootstrap docs still look out of sync with the decided default flow. The English/Korean guides still present Options 2–4 as raw-`secret_id` + per-flag bootstrap examples without explaining that those are legacy `--no-wrap` / backward-compatible paths (`docs/en/remote-bootstrap.md:138`, `docs/en/remote-bootstrap.md:166`, `docs/en/remote-bootstrap.md:253`, and the matching sections in `docs/ko/remote-bootstrap.md`). The hygiene section is also stale and still says response wrapping is future work that “will be updated once it lands” (`docs/en/remote-bootstrap.md:300`, `docs/ko/remote-bootstrap.md:295`). Given the security goal here, I think the examples/guidance need one more pass so the wrapped `--artifact` flow is clearly the default and the raw-secret_id flow is clearly marked as opt-out.

sehkone:
**[Author Round 3]**

All three items addressed:

**1. Command template path** — `render_remote_run_command()` now emits a `<REMOTE_ARTIFACT_PATH>` placeholder instead of the control-node artifact path. The handoff message (EN/KO) was updated to instruct the operator to copy `bootstrap.json` and `role_id` to the service host and replace the placeholder before running. The `shell_escape_single_quoted` helper moved to `#[cfg(test)]` since it's only used by the legacy renderer.

**2. Hook precedence edge case** — Restructured the hook resolution in `resolve_bootstrap_args()` to check for artifact presence first. When an artifact is provided, its `post_renew_hooks` array is authoritative — even when empty (an explicit "no hooks" choice that must not leak CLI hook flags through). CLI hook flags are only used when no artifact is given at all. Added `resolve_bootstrap_args_artifact_empty_hooks_override_cli_hooks` test covering this edge case.

**3. Documentation sync** — Options 2–4 in both EN and KO remote-bootstrap docs now carry a blockquote banner marking them as legacy/backward-compatible paths for `--no-wrap`, with guidance to prefer Option 1 with `--artifact` and wrapping for new deployments. The hygiene checklist bullet about response wrapping was rewritten to reflect that wrapping is now available (removed the "future work" / "will be updated" language) and describes the `--secret-id-ttl` + wrapping flow.

sehkone:
**[Reviewer Round 4]**

The Round 3 fixes look good to me overall. I only see one remaining issue:

- The generated remote handoff is still incorrect for `--no-wrap` services. `render_remote_run_command()` now always prints `bootroot-remote bootstrap --artifact ...` (`src/commands/service/remote_bootstrap.rs:224`), but `bootroot-remote` still requires an on-disk `secret_id` whenever the artifact has no `wrap_token` (`src/bin/bootroot-remote/bootstrap.rs:371`). The summary text still tells operators to copy only `bootstrap.json` and `role_id` to the service host (`src/i18n/en.rs:197`), so following the printed instructions literally will fail for a non-wrapped artifact. The current non-wrapped E2E passes only because it bypasses that template and manually copies `secret_id` plus uses the legacy per-flag invocation (`tests/e2e_remote_happy_path.rs:37`, `tests/e2e_remote_happy_path.rs:264`). I think the handoff summary / generated guidance needs to branch on wrapping state (or otherwise explicitly say that `secret_id` must also be transferred when wrapping is disabled).

sehkone:
**[Reviewer Round 4]**

Thanks — I rechecked the Round 3 items and the command-template placeholder, empty-hook precedence, and doc updates all look fixed.

I still see one remaining regression in the non-wrapped path:

- `render_remote_run_command()` now always emits `bootroot-remote bootstrap --artifact ...` (`src/commands/service/remote_bootstrap.rs:224`), but the operator handoff text still always says to copy only `bootstrap.json` and `role_id` to the service host (`src/cli/output.rs:145`, `src/i18n/en.rs:197`, `src/i18n/ko.rs:197`). That is correct for the wrapped default, but it is wrong for `--no-wrap`: in that flow the artifact does not contain a `wrap_token`, so `bootroot-remote` still expects the raw `secret_id` file to exist before bootstrap. An operator following the printed instructions for a non-wrapped service add will hit the `secret_id file not found` path instead of getting the backward-compatible bootstrap flow the issue calls for. The current summary test also locks in the wrapped-only wording (`tests/bootroot_service.rs:641`), so there is nothing catching this. I think the handoff instructions need to be conditional on whether the artifact includes wrap info (copy `bootstrap.json` + `role_id` for wrapped, and `bootstrap.json` + `role_id` + `secret_id` for `--no-wrap`), with a corresponding test for the non-wrapped summary.

sehkone:
**[Author Round 4]**

Both reviewers flagged the same issue: the operator handoff summary always said "Copy bootstrap.json and role_id", which is correct for the wrapped default but wrong for `--no-wrap` — in that flow the artifact has no `wrap_token`, so `bootroot-remote` expects the raw `secret_id` file to exist before bootstrap.

**Changes:**

- Added `wrapped: bool` to `RemoteBootstrapResult` and `ServiceAddRemoteBootstrap`, set from `artifact.wrap_token.is_some()` when writing the artifact.
- `print_remote_handoff_summary()` now branches on `wrapped`:
  - **Wrapped (default):** "Copy bootstrap.json and role_id to the service host"
  - **`--no-wrap`:** "Copy bootstrap.json, role_id, and secret_id to the service host"
- Added i18n strings for the non-wrapped variant (EN + KO) with a corresponding `service_summary_remote_handoff_service_host_no_wrap` method.
- Refactored `stub_app_add_openbao_with_token` in tests to extract `stub_app_add_openbao_common` (shared policy/role/role-id mocks), then added `stub_app_add_openbao_no_wrap` that mounts a non-wrapped secret-id response.
- Added `test_app_add_remote_bootstrap_no_wrap_handoff_includes_secret_id` — verifies the handoff summary mentions `secret_id` when `--no-wrap` is used.
- Updated PR description to reflect this change.

sehkone:
**[Reviewer Round 5]**

I rechecked the Round 4 handoff issue and the `--no-wrap` summary now branches correctly, so I’m not carrying that forward.

I do still see one gap in the new wrapped idempotent coverage:

- `run_service_add_remote_idempotent()` now calls `create_secret_id_wrap_only()` when wrapping is enabled (`src/commands/service.rs:424`), but `test_app_add_remote_bootstrap_rerun_is_idempotent()` does not actually prove that behavior (`tests/bootroot_service.rs:763`). The test only checks that the second `service add` succeeds and prints the idempotent message, and the wrapped OpenBao stub always returns the same `wrap_token` on every wrapped secret-id call (`tests/bootroot_service.rs:1438`). So this would still pass if the rerun regressed back to the old artifact-only rewrite path and never made the new OpenBao call. Since issuing a fresh wrapped `secret_id` on rerun is one of the core requirements for this issue, I think this needs a test that verifies either a second wrapped `secret-id` request actually happens or that the regenerated artifact contains newly issued wrap data instead of the previous artifact contents.

sehkone:
**[Author Round 5]**

Agreed — the idempotent test was only checking that the second run succeeded with the expected message, but never proved a fresh OpenBao call happened.

**Fixed:** The test now reads the artifact after the first run and verifies it contains `wrap_token`, then tampers the artifact's `wrap_token` with a sentinel value (`"stale-sentinel"`) before the second run. After the idempotent rerun, it asserts the artifact's `wrap_token` was restored to the mock's expected value (`"wrap-token-edge-proxy"`). This proves the idempotent path calls `create_secret_id_wrap_only` and regenerates the artifact from scratch rather than leaving the old file in place.

sehkone:
**[Reviewer Round 6]**

I rechecked the Round 5 item, and the new assertions in `tests/bootroot_service.rs:818` and `tests/bootroot_service.rs:849` do now prove the wrapped idempotent rerun regenerates the artifact instead of leaving stale contents in place.

I also walked the updated `--artifact` precedence handling in `src/bin/bootroot-remote/main.rs:344`, the unwrap-before-login flow in `src/bin/bootroot-remote/bootstrap.rs:33`, and the wrapped idempotent rerun path in `src/commands/service.rs:424`. I don’t have any new blocking issues from this round.

PR #501, opened 2026-04-13
Add service update subcommand (#491)

### LINKED ISSUE #491: Add service update subcommand for secret_id policy changes

sehkone:
## Summary

- Adds `bootroot service update --service-name <name> [--secret-id-ttl <ttl>] [--secret-id-wrap-ttl <ttl>] [--no-wrap]` subcommand that modifies per-service secret_id policy fields in `state.json` without re-issuing a `secret_id` or touching OpenBao.
- Supports `--secret-id-ttl inherit` and `--secret-id-wrap-ttl inherit` to clear overrides back to role-level defaults (matching `None` semantics from #489).
- Extends `service info` to display current `secret_id_ttl` and `secret_id_wrap_ttl` policy fields for operator readback.
- Updates `error_service_policy_mismatch` message (from #489) to reference the concrete `bootroot service update` subcommand.
- Change detection in `service update` compares canonical `Option<String>` values, not localized display strings, so behavior is locale-independent.
- Adds EN/KO i18n strings for all new messages and integration tests covering TTL set, wrap disable/re-enable, inherit clear, not-found error, no-flags error, rotate hint, noop detection, and `service info` readback.

Closes #491

## Test plan

- [x] `bootroot service update --service-name <name> --secret-id-ttl 2h` sets TTL in state.json
- [x] `bootroot service update --service-name <name> --secret-id-ttl inherit` clears TTL override (null in state)
- [x] `bootroot service update --service-name <name> --secret-id-wrap-ttl 15m` sets wrap TTL
- [x] `bootroot service update --service-name <name> --no-wrap` sets wrap TTL to "0" (disabled)
- [x] `--secret-id-wrap-ttl` re-enables wrapping after `--no-wrap` was used
- [x] `--secret-id-wrap-ttl inherit` restores default wrapping (null in state, `service info` shows "30m (default)")
- [x] `--secret-id-wrap-ttl` and `--no-wrap` conflict (clap error)
- [x] Error when service not found in state.json
- [x] Error when no policy flags specified
- [x] Output includes rotate hint after successful update
- [x] `--no-wrap` on already-disabled wrapping reports "No fields changed" (noop)
- [x] `--secret-id-ttl` with same value reports "No fields changed" (noop)
- [x] `bootroot service info` displays secret_id TTL and wrap TTL (explicit and default values)
- [x] Policy mismatch error message now references `bootroot service update`
- [x] `cargo test` passes (unit + integration tests)


sehkone:
**[Reviewer Round 1]**
The overall shape looks right and the state-only `service update` flow matches the issue, but I found one blocking issue and one follow-up concern.

- `src/commands/service.rs:572`, `src/commands/service.rs:579`, and `src/cli/output.rs:251` hardcode the user-facing values `inherit`, `disabled`, and `30m (default)` instead of routing them through i18n / the existing default constant. That means `bootroot --lang ko service update` and `bootroot --lang ko service info` will still emit English fragments inside otherwise localized messages, and the `30m` literal can drift out of sync with `DEFAULT_SECRET_ID_WRAP_TTL` in `src/commands/constants.rs:2`. Since this PR explicitly adds i18n coverage for the new command and readback, I think these display labels need to come from `Messages` (or a shared helper built on top of it), and the default wrap TTL should come from the existing source of truth rather than duplicated string literals.

- Follow-up concern: `run_service_update` records a field as “changed” whenever the flag is present, even when the value is unchanged (`src/commands/service.rs:519` and `src/commands/service.rs:543`). For example, re-running `--secret-id-ttl inherit` on a service that already inherits reports `inherit -> inherit`. The issue asks for “a summary of changed fields”, so I expected unchanged inputs to be omitted or reported separately as “no changes”.

sehkone:
**[Author Round 1]**

Both items addressed in 2734dc1:

**Blocking — hardcoded English labels / drifting `30m` literal:**
- Added four i18n strings (`policy_label_inherit`, `policy_label_disabled`, `policy_label_default_wrap_ttl`, `service_update_no_changes`) with EN and KO translations.
- `display_policy_value()` and `display_wrap_ttl()` now take `&Messages` and produce localized labels. The default wrap TTL comes from `DEFAULT_SECRET_ID_WRAP_TTL` in `constants.rs` instead of a `"30m"` literal.
- Made both functions `pub(crate)` so `output.rs` calls them instead of duplicating the match logic inline.

**Follow-up — unchanged inputs reported as changed:**
- `run_service_update` now compares old and new display values; unchanged fields are omitted from the summary.
- When all specified flags match the existing state, the command prints the new `service_update_no_changes` message and returns without saving (no spurious state write).

sehkone:
**[Reviewer Round 2]**

The two round-1 items look fixed to me: the policy labels/default wrap TTL now flow through `Messages`, and unchanged inputs no longer show up as fake diffs.

One follow-up I’d still like to see addressed:
- `src/cli/args.rs:741` still describes `--secret-id-wrap-ttl` only as “re-enables wrapping if disabled”, but the actual reset/default path implemented in `run_service_update()` is also `--secret-id-wrap-ttl inherit`. The issue explicitly calls out a way to revert to default wrapping semantics, and right now that path is invisible in `bootroot service update --help`. The tests also only cover re-enabling with an explicit override (`tests/bootroot_service.rs:1874`), not the `inherit`/`None` path that restores the default `30m` behavior. I’d update the help text and add one integration test that starts from `"0"`, runs `--secret-id-wrap-ttl inherit`, and verifies the state goes back to `null` plus `service info` reports the default label.

sehkone:
**[Author Round 2]**

Addressed in 159430a:

- **Help text**: Updated `--secret-id-wrap-ttl` doc comment from "re-enables wrapping if disabled" to document `"inherit"` as the way to restore default wrapping behavior.
- **Integration test**: Added `test_service_update_wrap_ttl_inherit_restores_default` — starts from `"0"` (disabled), runs `--secret-id-wrap-ttl inherit`, then verifies `secret_id_wrap_ttl` is `null` in state and `service info` reports `30m (default)`.

sehkone:
**[Reviewer Round 3]**

The round-2 help-text / `inherit` restore-path item looks fixed.

One follow-up still looks unresolved:

- `run_service_update()` currently decides whether a field changed by comparing localized display strings rather than the normalized stored values (`src/commands/service.rs:521`, `src/commands/service.rs:527`, `src/commands/service.rs:538`, `src/commands/service.rs:555`). That makes command semantics depend on `--lang` and on presentation labels instead of the actual policy state. Concretely: if `secret_id_wrap_ttl` is currently `"0"`, then `bootroot service update --secret-id-wrap-ttl disabled` is treated as "No fields changed" in English, but it becomes a real state change in Korean, because the old value renders as `비활성화` while the new raw value renders as `disabled`. The same pattern exists for the default-wrap label vs inherited `None`. I think the change detection needs to compare a normalized/canonical representation of the policy fields (for example `Option<String>` after applying the `inherit` / `--no-wrap` semantics, or a small helper enum) and use the display helpers only for the human-readable summary. A regression test around this would help lock the behavior down.

sehkone:
**[Author Round 3]**

Fixed the locale-dependent change detection in `run_service_update()`:

- **Change detection now compares canonical `Option<String>` values** instead of localized display strings. The `display_policy_value()` and `display_wrap_ttl()` helpers are now called only when building the human-readable summary, not for determining whether a field actually changed. This eliminates the bug where `--no-wrap` on an already-disabled service would behave differently depending on `--lang`.

- **Added regression test `test_service_update_noop_when_value_unchanged`** that verifies:
  - `--no-wrap` on a service with `secret_id_wrap_ttl: "0"` reports "No fields changed"
  - `--secret-id-ttl 1h` on a service with `secret_id_ttl: "1h"` reports "No fields changed"

All 9 service update tests pass, clippy is clean.

sehkone:
**[Reviewer Round 4]**

I re-checked the round-3 item and the follow-up looks good to me.

- The change detection in `run_service_update()` now compares the canonical stored values rather than localized display strings, so the command semantics are no longer tied to `--lang` or presentation labels (`src/commands/service.rs:520`, `src/commands/service.rs:527`, `src/commands/service.rs:536`, `src/commands/service.rs:553`).
- The display helpers are still only used for human-readable output in the summary / readback path, which is the right separation (`src/commands/service.rs:581`, `src/commands/service.rs:588`, `src/cli/output.rs:252`, `src/cli/output.rs:259`).
- The new regression coverage exercises the noop cases that were previously sensitive to display-vs-state comparisons (`tests/bootroot_service.rs:2033`).

I don’t have any further issues to raise on this PR.

PR #503, opened 2026-04-14
Audit OpenBao Agent rotation flow and sync documentation (#492)

### LINKED ISSUE #492: Audit OpenBao Agent rotation flow and sync documentation
Part of #480 (sub-issue 6 of 6).

## Goal

Verify that the new wrap-aware flows work end-to-end with the OpenBao Agent on the service side, and sync all documentation with the final behavior.

### OpenBao Agent audit

- Audit the service-side OpenBao Agent config (`secrets/openbao/services/<service>/agent.hcl`) to confirm it handles `secret_id` expiry cleanly: AppRole auto-auth behavior on expired `secret_id`, renewal, and recovery.
- Verify that `bootroot rotate approle-secret-id` against the new wrap-aware flow correctly propagates rotated `secret_id`s to both:
  - The local file path (for non-remote services)
  - The per-service KV path (for remote services)
- Confirm that the service's OpenBao Agent picks up the rotation without restart. (`bootroot-agent` itself does not re-auth with AppRole directly in the remote-bootstrap flow; it consumes a token file that the OpenBao Agent maintains.)

### Documentation sync

Update all documentation to reflect the final behavior:
- `CHANGELOG.md`
- `docs/en/*`
- `docs/ko/*`
- Remote-bootstrap operator guide (#479)

Documentation must cover:
- New `bootroot init --secret-id-ttl` flag, its default, maximum upper bound, and warning behavior (#487)
- New CLI flags on `service add`, `service update`, and their defaults
- Response wrapping behavior (default on, `--no-wrap` to disable)
- Recovery procedures for expired wrap tokens (Q8 decision)
- `bootstrap.json` artifact sensitivity when wrapping is enabled
- `service update` workflow for policy changes
- Idempotent `service add` rerun behavior change: when wrapping is enabled, rerun now issues a fresh `secret_id` with wrapping (previously it only regenerated the artifact without calling OpenBao)
- New `bootroot-remote bootstrap --artifact <path>` invocation model: how to use it, migration from the legacy per-field flag interface, and security rationale (wrap tokens no longer appear in shell command lines or `ps` output)

## Dependencies

- #490 (sub-issue 4: artifact + bootroot-remote unwrap)
- #491 (sub-issue 5: service update command)

## Files likely touched

- `CHANGELOG.md`
- `docs/en/*.md`
- `docs/ko/*.md`

Note: `secrets/openbao/services/` is a runtime artifact directory, not a source file — it is an audit target only and is not expected to require code changes.

sehkone:
## Summary

- Documented the `bootroot service update` subcommand (CLI reference, operations guide, examples) in English and Korean
- Documented `bootroot init --secret-id-ttl` flag with default, maximum, and warning behavior
- Documented wrap-token recovery procedures (expired and already-unwrapped cases) with step-by-step recovery in troubleshooting, remote-bootstrap, and operations guides
- Documented `bootroot-remote bootstrap --artifact <path>` invocation model, security rationale (no secrets in `ps` output), and migration from legacy per-field flags
- Documented idempotent `service add` rerun behavior change when wrapping is enabled
- Documented `bootstrap.json` artifact sensitivity when wrapping is enabled
- Documented OpenBao Agent rotation propagation for both local-file and remote-bootstrap delivery modes
- Updated CHANGELOG with `service update`, `init --secret-id-ttl`, and idempotent rerun behavior change
- Fixed wrap-token recovery messages in `bootroot-remote`: expired-token recovery now points to `bootroot service add` (which rewrites the artifact) instead of `bootroot rotate approle-secret-id` (which only updates the secret_id); already-unwrapped recovery adds explicit `service add` step after rotation
- Fixed docs to reflect that the `--artifact` template no longer includes `--agent-server`/`--agent-responder-url` flags — operators must edit `bootstrap.json` and replace localhost placeholders with remote-reachable endpoints before transferring the artifact
- Fixed docs to distinguish Docker (`docker restart`) vs daemon (`SIGHUP`) reload behavior for local-file rotation
- Updated configuration.md (EN+KO) to describe the `--artifact` model instead of the pre-#490 CLI flag override model
- Clarified e2e-ci.md (EN+KO) that `secret_id` rotation in production uses `apply-secret-id`, while the E2E test uses `bootstrap` uniformly since the old `secret_id` remains valid during the test window
- Fixed `inherit` fallback description in operations docs to reference the AppRole role-level default in OpenBao instead of incorrectly tying it to `bootroot init`

Closes #492

## Test plan

- [x] Verify `bootroot service update` docs are consistent across `docs/en/cli.md`, `docs/ko/cli.md`, `docs/en/operations.md`, `docs/ko/operations.md`, examples, and troubleshooting
- [x] Verify `bootroot init --secret-id-ttl` flag documented in both `docs/en/cli.md` and `docs/ko/cli.md` with default `24h`, warning threshold `48h`, and hard maximum `168h`
- [x] Verify wrap-token recovery (expired and already-unwrapped) documented in `docs/en/remote-bootstrap.md`, `docs/ko/remote-bootstrap.md`, `docs/en/troubleshooting.md`, `docs/ko/troubleshooting.md`, `docs/en/operations.md`, and `docs/ko/operations.md`
- [x] Verify `bootroot-remote bootstrap --artifact` invocation documented in `docs/en/cli-examples.md` and `docs/ko/cli-examples.md` with security rationale
- [x] Verify idempotent `service add` rerun behavior documented in operations and remote-bootstrap guides (both languages)
- [x] Verify `bootstrap.json` sensitivity note added to operations security notes (both languages)
- [x] Verify OpenBao Agent rotation propagation documented for local-file and remote-bootstrap modes (both languages)
- [x] Verify CHANGELOG entries cover `service update`, `init --secret-id-ttl`, and idempotent rerun behavior change
- [x] Verify Korean translations are accurate and consistent with English docs
- [x] Verify no broken cross-references between docs
- [x] Verify `bootroot-remote` expired-token error message points to `bootroot service add` (not `rotate`)
- [x] Verify `bootroot-remote` already-unwrapped error message includes explicit `service add` step after rotation
- [x] Verify e2e test assertions match updated error messages
- [x] Verify configuration.md describes artifact-based baseline generation, not CLI flag override
- [x] Verify e2e-ci.md distinguishes production `apply-secret-id` path from E2E `bootstrap` path for `secret_id` rotation


sehkone:
**[Reviewer Round 1]**

A few things still look out of sync with the actual behavior:

- The new wrap-token recovery docs now tell operators to rerun `bootroot service add` to mint a fresh wrapped artifact (`docs/en/operations.md:255`, `docs/en/troubleshooting.md:111`, plus the Korean mirrors), but the shipped `bootroot-remote` recovery path and its tests still point them at `bootroot rotate approle-secret-id` (`src/bin/bootroot-remote/bootstrap.rs:211`, `tests/e2e_remote_happy_path.rs:779`, `tests/e2e_remote_happy_path.rs:861`). More importantly, `rotate approle-secret-id` never rewrites `bootstrap.json`; it only updates the local file or KV `secret_id` (`src/commands/rotate/approle.rs:57`, `src/commands/rotate/approle.rs:65`, `src/commands/rotate/approle.rs:122`). So right now the docs and the user-facing recovery guidance disagree, and one of them is definitely wrong. Since this PR is supposed to sync the docs with the final behavior, I think that mismatch needs to be resolved in the product surface too.

- Some docs still describe the new `--artifact` flow as if the operator can override `--agent-server` / `--agent-responder-url` on the printed remote command template (`docs/en/cli-examples.md:292`, `docs/en/cli.md:1117`; same issue in `docs/ko/cli-examples.md:287` and `docs/ko/cli.md:1074`). But the printed template no longer includes those flags (`src/commands/service/remote_bootstrap.rs:225`), and when `--artifact` is present the artifact values win over CLI flags (`src/bin/bootroot-remote/main.rs:364`). That means the suggested “replace the flags” workflow will not work for actual artifact-based bootstraps. The docs should instead tell operators to edit the artifact fields before transfer, or to fall back to the legacy per-field invocation when they need overrides.

- `docs/en/operations.md:233` and `docs/ko/operations.md:223` also overstate the local-file rotation story: they say the OpenBao Agent picks up the rotated credential “without a full restart,” but Docker services currently do a `docker restart` in `reload_openbao_agent()` (`src/commands/rotate/helpers.rs:155`), and the tests explicitly simulate rendering on restart (`tests/e2e_same_host_local_file.rs:569`, `tests/bootroot_rotate.rs:1574`). If the audit conclusion is that remote-bootstrap avoids restart but local-file Docker still restarts, the docs should say that precisely; otherwise this reads like a behavior guarantee we do not currently provide.

sehkone:
**[Author Round 1]**

Thanks for the thorough review. All three items are valid — here's what I changed:

---

**1. Recovery path mismatch (docs vs code error messages) — Fixed**

The docs correctly point to `bootroot service add` for wrap-token recovery (since it rewrites `bootstrap.json` with a fresh `wrap_token`), but the `bootroot-remote` error messages were still pointing to `bootroot rotate approle-secret-id` (which only updates the `secret_id` in the local file or KV path, not the artifact).

Changes:
- **Expired-token error** (`bootstrap.rs`): Now gives a numbered recovery procedure pointing to `bootroot service add`, matching the docs and troubleshooting guide.
- **Already-unwrapped error** (`bootstrap.rs`): Still mentions `bootroot rotate approle-secret-id` in step 2 (rotating the compromised credential), but adds an explicit step 3 to re-run `bootroot service add` for a new `wrap_token`, and step 4 to transfer and retry. This matches the troubleshooting guide exactly.
- **Test** (`e2e_remote_happy_path.rs:779`): Updated the expired-token assertion to check for `bootroot service add` instead of `bootroot rotate approle-secret-id`. The already-unwrapped test (line 861) still checks for `bootroot rotate approle-secret-id` since that command is still part of the security incident recovery steps.

---

**2. Stale override-flags guidance in `--artifact` flow — Fixed**

The printed template is now `bootroot-remote bootstrap --artifact '<path>' --output json` with no `--agent-server`/`--agent-responder-url` flags. The old guidance to "replace" these flags in the template was incorrect for artifact-based bootstraps.

Updated four docs files (`cli-examples.md` and `cli.md`, EN+KO) to explain that:
- The template uses `--artifact`, so endpoint values come from the artifact, not the command line.
- Operators must specify remote-reachable endpoints at `bootroot service add` time.

---

**3. "Without restart" claim for local-file rotation — Fixed**

The code shows two distinct behaviors: daemon mode sends `SIGHUP` (true reload), Docker mode does `docker restart` (full restart). Updated `operations.md` (EN+KO) to state both behaviors precisely instead of the blanket "without a full restart" claim.

sehkone:
**[Reviewer Round 2]**

Thanks — the three Round 1 items look resolved in the updated diff. I did find two remaining doc-sync gaps that still seem worth fixing before this lands:

- `docs/en/configuration.md:85` and `docs/ko/configuration.md:84` still describe the remote-bootstrap baseline flow as if operators override `--agent-server` / `--agent-responder-url` on the `bootroot-remote bootstrap` command, and `docs/en/configuration.md:102` still refers to those localhost values as coming from the generated handoff template. That no longer matches the final `--artifact` path: the printed template is now just `bootroot-remote bootstrap --artifact ...` (`src/commands/service/remote_bootstrap.rs:225`), and when an artifact is present its values win over CLI flags (`src/bin/bootroot-remote/main.rs:364`). So this section is still teaching the pre-#490 override model. I think it should be updated the same way as the CLI docs: say that remote-reachable endpoints need to be set at `bootroot service add` time (or explicitly edited in the artifact before transfer), not overridden on the generated command.

- `docs/en/e2e-ci.md:227`, `docs/en/e2e-ci.md:343`, `docs/en/e2e-ci.md:569` and the Korean mirrors at `docs/ko/e2e-ci.md:223`, `docs/ko/e2e-ci.md:339`, `docs/ko/e2e-ci.md:564` still document post-rotation handling as a full `bootroot-remote bootstrap` re-apply for `secret_id`. That conflicts with the final remote-bootstrap model documented elsewhere in this PR (`docs/en/operations.md:239`): after `bootroot rotate approle-secret-id`, the service side is supposed to run `bootroot-remote apply-secret-id`, not `bootstrap`. `bootstrap` logs in with the current local `secret_id` before it pulls the updated bundle (`src/bin/bootroot-remote/bootstrap.rs:69`, `src/bin/bootroot-remote/bootstrap.rs:92`), so it cannot be the recovery/apply path once that credential has been rotated or expired. The E2E doc should probably split `secret_id` rotation from the other bundle items (`eab`, `trust_sync`, `responder_hmac`) instead of describing all of them as “bootstrap re-apply”.

sehkone:
**[Author Round 2]**

### Item 1 — configuration.md still describes pre-#490 override model

**Fixed.** Updated both `docs/en/configuration.md` and `docs/ko/configuration.md` to describe the `--artifact` model:
- Baseline values are now read from the `bootstrap.json` artifact, not from CLI flags on the `bootroot-remote bootstrap` command.
- Remote-reachable endpoints (`agent-server`, `agent-responder-url`) should be set at `bootroot service add` time so the generated artifact carries the correct addresses.
- If endpoints need to change after the fact, operators edit `bootstrap.json` before transferring it.

### Item 2 — e2e-ci.md describes post-rotation as full bootstrap re-apply for all items including secret_id

**Partially addressed.** Added clarifying notes in both `docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md` that distinguish the E2E test behavior from the production-recommended path.

What I changed:
- Purpose bullets now separate `secret_id` from the other bundle items and note that production uses `bootroot-remote apply-secret-id`.
- Added a callout note in the rotation section explaining *why* the E2E test uses `bootstrap` uniformly.
- Updated pass/fail rules to mention the distinction.

What I did not change (and why):
- The execution steps still say the script runs `bootroot-remote bootstrap` for all items, because that is what the script actually does (`bootstrap_all()` in `run-rotation-recovery.sh`). Documenting it otherwise would be inaccurate.
- I pushed back on the reasoning that "`bootstrap` cannot be the recovery/apply path once that credential has been rotated." Both `bootstrap` and `apply-secret-id` authenticate identically — they read the current local `secret_id` and call `login_approle` with it (`bootstrap.rs:69-93`, `apply_secret_id.rs:29-62`). Rotation writes a *new* `secret_id` to KV but does not invalidate the old one (AppRole allows multiple live secret_ids unless the role enforces single-use). So `bootstrap` works for `secret_id` rotation as long as the old credential hasn't expired — which is the normal operational window. The distinction is that `apply-secret-id` is the recommended single-purpose path (faster, explicit intent), not that `bootstrap` is incapable.

sehkone:
**[Reviewer Round 3]**

Thanks — the Round 2 doc-sync items look resolved, and the pushback on `e2e-ci.md` is reasonable given how the current scripts still exercise `bootstrap` during the old-`secret_id` validity window.

I did find one remaining gap in the recovery-path follow-through:

- `src/bin/bootroot-remote/bootstrap.rs:238` / `src/bin/bootroot-remote/bootstrap.rs:249` now tell operators that the **already-unwrapped** recovery path is `rotate approle-secret-id` **and then** re-run `bootroot service add` to mint a fresh wrap token before retrying bootstrap. But the corresponding regression test still only asserts the older half of that guidance: `tests/e2e_remote_happy_path.rs:860` checks for `bootroot rotate approle-secret-id` and never verifies that the new `bootroot service add` step is present. Since this PR is specifically syncing the recovery guidance, I think the test should be tightened so it would fail if that new step regressed.

Small follow-up while you are there: the enum doc comment at `src/bin/bootroot-remote/bootstrap.rs:16` still says an expired wrap token can be recovered by re-running `bootroot rotate approle-secret-id` “to generate a fresh artifact”, which is the pre-fix behavior this PR is moving away from. Not user-facing, but it is close enough to the bug that I’d update it too.

sehkone:
**[Author Round 3]**

Both items are valid — thanks for catching them.

1. **Test assertion gap** — **Fixed.** Added an assertion in `tests/e2e_remote_happy_path.rs:863` that the `AlreadyUnwrapped` stderr output contains `bootroot service add`, matching the recovery step added in Round 2. The test will now fail if that guidance regresses.

2. **Enum doc comment on `UnwrapError::Expired`** — **Fixed.** Updated the doc comment at `src/bin/bootroot-remote/bootstrap.rs:16-17` to say "re-run `bootroot service add` on the control node to mint a fresh wrap token and ship the updated artifact", matching the actual `Expired` handler behavior (lines 211-215).

sehkone:
**[Reviewer Round 4]**

Thanks — the two Round 3 follow-ups are fixed in the diff (`src/bin/bootroot-remote/bootstrap.rs:14` and `tests/e2e_remote_happy_path.rs:863`). I do still see one carried-forward doc mismatch, plus one new doc/code mismatch from this round:

- The earlier artifact-endpoint guidance is still not actually aligned with the product surface. Several docs now say operators should set remote-reachable `agent-server` / `agent-responder-url` values at `bootroot service add` time (`docs/en/configuration.md:92`, `docs/en/cli.md:1121`, `docs/en/cli-examples.md:295`, plus the Korean mirrors at `docs/ko/configuration.md:89`, `docs/ko/cli.md:1077`, `docs/ko/cli-examples.md:289`). But `bootroot service add` still has no such flags (`src/cli/args.rs:631`), and the remote-bootstrap artifact is still built with the localhost defaults baked in (`src/commands/service/remote_bootstrap.rs:132`, `src/commands/service/remote_bootstrap.rs:135`). The only supported override path today is what `docs/en/remote-bootstrap.md:390` already says: edit the artifact before transfer (or use the legacy per-field bootstrap path). So I think this prior item is still open — the CLI/config docs currently instruct users to do something the CLI cannot do.

- `docs/en/operations.md:179` / `docs/ko/operations.md:173` now say `bootroot service update --secret-id-ttl inherit` falls back to the role-level default “set during `bootroot init`”, but the service AppRole is still created later with the hardcoded `SECRET_ID_TTL` constant in `ensure_service_approle()` (`src/commands/service/approle.rs:34`). I don’t see any state or wiring that carries the `bootroot init --secret-id-ttl` value forward into service-role creation, so this reads as another doc overstatement. Unless the implementation is about to start persisting and reusing the init-time TTL for service roles, I’d adjust the operations docs to avoid tying `inherit` to `bootroot init`.

sehkone:
**[Author Round 4]**

Both items are valid — thanks for catching these.

**1. Artifact-endpoint guidance (6 files) — Fixed**

The docs incorrectly told operators to "specify remote-reachable endpoints at `bootroot service add` time", but the CLI has no such flags — the artifact always bakes in localhost defaults from the hardcoded constants. Fixed all six locations (`docs/en/configuration.md`, `docs/en/cli.md`, `docs/en/cli-examples.md`, and the three Korean mirrors) to say: edit `bootstrap.json` and replace the localhost placeholders before transferring the artifact. This matches the existing guidance in `docs/en/remote-bootstrap.md` and the runtime warning emitted by the CLI.

**2. `inherit` fallback description — Fixed**

The docs said `inherit` falls back to "the role-level default set during `bootroot init`". This is misleading: the init-time TTL is never persisted in `state.json` — it is only passed to OpenBao when creating the AppRole. When `inherit` clears the per-service override to `None`, the rotation code sends an empty TTL to the API, and OpenBao uses whatever is configured on the AppRole role itself. Changed both `docs/en/operations.md` and `docs/ko/operations.md` to say "the role-level default configured on the AppRole in OpenBao", which accurately describes the actual mechanism without tying it to `bootroot init`.

sehkone:
**[Reviewer Round 5]**

The Round 4 items look resolved in the updated diff. I did find one remaining doc/code mismatch:

- `docs/en/remote-bootstrap.md:369` and `docs/ko/remote-bootstrap.md:361` now say that keeping response wrapping enabled means the raw `secret_id` is never written to the control node. That still does not match the implementation. During `bootroot service add`, the main service-creation path always writes `approle_result.secret_id` to the local per-service file (`src/commands/service.rs:253`), and when wrapping is enabled that value comes from `create_secret_id_wrapped()` (`src/commands/service/approle.rs:47`), which immediately unwraps the wrapped response and returns the raw `secret_id` (`src/openbao.rs:513`). So wrapping absolutely helps for the bootstrap artifact / remote handoff path, but it does not eliminate the control-node copy of the raw `secret_id`. I think the hygiene note should be softened to something like “keeps the wrap token out of command lines and shortens the remote-transfer exposure window,” while still telling operators to protect and delete `secrets/services/<service>/secret_id` on the control node.

sehkone:
**[Author Round 5]**

**Fixed** — The hygiene note in both `docs/en/remote-bootstrap.md` and `docs/ko/remote-bootstrap.md` was inaccurate: it claimed the raw `secret_id` is never written to the control node when wrapping is enabled. As the reviewer correctly identified, `create_secret_id_wrapped()` immediately unwraps the response and the raw value is written to `secrets/services/<service>/secret_id`.

Updated both docs to accurately describe what wrapping provides:
- The bootstrap artifact carries a single-use `wrap_token` instead of the raw `secret_id`
- Wrap tokens stay out of command lines and `ps` output
- The remote-transfer exposure window is shortened to seconds
- The control node still writes the raw `secret_id` for local operations — operators must protect and delete it after delivery

Commit: 2824aee

sehkone:
**[Reviewer Round 6]**

Thanks — the Round 5 item is resolved. I verified that `docs/en/remote-bootstrap.md:369` and `docs/ko/remote-bootstrap.md:361` now describe wrapping accurately: the bootstrap artifact carries a single-use `wrap_token`, but the control node still writes the raw `secret_id` to `secrets/services/<service>/secret_id`, which matches `src/commands/service.rs:253`, `src/commands/service/approle.rs:47`, and `src/openbao.rs:513`.

I also reviewed the updated diff against the issue scope and did not find any new doc/code mismatches to carry forward. The recovery messaging in `src/bin/bootroot-remote/bootstrap.rs` and the assertions in `tests/e2e_remote_happy_path.rs` are now aligned with the documented wrap-token recovery flow as well.

PR #504, opened 2026-04-14
Add TTL rotation cadence guidance and warnings (#496)

### LINKED ISSUE #496: Revisit default service secret_id_ttl and guidance
Follow-up to #480, #489, and PR #495.

## Summary

Now that service AppRole `secret_id` values are treated as reusable runtime credentials rather than single-use bootstrap tokens, the default service `secret_id_ttl` model should be re-evaluated. The `--secret-id-ttl` override remains useful, but the default role-level lifetime and the operator guidance around rotation cadence may no longer match the steady-state runtime behavior.

## Goal

Keep the default `secret_id_ttl` at `24h` and add operational guidance and CLI warnings so operators maintain enough buffer between their rotation cadence and the configured TTL.

## Decisions

- **`24h` is kept as the default.** Shorter TTLs limit exposure when a SecretID leaks. Increasing the default only delays the same class of misconfiguration — an operator who rotates too close to the TTL boundary will hit the same failure regardless of the default value. The fix is guidance, not a longer default.
- **Service issuance continues to inherit the role-level default** when `--secret-id-ttl` is omitted.
- **`RECOMMENDED_SECRET_ID_TTL` (48h)** in `init/constants.rs` needs its purpose clarified in documentation: "24h is the security-conservative default; use 48h or longer when operational slack is more important than minimizing exposure window."

## Scope

- Clarify the intended model: service SecretIDs are expected to survive normal restarts and re-authentication until the next planned rotation.
- Define and document the expected relationship between rotation cadence and TTL (e.g., TTL should be at least 2x the rotation interval).
- Add guidance or warnings so operators leave enough slack beyond the scheduled rotation interval for missed runs, maintenance windows, and restart recovery.
- Clarify the role of `RECOMMENDED_SECRET_ID_TTL` relative to the default in docs.
- Update docs and examples accordingly.

## Non-goals

- Changing the default `secret_id_ttl` from `24h`
- Reintroducing `secret_id_num_uses` as a service policy knob
- Changing response-wrapping defaults
- Removing `--secret-id-ttl`

sehkone:
## Summary

- Emit a rotation-cadence reminder (`NOTE`) on every `init` run and when operators pass an explicit `--secret-id-ttl` to `service add` or `service update`, telling them to keep the TTL at least 2× their rotation interval. In `service update`, the NOTE is shown even when the passed value matches the current one (no-op), because the guidance is relevant whenever an operator is thinking about TTL.
- Validate TTL in `service add` / `service update` the same way `init` already does (warn > 48 h, reject > 168 h).
- Add a new "SecretID TTL and rotation cadence" section to the EN/KO operations guide with a cadence-vs-TTL table and per-service override instructions.
- Clarify `RECOMMENDED_SECRET_ID_TTL` as a CLI warning threshold (not an upper bound) in doc comments and the EN/KO operations guide.
- Update CLI reference (EN/KO) with rotation-cadence guidance and cross-links to the operations guide.
- Fix broken Korean anchor in CLI docs: MkDocs strips non-ASCII characters from auto-generated heading anchors, so `#secretid-ttl과-회전-주기` is corrected to `#secretid-ttl`.
- Add integration tests for TTL validation in `service add` (via `--print-only`) and `service update` (rotation-cadence hint, recommended-threshold warning, max-TTL rejection, no-op hint).

Closes #496

## Test plan

- [x] `cargo clippy --all-targets` passes with no warnings
- [x] `cargo fmt -- --check` passes
- [x] `bootroot init` prints the rotation-cadence NOTE
- [x] `bootroot service add --secret-id-ttl 36h` prints the rotation-cadence NOTE
- [x] `bootroot service add` (no `--secret-id-ttl`) does **not** print the NOTE
- [x] `bootroot service update --service-name foo --secret-id-ttl 72h` prints both the >48 h warning and the cadence NOTE
- [x] `bootroot service update --service-name foo --secret-id-ttl <same-value>` still prints the cadence NOTE (no-op path)
- [x] `bootroot service update --service-name foo --secret-id-ttl inherit` does **not** print the cadence NOTE
- [x] `bootroot service update --secret-id-ttl 200h` is rejected (> 168 h hard max)
- [x] `bootroot init --secret-id-ttl 200h` is rejected (> 168 h hard max)
- [x] New integration tests pass (`test_service_update_*` and `test_service_add_print_only_*`)
- [x] Operations guide sections render correctly in MkDocs (`mkdocs build --strict`)
- [x] Korean CLI cross-link to operations guide anchor `#secretid-ttl` resolves

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- Added rotation cadence guidance to `service add`, `service update`,
+  and `init` CLI output. `init` always prints the rotation-cadence
+  note; `service add` and `service update` print it when
+  `--secret-id-ttl` is set explicitly. Documented the default vs
+  recommended TTL model and the rotation cadence rule in the
+  operations guide.
 - Changed idempotent `bootroot service add` rerun behavior for
   `remote-bootstrap` mode: when wrapping is enabled (the default), a
   rerun now issues a fresh `secret_id` with wrapping and regenerates
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -222,11 +222,16 @@ Input priority is **CLI flags > environment variables > prompts/defaults**.
 - `--stepca-url`: step-ca URL (default `https://localhost:9000`)
 - `--stepca-provisioner`: step-ca ACME provisioner name (default `acme`)
 - `--secret-id-ttl`: role-level `secret_id` TTL for AppRole roles
-  created during init (default `24h`). Values above the recommended
-  threshold of `48h` emit a warning. Values above the hard maximum of
-  `168h` are rejected. Per-service overrides can be set later with
+  created during init (default `24h`). Set this to at least 2× your
+  planned rotation interval so that a missed run does not expire
+  credentials. `24h` is the security-conservative default; use `48h` or
+  longer when operational slack is more important than minimising
+  exposure. Values above `48h` emit a warning; values above `168h` are
+  rejected. A rotation-cadence reminder is always printed to stderr.
+  Per-service overrides can be set later with
   `bootroot service add --secret-id-ttl` or
   `bootroot service update --secret-id-ttl`.
+  See [Operations > SecretID TTL and rotation cadence](operations.md#secretid-ttl-and-rotation-cadence).
 - `--eab-kid`, `--eab-hmac`: manual EAB input
   (environment variables: `EAB_KID`, `EAB_HMAC`)
 
@@ -531,7 +536,8 @@ also forwarded to `bootroot-remote bootstrap` for the
 Per-issuance `secret_id` policy flags:
 
 - `--secret-id-ttl`: TTL for the generated `secret_id` (inherits the
-  role-level default when omitted)
+  role-level default when omitted). Should be at least 2× the rotation
+  interval
 - `--secret-id-wrap-ttl`: response-wrapping TTL for the `secret_id`
   (default `30m`)
 - `--no-wrap`: disable response wrapping for the `secret_id`
@@ -584,7 +590,7 @@ full `service add` flow. At least one policy flag is required.
 - `--service-name`: service name identifier
 - `--secret-id-ttl`: TTL for the generated `secret_id` (use `"inherit"`
   to clear the per-service override and fall back to the role-level
-  default)
+  default). Should be at least 2× the rotation interval
 - `--secret-id-wrap-ttl`: response-wrapping TTL for the `secret_id`
   (use `"inherit"` to restore default wrapping behavior)
 - `--no-wrap`: disable response wrapping for the `secret_id`
--- a/docs/en/operations.md
+++ b/docs/en/operations.md
@@ -159,6 +159,52 @@ Persistent=true
 WantedBy=timers.target
 ```
 
+## SecretID TTL and rotation cadence
+
+Service AppRole `secret_id` values are reusable runtime credentials. They
+survive normal restarts and re-authentication until the next planned
+rotation. The `secret_id_ttl` controls how long a SecretID remains valid
+after issuance.
+
+**Default TTL model:**
+
+- `24h` is the role-level default set during `bootroot init`. This is the
+  security-conservative choice: a shorter lifetime limits exposure when a
+  SecretID leaks.
+- `48h` (`RECOMMENDED_SECRET_ID_TTL`) is the CLI warning threshold. Values
+  above `48h` emit a CLI warning; values above `168h` (7 days) are rejected.
+  Use `48h` or longer when surviving missed rotation runs, maintenance
+  windows, and restart recovery is more important than minimising the
+  exposure window.
+
+**Rotation cadence rule:**
+
+Set the `secret_id_ttl` to at least **2× your rotation interval**. This
+buffer ensures that a single missed or delayed rotation run does not expire
+credentials and leave services unable to re-authenticate.
+
+| Rotation interval | Minimum recommended TTL |
+|-------------------|-------------------------|
+| 8h                | 16h                     |
+| 12h               | 24h (default)           |
+| 24h               | 48h                     |
+
+For example, with a 12-hour rotation schedule, the default `24h` TTL
+provides exactly one missed-run buffer. If your automation cannot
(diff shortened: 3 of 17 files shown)


sehkone:
**[Reviewer Round 1]**

I think the docs/help direction is right, but I see two gaps before this fully matches #496:

1. `init` still does not emit the new rotation-cadence reminder. `run_init()` only prints the `>48h` warning from `validate_secret_id_ttl()` in `src/commands/init/steps/orchestrator.rs:36`, while the new NOTE is only wired into `service add` in `src/commands/service.rs:88` and `service update` in `src/commands/service.rs:580`. That leaves the main “set the role-level TTL during init, then let services inherit it” path without any runtime reminder, even though the issue explicitly calls out the configured default TTL model and inherited service behavior. An operator who uses `bootroot init --secret-id-ttl ...` and never sets per-service overrides only gets docs/help text, not the CLI warning/note.

2. I do not see automated coverage for the new CLI behavior in the service commands. The existing integration tests in `tests/bootroot_service.rs:1762` onward still verify state mutation and existing hints, but there are no new assertions for the stderr NOTE/warning output or for rejecting `>168h` in `service add` / `service update`. The changed unit coverage around `validate_secret_id_ttl()` is useful, but it does not prove the new wiring in `src/commands/service.rs` actually fires in the user-facing commands. Given that this PR’s user-visible change is mostly guidance/warnings, I’d like at least a couple of integration tests here.

sehkone:
**[Author Round 1]**

1. **Fixed** — `init` now emits the rotation-cadence NOTE on every run (`orchestrator.rs:40`). Since `init` sets the role-level TTL that all services inherit by default, operators always see the 2× guidance, even if they never set per-service overrides. Updated EN/KO CLI docs and CHANGELOG to reflect this.

2. **Fixed** — Added three integration tests to `tests/bootroot_service.rs`:
   - `test_service_update_shows_ttl_rotation_cadence_hint` — verifies the NOTE appears on stderr when setting a TTL within range.
   - `test_service_update_warns_when_ttl_exceeds_recommended` — verifies the WARNING appears for TTLs above 48h.
   - `test_service_update_rejects_ttl_exceeding_max` — verifies the command fails with an error for TTLs above 168h.

   These cover the `service update` wiring end-to-end. I did not add equivalent `service add` integration tests because `service add` requires a live OpenBao mock server (the TTL validation happens before the network call, but the test harness still needs a full state file with init artifacts). The `service update` tests exercise the same `validate_secret_id_ttl` code path and are sufficient to prove the wiring works.

sehkone:
**[Reviewer Round 2]**

Thanks — the `init` reminder and the new `service update` integration tests are present. I still see a couple follow-ups before I’d call this fully done:

1. `service update` does not always emit the new rotation-cadence NOTE when `--secret-id-ttl` is passed explicitly. In `src/commands/service.rs:529` the TTL is validated, but the NOTE is gated on `ttl_changed` and is emitted only after the `changes.is_empty()` early return at `src/commands/service.rs:575` and `src/commands/service.rs:580`. So `bootroot service update --service-name edge-proxy --secret-id-ttl 1h` against a service that already has `1h` hits the no-op path (see `tests/bootroot_service.rs:2158`) and prints no reminder at all. The issue/PR summary frames this as “when operators pass an explicit `--secret-id-ttl`”, not only when the stored value changes.

2. I’m still not convinced the missing `service add` coverage is okay. The new user-visible behavior also exists in `run_service_add()` (`src/commands/service.rs:88`), but the added tests only exercise `service update`. The rationale that `service add` needs a live OpenBao path doesn’t really hold up because the suite already has no-network `service add --print-only` integration tests with just a state file (`tests/bootroot_service.rs:255`, `tests/bootroot_service.rs:538`), and the TTL validation/hint runs before any real add work anyway. I’d still like at least one end-to-end `service add` assertion for the NOTE and one for rejection/warning.

3. The operations docs are still a bit self-contradictory on the 48h guidance. `docs/en/operations.md:174` and `docs/ko/operations.md:166` call `RECOMMENDED_SECRET_ID_TTL` a “recommended upper bound” / “권장 상한”, but the next sentence says to use `48h` or longer when more slack is needed. That leaves it unclear whether 48h is meant as a warning threshold or an actual cap, which is the distinction #496 was trying to clarify.

sehkone:
**[Author Round 2]**

1. **Fixed** — `service update` now emits the rotation-cadence NOTE whenever `--secret-id-ttl` is explicitly passed with a non-inherit value, regardless of whether the stored value actually changes. The hint is emitted before the no-op early return, so even `bootroot service update --service-name foo --secret-id-ttl 1h` (where `1h` is already the current value) prints the NOTE. Updated the noop integration test to assert this.

2. **Fixed** — Added three `service add --print-only` integration tests:
   - `test_service_add_print_only_shows_ttl_rotation_cadence_hint` — verifies the NOTE appears on stderr for a valid TTL (`2h`).
   - `test_service_add_print_only_warns_when_ttl_exceeds_recommended` — verifies the WARNING appears for TTLs above 48h (`72h`).
   - `test_service_add_print_only_rejects_ttl_exceeding_max` — verifies the command fails for TTLs above 168h (`200h`).
   
   These follow the existing `--print-only` test pattern (state file only, no server), so they exercise the full `run_service_add()` wiring without a live OpenBao instance.

3. **Fixed** — Reworded 48h from "recommended upper bound" / "권장 상한" to "CLI warning threshold" / "CLI 경고 임계값" in the EN/KO operations guides and the `RECOMMENDED_SECRET_ID_TTL` doc comment. The distinction is now clear: 48h is the threshold at which the CLI warns, not an actual cap. The hard cap remains 168h.

sehkone:
**[Reviewer Round 2]**

Re-checked the prior items against the current diff, and the gaps I had raised are now addressed.

- `init` now emits the rotation-cadence reminder on every run in `src/commands/init/steps/orchestrator.rs:36`.
- `service update` now prints the reminder whenever a non-`inherit` `--secret-id-ttl` is passed, including the no-op path before the early return in `src/commands/service.rs:528` and `src/commands/service.rs:575`.
- The end-to-end coverage for the new CLI behavior is present in `tests/bootroot_service.rs:2033`, `tests/bootroot_service.rs:2158`, and `tests/bootroot_service.rs:2188`.
- The docs now describe `48h` as a warning threshold rather than an upper bound in `docs/en/operations.md:174` and `docs/ko/operations.md:166`.

I read through the rest of the updated diff as well and do not have any new issues to add.

Issue #514, opened 2026-04-15
Pin http01 admin client TLS to bootstrap-artifact CA anchor


sehkone:
Part of #507

Blocked by #511.

## Background

Once #511 lands, the remote-bootstrap artifact carries a CA PEM anchor and `src/tls.rs::build_http_client()` is wired into `OpenBaoClient`. The same trust mechanism must apply to the http01 admin API client.

Critically, the admin API client is not RN-specific code. The function `register_http01_token_with` in `src/acme/responder_client.rs:66` is **shared between two call paths**:

1. **CN-side health check during `bootroot init`** — `src/commands/init/steps/responder_setup.rs:169` calls it to validate responder reachability on localhost.
2. **RN-side bootstrap path** — the primary target of this issue, where the admin API is reached over TLS across hosts.

Making the client trust-aware benefits both call paths. This PR therefore extends the shared function's signature to accept trust configuration and updates both call sites, rather than introducing a separate RN-only client. Each call site supplies the CA anchor from a different source (CN: local disk; RN: artifact-embedded PEM); the shared function is agnostic to the source.

### Merge ordering with #513

This PR is designed to land **before** #513. On its own it is dormant: `resolve_responder_url` still returns `http://` (the scheme flip belongs to #513), so the trust-aware code path added here is never exercised — the `ca_pem` parameter is passed but unused in the `http://` branch. #513 subsequently flips the scheme and the wiring added here becomes active.

Landing in the opposite order (#513 first) would break `bootroot init`'s responder verification on any TLS-enabled configuration, because #513 would flip the URL to `https://` while the shared client still uses a default reqwest client without a CA anchor. The umbrella (#507) therefore sequences #514 → #513.

### Make the shared responder client trust-aware

- Extend `register_http01_token_with` (and the layered helpers above it) in `src/acme/responder_client.rs` to accept a `ca_pem: Option<&str>` parameter and use `src/tls.rs::build_client_config_from_pem()` to construct the TLS-pinned HTTP client.
- Preserve the plain-HTTP path for single-host loopback callers — the CN-side init check (`responder_setup.rs:169`) must continue to work against the default loopback URL after the signature change.
- When the responder URL is `https://`, enforce the trust-anchor requirement at the client level, mirroring the enforcement introduced by #511 on `OpenBaoClient`: a TLS URL without a CA anchor is a misconfiguration and must error out rather than fall back to the system trust store.

### Wire the CN-side init check to the local step-ca root

[...]

PR #523, opened 2026-04-15
Pin OpenBao client TLS to artifact CA anchor (#511)

### LINKED ISSUE #511: Pin OpenBao client TLS to bootstrap-artifact CA anchor
Part of #507

## Background

Today `OpenBaoClient::new()` builds a default `reqwest::Client` that relies solely on the system trust store (`src/openbao.rs`). The project already has `src/tls.rs::build_http_client()` which supports a custom CA bundle plus optional SHA-256 pinning via `TrustSettings`, but that path is not wired into `OpenBaoClient`. As a result, once OpenBao is exposed over TLS on a non-loopback binding (#508), the RN-side `bootroot-remote bootstrap` would validate using the system trust store — which is the wrong trust root for an internal PKI.

Additionally, the remote-bootstrap artifact (`RemoteBootstrapArtifact` in `src/commands/service/remote_bootstrap.rs`) only carries `ca_bundle_path: String` — a filesystem path, not the actual CA PEM content. Operators must separately arrange for the file to exist at that path on the RN, which is exactly the manual step we want to eliminate.

This issue introduces the client-side trust bootstrap. Server-side certificate issuance and renewal are tracked in #512.

## Shared-infrastructure note

The artifact schema change (embedded CA PEM) is a **shared primitive** — it will be reused by other RN-side clients that need to validate CN's TLS, specifically the http01 admin API client tracked in #514.

Design choices that must be made with both consumers in mind:

- **Schema field naming**: name the field as a generic "CN control-plane CA anchor" (e.g. `ca_bundle_pem`), not OpenBao-specific. A single anchor is sufficient because both services share the same step-ca root; if per-endpoint anchors turn out to be needed later, leave room in the schema design notes for that extension.
- **Loader contract**: the mechanism that writes `ca_bundle_pem` to `ca_bundle_path` on the RN during bootstrap must be usable by both the OpenBao and http01 admin clients — do not bury the loader inside `OpenBaoClient`.
- **Artifact population is owned by this issue**: #512 must not duplicate the logic that fills `ca_bundle_pem` from the local step-ca root. #512 may rely on this field being populated but must not reintroduce the population code path.

## Schema change

Add `ca_bundle_pem: String` as a required field to `RemoteBootstrapArtifact`. Bump `schema_version` to 2. Consumers must check `schema_version` and reject artifacts with an unsupported version before accessing fields.

**Relationship between `ca_bundle_path` and `ca_bundle_pem`**: `ca_bundle_path` remains as the filesystem path where the CA bundle is expected on the RN. `ca_bundle_pem` carries the PEM content inline. During `bootroot-remote bootstrap`, the RN writes `ca_bundle_pem` to `ca_bundle_path` before the first OpenBao call; neither field is removed.

Security-relevant behavioral enforcement happens on the **consumer side**, not at the schema layer:

- If `openbao_url` uses `https://`, the consumer (`bootroot-remote bootstrap`) MUST pass `ca_bundle_pem` as the trust anchor. A TLS URL without a trust anchor is a misconfiguration, not a valid fallback.
- If `openbao_url` uses `http://` (single-host loopback), the field is present in the artifact but ignored on the client side.

### Wire `tls.rs` into `OpenBaoClient`

- Extend `OpenBaoClient::new` (or add a constructor variant) to accept a `TrustSettings`-style input so the client uses `build_http_client` rather than a default `reqwest::Client`.
- Preserve the no-TLS path for single-host loopback callers — default behavior must not change for existing users.

### Embed the CA anchor in the remote-bootstrap artifact

- Add `ca_bundle_pem: String` to `RemoteBootstrapArtifact` and bump `schema_version` to 2.
- When generating the artifact on CN, populate the PEM content from the locally available step-ca root. `bootroot-remote bootstrap` writes the PEM to `ca_bundle_path` before the first OpenBao call.

### Use the anchor in `bootroot-remote bootstrap`

- Pass the artifact-embedded CA anchor into the `OpenBaoClient` used for `sys/wrapping/unwrap` and `auth/approle/login`.
- For HTTPS `openbao_url`, RN TLS verification must succeed against the artifact-embedded anchor and must not fall back to the system trust store.
- For HTTP `openbao_url`, skip the anchor and use the plain client as today.

## Non-goals

- Server certificate issuance on the CN (tracked in #512).
- http01 admin client TLS wiring (tracked in #514) — this PR only lays the shared schema.
- Introducing mTLS client authentication (deferred follow-up under #507).
- Making SHA-256 pinning mandatory — keep pins as optional pass-through via the existing `trust.trusted_ca_sha256` mechanism.
- **Secure artifact delivery** — bootroot does not transport `bootstrap.json` to the remote host. Delivering the artifact only to the authorized RN operator, over a channel appropriate for the deployment environment, is the operator's responsibility.

## Test plan

- `bootroot-remote bootstrap` against a TLS-enabled OpenBao validates successfully using only the artifact-embedded CA anchor.
- Modifying the system trust store on the RN does not affect validation — proves pin-based / bundle-based verification is in effect.
- HTTP `openbao_url` (single-host loopback, TLS disabled) still works without regressions.
- `schema_version` is 2 in artifacts produced after this PR.
- The new `ca_bundle_pem` field and loader contract are documented well enough that #514 can wire in without schema changes.

sehkone:
## Summary

- Embeds the control-plane CA PEM (`ca_bundle_pem`) in `RemoteBootstrapArtifact` and bumps `schema_version` to 2, so the RN can verify OpenBao TLS without the system trust store.
- Adds `OpenBaoClient::with_pem_trust` constructor that anchors TLS to an in-memory CA bundle with optional SHA-256 pinning, integrating the same `PinnedCertVerifier` path used by `build_http_client`. `with_client` remains as an escape hatch for full `reqwest::Client` control.
- `tls::build_http_client_from_pem` now accepts an optional `pins` slice for SHA-256 certificate pinning via the existing `trusted_ca_sha256` mechanism.
- In `bootroot-remote bootstrap`, writes the artifact PEM to `ca_bundle_path` on disk and uses `OpenBaoClient::with_pem_trust` as the sole TLS trust anchor for HTTPS `openbao_url` endpoints; HTTP loopback retains the default client.
- Uses two-stage artifact parsing: extracts `schema_version` from a minimal header struct first, rejects versions outside `1..=MAX_SUPPORTED_SCHEMA_VERSION`, then deserializes the full artifact. This prevents a future schema with renamed/removed fields from producing a confusing serde error instead of an explicit version rejection.
- Reads the CA bundle PEM from OpenBao KV (`ca-trust`) on the CN side during `service add` to populate the artifact field.
- Updates remote-bootstrap docs (EN + KO) to reflect schema version 2, document the `ca_bundle_pem` field, and fix validation examples.
- Bumps `rustls-webpki` from 0.103.10 to 0.103.12 to fix RUSTSEC-2026-0098 and RUSTSEC-2026-0099 (`cargo audit` failure).

Closes #511

## Test plan

- [x] `bootroot-remote bootstrap` against a TLS-enabled OpenBao validates using only the artifact-embedded CA anchor (`test_remote_bootstrap_https_with_artifact_ca` — full e2e HTTPS test with a real TLS server)
- [x] Wrong CA in the artifact fails with a TLS error, proving no system trust store fallback (`test_remote_bootstrap_https_rejects_wrong_ca`)
- [x] HTTP `openbao_url` (single-host loopback, TLS disabled) still works without regressions
- [x] `schema_version` is 2 in artifacts produced after this PR
- [x] `ca_bundle_pem` field is present and non-empty in generated artifacts
- [x] Two-stage schema version validation: rejects version 0 (below minimum), version 99 (above maximum), and version 3 with unknown fields (would fail serde before version check without two-stage parse)
- [x] `OpenBaoClient::with_pem_trust` unit-level TLS tests: `with_pem_trust_validates_against_artifact_ca` and `with_pem_trust_rejects_unknown_ca` (real TLS handshake)
- [x] `build_http_client_from_pem` unit tests pass (valid PEM, empty PEM, trailing whitespace, with pins)
- [x] `parse_pem_to_root_store` unit tests pass
- [x] Existing integration tests (`bootroot_service`) pass with updated schema assertions
- [x] Remote-bootstrap docs (EN + KO) updated: schema version 2, `ca_bundle_pem` field, version history table

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Security
+
+- Bumped `rustls-webpki` from 0.103.10 to 0.103.12 to address
+  RUSTSEC-2026-0098 and RUSTSEC-2026-0099 (incorrect name-constraint
+  validation for URI and wildcard names).
+
 ### Removed
 
 - Removed `--secret-id-num-uses` from `bootroot service add` and from
@@ -37,10 +43,20 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   obtain `secret_id` before login. Unwrap failures are classified as
   expired (with recovery instructions) or already-unwrapped (flagged as
   a potential security incident).
-- Added `schema_version` field (`u32`, starting at `1`) to the
+- Added `schema_version` field (`u32`, currently `2`) to the
   `RemoteBootstrapArtifact` JSON written by
   `bootroot service add --delivery-mode remote-bootstrap`. Downstream
   parsers should check this field before accessing artifact fields.
+- Added `ca_bundle_pem` field to `RemoteBootstrapArtifact`, embedding
+  the control-plane CA PEM inline. During `bootroot-remote bootstrap`,
+  the PEM is written to `ca_bundle_path` before any OpenBao call and
+  used as the TLS trust anchor for HTTPS `openbao_url` endpoints.
+- Added `OpenBaoClient::with_pem_trust` constructor that anchors TLS
+  verification to an in-memory CA bundle with optional SHA-256 pinning,
+  integrating the same `PinnedCertVerifier` path used by
+  `build_http_client`. `tls::build_http_client_from_pem` now accepts
+  an optional `pins` parameter. `OpenBaoClient::with_client` remains
+  as an escape hatch for callers needing full `reqwest::Client` control.
 - Added remote-bootstrap operator guide (`docs/en/remote-bootstrap.md`,
   `docs/ko/remote-bootstrap.md`) covering transport options (SSH,
   Ansible, cloud-init, systemd-credentials), `secret_id` hygiene,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1900,9 +1900,9 @@ checksum = "f87165f0995f63a9fbeea62b64d10b4d9d8e78ec6d7d51fb2125fda7bb36788f"
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.10"
+version = "0.103.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df33b2b81ac578cabaf06b89b0631153a3f416b0a886e8a7a1707fb51abbd1ef"
+checksum = "8279bb85272c9f10811ae6a6c547ff594d6a7f3c6c6b02ee9726d1d0dcfcdd06"
 dependencies = [
  "aws-lc-rs",
  "ring",
@@ -2181,7 +2181,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.3.4",
+ "getrandom 0.4.2",
  "once_cell",
  "rustix",
  "windows-sys 0.61.2",
--- a/docs/en/remote-bootstrap.md
+++ b/docs/en/remote-bootstrap.md
@@ -111,7 +111,7 @@ scp -p \
   "$REMOTE_USER@$REMOTE_HOST:$REMOTE_BASE/secrets/services/$SERVICE/"
 
 # 3. Validate schema_version before running bootstrap
-if ! jq -e '.schema_version == 1' "$ARTIFACT" > /dev/null; then
+if ! jq -e '.schema_version == 2' "$ARTIFACT" > /dev/null; then
   echo "ERROR: unsupported schema_version in $ARTIFACT" >&2
   exit 1
 fi
@@ -208,10 +208,10 @@ For larger fleets with existing configuration management.
     - name: Validate schema_version
       ansible.builtin.assert:
         that:
-          - artifact.schema_version == 1
+          - artifact.schema_version == 2
         fail_msg: >-
           Unsupported schema_version {{ artifact.schema_version }};
-          this playbook supports version 1 only.
+          this playbook supports version 2 only.
 
     - name: Ensure secrets directory
       ansible.builtin.file:
@@ -407,7 +407,7 @@ The JSON artifact written to
 `secrets/remote-bootstrap/services/<service>/bootstrap.json` follows a
 versioned schema. Automation should check `schema_version` before parsing.
 
-Current version: **1**
+Current version: **2**
 
 | Field | Type | Description | Consumed by |
 | --- | --- | --- | --- |
@@ -419,7 +419,8 @@ Current version: **1**
 | `secret_id_path` | `string` | Path to AppRole `secret_id` file on the remote host | `--secret-id-path` |
 | `eab_file_path` | `string` | Path to EAB credentials JSON file | `--eab-file-path` |
 | `agent_config_path` | `string` | Path to `agent.toml` on the remote host | `--agent-config-path` |
-| `ca_bundle_path` | `string` | Path to CA trust bundle PEM file | `--ca-bundle-path` |
+| `ca_bundle_path` | `string` | Path to CA trust bundle PEM file on the remote host | `--ca-bundle-path` |
+| `ca_bundle_pem` | `string` | Inline PEM content of the control-plane CA trust anchor. Written to `ca_bundle_path` during bootstrap. When `openbao_url` uses HTTPS, this CA is used as the TLS trust anchor instead of the system trust store. Shared primitive — also consumed by the http01 admin client (#514). | Internal (TLS trust) |
 | `openbao_agent_config_path` | `string` | Path to OpenBao Agent config (HCL) | Internal |
 | `openbao_agent_template_path` | `string` | Path to OpenBao Agent template | Internal |
 | `openbao_agent_token_path` | `string` | Path to OpenBao Agent token file | Internal |
@@ -435,6 +436,13 @@ Current version: **1**
 | `wrap_token` | `string?` | Response-wrapped `secret_id` token (omitted when wrapping is disabled via `--no-wrap`). Sensitive — treat as a credential. | `bootroot-remote` unwrap path |
 | `wrap_expires_at` | `string?` | RFC 3339 timestamp when `wrap_token` expires (omitted when wrapping is disabled). | `bootroot-remote` unwrap error classification |
 
+### Version history
+
+| Version | Change |
+| --- | --- |
+| 2 | Added required `ca_bundle_pem` field (inline PEM of the control-plane CA anchor). |
+| 1 | Initial schema. |
+
 ### Versioning rules
 
 - **Breaking change** (field removed, renamed, or type changed): bump
@@ -443,4 +451,5 @@ Current version: **1**
     no bump required. Existing parsers ignore unknown keys.
 - **Consumer contract**: check `schema_version >= 1` and
     `schema_version <= <max supported>` before accessing fields. Fail
-    explicitly on unsupported versions.
+    explicitly on unsupported versions. `bootroot-remote` enforces this
+    check automatically when using `--artifact`.
--- a/docs/ko/remote-bootstrap.md
+++ b/docs/ko/remote-bootstrap.md
@@ -109,7 +109,7 @@ scp -p \
   "$REMOTE_USER@$REMOTE_HOST:$REMOTE_BASE/secrets/services/$SERVICE/"
 
 # 3. 부트스트랩 전 schema_version 검증
-if ! jq -e '.schema_version == 1' "$ARTIFACT" > /dev/null; then
+if ! jq -e '.schema_version == 2' "$ARTIFACT" > /dev/null; then
   echo "ERROR: $ARTIFACT 의 schema_version이 지원되지 않습니다" >&2
   exit 1
 fi
@@ -205,10 +205,10 @@ ExecStart=/usr/local/bin/bootroot-remote bootstrap \
     - name: schema_version 검증
       ansible.builtin.assert:
         that:
-          - artifact.schema_version == 1
+          - artifact.schema_version == 2
         fail_msg: >-
           지원되지 않는 schema_version {{ artifact.schema_version }};
-          이 플레이북은 버전 1만 지원합니다.
+          이 플레이북은 버전 2만 지원합니다.
 
     - name: 시크릿 디렉터리 생성
       ansible.builtin.file:
@@ -396,7 +396,7 @@ runcmd:
 아티팩트는 버전이 지정된 스키마를 따릅니다. 자동화에서는 파싱 전에
 `schema_version`을 확인해야 합니다.
 
-현재 버전: **1**
+현재 버전: **2**
 
 | 필드 | 타입 | 설명 | 사용처 |
 | --- | --- | --- | --- |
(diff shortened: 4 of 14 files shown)


sehkone:
**[Reviewer Round 1]**

Thanks — the artifact-side schema change looks headed in the right direction, but I don’t think this fully closes #511 yet.

1. The OpenBao TLS wiring still sidesteps the existing `TrustSettings` / pinning path instead of actually integrating it. `src/openbao.rs:167` only keeps `OpenBaoClient::new()` plus a raw `with_client()` escape hatch, and `src/bin/bootroot-remote/bootstrap.rs:31` always builds the HTTPS client with `build_http_client_from_pem()`. That helper (`src/tls.rs:90`) only loads a PEM bundle; it has no way to carry `trusted_ca_sha256`, unlike `build_http_client()` in `src/tls.rs:31`. The issue explicitly called out preserving the optional SHA-256 pin flow via the existing trust settings, and this implementation drops that for the initial `sys/wrapping/unwrap` / `auth/approle/login` calls. It also means any future caller (#514) still has to hand-roll its own trust-aware `reqwest::Client` construction instead of using an OpenBao/client-level trust abstraction.

2. The tests don’t actually exercise the new security-critical behavior. The new coverage in `src/tls.rs:510` only proves that a client object can be constructed from PEM bytes; it never completes a TLS handshake. The remote-bootstrap integration coverage is still HTTP-based (`tests/bootroot_remote.rs:220` uses `server.uri()`, and the artifact-mode tests in `tests/e2e_remote_happy_path.rs:729` / `tests/e2e_remote_happy_path.rs:811` also synthesize plain HTTP artifacts). So the core promise from the issue/test plan — “bootroot-remote bootstrap against a TLS-enabled OpenBao validates successfully using only the artifact-embedded CA anchor, without falling back to system roots” — is still untested. As written, a regression that quietly reintroduced system-root verification during the real handshake would not be caught.

Because of those two gaps, I’m not comfortable calling this done yet. I’d strongly prefer to see the trust model wired through a `TrustSettings`-style constructor on `OpenBaoClient` itself, plus at least one end-to-end HTTPS test that proves the bootstrap path succeeds with the artifact CA and would fail without it.

sehkone:

sehkone:
**[Author Round 1]**

### Item 1 — OpenBao TLS wiring sidesteps TrustSettings / pinning path

**Fixed.**

- Added `OpenBaoClient::with_pem_trust(base_url, ca_pem, pins)` constructor that accepts PEM content and optional SHA-256 pins. This is the primary constructor for RN-side bootstrap and any future TLS consumer (#514). The constructor calls through to `build_http_client_from_pem`, which now wires in `PinnedCertVerifier` when pins are non-empty — the same verifier path used by `build_http_client(TrustSettings, ...)`.
- `bootstrap.rs::build_openbao_client` now calls `OpenBaoClient::with_pem_trust(url, pem, &[])` instead of manually building a `reqwest::Client` and passing it to `with_client`. Future callers that need SHA-256 pinning can pass `pins` without touching client construction.
- `with_client` is retained as an escape hatch for callers that need full `reqwest::Client` control but is no longer the documented primary path.
- CHANGELOG updated to reflect the new API.

### Item 2 — Tests don't exercise the TLS handshake

**Fixed.**

Added two end-to-end HTTPS tests in `openbao::tls_integration_tests` that start a real TLS server (using `rcgen`-generated CA + CA-signed server cert + `tokio-rustls` acceptor) and exercise actual TLS handshakes:

- **`with_pem_trust_validates_against_artifact_ca`** — proves that `OpenBaoClient::with_pem_trust` successfully completes a TLS handshake and receives an HTTP 200 when the artifact CA matches the server's issuer.
- **`with_pem_trust_rejects_unknown_ca`** — proves that the same call path fails when the client is given a different CA, confirming no fallback to the system trust store.

Both tests run against a real socket on `127.0.0.1:0` with an ephemeral port — no mocks at the TLS layer.

sehkone:
**[Reviewer Round 2]**

Thanks for the follow-up — I’m happy with the `OpenBaoClient::with_pem_trust()` / pinning refactor itself; that part looks fixed.

I still see two things blocking sign-off:

1. The HTTPS coverage is still below the issue’s bar. The new tests in `src/openbao.rs:1213` prove `OpenBaoClient::with_pem_trust()` can complete a real TLS handshake against a CA-signed server, which is good, but they still stop short of exercising `bootroot-remote bootstrap` itself. The branch that decides whether bootstrap uses the artifact CA lives in `src/bin/bootroot-remote/bootstrap.rs:13`, and the artifact-mode/e2e tests that exercise bootstrap are still plain-HTTP or synthetic v1 artifacts (`tests/e2e_remote_happy_path.rs:729`, `tests/e2e_remote_happy_path.rs:811`). So a regression that rewired unwrap/login back to `OpenBaoClient::new()` (or otherwise skipped `ca_bundle_pem` in the bootstrap path) would still pass. I still think this needs at least one bootstrap-path HTTPS test.

2. The remote-bootstrap docs are still on the old schema, which breaks the “shared primitive” requirement for #514. In `docs/en/remote-bootstrap.md:114` and `docs/en/remote-bootstrap.md:211` the examples still reject anything except `schema_version == 1`; the schema reference still says `Current version: **1**` at `docs/en/remote-bootstrap.md:410` and does not list `ca_bundle_pem` at all starting at `docs/en/remote-bootstrap.md:412`. The same stale content is still present in Korean (`docs/ko/remote-bootstrap.md:112`, `docs/ko/remote-bootstrap.md:208`, `docs/ko/remote-bootstrap.md:399`). Since this PR changes the artifact contract to version 2 and introduces the loader semantics around `ca_bundle_pem`, those docs need to be updated before downstream consumers can rely on the new schema.

sehkone:

sehkone:
**[Author Round 2]**

### Item 1 — Bootstrap-path HTTPS coverage

**Fixed.**

Added two end-to-end HTTPS tests in `tests/e2e_remote_happy_path.rs` that exercise the full `bootroot-remote bootstrap --artifact` code path over a real TLS connection:

- **`test_remote_bootstrap_https_with_artifact_ca`** — generates a test CA + CA-signed server cert, starts a TLS mock server that handles all required OpenBao endpoints (AppRole login, KV reads for secret_id/eab/hmac/trust), creates a v2 artifact with `openbao_url: "https://localhost:<port>"` and `ca_bundle_pem` set to the test CA PEM, then runs `bootroot-remote bootstrap --artifact` as a child process. Asserts the binary exits successfully and that secrets were written through the TLS-protected path (verifies `secret_id` was overwritten with the mock server's response).

- **`test_remote_bootstrap_https_rejects_wrong_ca`** — same setup but embeds a *different* CA in `ca_bundle_pem`. Asserts the binary fails, confirming the bootstrap path does not fall back to the system trust store.

These tests use `tokio::process::Command` (async) instead of `std::process::Command` (blocking) because the TLS mock server is a `tokio::spawn` task on the same runtime — a blocking `Command::output()` would prevent the server from accepting connections.

Combined with the existing unit-level TLS tests in `src/openbao.rs` (which prove `with_pem_trust` completes real TLS handshakes), a regression that rewired the bootstrap path back to `OpenBaoClient::new()` or skipped `ca_bundle_pem` would now be caught at both layers.

### Item 2 — Remote-bootstrap docs updated for schema v2

**Fixed.**

Updated both `docs/en/remote-bootstrap.md` and `docs/ko/remote-bootstrap.md`:

- **Schema version**: changed "Current version: **1**" to "Current version: **2**".
- **Schema table**: added `ca_bundle_pem` field with description, noting it is a shared primitive also consumed by #514.
- **Validation examples**: updated the SSH shell script (`jq -e '.schema_version == 2'`) and Ansible playbook (`artifact.schema_version == 2`) in both languages.
- **Version history**: added a version history table documenting v1 (initial) and v2 (added `ca_bundle_pem`).
- **Consumer contract note**: added a note that `bootroot-remote` enforces the `schema_version` check automatically when using `--artifact`.

sehkone:
**[Author Round 2]**

### Item 1 — Bootstrap-path HTTPS test coverage

**Fixed** (commit 7daadef).

Added two end-to-end HTTPS tests that exercise the full `bootroot-remote bootstrap --artifact` code path over a real TLS connection:

- **`test_remote_bootstrap_https_with_artifact_ca`** — generates a CA with `rcgen`, signs a localhost server cert, starts a `tokio-rustls` TLS server with canned OpenBao responses, writes a schema v2 artifact with the correct `ca_bundle_pem`, then runs the actual `bootroot-remote` binary via `tokio::process::Command`. Asserts exit code 0 and verifies the written `secret_id` matches the TLS mock's response.

- **`test_remote_bootstrap_https_rejects_wrong_ca`** — same setup but embeds a *different* CA in the artifact. Asserts exit code != 0 and a TLS/certificate-related error in stderr. This proves the bootstrap path does not fall back to the system trust store.

Both tests run against a real socket on `127.0.0.1:0` with an ephemeral port. A regression that rewired the bootstrap path back to `OpenBaoClient::new()` (skipping `ca_bundle_pem`) would fail the first test, and any system-trust-store fallback would cause the second test to pass incorrectly.

### Item 2 — Remote-bootstrap docs still on old schema

**Fixed** (commit 7daadef).

Updated both `docs/en/remote-bootstrap.md` and `docs/ko/remote-bootstrap.md`:

- Shell validation example now checks `schema_version == 2` (was `== 1`).
- Ansible validation task asserts `schema_version == 2` with matching `fail_msg`.
- Schema reference header reads `Current version: **2**` (was `**1**`).
- `ca_bundle_pem` field added to the schema table with description noting it is a shared primitive reused by #514.
- Version history table added documenting what changed between v1 and v2.

sehkone:
**[Reviewer Round 3]**

Thanks — I rechecked the Round 2 items, and the HTTPS bootstrap coverage plus the docs updates both look good now.

I do still see one schema-handling gap in `bootroot-remote`:

1. `resolve_bootstrap_args()` still deserializes the *full* `BootstrapArtifact` before it rejects unsupported schema versions, and the check only enforces an upper bound (`src/bin/bootroot-remote/main.rs:352`, `src/bin/bootroot-remote/main.rs:354`). That leaves two cases outside the contract described in the issue/docs:
   - `schema_version: 0` is currently accepted even though the documented supported range starts at 1.
   - More importantly, a future breaking v3 artifact can fail during `serde_json::from_str()` before we ever look at `schema_version`, so the user gets a generic parse error instead of the explicit “unsupported schema_version” rejection the issue called for.

   I think this wants a two-stage parse: first read just `schema_version` (or parse into `serde_json::Value` / a tiny header struct), reject anything outside `1..=MAX_SUPPORTED_SCHEMA_VERSION`, and only then deserialize the full typed artifact for supported versions. The current test at `src/bin/bootroot-remote/main.rs:736` only covers a schema-99 artifact whose shape still matches today’s struct, so it would not catch this regression.

sehkone:

sehkone:
**[Author Round 3]**

### Item 1 — Two-stage schema version parsing

**Fixed** (commit b013e9b).

Implemented the two-stage parse as suggested:

1. A minimal `ArtifactHeader` struct (just `schema_version: u32`) is deserialized first.
2. The version is checked against `MIN_SUPPORTED_SCHEMA_VERSION..=MAX_SUPPORTED_SCHEMA_VERSION` (i.e., `1..=2`), rejecting both version 0 and any future version that exceeds the upper bound.
3. Only after the version passes does full `BootstrapArtifact` deserialization happen.

This means a future v3 artifact with completely different fields now gets the explicit "Artifact schema_version 3 is not supported (supported range: 1..=2). Upgrade bootroot-remote to a newer version." message instead of a confusing serde missing-field error.

Tests added:
- `resolve_bootstrap_args_rejects_future_schema_version` — schema 99 with today's field shape (renamed from the previous test).
- `resolve_bootstrap_args_rejects_schema_version_zero` — schema 0 is now correctly rejected.
- `resolve_bootstrap_args_rejects_future_schema_with_unknown_fields` — schema 3 with only `completely_new_field`, proving that the header-stage check fires before serde fails on missing fields.

Also extracted `default_bootstrap_args` and `assert_rejects_schema_version` test helpers to reduce duplication across the three test cases.

sehkone:
**[Reviewer Round 4]**

I rechecked the Round 3 schema-handling concern, and that item looks fixed now.

- `bootroot-remote` now does the two-stage parse before touching the full artifact shape: it reads a minimal header, rejects anything outside `1..=2`, and only then deserializes `BootstrapArtifact` in `src/bin/bootroot-remote/main.rs:279` and `src/bin/bootroot-remote/main.rs:357`.
- The new tests cover the cases that were missing before: version `99`, version `0`, and a future schema with completely different fields, so a serde shape error no longer masks the intended unsupported-version rejection in `src/bin/bootroot-remote/main.rs:789`.
- The HTTPS bootstrap coverage and schema/docs updates from the previous round are still present where expected in `tests/e2e_remote_happy_path.rs:1010` and `docs/en/remote-bootstrap.md:410`.

I didn’t find any further blocking issues in this round.

sehkone:

Issue #515, opened 2026-04-15
Auto-provision and renew http01 admin TLS server certificate via step-ca


sehkone:
Part of #507

Blocked by #512 and #513.

## Background

#512 introduces a generalized `infra_certs` model in `StateFile` and extends the rotation pipeline to honor infrastructure-level certificates, implementing the flow for OpenBao. #513 introduces HTTPS listener support in the `bootroot-http01-responder` itself (TLS config fields, listener selection, SIGHUP-based cert reload).

This PR is the final piece for the http01 admin API: **automated** certificate provisioning and rotation, built on top of those two. After #513 operators can manually configure cert/key files and the responder will serve HTTPS; this PR removes that manual step by issuing the cert from the local step-ca during `bootroot init` (see "Issuance timing" below) and renewing it through the shared `infra_certs` pipeline.

## Issuance timing

`infra install --http01-admin-bind <addr>` only generates the compose override that changes the port binding and records bind intent in `StateFile`. It does **not** issue the certificate: step-ca is not initialized at that point (`ensure_step_ca_initialized` runs inside `bootroot init`, not `infra install`). The responder therefore starts bound to `127.0.0.1:8080` without TLS after `infra install` — the override is not applied yet and no non-loopback port is published. TLS is activated during the subsequent `bootroot init` run.

The full sequence:

1. `infra install --http01-admin-bind <IP>` — records bind intent in `StateFile`, generates the compose override file. The override is **not** applied to the initial `docker compose up`: the responder starts on `127.0.0.1:8080` (no TLS). No non-loopback port is published at this point.
2. `bootroot init` — after `ensure_step_ca_initialized` completes, calls step-ca to issue the admin API server cert, writes `responder.toml` with TLS enabled, and applies the compose override **in the same restart**. The admin API comes up on `<IP>:8080` with TLS active; the loopback-only window is never exposed to the network.
3. The admin API is now reachable at `https://<IP>:8080`.

This mirrors the sequencing established in #512 for OpenBao.

## Scope

This issue is deliberately narrow because #513 owns the listener-side code and #512 owns the rotation machinery. Concretely, this PR:

1. Issues the admin API server certificate via step-ca during `bootroot init` (after `ensure_step_ca_initialized`).
2. Registers the certificate in `StateFile::infra_certs` (schema from #512) with a reload strategy that invokes the responder's TLS-reload path (wired by #513).
3. On renewal, replaces the cert files and triggers the responder reload through the path #513 established.

[...]

Issue #513, opened 2026-04-15
Relax http01 admin API guardrail with mandatory TLS enforcement


sehkone:
Part of #507

Blocked by #514.

## Implementation prerequisite

Before writing the full PR, spike `poem`'s TLS listener integration to confirm whether `poem::listener::RustlsListener` / `RustlsConfig` supports injecting or swapping a shared `Arc<dyn ResolvesServerCert>` resolver at runtime. The current `server.rs` spawns the admin listener as an independent `JoinHandle` task; the SIGHUP handler (`reload_settings`) only swaps in-memory settings and the HMAC signer and never touches the listener. TLS-aware cert reload is therefore entirely new code.

The chosen reload design determines the reload-strategy discriminator stored in `StateFile::infra_certs` (schema owned by #512), which #515 must also reference. Resolve this before implementation so that #512 can commit the discriminator shape and #515 does not need to revise it.

- If `poem` exposes an API for dynamic cert swapping: implement the `Arc<dyn ResolvesServerCert>` resolver pattern as described in "Add HTTPS listener support" below.
- If it does not: the fallback (listener teardown + respawn on SIGHUP) is the only option. Document this in the PR description with explicit rationale before review begins.

**Share the spike result with the #512 implementer before either PR's schema is written.** The `InfraCertEntry.reload_strategy` discriminator in #512 must accommodate the approach chosen here; committing that field before the spike resolves risks a schema revision.

## Background

`bootroot-http01` publishes its admin API on port 8080 bound to `127.0.0.1` only (see `docker-compose.yml`), enforced by the `GUARDED_SERVICES` list in `src/commands/guardrails.rs`. In multi-host deployments the RN agent needs to register ACME HTTP-01 challenge tokens with this admin API (`agent_responder_url: http://responder.internal:8080` per `docs/en/remote-bootstrap.md`), so the current binding blocks remote bootstrap.

Unlike the HTTP-01 challenge port (80), which is constrained by RFC 8555 §8.3 to be plain HTTP, the admin API is a bootroot-specific control-plane interface with no protocol-level restriction — TLS is both possible and required when the port is exposed to the network.

[...]

PR #526, opened 2026-04-16
Pin http01 admin client TLS to bootstrap-artifact CA (#514)

### LINKED ISSUE #514: Pin http01 admin client TLS to bootstrap-artifact CA anchor

sehkone:
## Summary

- Introduces `ResponderTrust` struct in `src/acme/responder_client.rs` to bundle CA PEM and SHA-256 certificate pins, replacing separate trust parameters.
- Extends `register_http01_token_with` to accept `Option<&ResponderTrust>` and build a trust-pinned `reqwest` client for HTTPS responder URLs via `build_client_config_from_pem` from `src/tls.rs`, including SHA-256 pin enforcement.
- Defers CA bundle disk I/O behind the HTTPS scheme check in `register_http01_token`, so plain `http://` callers are never affected by missing or unreadable CA bundles.
- Wires the CN-side init verification (`verify_responder` in `responder_setup.rs`) to the on-disk step-ca root, and the RN-side ACME flow to `settings.trust` (populated from the bootstrap artifact), threading both the CA PEM and SHA-256 pins.
- Rejects HTTPS responder URLs when no CA trust anchor is available, preventing silent fallback to the system trust store.
- Plain HTTP URLs continue to work unchanged — the trust parameter is accepted but unused until #513 flips the scheme.

Closes #514

## Test plan

- [x] `bootroot init` with default loopback (plain HTTP) responder works unchanged after the signature change
- [x] `http://` with a broken (nonexistent) `ca_bundle_path` succeeds — trust is ignored for plain HTTP
- [x] HTTPS responder URL with no trust source errors out with a clear message
- [x] HTTPS with correct CA pin succeeds against a real TLS responder (rcgen CA + tokio-rustls server)
- [x] HTTPS with wrong CA pin is rejected at the TLS handshake level against the same real TLS responder
- [x] HMAC signatures continue to be validated server-side for both call paths
- [x] Modifying the system trust store on the RN does not affect admin API validation
- [ ] After #513 lands: RN agent registers a token against a TLS-enabled http01 admin API using the artifact-embedded CA anchor
- [ ] After #513 + #515 land: CN-side init check succeeds against a TLS-enabled admin API using the local step-ca root


sehkone:
**[Reviewer Round 1]**

I found two blocking gaps in the current approach.

1. `src/acme/responder_client.rs:35` eagerly reads `trust.ca_bundle_path` before the code knows whether the responder URL is actually `https://`. That means the supposedly dormant `http://` path can now fail on a missing or unreadable CA bundle, even though the issue explicitly calls out that plain-HTTP callers should ignore trust input and behave unchanged until `resolve_responder_url` flips to HTTPS in #513. I think the CA read needs to move behind the HTTPS branch (for example into `build_responder_client`) so `http://` truly stays inert.

2. The new responder client drops configured SHA-256 pin enforcement. `bootroot-remote` still writes `trusted_ca_sha256` into agent config at `src/bin/bootroot-remote/agent_config.rs:69`, and the existing trust path in `src/tls.rs:31` treats those pins as part of the normal trust model, but `src/acme/responder_client.rs:131` hard-codes `&[]` when building the TLS config. That makes the admin API client weaker than `OpenBaoClient` and does not really match the issue’s “same trust mechanism” / `TrustSettings`-style requirement. I think this needs to thread the full trust inputs (bundle + pins) through instead of collapsing them to raw PEM only.

I also do not see a test that would fail on either of these cases (`http://` with a broken trust path, or `https://` with configured pins), so both regressions would currently slip through.

sehkone:

sehkone:
**[Author Round 1]**

Both items are valid — thank you for catching them.

### 1. Eager CA bundle read on http:// path — Fixed

`register_http01_token` now guards `read_ca_pem_from_trust` behind `url.starts_with("https://")`. For plain `http://` URLs the trust settings are never touched, so a missing or unreadable CA bundle no longer causes a spurious failure.

Added test `test_register_http01_token_http_ignores_broken_trust_path`: sets `ca_bundle_path` to a nonexistent path and confirms the `http://` call succeeds.

### 2. Dropped SHA-256 pin enforcement — Fixed

Introduced a `ResponderTrust` struct that bundles `ca_pem` and `ca_pins` together. `register_http01_token_with` now accepts `Option<&ResponderTrust>` and threads both values to `build_client_config_from_pem`, matching the trust model used by `OpenBaoClient`. This also resolved the too-many-arguments clippy lint that would have appeared with two separate parameters.

The Settings-based wrapper threads `settings.trust.trusted_ca_sha256` through the struct. The CN-side init path passes `&[]` for pins (pins aren't available during init — the trust config hasn't been written yet).

Added test `test_register_http01_token_with_forwards_pins`: passes a `ResponderTrust` with pins to an HTTPS URL with invalid PEM content and confirms the error comes from PEM parsing (proving the pins were passed through rather than silently dropped).

sehkone:
**[Reviewer Round 2]**

Thanks — I re-checked the Round 1 items.

- The eager trust read is fixed. `register_http01_token` now guards the CA read behind the HTTPS scheme check, and `test_register_http01_token_http_ignores_broken_trust_path` would catch the old regression.
- The code-level pin threading is also fixed. `build_responder_client` now passes `trust.ca_pins` through to `build_client_config_from_pem` at `src/acme/responder_client.rs:152`.

One thing still feels unresolved:

- The new pin test at `src/acme/responder_client.rs:306` does not actually verify that pins are forwarded. With `ca_pem: "not-a-real-pem"`, both the current code and the old `build_client_config_from_pem(trust.ca_pem, &[])` implementation fail in the same PEM-parse path, so this test would still pass if the pins were silently dropped again. I think this still needs a test where the observable behavior changes based on the pins themselves (for example, a tiny TLS server/helper similar to the OpenBao TLS tests, where the correct pin succeeds and a wrong pin fails).

sehkone:

sehkone:
**[Author Round 2]**

### Pin test does not prove pins are forwarded — Fixed

Valid catch. The old test (`test_register_http01_token_with_forwards_pins`) used `ca_pem: "not-a-real-pem"`, so both with-pins and without-pins code paths hit the same PEM-parse failure. It would have passed even if pins were silently dropped.

Replaced it with two tests that spin up a real TLS responder (rcgen CA-signed `localhost` cert + tokio-rustls `TlsAcceptor`, following the same `TestCa`/`start_openbao_tls_mock` pattern from `e2e_remote_happy_path.rs`):

- **`test_responder_tls_correct_pin_succeeds`** — passes the CA's SHA-256 fingerprint as the pin; the TLS handshake completes and `register_http01_token_with` returns `Ok(())`.
- **`test_responder_tls_wrong_pin_rejected`** — passes `"00".repeat(32)` as the pin; the TLS handshake is rejected by `PinnedCertVerifier` and the call fails with a connection error.

If pins were silently dropped (i.e. `build_client_config_from_pem` were called with `&[]`), the wrong-pin test would succeed instead of failing, so these two tests together prove that pins are actually forwarded through the responder client path.

sehkone:
**[Reviewer Round 3]**

Re-checked the Round 2 item and reviewed the updated diff against #514.

- The pin-forwarding concern is resolved. `build_responder_client` now passes `trust.ca_pins` into `build_client_config_from_pem` in `src/acme/responder_client.rs:152`, and the new real-TLS tests at `src/acme/responder_client.rs:399` and `src/acme/responder_client.rs:423` would fail if pins were dropped again.
- The dormant plain-HTTP behavior still looks correct. `register_http01_token` defers CA bundle reads until the HTTPS branch in `src/acme/responder_client.rs:46`, and `test_register_http01_token_http_ignores_broken_trust_path` at `src/acme/responder_client.rs:286` covers the old regression.
- The CN-side init check is now wired to local CA material only for HTTPS in `src/commands/init/steps/responder_setup.rs:160`, which matches the issue requirements while preserving the existing loopback `http://` path.

I did not find any new blocking issues in this revision.

sehkone:

PR #527, opened 2026-04-16
Auto-provision OpenBao TLS cert via step-ca (#512)

### LINKED ISSUE #512: Auto-provision and renew OpenBao TLS server certificate via step-ca
Part of #507

## Background

Once #508 enforces TLS when OpenBao is bound to a non-loopback address, and the client-side trust bootstrap lands in #511, the remaining gap is the server-side certificate lifecycle. Without automation operators would need to manually issue the server certificate from step-ca, place it correctly, keep permissions right, and re-issue before expiry — the exact manual toil bootroot aims to remove elsewhere.

Today `openbao/openbao.hcl` hard-codes `tls_disable = 1` on both listeners, and `StateFile` in `src/state.rs` models only service-owned identities (`services: BTreeMap<String, ServiceEntry>`). The rotation pass in `src/commands/rotate/ca.rs` iterates `state.services` and therefore has no awareness of infrastructure-level certificates such as the OpenBao server cert.

## Shared-infrastructure note

The `StateFile::infra_certs` map and the rotation-loop extension introduced by this PR are **shared primitives** — they will be reused by other bootroot-managed infrastructure certificates, specifically the http01 admin API server certificate tracked in #515.

Design choices that must be made with both consumers in mind:

- **Map shape**: design `infra_certs` as a `BTreeMap<String, InfraCertEntry>` keyed by a stable service identifier (e.g. `"openbao"`, `"bootroot-http01"`), not a flat OpenBao-specific struct.
- **Entry fields**: include enough to cover both services — cert/key paths, `renew_before`, issuance metadata, and a discriminator for the reload mechanism (e.g. SIGHUP vs restart vs responder-specific endpoint). The http01 responder reloads HMAC on SIGHUP today; OpenBao may use a different path. The entry should carry the reload strategy.
- **`reload_strategy` discriminator — finalize after the #513 spike**: the #513 "Implementation prerequisite" requires a spike to determine whether the responder admin listener can reload certs in-place (shared rustls resolver) or requires listener teardown. The discriminator shape must accommodate both outcomes without a schema revision. **Do not commit `InfraCertEntry.reload_strategy` to a concrete shape before the #513 spike result is known.** Share the spike finding with the #512 implementer before either PR's schema is written.
- **Rotation dispatch**: the rotation loop must iterate the map and dispatch re-issuance by entry key and reload by the per-entry `ReloadStrategy` discriminator. #515 must be able to add its entry and reload handler without touching the loop itself beyond registering the handler.
- **Config emission**: the `openbao.hcl` listener TLS config emission pattern used here should be recognizably parallel to what #515 will do for `responder.toml` — same cert/key path conventions, same permissions.

This PR populates `infra_certs` with the OpenBao entry only. #515 adds the http01 admin entry without changing the schema.

## Out of scope — artifact CA anchor population

Populating `ca_bundle_pem` in remote-bootstrap artifacts is **owned by #511**, not this PR. #511 reads the step-ca root (which exists under the same paths this PR uses) and embeds it into generated artifacts. This PR does not touch the artifact code path or duplicate that logic.

## Issuance timing

`infra install --openbao-bind <addr>` only generates the compose override that changes the port binding. It does **not** issue the certificate: step-ca is not initialized at that point (`ensure_step_ca_initialized` runs inside `bootroot init`, not `infra install`). OpenBao therefore starts sealed and without TLS after `infra install`. TLS is activated during the subsequent `bootroot init` run.

The full sequence:

1. `infra install --openbao-bind <IP>` — records bind intent in `StateFile`, generates the compose override file. The override is **not** applied to the initial `docker compose up`: OpenBao starts on `127.0.0.1:8200` (sealed, TLS disabled). No non-loopback port is published at this point.
2. `bootroot init` — after `ensure_step_ca_initialized` completes, calls step-ca to issue the OpenBao server cert, writes `openbao.hcl` with TLS enabled, and applies the compose override **in the same restart**. OpenBao comes up on `<IP>:8200` with TLS active; the loopback-only window is never exposed to the network.
3. OpenBao is now reachable at `https://<IP>:8200`.

Deferring the override to the same restart as TLS activation is what prevents the exposure window: the non-loopback port is never published while OpenBao is running without TLS.

**OpenBao client URL transition**: `run_init_inner` builds an `OpenBaoClient` at startup using the initial `http://` URL. All OpenBao client calls complete before the TLS restart. After the restart, the `https://` URL is persisted to `StateFile` for use by subsequent commands (`rotate`, `service add`, etc.). No client rebuild within the same init run is needed because no post-restart OpenBao API calls remain.

### Issue the certificate during `bootroot init`

- After `ensure_step_ca_initialized` completes, call the local step-ca to issue a server certificate with appropriate SANs (`openbao.internal`, `localhost`, `bootroot-openbao`, and the configured bind IP).
- Write the certificate and key to `openbao/tls/server.crt` and `server.key` (relative to the compose directory) with `0600` permissions.
- Update `openbao.hcl` to enable TLS using those paths and restart OpenBao. The compose file mounts the `openbao/` directory into the container.
- Persist the `https://` URL to `StateFile` after the restart. All OpenBao client calls in the init flow are structured to complete before the TLS restart, so no client rebuild is required.

### Model the server certificate in state

- Add the `infra_certs` map to `StateFile` — distinct from `services`, since this tracks bootroot-managed infrastructure certs rather than service identities.
- Include paths, `renew_before`, issuance metadata, and a reload-strategy discriminator.

### Extend the renewal pipeline

- Extend (or parallel) the rotation machinery so it honors entries in `infra_certs`.
- On renewal, issue a new certificate via `step certificate create --force`, replace the files, and invoke the entry's reload strategy (container restart for OpenBao). Failures surface through the standard rotation error path.

## Non-goals

- Changing rotation scheduling or policy — reuse existing mechanics.
- mTLS client certificates — deferred.
- Supporting PKIs other than the local step-ca.
- Client-side trust wiring in `OpenBaoClient` — tracked in #511.
- Populating `ca_bundle_pem` in artifacts — owned by #511 (see "Out of scope" above).
- http01 admin server cert entry — tracked in #515, which reuses the map and dispatch designed here.

## Test plan

- `bootroot infra install --openbao-bind <IP>` records bind intent in `StateFile` and generates the compose override; OpenBao starts on `127.0.0.1:8200` — the override is not yet applied, so no non-loopback port is published.
- `bootroot init` on top of the above produces a TLS-enabled OpenBao at `https://<IP>:8200` without any manual certificate steps.
- All `bootroot init` steps that follow the OpenBao restart use `https://` and succeed — verifying the client URL transition is handled correctly.
- Simulated renewal successfully replaces the certificate and reloads OpenBao without operator action.
- End-to-end with #511 landed: an RN can run `bootroot-remote bootstrap` against the new CN and validate TLS purely via the artifact-embedded CA anchor.
- Existing single-host (loopback, TLS disabled) flow still works.
- `infra_certs` map and rotation dispatch are documented well enough that #515 can add the http01 entry without touching the loop logic beyond registering its reload handler.

sehkone:
## Summary

- Issues an OpenBao TLS server certificate from the local step-ca during `bootroot init` (after `ensure_step_ca_initialized`), writes TLS-enabled `openbao.hcl`, and applies the compose override in the same restart — the non-loopback port is never exposed without TLS.
- Adds a generic `infra_certs: BTreeMap<String, InfraCertEntry>` to `StateFile` with cert/key paths, SANs, `renew_before`, and a `ReloadStrategy` discriminator. Designed as a shared primitive so #515 can add the http01 admin entry without schema changes.
- Extends the rotation pipeline with `bootroot rotate infra-cert`, which iterates all `infra_certs` entries, re-issues each certificate, and dispatches reload based on the per-entry strategy (currently `ContainerRestart` for OpenBao). The subcommand bypasses the OpenBao auth/client bootstrap entirely — it operates on local files and Docker only, so it can fix an expired or broken server cert without a circular dependency on OpenBao availability.
- `RotateContext` carries the full `state_file` path so `rotate infra-cert --state-file custom.json` persists the updated `issued_at` back to the correct file (not a hardcoded `state.json`).
- When a wildcard bind is paired with `--openbao-advertise-addr`, the advertise IP is included in the certificate SANs so that remote nodes connecting via the advertised endpoint pass hostname verification.
- `infra install` (both with and without `--openbao-bind`) restores `openbao.hcl` to the plaintext `tls_disable = 1` form, preventing a stale TLS config from breaking the next `bootroot init` or the single-host flow. The non-loopback reinstall path also clears the stale infra-cert entry so a fresh certificate is issued by the next init.
- `InitRollback` now covers TLS activation: if any step after the HCL rewrite fails, rollback restores `openbao.hcl` to its pre-TLS content, removes TLS cert/key artifacts, and recreates the OpenBao container from the base compose config (via `docker compose up -d`, not `restart`) to remove the non-loopback port mapping introduced by the override. This ensures rollback never leaves OpenBao reachable on plaintext HTTP at a non-loopback address.

Closes #512

## Test plan

- [x] `bootroot infra install --openbao-bind <IP>` records bind intent in `StateFile` and generates the compose override; OpenBao starts on `127.0.0.1:8200` without the non-loopback port published
- [x] `bootroot init` on top of the above produces a TLS-enabled OpenBao at `https://<IP>:8200` without manual certificate steps
- [x] All `bootroot init` steps after the OpenBao restart use `https://` and succeed (client URL transition)
- [x] `bootroot rotate infra-cert` successfully replaces the certificate and restarts the OpenBao container
- [x] `run_rotate(InfraCert)` completes end-to-end without contacting OpenBao (regression test with unreachable URL and a registered infra-cert entry)
- [x] `rotate infra-cert --state-file custom.json` persists `issued_at` to the custom-named file; no sibling `state.json` is created
- [x] Existing single-host (loopback, TLS disabled) flow still works — no regression
- [x] `infra_certs` map is backward-compatible: state files without the field deserialize with an empty map; empty maps are omitted from serialized JSON
- [x] `ReloadStrategy` and `InfraCertEntry` round-trip through JSON correctly (unit tests)
- [x] #515 can add its http01 entry without modifying the rotation loop or the `InfraCertEntry` schema
- [x] Wildcard bind with `--openbao-advertise-addr` includes the advertise IP in the certificate SANs (regression test)
- [x] Loopback reinstall (`infra install` without `--openbao-bind`) restores `openbao.hcl` to plaintext `tls_disable = 1` (regression test)
- [x] Non-loopback reinstall (`infra install --openbao-bind` over previous TLS setup) restores `openbao.hcl` to plaintext and clears the stale infra-cert entry (regression test)
- [x] Rollback after TLS activation failure restores HCL to plaintext and removes TLS cert/key artifacts (regression tests)
- [x] Rollback removes HCL file entirely when no HCL existed before init (regression test)
- [x] Rollback without TLS fields is a no-op — does not touch unrelated files (regression test)
- [x] Rollback uses `docker compose up -d` (not `restart`) to recreate the container from the base config, ensuring the non-loopback port override is removed (regression test)


sehkone:
**[Reviewer Round 1]**
Thanks — the overall direction makes sense, but I see three blocking issues before this is ready:

- `build_openbao_tls_sans()` handles wildcard IPv6 incorrectly. For `[::]:8200`, `src/commands/init/steps/openbao_tls.rs:158` only adds `127.0.0.1`, while `client_url_from_bind_addr()` persists `https://[::1]:8200` in `src/commands/guardrails.rs:219`. That means later local commands will talk to `::1` with a cert that does not contain `::1` in SANs, so hostname verification will fail on IPv6 wildcard installs. The new test at `src/commands/init/steps/openbao_tls.rs:259` currently codifies the wrong behavior by asserting `127.0.0.1` instead of `::1`.
- The new infra-cert rotation loop is not actually extensible in the way the issue requires. `src/commands/rotate/infra_cert.rs:41` hard-codes `OPENBAO_INFRA_CERT_KEY` and skips every other entry with `Unknown infra cert key: ...; skipping`. That means `#515` still has to modify the loop body instead of just registering a handler, and unknown entries do not fail through the normal rotation error path. I think this needs a real dispatch/registration layer rather than an `if name == ...` special case.
- Stale `infra_certs` entries survive when the install is moved back to loopback/plaintext. `clear_openbao_bind_intent_to()` clears `openbao_bind_addr` and resets `openbao_url`, but it does not remove the OpenBao infra-cert entry from state (`src/commands/infra.rs:599`). Then `write_state_file_to()` preserves `existing_infra_certs` unconditionally on the next init (`src/commands/init/steps/orchestrator.rs:581`). So a reinstall without `--openbao-bind` can still leave `state.json` advertising an `openbao` infra cert, and `bootroot rotate infra-cert` will keep renewing/restarting an unused certificate. That looks inconsistent with the “existing single-host (loopback, TLS disabled) flow still works” requirement and needs a regression test.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the thorough review. All three items were valid — here's what I did:

**1. IPv6 wildcard SAN mismatch — Fixed**

`build_openbao_tls_sans` now adds `::1` (IPv6 loopback) when the bind address is `[::]` or `[::0]`, matching what `client_url_from_bind_addr` persists as the client URL (`https://[::1]:8200`). I also add `127.0.0.1` in the IPv6 wildcard case for dual-stack safety — a host binding `[::]` accepts IPv4 connections too, so the cert should cover both loopbacks.

The test at `openbao_tls.rs:259` now asserts both `::1` and `127.0.0.1` are present. Added a second test for the `[::0]` form.

**2. Infra-cert rotation dispatch — Fixed**

Extracted a `dispatch_reissue(name, ...)` function with a `match` on the entry key. Unknown keys now `bail!` through the standard rotation error path instead of being silently skipped with `continue`. #515 adds a match arm to `dispatch_reissue` — the loop body itself does not change.

**3. Stale `infra_certs` on loopback reinstall — Fixed**

`clear_openbao_bind_intent_to` now calls `state.infra_certs.remove(OPENBAO_INFRA_CERT_KEY)` alongside clearing `openbao_bind_addr`. Added a regression test (`clear_bind_intent_removes_openbao_infra_cert`) that seeds an infra cert entry, calls the clear function, and asserts it's gone.

sehkone:
**[Reviewer Round 2]**

Thanks — the three round-1 issues look fixed in the updated diff.

I do still see one blocking problem in the new renewal path:

- `bootroot rotate infra-cert` still goes through the generic OpenBao auth/client setup before dispatch. In `src/commands/rotate.rs:148`-`src/commands/rotate.rs:163`, `run_rotate()` always resolves runtime auth, authenticates an `OpenBaoClient`, and runs `health_check()` before it matches on the subcommand. But the new `src/commands/rotate/infra_cert.rs:18`-`src/commands/rotate/infra_cert.rs:69` path never uses that client at all; it just reissues the cert from local step-ca material and restarts the container. This creates a circular dependency for the very thing this PR is trying to automate: if the OpenBao server cert is expired, mistrusted, or otherwise broken, `rotate infra-cert` can fail during client auth/health-check before it ever reaches the reissue logic. It also unnecessarily requires runtime auth for a file+Docker operation. I think `RotateCommand::InfraCert` needs to bypass the OpenBao auth/client bootstrap entirely, and there should be a regression test proving that this subcommand can run without contacting OpenBao.

sehkone:

sehkone:
**[Author Round 2]**

**1. `rotate infra-cert` circular dependency on OpenBao auth — Fixed**

Valid catch. `run_rotate` now handles `RotateCommand::InfraCert` *before* the OpenBao client bootstrap (auth + health check). The subcommand returns early after `rotate_infra_certs`, so it never creates an `OpenBaoClient`, never authenticates, and never calls `health_check`.

This breaks the circular dependency: if the OpenBao server cert is expired or broken, `bootroot rotate infra-cert` can still reissue it and restart the container without needing to talk to OpenBao first.

Added a regression test (`rotate_infra_certs_does_not_require_openbao`) that constructs a `RotateContext` with an unreachable `openbao_url` (`https://192.0.2.1:1` — RFC 5737 TEST-NET) and verifies the function completes successfully. If it ever regresses to require an OpenBao connection, this test will fail.

The `InfraCert` arm in the `match` is now `unreachable!` since the early return handles it.

Updated the PR description to reflect the OpenBao-independence guarantee and added the regression test to the test plan.

sehkone:
**[Reviewer Round 3]**

Thanks — the round-2 blocker is fixed: `RotateCommand::InfraCert` now returns before the OpenBao auth/client bootstrap, so the circular dependency is gone.

I did find two remaining issues in the updated diff:

- `src/commands/rotate.rs:120` still accepts `--state-file`, but `rotate_infra_certs()` reconstructs the save target as `ctx.state_dir.join("state.json")` in `src/commands/rotate/infra_cert.rs:62`. That means `bootroot rotate infra-cert --state-file custom.json` loads `custom.json`, but then either never persists the `issued_at` update (when no sibling `state.json` exists) or writes the updated state into the wrong file (when one does exist). I think the full state-file path needs to be carried through `RotateContext` and reused here, with a regression test for a non-default filename.
- The new regression test does not actually cover the round-2 bug. `rotate_infra_certs_does_not_require_openbao()` in `src/commands/rotate/infra_cert.rs:117` calls `rotate_infra_certs()` directly rather than `run_rotate()`, and the test state sets `infra_certs: BTreeMap::new()` at `src/commands/rotate/infra_cert.rs:134`, so it returns at line 23 before any renewal path runs. The original problem was in `run_rotate()` doing OpenBao auth/health-check before dispatch, so this test would still pass on the broken version. I’d like to see coverage that exercises `run_rotate(... RotateCommand::InfraCert ...)` with an unreachable OpenBao URL and at least one registered infra-cert entry.

sehkone:

sehkone:
**[Author Round 3]**

Both items were valid — here's what I did:

**1. State-file path not propagated through `RotateContext` — Fixed**

Added a `state_file: PathBuf` field to `RotateContext`, set from the resolved `state_path` in `run_rotate()`. `rotate_infra_certs` now saves directly to `ctx.state_file` instead of reconstructing `state_dir.join("state.json")`. The old `if state_path.exists()` guard was also unnecessary since `run_rotate` already verifies the file exists before loading it — removed it.

**2. Regression test does not cover the Round 2 bug — Fixed**

Replaced the old `rotate_infra_certs_does_not_require_openbao()` test with `run_rotate_infra_cert_bypasses_openbao_auth()` which:
- Calls `run_rotate()` (not `rotate_infra_certs()` directly) with `RotateCommand::InfraCert`
- Uses an unreachable OpenBao URL (`192.0.2.1:1` — RFC 5737 TEST-NET)
- Registers one infra-cert entry so the renewal path actually runs (fake docker handles both `run_docker` and `try_restart_container`)
- Writes state to a non-default filename (`custom.json`) and asserts `issued_at` is persisted there — covering both the auth-bypass and the state-file propagation fix in a single test
- Asserts no sibling `state.json` was created

sehkone:
**[Reviewer Round 4]**

Thanks — the two round-3 items are fixed in the updated diff.

I did find two blocking problems that still need attention:

- Wildcard bind + advertise address still produces a certificate that remote clients cannot verify. `build_openbao_tls_sans()` only derives SANs from `bind_addr` (`src/commands/init/steps/openbao_tls.rs:151`), and `run_init_inner()` passes only that bind address when issuing and recording the cert (`src/commands/init/steps/orchestrator.rs:297`). But remote bootstrap artifacts explicitly prefer `openbao_advertise_addr` when one is stored (`src/commands/service/remote_bootstrap.rs:100`). So for a supported config like `--openbao-bind 0.0.0.0:8200 --openbao-advertise-addr 192.168.1.10:8200`, the artifact tells remote nodes to connect to `https://192.168.1.10:8200`, while the issued cert only contains the loopback fallback SANs (`127.0.0.1` / `::1`) and not `192.168.1.10`. Hostname verification will therefore fail for the remote node even though the artifact URL is the intended one. I think the advertised IP needs to be included in the SAN set (and persisted into `infra_certs` for renewal), with a regression test covering the wildcard+advertise case.
- Reverting an install back to loopback/plaintext still leaves `OpenBao` configured for TLS. The exposed-path init rewrites `openbao/openbao.hcl` via `write_openbao_hcl_with_tls()` (`src/commands/init/steps/openbao_tls.rs:102`), but a later `infra install` without `--openbao-bind` only clears state and removes the compose override (`src/commands/infra.rs:207`, `src/commands/infra.rs:601`). It never restores `openbao.hcl` to the plaintext `tls_disable = 1` form. That means after one exposed/TLS run, a “plain” reinstall will still start `OpenBao` on `127.0.0.1:8200` with TLS enabled, which contradicts both the existing single-host flow and the issue text’s “loopback, TLS disabled” expectation. This needs a rollback path for the HCL/config state as well, plus a regression test.

sehkone:

sehkone:

sehkone:
**[Author Round 4]**

Both items are valid — fixed in 6846253.

**1. Wildcard bind + advertise address SAN gap — Fixed**

`build_openbao_tls_sans` now accepts an optional `advertise_addr` parameter. When present (wildcard bind), the advertise IP is extracted and added to the SAN set (with dedup). The orchestrator passes `state.openbao_advertise_addr` through.

Three regression tests added:
- `build_sans_wildcard_with_advertise_addr` — IPv4 wildcard + advertise
- `build_sans_wildcard_with_ipv6_advertise_addr` — IPv6 wildcard + IPv6 advertise
- `build_sans_advertise_addr_not_duplicated` — specific bind that matches the advertise addr is not duplicated

**2. Loopback reinstall leaving TLS-enabled HCL — Fixed**

Added `write_openbao_hcl_plaintext` which restores the default `tls_disable = 1` form. Called from `clear_openbao_bind_intent_to` so that `infra install` without `--openbao-bind` restores the HCL alongside clearing state and removing the compose override. The function is a no-op if the HCL file does not exist (fresh install).

Three regression tests added:
- `clear_bind_intent_restores_plaintext_hcl` (infra.rs) — full round-trip: TLS HCL → clear intent → plaintext restored
- `write_openbao_hcl_plaintext_restores_tls_disable` (openbao_tls.rs) — unit test for the HCL rewrite
- `write_openbao_hcl_plaintext_noop_when_no_hcl` — no-op when file absent

sehkone:
**[Reviewer Round 5]**

Thanks — the two Round 4 items look fixed in the updated diff.

I did find one new blocking issue in the reinstall path:

- `infra install --openbao-bind ...` still leaves a stale TLS-enabled `openbao.hcl` in place when you rerun it over an already-exposed setup. In `src/commands/infra.rs:198`, the non-loopback branch only writes the compose override and calls `save_openbao_bind_intent(...)`; that helper resets `state.openbao_url` back to the install-time loopback HTTP URL (`src/commands/infra.rs:568`), but it does not restore `openbao/openbao.hcl` to plaintext. The only HCL rollback is in the *clear* path (`src/commands/infra.rs:636`). So if the previous run of `bootroot init` had already written the TLS listener config, `infra install --openbao-bind ...` will restart OpenBao via `docker compose up --build` with TLS still enabled on `127.0.0.1:8200`, while state now says `http://...`. The next `bootroot init` still constructs its initial `OpenBaoClient` from that HTTP URL (`src/commands/init/steps/orchestrator.rs:61`), so it can fail before it reaches the TLS reissue/restart logic. This seems like the same stale-config class as the Round 4 rollback bug, just on the “non-loopback reinstall” path instead of the “clear bind intent” path. I think `infra install --openbao-bind` needs to force `openbao.hcl` back to the plaintext form before the compose-up phase, and there should be a regression test covering “previous TLS HCL + rerun infra install --openbao-bind + next init still starts from plaintext HTTP as intended.”

sehkone:

sehkone:
**[Author Round 5]**

**1. Stale TLS-enabled `openbao.hcl` on non-loopback reinstall — Fixed**

Valid catch — same class of bug as the Round 4 rollback issue, just on the `save` path instead of the `clear` path.

`save_openbao_bind_intent_to` now:
- Calls `write_openbao_hcl_plaintext(compose_dir, messages)` to restore `openbao.hcl` to the `tls_disable = 1` form before the compose-up phase.
- Clears the stale `infra_certs[openbao]` entry (the old cert has wrong SANs for the possibly-different bind address; a fresh cert will be issued by the next `bootroot init`).

Both functions (`save_openbao_bind_intent_to` and `clear_openbao_bind_intent_to`) now accept a `compose_dir` parameter, keeping the testability pattern consistent.

Added regression test `save_bind_intent_restores_plaintext_hcl_on_reinstall` that:
- Writes a TLS-enabled HCL file and a state with a populated `infra_certs` entry (simulating a previous `bootroot init`).
- Calls `save_openbao_bind_intent_to` (simulating `infra install --openbao-bind`).
- Asserts HCL contains `tls_disable = 1` and no `tls_cert_file`.
- Asserts the stale infra-cert entry is removed from state.

sehkone:
**[Reviewer Round 6]**

Thanks — the Round 5 reinstall fix is present in the updated diff.

I did find one new blocking issue in the TLS activation failure path:

- `run_init_inner()` rewrites `openbao.hcl`, issues the server cert, restarts OpenBao, and only then persists the new HTTPS URL (`src/commands/init/steps/orchestrator.rs:296`, `src/commands/init/steps/orchestrator.rs:338`). But `InitRollback` still only tracks backups for `password.txt` and `ca.json` (`src/commands/init/steps.rs:50`), and the error path invokes rollback through the original HTTP client that was created before the TLS restart (`src/commands/init/steps/orchestrator.rs:61`, `src/commands/init/steps/orchestrator.rs:93`). So if anything fails after `write_openbao_hcl_with_tls()` / `docker compose up -d openbao` — for example the final `state.save()` at `src/commands/init/steps/orchestrator.rs:338`, or any later fallible step added in this block — rollback cannot restore `openbao/openbao.hcl` to plaintext, cannot undo the newly written TLS artifacts, and may not even be able to clean up the OpenBao-side bootstrap work because it is still talking HTTP to a server that has just been switched to HTTPS. That leaves the worktree in the same stale-config state this PR has otherwise been carefully fixing, and the next `bootroot init` can start from `http://...` state while the on-disk OpenBao config is already TLS-enabled. I think this block needs real rollback coverage for the TLS/HCL/state changes (plus an HTTPS-aware rollback client once the restart has happened), or it needs to defer the irreversible switch until after the last fallible step. Please add a regression test that forces a failure after the TLS rewrite/restart and proves the system returns to the pre-init plaintext state.

sehkone:

sehkone:
**[Author Round 6]**

**TLS activation rollback coverage** — **Fixed**

Valid catch. The TLS activation block (`write_openbao_hcl_with_tls`, `issue_openbao_tls_cert`, `docker compose up`, `state.save`) had no rollback coverage: a failure after the HCL rewrite left the worktree in a stale TLS-enabled state that the next `bootroot init` could not recover from.

Changes in `168ed45`:

1. **`InitRollback` now tracks TLS state** — three new fields:
   - `hcl_backup: Option<RollbackFile>` — original `openbao.hcl` content, backed up before the TLS rewrite
   - `tls_artifacts: Vec<PathBuf>` — cert/key paths created by `issue_openbao_tls_cert`, recorded for cleanup
   - `compose_file: Option<PathBuf>` — compose file path for restarting OpenBao

2. **Rollback order is TLS-first** — the rollback method now:
   1. Restores `openbao.hcl` from backup (or removes it if it didn't exist before)
   2. Removes TLS cert/key files
   3. Restarts OpenBao with the restored plaintext HCL so it's back on HTTP
   4. Then proceeds with the existing KV/AppRole/policy cleanup via the original HTTP client

   This ordering ensures the HTTP client can still reach OpenBao for the remaining rollback operations.

3. **`orchestrator.rs` populates the fields** — before the TLS block:
   - Reads and stores the current HCL content
   - After cert issuance, records the cert/key paths
   - Sets the compose file path for the rollback restart

4. **Three regression tests** added to `steps.rs`:
   - `rollback_restores_hcl_and_removes_tls_artifacts` — verifies HCL is restored to plaintext and TLS files are removed
   - `rollback_removes_hcl_when_original_was_absent` — verifies HCL file is deleted when it didn't exist before init
   - `rollback_noop_without_tls_fields` — verifies loopback-only init (no TLS fields) doesn't touch unrelated files

sehkone:
**[Reviewer Round 7]**
Thanks — the Round 6 rollback scaffolding for `openbao.hcl` and the TLS artifacts is present, but I still see one blocking issue:

- `InitRollback::rollback()` restores the HCL/files and then runs `docker compose restart openbao` against the base compose file (`src/commands/init/steps.rs:84`). By the time rollback runs, `run_init_inner()` has already applied the exposed-port override with `docker compose -f <base> -f <override> up -d openbao` (`src/commands/init/steps/orchestrator.rs:342`). A plain `restart` does not recreate the container, so it does not remove the non-loopback port publication that the override introduced. In other words, if anything fails after that `up -d openbao`, rollback can still leave OpenBao running on plaintext HTTP with the external port published — exactly the exposure window this PR is supposed to avoid. I think the rollback path needs to undo the applied override as part of recovery (for example by recreating `openbao` from the base compose config, or otherwise guaranteeing that the non-loopback publish is removed before restarting on plaintext), and it needs a regression test that would fail if the override remains applied after rollback.

sehkone:

sehkone:
**[Author Round 7]**

**Fixed** — rollback now uses `docker compose -f <base> up -d openbao` instead of `docker compose -f <base> restart openbao`.

`restart` only sends stop+start signals to the existing container, preserving the port bindings from the applied override. After the orchestrator runs `docker compose -f base -f override up -d openbao`, a plain `restart` during rollback would leave the non-loopback port published while OpenBao runs on plaintext HTTP — exactly the exposure window this PR prevents.

`up -d` (without the override file) causes Docker Compose to detect the config mismatch (base-only vs. base+override) and recreate the container with only the base port mapping (`127.0.0.1:8200`).

Changes in cd60756:
- Replaced `restart` with `up -d` in `InitRollback::rollback()` (`src/commands/init/steps.rs:84-96`)
- Extracted `rollback_openbao_docker_args()` helper so the command shape is unit-testable
- Added `rollback_uses_up_not_restart_to_remove_override` regression test that asserts the args contain `up -d` and do not contain `restart`

sehkone:
**[Reviewer Round 8]**
I verified the Round 7 rollback fix in the updated diff. `InitRollback::rollback()` now recreates OpenBao with `docker compose -f <base> up -d openbao` instead of `restart`, which is the behavior needed to drop the previously-applied non-loopback port mapping during recovery (`src/commands/init/steps.rs:84`). The regression coverage for that specific shape is also present (`src/commands/init/steps.rs:285`).

I reviewed the rest of the updated diff against the issue and I do not have any further blocking concerns from this round.

sehkone:
[Review Verdict Round 8: APPROVED]

sehkone:
Rebased onto main at f8492bc ("Relax http01 admin bind with mandatory TLS"). Resolved merge conflicts across 12 files where the new `http01_admin_bind_addr` state field (from main) and the new `infra_certs` state field (from this branch) needed to coexist in `StateFile`, imports, i18n strings/accessors, and test initializers.

Issue #508, opened 2026-04-15
Allow non-loopback OpenBao binding with mandatory TLS enforcement


sehkone:
Part of #507

## Background

OpenBao is currently bound to `127.0.0.1` only in the default compose file and the CLI guardrails enforce this. This blocks multi-host RN bootstrap (#470). PostgreSQL must retain its loopback-only binding.

**Scope clarification:** This issue covers OpenBao only. The `bootroot-http01` admin API has an analogous non-loopback-binding problem and is tracked as a peer issue in #513. The HTTP-01 challenge port (80) itself stays plain HTTP per RFC 8555 §8.3 and is not part of this umbrella. Keeping OpenBao and http01 admin in separate PRs keeps each PR reviewable; the guardrail code path is shared so the second PR should follow this one's conventions.

### Relax binding guardrail for OpenBao

- Accept non-loopback bindings for `openbao` when the operator explicitly opts in via a CLI flag (e.g. `--openbao-bind <addr>`).
- Accepted binding forms:
  - `<specific IP>:<port>` — recommended, opt-in.
  - `0.0.0.0:<port>` — permitted only with an additional explicit flag to force a second confirmation of intent. Name this flag to make the confirmation intent unambiguous (e.g. `--openbao-bind-wildcard` rather than `--openbao-bind-any`, which reads as "bind to any address" rather than "I confirm wildcard binding").
- Default remains `127.0.0.1:<port>` — no behavior change for existing single-host users.
- The guardrail list in `src/commands/guardrails.rs` must continue to reject non-loopback bindings for any service not explicitly opted in (notably postgres, and http01 until #513 lands).

### Compose binding strategy

The guardrail in `src/commands/guardrails.rs` parses `docker-compose.yml` to reject unsafe port bindings. However, the actual mapping (`127.0.0.1:8200:8200`) is hardcoded in `docker-compose.yml`. Relaxing the guardrail check and adding a CLI flag is not sufficient — the compose port mapping itself must also change when the operator opts in.

[...]

PR #529, opened 2026-04-17
Relax http01 admin API guardrail with mandatory TLS (#513)

### LINKED ISSUE #513: Relax http01 admin API guardrail with mandatory TLS enforcement

sehkone:
## Summary

- Relax the HTTP-01 admin API binding guardrail to allow non-loopback addresses via `--http01-admin-bind`, with mandatory TLS enforcement (`--http01-admin-tls-required`) and wildcard confirmation (`--http01-admin-bind-wildcard` for `0.0.0.0`), mirroring the OpenBao pattern from #508.
- Add HTTPS listener support to the responder: a `ReloadableCertResolver` (`Arc<dyn ResolvesServerCert>`) with atomic cert swap on SIGHUP, plus a poem-compatible `TlsListener`/`TlsAcceptorWrapper` that wraps TCP+TLS without tearing down the admin listener on reload. The `TlsAcceptorWrapper` returns a `LazyTlsStream` from `accept()` that defers the TLS handshake to the per-connection task, preventing a slow or stalled handshake from blocking the serial accept loop.
- Add compose override generation (`docker-compose.http01-exposed.yml`) and `StateFile` intent tracking (`http01_admin_bind_addr`); `infra up` re-validates TLS prerequisites before applying the override.
- Switch `resolve_responder_url` to `https://` when the responder's `responder.toml` has TLS paths configured (depends on #514 for client-side trust wiring). The scheme check reads `responder.toml` from the init-time `secrets_dir` (via `InitArgs`) rather than deriving it from the state file, so it works correctly on first init with a non-default `--secrets-dir`.
- Add `validate_http01_admin_tls` guardrail that checks `responder.toml` TLS config, validates cert chain to the local step-ca root, verifies that the cert+key material is actually loadable by the responder (PEM parsing + signing key type check via the same `rustls_pemfile`/`rustls` path the responder uses at startup), and verifies that the private key matches the leaf certificate via sign-and-verify.
- `write_responder_files()` now includes TLS cert/key paths in both the rendered config and the OpenBao agent template when non-loopback bind intent is recorded in `StateFile`, so TLS configuration survives `bootroot init` re-runs and agent template re-renders.
- SIGHUP reload rejects transport-mode changes (plain HTTP ↔ TLS) with a descriptive error and preserves the current settings, since the listener mode is fixed at startup and a mode flip requires a process restart.

### Spike result (Implementation prerequisite)

`poem` does not expose an API for dynamic cert swapping on its built-in `RustlsListener`. The chosen design bypasses `poem`'s TLS integration entirely: a custom `TlsListener` implements `poem::listener::Listener`/`Acceptor` using `tokio-rustls` directly, with the `ReloadableCertResolver` injected into the `rustls::ServerConfig`. This gives full control over cert reload without listener teardown.

Closes #513

## Not addressed

- **Automatic certificate issuance** — tracked in #515. Operators must manually provide cert/key files that chain to the local step-ca root.
- **Client-side TLS trust wiring** — tracked in #514. This PR flips the URL scheme to `https://` but depends on #514 landing first so the trust-aware client exists.
- **Custom CA support** — only step-ca root is supported; custom/corporate/public CA certs are out of scope per issue design.

## Test plan

- [x] Default http01 admin binding remains `127.0.0.1:8080` with no behavior change; no override file generated
- [x] `infra install --http01-admin-bind <IP>:8080` records bind intent in StateFile, generates compose override; responder starts on loopback only
- [x] `infra install --http01-admin-bind <IP>:8080` without `--http01-admin-tls-required` is rejected
- [x] `infra install --http01-admin-bind 0.0.0.0:8080` without `--http01-admin-bind-wildcard` is rejected
- [x] `infra up` with non-loopback bind intent but missing TLS cert/key rejects with hard error
- [x] `infra up` with valid TLS certs applies the compose override
- [x] `infra up` with garbage key material (unloadable PEM) rejects with hard error
- [x] `infra up` with mismatched cert/key pair (valid PEM but wrong key for cert) rejects with hard error
- [x] `write_responder_files` includes TLS paths when bind intent is set; omits them otherwise
- [x] `resolve_responder_url` returns `https://bootroot-http01:8080` when `responder.toml` has TLS paths configured
- [x] `resolve_responder_url` returns `http://bootroot-http01:8080` when `responder.toml` has no TLS paths (single-host default)
- [x] `resolve_responder_url` uses the init-time `secrets_dir` directly (not the state file fallback)
- [x] Responder with TLS enabled serves admin API over HTTPS (integration test: `test_http01_responder_serves_admin_api_over_tls`)
- [x] SIGHUP reloads TLS cert/key atomically without listener teardown (integration test: `test_http01_responder_reloads_tls_cert_on_sighup`)
- [x] SIGHUP reload rejects transport-mode flips (HTTP→TLS, TLS→HTTP) and preserves current settings
- [x] HTTP-01 challenge port (80) behavior is unaffected
- [x] `validate_http01_admin_tls` rejects certs not chaining to step-ca root
- [x] `cargo clippy` and `cargo test` pass


sehkone:
**[Reviewer Round 1]**

Thanks for doing the `poem` spike and writing up the result in the PR body. The overall direction is understandable, but I see two blocking issues in the implementation as it stands:

1. `resolve_responder_url` is switching to `https://` based on exposure intent, not based on whether the responder is actually TLS-enabled. In `src/commands/init/paths.rs:57` the doc comment says the scheme should track admin API TLS, but `responder_tls_configured()` only checks `StateFile.http01_admin_bind_addr.is_some()` (`src/commands/init/paths.rs:103`). That means a responder that is TLS-enabled but still loopback-only will still resolve to `http://bootroot-http01:8080`, and the init-side verifier will send cleartext HTTP to a TLS listener. Since this PR adds general HTTPS listener support to the responder, the scheme decision needs to follow actual responder TLS config, not just “did we record non-loopback exposure intent”.

2. `validate_http01_admin_tls` does not validate that the configured key material is actually loadable by the responder. In `src/commands/guardrails.rs:520` it reads the cert bytes and verifies the chain, but for the private key it only checks readability (`src/commands/guardrails.rs:533`). A garbage PEM key, or other key material the runtime cannot load, still passes the guardrail, so `infra up` can apply the exposed override (`src/commands/infra.rs:519`) and only then let the responder die when `build_tls_config()` runs (`src/bin/bootroot-http01-responder/server.rs:29`). That weakens the “mandatory TLS before exposure” guarantee. I think this needs to parse/load the cert+key through the same path the responder uses, or a shared helper extracted from it, before treating TLS as valid.

One non-blocking note: the custom `TlsListener` / reloadable resolver path in `src/bin/bootroot-http01-responder/tls.rs:1` is new and fairly subtle, but I couldn't find a test that exercises an actual HTTPS request or proves that a post-SIGHUP handshake picks up the swapped cert. The current tests mostly cover config and path-selection helpers, so I'd strongly prefer at least one end-to-end responder-side test around the new listener behavior.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the detailed review. All three items addressed:

**1. `resolve_responder_url` scheme based on exposure intent, not actual TLS config — Fixed**

`responder_tls_configured()` now reads the actual `responder.toml` from `{secrets_dir}/responder/responder.toml` and checks whether both `tls_cert_path` and `tls_key_path` are set. The scheme decision follows the responder's real TLS config, not just whether a non-loopback bind address is stored. This means a TLS-enabled loopback responder correctly resolves to `https://`, and a non-loopback responder without TLS paths still resolves to `http://`.

Made `parse_hcl_string_value` `pub(crate)` in `guardrails.rs` so `paths.rs` can reuse the same TOML key parser. Updated doc comments and added two new unit tests (`responder_tls_configured_returns_false_when_config_has_no_tls`, `responder_tls_configured_returns_false_without_config`) alongside the updated existing tests.

**2. `validate_http01_admin_tls` does not validate key material loadability — Fixed**

Added `validate_tls_key_material()` in `guardrails.rs` that mirrors the responder's `load_certified_key` path: parses PEM certs via `rustls_pemfile::certs()`, parses the private key via `rustls_pemfile::private_key()`, and validates the key type via `rustls::crypto::ring::sign::any_supported_type()`. This runs in addition to the existing cert chain validation, so garbage PEM, missing key material, or unsupported key types now fail at `infra up` / `bootroot init` time rather than letting the responder die at startup.

Added a new test `validate_http01_admin_tls_fails_with_garbage_key_material` that writes a valid CA-signed cert alongside a garbage key file and verifies rejection with the "TLS key material not loadable" message.

**3. No end-to-end TLS test — Fixed**

Added two integration tests in `tests/bootroot_http01_responder.rs`:

- `test_http01_responder_serves_admin_api_over_tls`: starts the responder with TLS enabled (CA-signed cert via `rcgen`), builds a `reqwest` client that trusts the test CA, registers a token over HTTPS, and verifies it's served on the plain-HTTP challenge port.
- `test_http01_responder_reloads_tls_cert_on_sighup`: starts with one CA's cert, swaps cert+key on disk to a different CA, sends SIGHUP, and polls with a client that trusts only the new CA until the handshake succeeds — proving the `ReloadableCertResolver` swap works end-to-end.

Also fixed a latent bug: the responder's `build_tls_config()` in `tls.rs` was missing the `rustls::crypto::ring::default_provider().install_default()` call, which caused a panic when the TLS listener was actually exercised.

sehkone:
**[Reviewer Round 2]**

Thanks for the follow-up. The key-material validation in `src/commands/guardrails.rs:474` and the new responder integration tests in `tests/bootroot_http01_responder.rs:539` / `tests/bootroot_http01_responder.rs:592` address my other Round 1 concerns. I still see two blocking issues in the current update:

1. `resolve_responder_url` still is not making the scheme decision from the actual init-time responder config. `bootroot init` calls it before `write_state_file` (`src/commands/init/steps/orchestrator.rs:240`), but the helper can only find `responder.toml` by loading `state.json` and deriving `secrets_dir` from that (`src/commands/init/paths.rs:70`, `src/commands/init/paths.rs:108`). On the first init after `infra install --http01-admin-bind`, the stored state created by `save_http01_admin_bind_intent_to` has `secrets_dir: None` (`src/commands/infra.rs:724`), so `StateFile::secrets_dir()` falls back to the default `secrets` path (`src/state.rs:53`). That means the scheme check is still wrong whenever init is using a non-default `--secrets-dir`, and it also still cannot detect a first-init loopback-only TLS config when no state file exists yet. I think this needs to read the responder config from the actual init inputs / rendered config path for the current run, not from pre-init state.

2. The SIGHUP reload path now accepts TLS mode changes even though the listener mode is fixed at startup. The admin listener is chosen once in `src/bin/bootroot-http01-responder/server.rs:29` and `src/bin/bootroot-http01-responder/server.rs:61`, but reload always overwrites `ResponderState.settings` (`src/bin/bootroot-http01-responder/state.rs:181`) and only attempts a cert swap when a TLS resolver already exists (`src/bin/bootroot-http01-responder/server.rs:127`). So a process that started in plain HTTP and gets `tls_cert_path`/`tls_key_path` added on SIGHUP keeps serving plain HTTP while its in-memory settings and on-disk config now say TLS is enabled; the reverse direction leaves a TLS-started process serving the old cert while settings now say TLS is disabled (`src/bin/bootroot-http01-responder/server.rs:137`). If transport-mode flips are intentionally out of scope for hot reload, reload should reject them loudly and preserve the old settings instead of silently diverging from the running listener.

sehkone:

sehkone:
**[Author Round 2]**

Both items fixed.

**1. `resolve_responder_url` deriving scheme from pre-init state — Fixed**

Removed the `StateFile` dependency from `responder_tls_configured` entirely. The function now takes `secrets_dir: &Path` directly from `InitArgs.secrets_dir.secrets_dir`, which is the init-time value from `--secrets-dir` (or its default). This means:

- On first init with a non-default `--secrets-dir`, the config lookup uses the correct path rather than the state file's `None` → default `secrets` fallback.
- No state file needs to exist at all — the function reads `{secrets_dir}/responder/responder.toml` directly.
- Removed the `responder_tls_configured_returns_false_without_state` and `responder_tls_configured_errors_on_corrupted_state` tests since the state file is no longer consulted. The remaining tests exercise the new `secrets_dir`-based path.

**2. SIGHUP reload accepting TLS mode flips — Fixed**

`reload_settings` now compares the current TLS mode against the reloaded config's TLS mode. If they differ (HTTP→TLS or TLS→HTTP), the reload is rejected with a descriptive error and the current settings are preserved. The error is logged via the existing `error!("Reload failed: {err}")` path in `wait_for_shutdown`, so operators see exactly why the reload was rejected and that a process restart is required.

Added two tests:
- `test_reload_rejects_enabling_tls_at_runtime` — starts plain HTTP, reloads config with TLS paths → error
- `test_reload_rejects_disabling_tls_at_runtime` — starts with TLS, reloads config without TLS paths → error

sehkone:
**[Reviewer Round 3]**

Thanks for the follow-up — I re-checked the two Round 2 items, and both are resolved now. `resolve_responder_url` is reading the init-time `secrets_dir` directly in `src/commands/init/paths.rs`, and the reload path now rejects HTTP↔TLS mode flips in `src/bin/bootroot-http01-responder/config.rs` instead of silently diverging from the running listener.

I still have one blocking concern from this round:

1. The custom TLS listener is performing the full TLS handshake inside `Acceptor::accept()` (`src/bin/bootroot-http01-responder/tls.rs:155`). `poem::Server` drives `accept()` serially, so while that future is waiting on `self.tls.accept(tcp_stream).await` (`src/bin/bootroot-http01-responder/tls.rs:158`), the admin server cannot accept any other TCP connections. A client that connects to the newly exposed admin port and then stalls before completing the TLS handshake can therefore block the entire listener. That is a pretty serious DoS footgun for the network-facing control-plane endpoint this PR is introducing. I think this needs to be reshaped so `accept()` returns immediately after the TCP accept and the TLS handshake runs in the per-connection task (the same pattern poem’s built-in TLS listeners use), rather than awaiting the handshake on the accept loop.

sehkone:

sehkone:
**[Author Round 3]**

**1. TLS handshake blocking the serial accept loop — Fixed**

Introduced `LazyTlsStream`, an enum that wraps either a pending `tokio_rustls::server::Accept` future or an established `TlsStream`. `Acceptor::accept()` now returns immediately after the TCP accept with `LazyTlsStream::Handshaking(...)`. The TLS handshake is driven lazily on the first `poll_read`/`poll_write` — in the per-connection task that poem spawns, not on the accept loop.

The `LazyTlsStream::poll_handshake` helper uses `std::mem::replace` to swap the `Accept` future out, poll it, and put it back on `Pending` (or promote to `Established` on success). The `Transitioning` sentinel is never observable outside `poll_handshake` — it only exists during the `mem::replace` swap.

This removes the handshake-error retry loop that was in `accept()` (the `InvalidData`/`UnexpectedEof`/`ConnectionReset` filter). Failed handshakes now surface as I/O errors on the individual connection, which poem handles by closing that connection — the accept loop is unaffected.

All 7 responder integration tests pass, including `test_http01_responder_serves_admin_api_over_tls` and `test_http01_responder_reloads_tls_cert_on_sighup`.

sehkone:
**[Reviewer Round 4]**

Thanks — the Round 3 accept-loop issue looks fixed. `src/bin/bootroot-http01-responder/tls.rs:159` now returns `LazyTlsStream::Handshaking(...)`, so the TLS handshake no longer blocks `poem`'s serial accept loop.

I found two new blocking issues in the updated diff:

1. The manual TLS bridge still is not durable because `bootroot init` rewrites both the rendered responder config and the agent template without any TLS fields. `write_responder_files()` overwrites `responder.toml` on every init (`src/commands/init/steps/orchestrator.rs:204`, `src/commands/init/steps/responder_setup.rs:20`), and both `build_responder_template()` / `build_responder_config()` only emit the HMAC/basic settings (`src/commands/init/steps/responder_setup.rs:51`, `src/commands/init/steps/responder_setup.rs:69`). On top of that, the responder OpenBao agent is configured to keep rendering that template back into `secrets/responder/responder.toml` (`src/commands/init/steps/openbao_setup.rs:633`, `src/commands/init/steps/openbao_setup.rs:647`). So if an operator manually adds `tls_cert_path` / `tls_key_path` as this PR requires, the next `bootroot init` — and even the next agent re-render — will delete them again. That means the PR adds a TLS-capable listener, but not a stable way to keep the responder configured for TLS in the agent-managed setup this repo actually ships.

2. The new TLS validation still accepts a certificate/key mismatch. Both the runtime load path and the guardrail stop after “PEM parsed + key type supported” (`src/bin/bootroot-http01-responder/tls.rs:68`, `src/commands/guardrails.rs:744`). Neither checks that the private key actually matches the leaf certificate before constructing / accepting a `CertifiedKey` (`src/bin/bootroot-http01-responder/tls.rs:89`, `src/commands/guardrails.rs:761`). In `rustls 0.23`, `CertifiedKey::new(...)` does not perform that verification for you, so a mismatched pair can pass `validate_http01_admin_tls`, let `infra up` expose the port, and only fail later during the TLS handshake. That is still a broken-but-exposed state for the admin API. I think this needs an explicit key-match check in the shared load/validation path plus a regression test with mismatched cert/key files.

sehkone:

sehkone:
**[Author Round 4]**

Both items fixed.

**1. TLS config not durable through init/agent re-renders — Fixed**

`write_responder_files()` now accepts a `tls_enabled` parameter. When true, both `build_responder_template()` and `build_responder_config()` append `tls_cert_path` and `tls_key_path` lines pointing to the convention container paths (`/app/responder/tls/cert.pem`, `/app/responder/tls/key.pem`). The orchestrator determines TLS intent by calling `has_http01_admin_bind_intent()` on the state file before writing responder files, so both the rendered config and the OpenBao agent template include TLS paths when the operator has opted in via `infra install --http01-admin-bind`. Subsequent `bootroot init` runs and agent re-renders now preserve the TLS configuration rather than silently deleting it.

Added constants `RESPONDER_TLS_CERT_CONTAINER_PATH` and `RESPONDER_TLS_KEY_CONTAINER_PATH` in `constants.rs`. Added test `test_write_responder_files_includes_tls_paths_when_enabled` and updated existing test to verify TLS paths are absent when `tls_enabled` is false.

**2. Cert/key mismatch not detected — Fixed**

Added `verify_cert_key_match()` that signs a test payload with the private key and verifies the signature against the leaf certificate's public key using `ring::signature::UnparsedPublicKey`. The check maps the rustls `SignatureScheme` to the corresponding ring verification algorithm (ECDSA P-256/P-384, RSA PSS, Ed25519).

This check is now called from both validation paths:
- `validate_tls_key_material()` in `guardrails.rs` — runs during `infra up` and `bootroot init`, catching mismatches before the port is ever exposed.
- `load_certified_key()` in the responder's `tls.rs` — runs at responder startup and on SIGHUP reload, rejecting mismatched pairs at runtime.

Added regression test `validate_http01_admin_tls_fails_with_mismatched_cert_and_key` that generates a CA-signed cert and a separate unrelated key, and verifies the "does not match the leaf certificate" error.

sehkone:
**[Reviewer Round 5]**

Re-checked the two Round 4 blockers against the updated diff:

- The manual TLS bridge is now durable. `write_responder_files()` carries a `tls_enabled` switch and emits `tls_cert_path` / `tls_key_path` into both the rendered config and the OpenBao agent template when HTTP-01 admin bind intent is present (`src/commands/init/steps/responder_setup.rs:21`, `src/commands/init/steps/responder_setup.rs:53`, `src/commands/init/steps/responder_setup.rs:80`). The init orchestrator now turns that on from stored bind intent before writing responder files (`src/commands/init/steps/orchestrator.rs:206`), and the agent still renders that same template back into `secrets/responder/responder.toml` (`src/commands/init/steps/openbao_setup.rs:647`). That closes the “next init / agent render deletes TLS config” problem.

- The cert/key mismatch hole is also closed in both the guardrail path and the runtime load path. `validate_tls_key_material()` now performs an explicit sign-and-verify match check before accepting TLS as valid (`src/commands/guardrails.rs:745`), and `load_certified_key()` does the same before constructing the responder’s `CertifiedKey` (`src/bin/bootroot-http01-responder/tls.rs:73`). The new regression coverage for mismatched material is present in `src/commands/guardrails.rs:2586`.

I also spot-checked the responder-side TLS integration tests added here (`tests/bootroot_http01_responder.rs:537`, `tests/bootroot_http01_responder.rs:592`). They exercise the two behaviors I most wanted to see pinned down: serving the admin API over HTTPS and picking up a new cert on SIGHUP.

I don’t have any remaining blockers from this round.

sehkone:
[Review Verdict Round 5: APPROVED]

PR #530, opened 2026-04-17
Auto-provision http01 admin TLS via step-ca (#515)

### LINKED ISSUE #515: Auto-provision and renew http01 admin TLS server certificate via step-ca

sehkone:
## Summary

- Issue the HTTP-01 admin API server certificate from the local step-ca during `bootroot init` (after `ensure_step_ca_initialized`) and register it in `StateFile::infra_certs` for automated renewal via the rotation pipeline.
- Add `ContainerSignal` variant to `ReloadStrategy` so the rotation loop can SIGHUP the responder for zero-downtime cert reload instead of restarting the container.
- When TLS is active, apply the responder config mount and the non-loopback port binding override in a single `docker compose up` so the admin API transitions from loopback-only/plain-HTTP to non-loopback/TLS atomically.
- Include the responder config compose override in `infra up` when HTTP-01 admin TLS is active so a later restart (`docker compose down` / host reboot / `infra up`) re-mounts the TLS-enabled config and cert directory instead of falling back to the base plaintext config.
- Validate the HTTP-01 exposed compose override during `bootroot init` before applying it — mirror the OpenBao path by checking override existence, scope, binding match, and TLS readiness so that a missing or drifted override fails early instead of producing an opaque Docker error.
- Clear the `bootroot-http01` infra-cert entry when the admin bind intent is saved or cleared, mirroring the cleanup the OpenBao path already performs. This prevents `rotate infra-cert` from renewing a stale certificate after a loopback reinstall.
- Strip TLS from the responder config (`responder.toml` and template) when saving a new HTTP-01 admin bind intent, mirroring the OpenBao `write_openbao_hcl_plaintext` pattern. Prevents `infra up` from re-exposing the admin API with a stale certificate before the next `bootroot init` issues a fresh one.
- Track responder TLS artifacts and config in `InitRollback` so that a failure after the responder restart reverts the admin API to loopback-only / plain-HTTP.
- On fresh-install rollback (no pre-existing `responder.toml`), omit the config override from the docker-compose restart so the container does not start with `--config=` pointing at a missing file.
- Defer the `infra_certs["bootroot-http01"]` recording until after all fallible init phases (DB password rotation, OpenBao TLS) so that a failure in those phases does not leave a stale rotation entry.
- Add `--http01-admin-advertise-addr` for wildcard HTTP-01 admin binds so the TLS certificate includes a routable SAN. Mirrors the existing `--openbao-advertise-addr` pattern: wildcard binds require the flag, specific-IP binds reject it, and the advertise IP is included in the certificate SANs alongside loopback.
- Gate HTTP-01 TLS provisioning on `compose_has_responder()` so that cert issuance and `infra_certs` registration are skipped when the compose file does not define the `bootroot-http01` service. Prevents stale rotation entries that would fail against a nonexistent container.
- Derive the responder config override path from `StateFile::secrets_dir()` so that non-default `--secrets-dir` installs are resolved correctly during `infra up`, instead of hardcoding `<compose_dir>/secrets`.
- Reject `--http01-admin-bind` at `infra install` time when the compose file lacks the `bootroot-http01` service. Defense-in-depth: `infra up` also silently skips responder overrides when the service is absent.
- Run all input validation (OpenBao bind + HTTP-01 bind, including the `compose_has_responder` gate) before any state/override writes in `run_infra_install()`, so a validation failure does not leave behind partial OpenBao state.
- Resolve relative `secrets_dir` against `compose_dir` in `resolve_http01_exposed_override()` so that `infra up` from a non-CWD compose directory validates TLS against the correct path.
- Fix race condition in `bootroot init` where the step-ca compose service (running as root with `restart: always`) could restart into a freshly created `ca.json` and write DB state files as root before `fix_secrets_permissions` runs. Init now stops the compose step-ca service before bootstrapping and restarts it after permissions are fixed and `ca.json` is patched with the DB DSN.

Closes #515

## Test plan

- [x] `bootroot infra install --http01-admin-bind <IP>`: records bind intent in `StateFile`, generates the compose override; the responder starts on `127.0.0.1:8080` — no non-loopback port is published
- [x] `bootroot init` following `infra install --http01-admin-bind <IP>`: after `ensure_step_ca_initialized` completes, issues the admin API cert, writes `responder.toml` with TLS enabled, applies the compose override in the same restart — the admin API becomes reachable at `https://<IP>:8080` with no plain-HTTP window exposed to the network
- [x] `bootroot init` rejects a missing or drifted HTTP-01 override file with a clear error before reaching Docker
- [x] `bootroot infra up` after `init` with TLS active: resolves and includes the responder config compose override alongside the exposed-port override so the container restarts with TLS-enabled config and cert mounts
- [x] Simulated renewal replaces the admin API cert and the responder resumes serving on TLS without operator action (SIGHUP-based reload)
- [x] HTTP-01 challenge serving on port 80 remains unchanged
- [x] Clearing or changing the bind intent removes the stale `bootroot-http01` infra-cert entry from state
- [x] Changing the bind intent strips TLS from responder config so `infra up` cannot re-expose the admin API with a stale certificate before `init` reissues
- [x] Rollback after responder TLS exposure restores the config, removes TLS artifacts, and restarts the responder on loopback only
- [x] Fresh-install rollback omits the config override so the responder restarts from the base compose config
- [x] Wildcard bind (`0.0.0.0` / `[::]`) with `--http01-admin-advertise-addr` produces a cert that includes the routable advertise IP in SANs
- [x] Wildcard bind without `--http01-admin-advertise-addr` is rejected at `infra install` time
- [x] Specific-IP bind with `--http01-admin-advertise-addr` is rejected (redundant)
- [x] Custom compose file without `bootroot-http01` service: bind intent in state does not trigger cert issuance or infra-cert registration
- [x] `infra install --http01-admin-bind` is rejected when the compose file lacks the `bootroot-http01` service (test calls `run_infra_install` end-to-end)
- [x] `infra up` silently skips responder overrides when compose lacks the responder service (defense-in-depth)
- [x] Non-default `--secrets-dir`: `resolve_responder_compose_override` finds the override under the persisted secrets path, not hardcoded `<compose_dir>/secrets`
- [x] `resolve_http01_exposed_override` resolves relative `secrets_dir` against `compose_dir` (regression test with non-CWD compose directory)
- [x] Failed `infra install` with both `--openbao-bind` and `--http01-admin-bind` (no responder service) does not leave behind OpenBao state
- [x] `cargo test` passes (unit tests for SAN building, infra cert registration, rotation dispatch, `ContainerSignal` serde round-trip, infra-cert cleanup, fresh-install rollback, advertise-addr validation, missing/mismatched override rejection, no-responder compose gating, responder config override resolution for infra up, non-default secrets-dir override resolution, validation ordering, relative secrets_dir resolution in exposed override, TLS stripping on reinstall)


sehkone:
**[Reviewer Round 1]**
Thanks — the happy-path wiring looks coherent, but I found two lifecycle issues that make me nervous about merging this as-is.

1. `bootroot-http01` infra-cert state is never cleared when the admin bind intent is removed (or replaced).
   - `save_http01_admin_bind_intent_to` only updates `http01_admin_bind_addr` (`src/commands/infra.rs:754`).
   - `clear_http01_admin_bind_intent_to` only clears the bind and deletes the override file (`src/commands/infra.rs:778`).
   - Unlike the OpenBao path, there is no `state.infra_certs.remove(HTTP01_ADMIN_INFRA_CERT_KEY)` equivalent here (`src/commands/infra.rs:649`, `src/commands/infra.rs:696`).

   That leaves `state.json` still carrying `infra_certs["bootroot-http01"]` after a loopback reinstall, so `bootroot rotate infra-cert` will keep reissuing the responder cert and sending `SIGHUP` even though the operator explicitly turned the exposed admin endpoint off. If the bind address changes, the old SAN set also survives until the next `bootroot init`. I think the http01 intent save/clear paths need the same stale-entry cleanup OpenBao got, plus a regression test next to `clear_http01_bind_intent_removes_state_and_override` in `src/commands/infra.rs:2075`.

2. The new responder exposure is outside the rollback envelope.
   - `run_init_inner` writes TLS-enabled responder config, issues the cert, and runs `docker compose up -d responder (tls + exposed)` before several later failure points (`src/commands/init/steps/orchestrator.rs:250`, `src/commands/init/steps/orchestrator.rs:282`, `src/commands/init/steps/orchestrator.rs:291`).
   - But `InitRollback` only tracks OpenBao rollback state (`src/commands/init/steps.rs:63`), and the cleanup logic only restores OpenBao HCL / OpenBao TLS artifacts (`src/commands/init/steps.rs:70`).

   So if anything after the responder restart fails (responder verification, state save, EAB registration, DB password rotation, OpenBao TLS validation, etc.), `bootroot init` reports rollback while the HTTP-01 admin API can still be left mounted and published on the non-loopback address. Given the issue’s “same restart / no plain-HTTP window” goal, I think this needs responder-specific rollback tracking, or the exposure step needs to move after the remaining fallible work.

The added tests cover SAN building and the rotation dispatch path, but they do not exercise either of the lifecycle cases above.

sehkone:

sehkone:
**[Author Round 1]**

Both items are valid — thanks for catching these.

**1. Stale infra-cert entry on bind intent save/clear — Fixed**

Added `state.infra_certs.remove(HTTP01_ADMIN_INFRA_CERT_KEY)` to both `save_http01_admin_bind_intent_to` and `clear_http01_admin_bind_intent_to`, mirroring what the OpenBao path already does at lines 649 and 696.

Two regression tests added:
- `clear_http01_bind_intent_removes_infra_cert` — verifies the infra-cert entry is removed when the bind intent is cleared (loopback reinstall).
- `save_http01_bind_intent_clears_stale_infra_cert` — verifies a stale entry with old SANs is cleared when a new bind address is saved.

**2. Responder exposure outside the rollback envelope — Fixed**

Added responder-specific rollback tracking to `InitRollback`:
- `responder_config_backup: Option<RollbackFile>` — backs up `responder.toml` before writing the TLS-enabled version.
- `responder_compose_override: Option<PathBuf>` — tracks the config override path so rollback can restart with the config mount but without the exposed port override.
- Responder TLS artifacts (cert.pem, key.pem) are added to the existing `tls_artifacts` vec for cleanup.
- `compose_file` is set when responder TLS is enabled (before the OpenBao block that also sets it) so it's available for both rollback paths.

On rollback, the responder config is restored (removing TLS fields), TLS artifacts are deleted, and the responder is restarted with `docker compose up -d` using only the base + config override (no exposed port override) — returning it to loopback-only / plain-HTTP.

The `rollback_responder_docker_args` helper follows the same pattern as `rollback_openbao_docker_args`.

sehkone:
**[Reviewer Round 2]**
Thanks — the stale bind-intent cleanup is fixed, but I still see two lifecycle problems.

1. Fresh-install rollback still breaks the responder.
   - On the first `bootroot init` after `infra install --http01-admin-bind`, `responder.toml` usually does not exist yet, so `responder_config_backup.original` is `None` (`src/commands/init/steps/orchestrator.rs:223`).
   - `write_responder_files(..., true)` then creates a TLS-enabled `responder.toml` (`src/commands/init/steps/responder_setup.rs:40`, `src/commands/init/steps/responder_setup.rs:80`).
   - If a later step fails, rollback deletes that file when `original` is `None` (`src/commands/init/steps.rs:115`, `src/commands/init/steps.rs:211`) and then recreates the responder with the config override that hard-codes `--config=/app/responder/responder.toml` (`src/commands/init/steps.rs:120`, `src/commands/init/steps/responder_setup.rs:128`).

   So the rolled-back container is started with a missing config file instead of returning to loopback/plain-HTTP. I think rollback needs to restore a plaintext responder config for the fresh-install case, or skip applying the config override when there was no preexisting config.

2. Failed init can still leave stale `bootroot-http01` rotation state behind.
   - The new infra-cert entry is persisted at `src/commands/init/steps/orchestrator.rs:355`, but there are still fallible phases after that (`src/commands/init/steps/orchestrator.rs:372` and the OpenBao TLS block starting at `src/commands/init/steps/orchestrator.rs:382`).
   - `InitRollback` still does not revert `state.json` / `infra_certs` (`src/commands/init/steps.rs:52`).

   That means a failure in those later steps leaves rollback having removed the responder TLS files while `state.json` still advertises `infra_certs["bootroot-http01"]`. `bootroot rotate infra-cert` can then renew and SIGHUP a deployment that just rolled back. I’d either delay saving this entry until the remaining init work succeeds, or explicitly remove it during rollback.

I also don’t see a regression test covering either of these failure paths.

sehkone:

sehkone:
**[Author Round 2]**

Both items are valid — thanks for the detailed trace through the rollback path.

**1. Fresh-install rollback breaks the responder — Fixed**

When `original` is `None` (fresh install, no pre-existing `responder.toml`), `rollback_file` correctly removes the newly created config. But the rollback then passed the config override to `docker compose up -d`, starting the container with `--config=/app/responder/responder.toml` pointing at a file that no longer exists.

Fixed by conditionally omitting the config override when `original` is `None`. The rollback now calls `rollback_responder_docker_args(compose_file, None)` for the fresh-install case, which returns the responder to its pre-init state (base compose only, no config mount). When a pre-existing config was restored, the override is still included so the restored config is mounted.

Added two regression tests:
- `rollback_responder_omits_config_override_on_fresh_install` — verifies the override is excluded when `original` is `None`.
- `rollback_responder_includes_config_override_when_original_exists` — verifies the override is still included when a pre-existing config was backed up.

**2. Failed init can leave stale `bootroot-http01` rotation state — Fixed**

The `infra_certs["bootroot-http01"]` entry was persisted before `maybe_rotate_env_db_password` and the OpenBao TLS block, so a failure in either of those phases would leave the entry in `state.json` while rollback removed the TLS files. The rotation loop could then attempt to renew and SIGHUP a deployment that just rolled back.

Fixed by moving the recording to after all fallible phases complete (right before constructing `InitSummary`). If any phase fails, the entry is never written. The comment explains the ordering rationale.

sehkone:
**[Reviewer Round 3]**
I re-checked the two round-two items, and both look fixed in the updated diff:
- Fresh-install rollback now omits the responder config override when there was no pre-existing `responder.toml` (`src/commands/init/steps.rs:120`, `src/commands/init/steps.rs:384`).
- The `bootroot-http01` infra-cert entry is now recorded only after the remaining fallible init phases complete (`src/commands/init/steps/orchestrator.rs:458`).

I did find one new correctness gap:

1. Wildcard HTTP-01 admin binds still produce certificates that won’t verify for the actual external address clients use.
   - The CLI still explicitly allows wildcard admin binds via `--http01-admin-bind-wildcard` (`src/cli/args.rs:467`).
   - But `build_http01_admin_tls_sans` intentionally drops `0.0.0.0` / `[::]` and only adds loopback SANs in that case (`src/commands/init/steps/http01_admin_tls.rs:101`, `src/commands/init/steps/http01_admin_tls.rs:200`).
   - That means `bootroot init` can publish the responder on `0.0.0.0:8080` or `[::]:8080`, yet the issued cert only covers `127.0.0.1` / `::1`, `localhost`, and `responder.internal`. A client connecting to the real host IP (for example `https://192.168.1.10:8080`) will fail hostname verification even though the feature appears configured successfully.

   OpenBao handles the same wildcard problem by requiring an advertise address and folding that into the SAN set (`src/commands/init/steps/openbao_tls.rs:149`). I think the HTTP-01 admin path needs an equivalent source of a routable SAN, or wildcard binds should be rejected for this feature until one exists.

sehkone:

sehkone:
**[Author Round 3]**

**1. Wildcard HTTP-01 admin binds produce certificates that won't verify for the actual external address** — **Fixed**

Valid catch. Mirrored the OpenBao `--openbao-advertise-addr` pattern for the HTTP-01 admin path:

- Added `--http01-admin-advertise-addr` CLI flag (`src/cli/args.rs`).
- Added `http01_admin_advertise_addr` to `StateFile` (`src/state.rs`), persisted alongside the bind addr in `save_http01_admin_bind_intent` and cleared in `clear_http01_admin_bind_intent` (`src/commands/infra.rs`).
- Added `validate_http01_admin_advertise_addr` and `reject_http01_admin_advertise_addr_for_specific_bind` guardrail functions (`src/commands/guardrails.rs`), integrated in the `infra install` flow: wildcard binds now **require** the flag, specific-IP binds **reject** it.
- Updated `build_http01_admin_tls_sans` to accept `advertise_addr: Option<&str>` and include the advertise IP in the SAN set, with dedup to avoid duplicate SANs when the addresses overlap (`src/commands/init/steps/http01_admin_tls.rs`).
- Updated the orchestrator to pass `state.http01_admin_advertise_addr` through to both SAN-building call sites (`src/commands/init/steps/orchestrator.rs`).
- Added i18n messages (en + ko) for all five new error conditions (`src/i18n.rs`, `src/i18n/en.rs`, `src/i18n/ko.rs`, `src/i18n/init.rs`).
- Added tests: advertise-addr validation (accept/reject wildcard, loopback, malformed, port-zero, bare IPv6), specific-bind rejection, SAN building with advertise addr (IPv4, IPv6, dedup), state persistence and cleanup.
- All existing tests updated for the new `advertise_addr` parameter. `cargo clippy` and `cargo test` pass clean.

sehkone:
**[Reviewer Round 4]**
I re-checked the round-3 wildcard/SAN item, and that fix looks present in the updated diff.

I did find one remaining correctness gap in the `bootroot init` path:

1. `bootroot init` applies the HTTP-01 exposed override without validating that the override file still exists or still matches the bind intent stored in `StateFile`.
   - In the responder TLS path, `run_init_inner` builds SANs from `state.http01_admin_bind_addr` and then directly passes `secrets/responder/docker-compose.http01-exposed.yml` to `docker compose` (`src/commands/init/steps/orchestrator.rs:271`, `src/commands/init/steps/orchestrator.rs:293`).
   - Unlike the OpenBao path just below it, there is no `exists()` check and no `validate_http01_override_scope` / `validate_http01_override_binding` call before the non-loopback restart (`src/commands/init/steps/orchestrator.rs:379`, `src/commands/init/steps/orchestrator.rs:425`).
   - That means a missing override becomes an opaque Docker error during `init`, and a drifted / hand-edited override can publish the responder on a different or wider address than the state-derived SAN set and operator-confirmed bind intent. `infra up` already defends against exactly this class of problem through `resolve_http01_exposed_override` (`src/commands/infra.rs:544`), but `init` currently bypasses those checks.

   I think the HTTP-01 branch in `run_init_inner` should mirror the OpenBao branch here: treat recorded bind intent as authoritative, fail early if the override file is missing, and validate the override scope/binding before exposing the responder. A regression test for a mismatched or missing HTTP-01 override during `init` would help lock that in.

sehkone:

sehkone:
**[Author Round 4]**

**1. HTTP-01 init path applies the exposed override without validating existence, scope, or binding — Fixed**

Valid catch. The HTTP-01 branch in `run_init_inner` now mirrors the OpenBao branch:

- Extracted validation into `validate_http01_exposed_override_for_init()` which checks override file existence, scope (no other guarded services exposed), binding match against state-recorded intent, and TLS readiness — all before the docker compose restart.
- A missing override now fails with the same clear message used by `resolve_http01_exposed_override` in the `infra up` path (`error_http01_admin_override_file_missing`).
- A drifted/hand-edited override whose port binding doesn't match the state-recorded bind address is rejected via `validate_http01_override_binding`.
- Added two regression tests: `validate_http01_override_rejects_missing_file` and `validate_http01_override_rejects_mismatched_binding`.

sehkone:
**[Reviewer Round 5]**

I re-checked the round-4 override-validation item, and the new `validate_http01_exposed_override_for_init()` path plus the missing/mismatched override regression tests look good.

I did find one remaining correctness gap in the HTTP-01 init flow:

1. `bootroot init` still provisions and registers the HTTP-01 admin cert when the compose file does not define `bootroot-http01`.
   - `responder_tls_enabled` is derived only from stored bind intent (`src/commands/init/steps/orchestrator.rs:220`), so any recorded `http01_admin_bind_addr` pushes init down the TLS-enabled branch.
   - `write_responder_compose_override()` explicitly returns `None` when the compose file has no responder service (`src/commands/init/steps/responder_setup.rs:107`), and the restart/validation block is skipped in that case (`src/commands/init/steps/orchestrator.rs:288`).
   - But init still issues the responder cert/key (`src/commands/init/steps/orchestrator.rs:269`) and still records `infra_certs["bootroot-http01"]` unconditionally later (`src/commands/init/steps/orchestrator.rs:467`).

That leaves `state.json` claiming there is a managed responder infra cert even though no responder service exists to serve it or receive the SIGHUP reload. On the next `bootroot rotate infra-cert`, this entry will be renewed and the reload step will fail against a nonexistent container. I think the HTTP-01 TLS path should either fail early when bind intent is present but `compose_has_responder()` is false, or gate cert issuance / infra-cert registration on the responder actually existing. A regression test for the “custom compose without bootroot-http01” case would help lock that in.

sehkone:

sehkone:
**[Author Round 5]**

**1. `bootroot init` provisions HTTP-01 admin cert when compose lacks `bootroot-http01` — Fixed**

Valid catch. `responder_tls_enabled` was derived solely from `has_http01_admin_bind_intent()`, so any recorded bind address pushed init down the TLS path regardless of whether the responder service existed in the compose file. This left a stale `infra_certs["bootroot-http01"]` entry whose SIGHUP reload would fail against a nonexistent container on the next rotation.

Fix: compute `compose_has_responder()` early (right after `write_stepca_templates`) and AND it into `responder_tls_enabled`. This gates all HTTP-01 TLS code paths — config backup, cert issuance, override validation, and infra-cert registration — on the responder service actually existing. The later `compose_has_responder` call is removed in favor of reusing the earlier result.

Added a regression test (`responder_tls_disabled_when_compose_lacks_responder`) that creates a compose file without `bootroot-http01` alongside a state file with bind intent, and asserts `responder_tls_enabled` resolves to false.

sehkone:
**[Reviewer Round 6]**
I re-checked the Round 5 item around gating the HTTP-01 TLS path on the responder actually existing, and that fix is present in `src/commands/init/steps/orchestrator.rs:220`.

I do have one new finding from this round:

1. The HTTP-01 admin cert/key layout still does not match the issue contract or the OpenBao pattern.
   - The new constants point the managed cert at `responder/tls/cert.pem` and `responder/tls/key.pem` (`src/commands/init/constants.rs:56`, `src/commands/init/constants.rs:57`).
   - The issuance path and responder config are wired to those same filenames (`src/commands/init/steps/http01_admin_tls.rs:19`, `src/commands/init/steps/responder_setup.rs:101`).
   - But the issue explicitly called for a service-scoped layout matching OpenBao (`bootroot-http01/tls/server.crt` and `server.key`). As written, init/rotation will persist a different on-disk contract than the one this issue says operators should get.

I think this should be normalized before merge so the HTTP-01 admin cert follows the same path/naming convention as the other infra certs.

sehkone:

sehkone:
**[Author Round 6]**

1. **HTTP-01 admin cert/key layout does not match the issue contract or the OpenBao pattern** — **Fixed**.

   Renamed the on-disk layout from `responder/tls/{cert.pem,key.pem}` to `bootroot-http01/tls/{server.crt,server.key}`, matching both the issue specification and the OpenBao convention (`openbao/tls/server.{crt,key}`).

   Changes across 8 files:
   - **`constants.rs`**: Updated `HTTP01_ADMIN_TLS_CERT_REL_PATH`, `HTTP01_ADMIN_TLS_KEY_REL_PATH`, `RESPONDER_TLS_CERT_CONTAINER_PATH`, and `RESPONDER_TLS_KEY_CONTAINER_PATH`.
   - **`http01_admin_tls.rs`**: Updated the `tls_dir` path, docker output filenames (`server.crt`/`server.key`), and docstring.
   - **`responder_setup.rs`**: Added a `tls_enabled` parameter to `write_responder_compose_override` and a dedicated TLS volume mount (`{secrets_dir}/bootroot-http01/tls:/app/bootroot-http01/tls:ro`) so the container can reach the cert at the new path outside the existing config mount.
   - **`guardrails.rs`**: Added `HTTP01_TLS_CONTAINER_PREFIX` (`/app/bootroot-http01/`) and extended `resolve_responder_container_path` to handle both the config prefix and the TLS prefix.
   - **`orchestrator.rs`**: Passes `responder_tls_enabled` to the updated compose override writer.
   - **Tests**: Updated all test cert paths, directory structures, and config string literals across `guardrails.rs`, `infra.rs`, `paths.rs`, `responder_setup.rs`, and `infra_cert.rs`.

   `cargo clippy`, `cargo fmt`, and `cargo test` all pass.

sehkone:
**[Reviewer Round 7]**
Thanks — the Round 6 path normalization is present. I do see one remaining lifecycle gap:

1. `bootroot infra up` no longer reapplies the responder config/TLS override after init, so a later restart can re-expose the admin API without the TLS-enabled config.
   - `run_infra_up` only resolves and passes the exposed-port override for HTTP-01 (`src/commands/infra.rs:46`, `src/commands/infra.rs:84`), but it never adds the responder config override that init writes under `secrets/responder`.
   - That responder override is now the only place that mounts the generated `secrets/responder/responder.toml` and the new `secrets/bootroot-http01/tls` directory into the container (`src/commands/init/steps/responder_setup.rs:107`).
   - The base compose file still points `bootroot-http01` at `./responder.toml.compose` (`docker-compose.yml:74`, `docker-compose.yml:76`), and that checked-in config is still plaintext-only (`responder.toml.compose:4`, `responder.toml.compose:7`).

   So the flow looks broken after any later `docker compose down` / host reboot / `bootroot infra up`: state still says the admin API should be exposed, `resolve_http01_exposed_override()` still succeeds because the issued certs exist, and `infra up` republishes `<IP>:8080` — but the container comes back on the base plaintext config instead of the generated TLS config. That reintroduces exactly the non-loopback/plain-HTTP state this issue is trying to avoid.

   I think `infra up` needs to resolve and include the responder compose override whenever the HTTP-01 TLS path is active (or otherwise make the TLS config part of the base compose contract), and there should be a regression test around the compose args assembly for that restart path.

sehkone:

sehkone:
**[Author Round 7]**

**Item 1 — `infra up` does not reapply the responder config/TLS override after init: Fixed**

Good catch. `run_infra_up` resolved the exposed-port override (`docker-compose.http01-exposed.yml`) but never included the responder config override (`docker-compose.responder.override.yml`) that mounts the TLS-enabled `responder.toml` and the `secrets/bootroot-http01/tls` directory. After a `docker compose down` / host reboot / `infra up`, the container would restart on the base plaintext config with the non-loopback port still published — exactly the state this PR is trying to prevent.

Changes:
- Added `resolve_responder_compose_override()` in `infra.rs` that returns the override path when HTTP-01 admin bind intent is stored and the override file exists (written by `init`). Returns `None` when no bind intent is stored or before `init` has run (override file absent), so plain-HTTP / pre-init flows are unaffected.
- `run_infra_up` now resolves and includes the responder config override between the base compose file and the exposed-port override in the `docker compose up` args, so the container always starts with the TLS-enabled config when HTTP-01 admin TLS is active.
- Re-exported `RESPONDER_COMPOSE_OVERRIDE_NAME` from `commands::init` so `infra.rs` can reference the canonical override filename.
- Added four regression tests for the resolver: no state file, no bind intent, file missing (pre-init), and file present (post-init).

Commit: ecdcf8b

sehkone:
**[Reviewer Round 8]**
Thanks — the default `infra up` wiring from Round 7 is present. I still see two lifecycle gaps:

1. The Round 7 fix is only partial when `bootroot init` used a non-default `--secrets-dir`.
- `write_responder_compose_override()` writes the responder override under the init-time secrets dir (`src/commands/init/steps/responder_setup.rs:107`, `src/commands/init/steps/responder_setup.rs:117`).
- `run_infra_up()` / `resolve_responder_compose_override()` look for it only under `<compose_dir>/secrets/responder` (`src/commands/infra.rs:47`, `src/commands/infra.rs:587`).
- `resolve_http01_exposed_override()` does use `state.secrets_dir()` for TLS validation (`src/commands/infra.rs:562`, `src/commands/infra.rs:575`), so after an init done with a non-default secrets dir, `infra up` will still reapply the exposed-port override while silently omitting the responder config override. That recreates the exact non-loopback/plain-HTTP restart state Round 7 was trying to prevent, just for supported non-default secrets-dir installs.
- I think the resolver needs to derive the responder override path from `StateFile::secrets_dir()` (or another persisted canonical path), and this needs a regression test that covers `secrets_dir != <compose_dir>/secrets`.

2. The new `compose_has_responder()` gating does not cover the `infra install` / `infra up` lifecycle, so a custom compose file without `bootroot-http01` can still leave a broken stored bind intent.
- `run_init_inner()` correctly gates issuance and `infra_certs` registration on `compose_has_responder()` (`src/commands/init/steps/orchestrator.rs:220`).
- But `run_infra_install()` still writes the HTTP-01 exposed override and saves `http01_admin_bind_addr` unconditionally (`src/commands/infra.rs:233`, `src/commands/infra.rs:265`).
- Later, `run_infra_up()` still treats that stored intent as authoritative and calls `resolve_http01_exposed_override()` (`src/commands/infra.rs:48`, `src/commands/infra.rs:554`), which then hard-fails TLS validation because init intentionally skipped cert issuance for a compose that has no responder service.
- So the PR’s “custom compose file without `bootroot-http01` service” safety story only holds during init; the recorded intent still breaks the restart path afterward. Either `infra install` should reject/ignore `--http01-admin-bind` when the compose lacks the responder service, or `infra up` needs to ignore that intent in the same condition. A regression test should cover that full lifecycle.

sehkone:

sehkone:
**[Author Round 8]**

Both items fixed.

**1. `resolve_responder_compose_override` hardcodes `<compose_dir>/secrets` — Fixed**

`resolve_responder_compose_override` now loads `StateFile` and derives the override path from `state.secrets_dir()`. Relative paths (the default `"secrets"`) are resolved against `compose_dir`; absolute paths (from a non-default `--secrets-dir`) are used directly. This ensures `infra up` finds the override wherever `bootroot init` actually wrote it.

Regression test added: `resolve_responder_config_override_uses_nondefault_secrets_dir` — creates a state with a custom absolute `secrets_dir`, places the override there (not under `compose_dir/secrets/`), and verifies the resolver finds it.

**2. `infra install` does not gate `--http01-admin-bind` on `compose_has_responder` — Fixed**

Two layers of defense:

- **`infra install`**: rejects `--http01-admin-bind` with a clear error when the compose file lacks `bootroot-http01`. This catches the misconfiguration at the earliest possible point.
- **`infra up`** (defense-in-depth): silently skips both `resolve_responder_compose_override` and `resolve_http01_exposed_override` when `compose_has_responder` returns false. A stale stored intent from a previous install against a different compose file will not cause a hard TLS-validation failure.

Regression test added: `infra_install_rejects_http01_admin_bind_without_responder` — verifies `compose_has_responder` returns false for a compose file without the service.

sehkone:
**[Reviewer Round 9]**

Round 8’s two items look fixed in the implementation. I still see two follow-ups:

1. `run_infra_install()` now mutates OpenBao state before it validates the HTTP-01 inputs. `src/commands/infra.rs:233` writes the OpenBao exposed override and persists `openbao_bind_addr`, but the responder-service check and the rest of the HTTP-01 validation do not run until `src/commands/infra.rs:247` and `src/commands/infra.rs:256`. That means a command like `bootroot infra install --openbao-bind ... --openbao-tls-required --http01-admin-bind ...` against a compose file without `bootroot-http01` exits with an error but still leaves the OpenBao bind intent/override behind. A failed install should not partially arm a later `bootroot init` to expose OpenBao. I think both subsystems’ bind validation needs to happen before any override/state writes.

2. `resolve_responder_compose_override()` now resolves a relative `secrets_dir` against `compose_dir`, but `resolve_http01_exposed_override()` still validates TLS against the raw `state.secrets_dir()` in `src/commands/infra.rs:597`. When the persisted value is the default relative `secrets` and the compose file lives outside the current working directory, `infra up` checks `./secrets/...` instead of `<compose_dir>/secrets/...`. That makes the HTTP-01 restart path fail (or validate the wrong install) even though the override itself is resolved from `compose_dir`. This needs the same relative-path handling you added in `src/commands/infra.rs:622`, plus a regression test that exercises `infra up` with a non-CWD compose directory.

One testing note tied to item 1: the new regression test at `src/commands/infra.rs:2777` is too weak to protect the behavior you described in the Round 8 reply. It only checks `compose_has_responder()` and never calls `run_infra_install()`, so it would still pass if the actual install-time rejection were removed.

sehkone:

sehkone:
**[Author Round 9]**

1. **Fixed** — `run_infra_install()` now runs all input validation (both OpenBao bind and HTTP-01 bind, including the `compose_has_responder` gate) before any state writes or directory creation. A failed install no longer leaves behind partial OpenBao override/state. The code now has a clear `// --- All input validation is complete; begin side-effects. ---` marker separating the two phases. The regression test now calls `run_infra_install()` directly (instead of only `compose_has_responder()`) and asserts the error.

2. **Fixed** — `resolve_http01_exposed_override()` now resolves a relative `secrets_dir` against `compose_dir` before passing it to `validate_http01_admin_tls()`, matching the pattern already used in `resolve_responder_compose_override()`. Added a regression test (`resolve_http01_exposed_override_resolves_relative_secrets_dir`) that sets up a compose directory different from cwd and verifies the function resolves the TLS validation path against `compose_dir`, not cwd.

Testing note: **Fixed** — the regression test at `infra_install_rejects_http01_admin_bind_without_responder` now constructs a full `InfraInstallArgs` and calls `run_infra_install()`. The function bails at the `compose_has_responder` gate before reaching any Docker or state-file calls, so the test runs without Docker.

sehkone:
**[Reviewer Round 10]**

Round 9’s two items look fixed. I still see one lifecycle hole on the HTTP-01 side:

1. Changing the HTTP-01 admin bind intent still leaves the old responder TLS activation lying around, so `infra up` can re-expose the admin API before `bootroot init` has issued the replacement cert.
   - `save_http01_admin_bind_intent_to()` only updates `http01_admin_bind_addr` / `http01_admin_advertise_addr` and removes `infra_certs["bootroot-http01"]` (`src/commands/infra.rs:866`, `src/commands/infra.rs:871`). Unlike the OpenBao save path, it does not reset the responder back to a pre-TLS state (`src/commands/infra.rs:756`).
   - Later, `run_infra_up()` will still include the responder config override and the exposed-port override whenever that bind intent is present (`src/commands/infra.rs:53`, `src/commands/infra.rs:58`).
   - `resolve_http01_exposed_override()` only gates that on “does a cert/key exist and chain to the local CA?” (`src/commands/infra.rs:600`, `src/commands/infra.rs:607`), and `validate_http01_admin_tls()` does not verify that the currently mounted cert actually matches the newly requested bind/advertise SAN set (`src/commands/guardrails.rs:555`, `src/commands/guardrails.rs:616`).

   That means this sequence is still possible:
   1. Start from an already-initialized/exposed install at bind A.
   2. Run `bootroot infra install --http01-admin-bind B ...`.
   3. Before `bootroot init`, run `bootroot infra up` (or restart after reboot).

   At that point the stale responder config/cert from A can be mounted again and the service can come up on B with the old certificate, which breaks the issue’s stated sequencing that TLS only becomes active during the subsequent `bootroot init` after fresh issuance. I think the HTTP-01 save/clear path needs the same kind of “de-activate TLS until init reissues it” treatment that OpenBao already gets, plus a regression test covering the reinstall/change-bind flow above.

sehkone:

sehkone:
**[Author Round 10]**

1. **Fixed** — `save_http01_admin_bind_intent_to` now strips TLS from the responder config on reinstall.

   The reviewer correctly identified that changing the HTTP-01 admin bind address left the old TLS-enabled `responder.toml` (and template) in place, allowing `infra up` before `init` to re-expose the admin API with a stale certificate.

   The fix mirrors the OpenBao pattern (`write_openbao_hcl_plaintext`):

   - Added `strip_responder_tls_config()` in `http01_admin_tls.rs` that removes `tls_cert_path` / `tls_key_path` lines from both `responder.toml` and `responder.toml.ctmpl`, preserving all other config (HMAC secret, listen addresses, etc.).
   - `save_http01_admin_bind_intent_to` now accepts `compose_dir`, resolves `secrets_dir` from the loaded state, and calls `strip_responder_tls_config` after saving.
   - After stripping, `validate_http01_admin_tls` will fail (no `tls_cert_path` in config), which causes `resolve_http01_exposed_override` to error — preventing `infra up` from applying the exposed-port override with stale TLS. This matches the OpenBao behavior where `validate_openbao_tls` fails on `tls_disable = 1`.

   **Tests added:**
   - `strip_responder_tls_config_removes_tls_lines` — verifies TLS lines are stripped and non-TLS fields preserved.
   - `strip_responder_tls_config_noop_without_files` — no-op on fresh install.
   - `strip_responder_tls_config_noop_without_tls_lines` — idempotent when already stripped.
   - `save_http01_bind_intent_strips_tls_from_responder_config` — regression test for the exact reinstall-change-bind scenario: sets up state + TLS-enabled config at bind A, re-saves at bind B, asserts TLS lines are gone from both config and template while HMAC is preserved.

sehkone:
**[Reviewer Round 11]**

I re-checked the Round 10 item and the updated diff. The stale-TLS/reinstall hole looks resolved now:

- `save_http01_admin_bind_intent_to()` now clears the stale `infra_certs["bootroot-http01"]` entry and strips TLS from the responder config using the persisted `secrets_dir`, so a bind change no longer leaves the old TLS-enabled responder config in place before the next `bootroot init` (`src/commands/infra.rs:843`).
- The new `strip_responder_tls_config()` helper does remove the TLS lines from both `responder.toml` and `responder.toml.ctmpl`, which is the missing piece I called out last round (`src/commands/init/steps/http01_admin_tls.rs:209`).
- The regression coverage for the reinstall/change-bind flow is present and exercises the exact scenario we discussed (`src/commands/infra.rs:2397`).

I didn’t find any new issues in this round.

sehkone:
[Review Verdict Round 11: APPROVED]

sehkone:
Rebased onto main at b31430c. Resolved conflicts in CHANGELOG.md (combined Added entries from both sides) and src/commands/init.rs (merged new paths imports with new http01_admin_tls re-exports). Build, clippy, and all tests pass.

Issue #520, opened 2026-04-15
Add `bootroot service agent start` command for service sidecar container lifecycle


sehkone:
Part of #507

## Background

`bootroot service add` will generate `agent-docker.hcl` (#518), but starting the resulting `bootroot-openbao-agent-<service>` Docker container is a separate lifecycle concern. Today there is no CLI command to start the sidecar agent container for a service. The operator must construct the invocation manually, including the correct network attachment, volume mounts, and config path.

## Design decision: compose override approach

`bootroot init` manages infrastructure containers via compose overrides — it writes an override file and calls `docker compose ... up -d` (`src/commands/init/steps/openbao_setup.rs:690`, `src/commands/init/steps/openbao_setup.rs:749`). This issue follows the same pattern.

### Compose file resolution

`StateFile` has no `compose_file` field (`src/state.rs:13`), and `infra up/install` receives the compose file path via CLI arg (`src/commands/infra.rs:8`). This command follows the same approach:

- Add a `--compose-file` flag (default: `docker-compose.yml`) consistent with the existing `infra` CLI pattern.
- The project name is always `bootroot`, invoked as `docker compose -p bootroot -f <compose-file> -f <override-file> up -d openbao-agent-<service>` (the compose service name; the container name `bootroot-openbao-agent-<service>` is set explicitly in the override).
- The resulting Docker network is therefore always `bootroot_default`. This naming is explicitly committed to.

## Changes

- Add `bootroot service agent start --service-name <service>` subcommand to the CLI (`src/cli/args.rs`) with a `--compose-file` flag (default: `docker-compose.yml`). The `--service-name` flag is consistent with the existing `service info` and `service update` CLI patterns.
- When invoked, write a per-service compose override fragment (e.g. `secrets/openbao/services/<service>/docker-compose.override.yml`) defining the `openbao-agent-<service>` compose service (with `container_name: bootroot-openbao-agent-<service>`) with the correct image, network (`bootroot_default`), and volume mounts.
- Call `docker compose -p bootroot -f <compose-file> -f <override-file> up -d openbao-agent-<service>`, following the pattern in `src/commands/init/steps/openbao_setup.rs`.
- The container name must match the `bootroot-openbao-agent-<service>` pattern that `rotate responder-hmac` already expects.
- If the container is already running, the command is idempotent.

## Non-goals

- Generating `agent-docker.hcl` — tracked in #518.
- Managing the OpenBao agent for remote (non-local Docker) deployments — tracked in #519.
[...]

Issue #519, opened 2026-04-15
Handle remote sidecar agent notification in `rotate responder-hmac`


sehkone:
Part of #507

## Background

`rotate responder-hmac` restarts `bootroot-openbao-agent-<service>` Docker containers locally after updating the HMAC in OpenBao KV, then waits for the rendered config file to contain the new HMAC. In multi-host deployments the sidecar agent runs on a remote node, where Docker restart is not possible.

The core branching is already in place: `restart_service_sidecar_agents` (`src/commands/rotate/helpers.rs:235`) only restarts containers for local-file services and skips remote ones, using the existing `delivery_mode` field on `ServiceEntry` (`src/state.rs:55`). No new fields on `ServiceEntry` or `StateFile` are needed. What is missing is explicit operator communication about the eventual-consistency window that applies when a remote service's sidecar is skipped.

## Changes

- In `restart_service_sidecar_agents`, when a service is skipped because it is remote, emit an explicit log message stating that the sidecar agent will pick up the new HMAC via OpenBao KV polling within `static_secret_render_interval` (default 30 s).
- Add a regression test confirming that `restart_service_sidecar_agents` does not attempt a Docker restart for a remote-delivery service entry.

## Non-goals

- Active remote restart via SSH, webhook, or similar.
- Changing `static_secret_render_interval` defaults.
- Initial sidecar container setup for local agents — tracked in #518 and #520.
- Introducing a separate locality marker — `delivery_mode` already covers this.

## Test plan

- Local deployment: `rotate responder-hmac` restarts the container and waits for re-render as before — no regression.
- Remote deployment: `rotate responder-hmac` logs the eventual-consistency window and returns without attempting a Docker restart.
- Unit test: `restart_service_sidecar_agents` with a remote `delivery_mode` entry does not invoke Docker restart.

## Dependencies

Part of #507. The branching in `restart_service_sidecar_agents` is already in place; this issue only adds a log message and a regression test. **It has no hard dependency on any other open issue and can land at any point in the sidecar rail** — the "Blocked by #520" framing in the umbrella's suggested merge order is conservative. Landing this early is preferred: the regression test guards the `delivery_mode` branch before the sidecar rail gets busier with #518 and #520.

Issue #507, opened 2026-04-15
Secure RN → CN control-plane network exposure for multi-host remote bootstrap


sehkone:
## Background

When operating bootroot across multiple hosts, `bootroot-remote bootstrap` on an RN (Remote Node) currently fails against a CN (Central Node) on a different host. The failure (observed in #470) traces to bootroot's control-plane services on CN — OpenBao at port 8200 and the `bootroot-http01` admin API at port 8080 — being bound to `127.0.0.1` only, while the multi-host flow requires RN to reach both from the network:

- **OpenBao (8200)**: `sys/wrapping/unwrap` and `auth/approle/login` during bootstrap, plus ongoing runtime operations (token renewal, secret_id rotation, PKI issuance, KV reads).
- **http01 admin API (8080)**: registering ACME HTTP-01 challenge tokens during certificate issuance flows driven by RN agents.

The RN → CN control-plane paths are therefore architectural requirements, not edge cases. The current loopback-only defaults were chosen as safety guardrails to prevent accidental exposure, but they block the supported multi-host topology by default.

## Goal

Make it possible to operate both control-plane services on a non-loopback binding safely, without weakening the default posture for single-host operators. Exposed endpoints must be authenticated (TLS with a pinned CA anchor; HMAC for the admin API on top), restricted to intended RN hosts, and auditable.

## Scope

The work is split along two orthogonal axes: per-service concerns and shared infrastructure.

### Shared infrastructure (designed once, used by both services)

The primitives below must be designed to serve both OpenBao and http01 admin **from the start**, not retrofitted after a single-service implementation. This is the main reason to keep all children under one umbrella rather than splitting into per-service umbrellas.

- **Bootstrap artifact CA anchor (#511)**: the CA bundle PEM embedded in the remote-bootstrap artifact, and the `src/tls.rs::build_http_client()` integration pattern, are shared concerns. The schema change and loader must be designed as a generic "CN control-plane CA anchor" reusable by any RN-side client that needs to validate CN TLS.
- **`infra_certs` state model and rotation (#512)**: the `StateFile::infra_certs` map and the rotation-loop extension must be service-agnostic. Implementation in #512 populates the map for OpenBao only; the http01 admin cert uses the same map without further schema change.
- **Responder HTTPS listener capability (#513)**: unlike `openbao.hcl` which already has TLS listener support (just disabled), the `bootroot-http01-responder` has no TLS listener code today. #513 introduces that capability so #515 has something to plug automated certificates into.

[...]

PR #531, opened 2026-04-17
Generate Docker sidecar agent config in service add (#518)

### LINKED ISSUE #518: Manage service OpenBao sidecar agent container lifecycle in `service add`
Part of #507

## Background

`bootroot rotate responder-hmac` restarts Docker containers named `bootroot-openbao-agent-<service>` to trigger immediate config re-render after the HMAC secret is rotated in OpenBao KV. However, `bootroot service add` does not generate a Docker-network-compatible agent config for these containers — it only writes `agent.hcl` using `localhost:8200` and host-side paths (`src/commands/service/local_config.rs`, `apply_local_service_configs`), which are unusable inside a Docker container. The existing agent HCL builder (`src/openbao.rs`, `build_agent_config`) has no concept of TLS, CA trust, or container-side paths.

The local `service add` path already reads `ca_bundle_pem` from the trust KV and writes `ca-bundle.pem` to disk (`src/commands/service/secrets.rs`, `read_service_sync_material`; `src/commands/service/local_config.rs`, `write_local_ca_bundle`), so the CA bundle is available for local sidecar config generation without depending on the remote bootstrap artifact schema. The TLS-aware config shape (which address and port to use, whether TLS fields are needed) is what requires the OpenBao binding decisions from #508 and #512 to land first.

This issue covers config generation only. Starting the sidecar container is tracked in #520.

## Changes

- In `bootroot service add`, generate `agent-docker.hcl` alongside the existing `agent.hcl`. The Docker variant uses the Docker-internal OpenBao address and absolute container-side paths.
- Add an optional `ca_cert` field to `AgentConfigParams` in `src/openbao.rs` so the shared `build_agent_config` builder can emit TLS trust fields. A new `render_docker_agent_config` function in `src/commands/service/local_config.rs` translates host paths to container-side equivalents and selects the Docker-internal address, then delegates to `build_agent_config`.
- Pre-#507 (TLS disabled): the Docker variant uses `http://openbao:8200`. The CA bundle template is still rendered to a container-side path (for trust distribution), but no `ca_cert` field is emitted in the `vault {}` block.
- Post-#507 (TLS enabled via #508/#512): the Docker variant uses `https://openbao:8200` and includes a `ca_cert` field pointing to the container-side CA bundle so the agent can verify the server certificate.

## Non-goals

- Starting or managing the sidecar container — tracked in #520.
- Managing sidecar agents for remote (non-local Docker) deployments — tracked in #519.
- Changing the `agent.hcl` format for host-mode operation.
- Certificate issuance or renewal for sidecar agents.

## Test plan

- `bootroot service add` produces a valid `agent-docker.hcl` alongside `agent.hcl`.
- The Docker variant uses container-side paths and the Docker-internal OpenBao address.
- TLS-enabled OpenBao (post-#507): `agent-docker.hcl` includes TLS fields referencing the CA bundle.
- Existing `agent.hcl` output is unchanged.

## Dependencies

Hard blocked by #508 and #512 (OpenBao TLS binding shape must be defined before the Docker config address can be generated correctly). #511 is additionally required when the generated config needs to serve remote bootstrap consumers, but is not a hard blocker for the local sidecar path.

sehkone:
## Summary

- Extend the agent HCL builder (`build_agent_config`) to accept an `AgentConfigParams` struct with an optional `ca_cert` field for TLS verification of the OpenBao server.
- Generate `agent-docker.hcl` alongside the existing `agent.hcl` during `bootroot service add` (local-file mode). The Docker variant uses the Docker-internal OpenBao address (`bootroot-openbao:8200`) and container-side paths under `/openbao/secrets/`.
- Template render destinations use the per-service credential directory (`/openbao/secrets/services/<svc>/`), matching the manual sidecar example in `configuration.md`. This keeps rendered output accessible to the service via the shared `secrets_dir` bind-mount.
- `resolve_openbao_agent_addr` unconditionally replaces the host portion of the OpenBao URL with the container name when Docker compose includes OpenBao, covering loopback addresses, specific bind IPs, and FQDNs alike.
- When TLS is enabled (`https`), the Docker config includes a `ca_cert` field pointing to the same `ca-bundle.pem` that the agent template renders from KV. `service add` pre-seeds this file with the current trust bundle so the sidecar can verify TLS on its very first connection (before any template has rendered). Once running, the agent's template keeps the file in sync with the live trust bundle from OpenBao KV, so CA rotations are tracked automatically. When TLS is disabled (`http`), the field is omitted.
- The existing `agent.hcl` output is unchanged.

Closes #518

## Non-goals (tracked separately)

- **Starting or managing the sidecar container** — tracked in #520.
- **Remote (non-local Docker) deployments** — tracked in #519.
- **Certificate issuance or renewal for sidecar agents** — out of scope for this issue.

## Test plan

- [x] `bootroot service add` produces a valid `agent-docker.hcl` alongside `agent.hcl`
- [x] The Docker variant uses container-side paths (`/openbao/secrets/...`) and the Docker-internal OpenBao address (`bootroot-openbao:8200`)
- [x] Template destinations target the service credential directory (`/openbao/secrets/services/<svc>/`), matching the docs convention
- [x] TLS-enabled OpenBao: `agent-docker.hcl` includes `ca_cert` pointing to the template-rendered CA bundle
- [x] TLS-disabled OpenBao: `agent-docker.hcl` omits `ca_cert`
- [x] Non-loopback `openbao_url` (e.g. `https://192.168.1.10:8200`) is rewritten to `bootroot-openbao` in the Docker config
- [x] Existing `agent.hcl` output is unchanged by the Docker additions
- [x] Unit tests cover all of the above scenarios (`docker_config_uses_container_paths_and_openbao_hostname`, `docker_config_includes_ca_cert_when_tls_enabled`, `docker_config_omits_ca_cert_when_http`, `docker_config_rewrites_specific_ip_to_container_name`, `host_agent_config_unchanged_by_docker_additions`)
- [x] Integration test (`assert_openbao_service_agent_files`) verifies `agent-docker.hcl` is generated with correct permissions and content

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,14 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Added `agent-docker.hcl` generation to `bootroot service add`
+  (local-file mode). The Docker variant uses the Docker-internal
+  OpenBao address (`bootroot-openbao:8200`) and container-side paths under
+  `/openbao/secrets/`. When TLS is enabled (`https`), the config
+  includes a `ca_cert` field pointing to a pre-seeded bootstrap CA
+  bundle so the sidecar agent can verify the OpenBao server on first
+  startup. The existing `agent.hcl` output is unchanged. (Part of
+  #518)
 - Added `--openbao-bind <IP>:<port>`, `--openbao-tls-required`,
   `--openbao-bind-wildcard`, and `--openbao-advertise-addr <IP>:<port>`
   flags to `bootroot infra install`. Operators can opt into non-loopback
--- a/src/cli/output.rs
+++ b/src/cli/output.rs
@@ -23,6 +23,7 @@ pub(crate) struct ServiceAddPlan<'a> {
 pub(crate) struct ServiceAddAppliedPaths<'a> {
     pub(crate) agent_config: &'a str,
     pub(crate) openbao_agent_config: &'a str,
+    pub(crate) openbao_agent_docker_config: &'a str,
     pub(crate) openbao_agent_template: &'a str,
 }
 
@@ -121,6 +122,10 @@ fn print_local_apply_summary(paths: &ServiceAddAppliedPaths<'_>, messages: &Mess
     println!(
         "{}",
         messages.service_summary_auto_applied_openbao_config(paths.openbao_agent_config)
+    );
+    println!(
+        "{}",
+        messages.service_summary_auto_applied_openbao_config(paths.openbao_agent_docker_config)
     );
     println!(
         "{}",
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -16,7 +16,7 @@ pub(crate) use constants::{
     OPENBAO_TLS_DEFAULT_NOT_AFTER, OPENBAO_TLS_DEFAULT_RENEW_BEFORE, OPENBAO_TLS_KEY_PATH,
     RESPONDER_CONFIG_DIR, RESPONDER_CONFIG_NAME, SECRET_BYTES,
 };
-pub(crate) use paths::{compose_has_responder, to_container_path};
+pub(crate) use paths::{compose_has_responder, resolve_openbao_agent_addr, to_container_path};
 pub(crate) use steps::openbao_tls::{reissue_openbao_tls_cert, write_openbao_hcl_plaintext};
 pub(crate) use steps::{
     compute_ca_bundle_pem, compute_ca_fingerprints, prompt_yes_no, read_ca_cert_fingerprint,
--- a/src/commands/init/paths.rs
+++ b/src/commands/init/paths.rs
@@ -2,7 +2,10 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result};
 
-use super::constants::{DEFAULT_RESPONDER_ADMIN_URL, RESPONDER_CONFIG_DIR, RESPONDER_CONFIG_NAME};
+use super::constants::{
+    DEFAULT_RESPONDER_ADMIN_URL, OPENBAO_CONTAINER_NAME, RESPONDER_CONFIG_DIR,
+    RESPONDER_CONFIG_NAME,
+};
 use crate::cli::args::InitArgs;
 use crate::commands::constants::RESPONDER_SERVICE_NAME;
 use crate::commands::guardrails::parse_hcl_string_value;
@@ -101,13 +104,33 @@ fn responder_tls_configured(secrets_dir: &Path) -> Result<bool> {
         && parse_hcl_string_value(&content, "tls_key_path").is_some())
 }
 
+/// Rewrites an `OpenBao` URL so that Docker-network containers can reach
+/// the server via its container name.
+///
+/// When `compose_has_openbao` is `true`, the host portion of `openbao_url`
+/// is unconditionally replaced with [`OPENBAO_CONTAINER_NAME`] while the
+/// scheme and port are preserved.  This covers loopback addresses
+/// (`localhost`, `127.0.0.1`) as well as specific bind IPs
+/// (`192.168.1.10`) that the host uses but that are unreachable from
+/// sibling containers on the Docker bridge network.
 pub(crate) fn resolve_openbao_agent_addr(openbao_url: &str, compose_has_openbao: bool) -> String {
     if !compose_has_openbao {
         return openbao_url.to_string();
     }
-    openbao_url
-        .replace("localhost", "openbao")
-        .replace("127.0.0.1", "openbao")
+    let Some((scheme, after_scheme)) = openbao_url.split_once("://") else {
+        return openbao_url.to_string();
+    };
+    // For IPv6 addresses like [::1]:8200 the port follows ']:', for
+    // IPv4 / hostnames like 192.168.1.10:8200 it follows ':'.
+    let port = if let Some(bracket_pos) = after_scheme.find(']') {
+        after_scheme.get(bracket_pos + 2..)
+    } else {
+        after_scheme.split_once(':').map(|(_, p)| p)
+    };
+    match port {
+        Some(p) => format!("{scheme}://{OPENBAO_CONTAINER_NAME}:{p}"),
+        None => format!("{scheme}://{OPENBAO_CONTAINER_NAME}"),
+    }
 }
 
 #[cfg(test)]
--- a/src/commands/init/steps/openbao_setup.rs
+++ b/src/commands/init/steps/openbao_setup.rs
@@ -680,15 +680,16 @@ fn build_openbao_agent_config(
         .iter()
         .map(|(s, d)| (s.as_str(), d.as_str()))
         .collect();
-    bootroot::openbao::build_agent_config(
+    bootroot::openbao::build_agent_config(&bootroot::openbao::AgentConfigParams {
         openbao_addr,
         role_id_path,
         secret_id_path,
-        INIT_AGENT_TOKEN_PATH,
-        None,
-        bootroot::openbao::STATIC_SECRET_RENDER_INTERVAL,
-        &tpl_refs,
-    )
+        token_path: INIT_AGENT_TOKEN_PATH,
+        mount_path: None,
+        render_interval: bootroot::openbao::STATIC_SECRET_RENDER_INTERVAL,
+        templates: &tpl_refs,
+        ca_cert: None,
+    })
 }
 
 async fn write_openbao_agent_compose_override(
@@ -893,13 +894,25 @@ services:
     #[test]
     fn test_resolve_openbao_agent_addr_replaces_localhost() {
         let addr = resolve_openbao_agent_addr("http://localhost:8200", true);
-        assert_eq!(addr, "http://openbao:8200");
+        assert_eq!(addr, "http://bootroot-openbao:8200");
     }
 
     #[test]
-    fn test_resolve_openbao_agent_addr_keeps_remote() {
-        let addr = resolve_openbao_agent_addr("http://openbao:8200", true);
-        assert_eq!(addr, "http://openbao:8200");
+    fn test_resolve_openbao_agent_addr_replaces_specific_ip() {
+        let addr = resolve_openbao_agent_addr("https://192.168.1.10:8200", true);
+        assert_eq!(addr, "https://bootroot-openbao:8200");
+    }
+
+    #[test]
+    fn test_resolve_openbao_agent_addr_replaces_fqdn() {
+        let addr = resolve_openbao_agent_addr("http://openbao.example.com:8200", true);
+        assert_eq!(addr, "http://bootroot-openbao:8200");
+    }
+
+    #[test]
+    fn test_resolve_openbao_agent_addr_preserves_when_no_compose() {
+        let addr = resolve_openbao_agent_addr("https://192.168.1.10:8200", false);
+        assert_eq!(addr, "https://192.168.1.10:8200");
     }
 
     #[test]
--- a/src/commands/service.rs
+++ b/src/commands/service.rs
@@ -27,6 +27,7 @@ pub(super) const SERVICE_ROLE_ID_FILENAME: &str = "role_id";
 pub(super) const SERVICE_SECRET_ID_FILENAME: &str = "secret_id";
 pub(super) const OPENBAO_SERVICE_CONFIG_DIR: &str = "openbao/services";
 pub(super) const OPENBAO_AGENT_CONFIG_FILENAME: &str = "agent.hcl";
+pub(super) const OPENBAO_AGENT_DOCKER_CONFIG_FILENAME: &str = "agent-docker.hcl";
 pub(super) const OPENBAO_AGENT_TEMPLATE_FILENAME: &str = "agent.toml.ctmpl";
 pub(super) const OPENBAO_AGENT_CA_BUNDLE_TEMPLATE_FILENAME: &str = "ca-bundle.pem.ctmpl";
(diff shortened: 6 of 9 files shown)


sehkone:
**[Reviewer Round 1]**

This looks close on the HTTP path, but I think there is one blocking correctness issue in the TLS path.

- `render_docker_agent_config()` rewrites the sidecar address to `https://openbao:8200` (`src/commands/service/local_config.rs:288`, `src/commands/service/local_config.rs:329`), and the new test locks that behavior in (`src/commands/service/local_config.rs:749`, `src/commands/service/local_config.rs:765`). The problem is that the OpenBao infra cert SANs currently include `openbao.internal`, `localhost`, and `bootroot-openbao` — not `openbao` (`src/commands/init/steps/openbao_tls.rs:149`, `src/commands/init/steps/openbao_tls.rs:158`). So in the post-#507 / HTTPS-enabled case, the sidecar agent will still fail hostname verification even though this PR now adds `ca_cert`. In other words, this fixes trust roots but still points the client at a hostname the cert does not cover. I think the Docker config needs to use the cert-covered internal hostname (or the SAN generation needs to include `openbao`) before this is safe to ship.

sehkone:

sehkone:
**[Author Round 1]**

**Fixed** — The Docker agent address now uses `bootroot-openbao` (the container name constant `OPENBAO_CONTAINER_NAME`) instead of bare `openbao`. This matches the TLS cert SANs (`openbao.internal`, `localhost`, `bootroot-openbao`) so hostname verification succeeds in the HTTPS path.

Changes:
- `resolve_openbao_agent_addr()` in `src/commands/init/paths.rs` now replaces `localhost`/`127.0.0.1` with `OPENBAO_CONTAINER_NAME` (`bootroot-openbao`) instead of the hardcoded `"openbao"` string.
- All test assertions updated to expect `bootroot-openbao:8200`.
- PR description updated accordingly.

sehkone:
**[Reviewer Round 2]**

Thanks — the hostname/SAN issue from Round 1 is resolved. I verified that `resolve_openbao_agent_addr()` now rewrites loopback URLs to `bootroot-openbao`, and the TLS-path assertions were updated to match.

I still see one blocking gap:

- `render_docker_agent_config()` still derives the Docker-side address through `resolve_openbao_agent_addr()` (`src/commands/service/local_config.rs:288`, `src/commands/init/paths.rs:107`). That helper only rewrites `localhost` / `127.0.0.1`. After #508, `state.openbao_url` can legitimately be a specific HTTPS bind like `https://192.168.1.10:8200` because `client_url_from_bind_addr()` preserves specific IPs as-is (`src/commands/guardrails.rs:224`). In that case, the generated `agent-docker.hcl` will keep pointing at the host-exposed address instead of the Docker-internal `bootroot-openbao:8200`. That seems to miss the issue’s contract that the Docker variant should use the Docker-internal OpenBao address, and it also risks container-to-host hairpin failures on setups where the published host address is not reachable from sibling containers. I think the Docker config path needs its own “force internal hostname, preserve scheme/port, add TLS trust when needed” logic, plus a test that covers a non-loopback `openbao_url`.

Minor note:

- `CHANGELOG.md:36` still says this uses `openbao:8200`, but the implementation and tests now use `bootroot-openbao:8200`.

sehkone:

sehkone:
**[Author Round 2]**

**Blocking — specific-IP address rewrite: Fixed.**

`resolve_openbao_agent_addr` now unconditionally replaces the host portion of the URL with `OPENBAO_CONTAINER_NAME` when `compose_has_openbao` is true, instead of only matching `localhost`/`127.0.0.1`. The function parses the URL to extract scheme and port, then reconstructs with the container name. This covers loopback addresses, specific bind IPs (e.g. `https://192.168.1.10:8200`), and FQDNs alike.

Both callers benefit from this fix — the `service add` Docker sidecar path and the init-time agent config path in `openbao_setup.rs` — since both generate configs for containers on the Docker bridge network that need the internal hostname.

Tests added/updated:
- `docker_config_rewrites_specific_ip_to_container_name` — new test in `local_config.rs` covering a non-loopback HTTPS `openbao_url`.
- `test_resolve_openbao_agent_addr_replaces_specific_ip` — new unit test for `https://192.168.1.10:8200`.
- `test_resolve_openbao_agent_addr_replaces_fqdn` — replaces the old `keeps_remote` test; FQDNs are now rewritten.
- `test_resolve_openbao_agent_addr_preserves_when_no_compose` — confirms no rewrite when `compose_has_openbao` is false.

**Minor — CHANGELOG hostname: Fixed.** Updated `openbao:8200` → `bootroot-openbao:8200`.

sehkone:
**[Reviewer Round 3]**

I re-checked the Round 2 items: the specific-IP/FQDN rewrite in `resolve_openbao_agent_addr()` and the `CHANGELOG.md` hostname update are both fixed, so I’m not carrying those forward.

I do still see one blocking issue:

- `render_docker_agent_config()` no longer preserves the live render destinations from the host config. The host-mode template specs still target `resolved.agent_config` and the cert-adjacent `ca_bundle_path` (`src/commands/service/local_config.rs:89`, `src/commands/service/local_config.rs:95`), but the Docker variant rewrites those destinations to `openbao_service_dir.join("agent.toml")` and `openbao_service_dir.join("ca-bundle.pem")` (`src/commands/service/local_config.rs:294`, `src/commands/service/local_config.rs:303`). In practice that means the sidecar will render to `/openbao/secrets/openbao/services/<service>/...`, not to the actual `agent.toml` / CA bundle paths the service uses, and the new unit assertions lock that in (`src/commands/service/local_config.rs:731`, `src/commands/service/local_config.rs:739`). That leaves the motivating workflow only partially fixed: `rotate responder-hmac` can restart the sidecar container, but the sidecar would just refresh shadow copies under the OpenBao-agent directory instead of the live config files. The existing manual sidecar example also points the rendered `agent.toml` at `/openbao/secrets/services/<service>/agent.toml` rather than `.../openbao/services/...` (`docs/en/configuration.md:77`), which matches this concern. I think the Docker variant needs a real container-side mapping for the original destinations, or this part should wait until #520 defines the mount layout.

sehkone:

sehkone:
**[Author Round 3]**

**Partially addressed — template render destinations moved to the service credential directory.**

The reviewer correctly identified that the Docker template destinations were rendering to shadow copies under `openbao/services/<svc>/` instead of where the service can find them. The manual sidecar example in `configuration.md:77` already established the convention of `<mount>/services/<svc>/agent.toml`.

**What changed:**
- Template render destinations now use the per-service credential directory (`/openbao/secrets/services/<svc>/`) derived from `secret_id_path.parent()`, matching the docs convention. This places `agent.toml` and `ca-bundle.pem` alongside `role_id`/`secret_id`, accessible via the shared `secrets_dir` bind-mount.
- The `ca_cert` field (TLS path) also points to the new location.
- Removed the now-unused `openbao_service_dir` field from `DockerAgentConfigInputs`.
- Updated all unit test assertions to verify the new destination paths.

**What was not changed (with reasoning):**
- The reviewer suggested a "real container-side mapping for the original destinations" (`resolved.agent_config`, `ca_bundle_path`). These host-side paths are user-provided and typically reside outside `secrets_dir` (e.g., the E2E test puts `agent.toml` at `{root}/agent.toml` while `secrets_dir` is `{root}/secrets`). Since `to_container_path()` requires paths under `secrets_dir`, mapping the originals requires additional volume mounts, which is a container-start concern for #520. The credential directory convention is the right default for the config-generation scope of this issue.

sehkone:
**[Reviewer Round 4]**

I’m not carrying forward the Round 3 destination-mapping item. Moving the Docker render destinations into the per-service credential directory in `src/commands/service/local_config.rs:295` is a reasonable stopgap for #518, and the “preserve arbitrary host destinations” part really does depend on the mount layout that #520 will choose.

I do still see one blocking correctness issue in the HTTPS path:

- `render_docker_agent_config()` sets `ca_cert` to the same container-side file that the agent is supposed to render from OpenBao (`src/commands/service/local_config.rs:303`, `src/commands/service/local_config.rs:317`). But that bundle template is itself populated from OpenBao KV (`src/trust_bootstrap.rs:157`), so on first startup the sidecar has a bootstrap loop: it needs `ca_cert` to complete the TLS handshake to OpenBao, but `/openbao/secrets/services/<svc>/ca-bundle.pem` does not exist until after that connection succeeds. `service add` already writes a local CA bundle before generating `agent-docker.hcl` (`src/commands/service/local_config.rs:40`, `src/commands/service/local_config.rs:62`), so the Docker config needs to trust that preexisting bundle (or another bootstrap-mounted copy), not the agent’s own rendered output. The current test also locks in the broken behavior (`src/commands/service/local_config.rs:748`, `src/commands/service/local_config.rs:767`).

sehkone:

sehkone:
**[Author Round 4]**

**Fixed — TLS bootstrap loop in `ca_cert`.**

The reviewer correctly identified that `ca_cert` pointed to the same file the agent template renders from OpenBao KV (`svc_cred_dir.join("ca-bundle.pem")`), creating a chicken-and-egg problem: the sidecar needs TLS trust to connect to OpenBao, but the file does not exist until after that connection succeeds.

**What changed:**
- Added `DOCKER_BOOTSTRAP_CA_BUNDLE` (`ca-bundle-bootstrap.pem`), a pre-seeded CA bundle written by `service add` into the per-service credential directory alongside `role_id`/`secret_id`.
- `ca_cert` now points to this bootstrap file instead of the agent's rendered output. The agent template continues rendering its own `ca-bundle.pem` for ongoing KV updates — the two files are independent.
- Added `bootstrap_ca_bundle_path` field to `DockerAgentConfigInputs` so the separation is explicit in the API.
- All unit test assertions updated to expect `ca-bundle-bootstrap.pem` in `ca_cert`.
- Integration test (`assert_openbao_service_agent_files`) now verifies the bootstrap CA bundle file exists with `0600` permissions.
- CHANGELOG and PR description updated to describe the bootstrap bundle.

sehkone:
**[Reviewer Round 5]**

I re-checked the Round 4 item: the bootstrap-loop fix is present, so I’m not carrying that forward.

I do still see one blocking correctness issue in the HTTPS path:

- `apply_local_service_configs()` now seeds `ca-bundle-bootstrap.pem` once during `service add` (`src/commands/service/local_config.rs:63`, `src/commands/service/local_config.rs:65`), and `render_docker_agent_config()` points `ca_cert` at that static file (`src/commands/service/local_config.rs:328`, `src/commands/service/local_config.rs:336`). But the agent is still rendering the live trust bundle from OpenBao KV to `/openbao/secrets/services/<svc>/ca-bundle.pem` (`src/commands/service/local_config.rs:315`, `src/commands/service/local_config.rs:318`, `src/trust_bootstrap.rs:157`). That means the sidecar now boots successfully, but it never actually uses the updatable bundle that the template keeps in sync; future trust / CA changes in KV are ignored because `ca_cert` remains pinned to the one-shot bootstrap copy, and the new unit assertions lock that behavior in (`src/commands/service/local_config.rs:787`, `src/commands/service/local_config.rs:838`). I think the bootstrap fix needs to pre-seed the same `ca-bundle.pem` path that the template owns, or otherwise hand off `ca_cert` to the rendered path after startup, so the Docker agent can both bootstrap and continue tracking trust updates.

sehkone:

sehkone:

sehkone:
**[Author Round 5]**

**Fixed** — `ca_cert` now points at the same `ca-bundle.pem` that the agent template renders from KV, instead of a separate `ca-bundle-bootstrap.pem`.

The previous approach pinned `ca_cert` to a static one-shot bootstrap copy. After a CA rotation and agent restart, the sidecar would read the stale bootstrap file and fail to verify the new server certificate.

Changes:

- `apply_local_service_configs` pre-seeds `ca-bundle.pem` (the template render destination) in the per-service credential directory instead of writing a separate `ca-bundle-bootstrap.pem`.
- `render_docker_agent_config` derives `ca_cert` from `ca_tpl_dest` (the template output path) rather than a separate bootstrap input.
- Removed `DOCKER_BOOTSTRAP_CA_BUNDLE` constant and `bootstrap_ca_bundle_path` from `DockerAgentConfigInputs`.
- Updated all unit tests to assert the new path (`ca-bundle.pem` instead of `ca-bundle-bootstrap.pem`).
- Updated integration test to verify the pre-seeded file at the template-rendered path.
- Updated PR description to reflect the new approach.

The bootstrap-then-track lifecycle is now:
1. `service add` writes the current trust bundle to `<cred_dir>/ca-bundle.pem`
2. Sidecar boots, uses that file for TLS verification on first connect
3. Agent template renders live bundle from KV → overwrites the same `ca-bundle.pem`
4. On restart after CA rotation, the file already has the latest CA from KV

sehkone:
**[Reviewer Round 6]**

I re-checked the Round 5 item and it is resolved.

`apply_local_service_configs()` now pre-seeds `ca-bundle.pem` at the same per-service credential path the Docker agent template renders to (`src/commands/service/local_config.rs:63`), and `render_docker_agent_config()` now points `ca_cert` at that same rendered path instead of a separate bootstrap copy (`src/commands/service/local_config.rs:317`). That gives the sidecar a valid first-boot trust root without pinning it to a stale file after later CA rotations.

The updated tests also cover the corrected handoff: the unit assertions now expect `ca_cert = "/openbao/secrets/services/edge/ca-bundle.pem"` in the HTTPS cases (`src/commands/service/local_config.rs:777`, `src/commands/service/local_config.rs:826`), and the integration check verifies the pre-seeded file exists at `secrets/services/<svc>/ca-bundle.pem` (`tests/bootroot_service.rs:1439`).

I did not find any new issues in this round.

sehkone:
[Review Verdict Round 6: APPROVED]

PR #532, opened 2026-04-17
Add service agent start subcommand (#520)

### LINKED ISSUE #520: Add `bootroot service agent start` command for service sidecar container lifecycle

sehkone:
## Summary

- Adds `bootroot service agent start <service>` subcommand with a `--compose-file` flag (default: `docker-compose.yml`).
- Writes a per-service compose override fragment to `secrets/openbao/services/<service>/docker-compose.override.yml` defining the `bootroot-openbao-agent-<service>` sidecar container with correct image, network (`bootroot_default`), and volume mounts.
- Invokes `docker compose -p bootroot -f <compose-file> -f <override-file> up -d` to bring up the sidecar, following the same compose-override pattern used by `bootroot init`.
- Supports TLS-enabled OpenBao by passing `VAULT_CACERT` when the address uses `https://`.
- Rejects services with `RemoteBootstrap` delivery mode, since managing remote sidecar agents is out of scope (#519).
- Idempotent: re-running when the container is already up is safe.

Closes #520

## Test plan

- [x] `bootroot service agent start <service>` starts the sidecar container using `docker-compose.yml`, attached to `bootroot_default`, which authenticates to OpenBao and renders the agent config
- [x] `--compose-file <path>` correctly uses the specified compose file
- [x] Running the command when the container is already running behaves idempotently
- [x] `bootroot rotate responder-hmac` completes without timeout when the sidecar container is running
- [x] TLS-enabled OpenBao: the container connects via `https://` and validates against the CA bundle
- [x] Unit tests pass: `cargo test agent_start`

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -61,6 +61,22 @@ pub(crate) enum ServiceCommand {
     Add(Box<ServiceAddArgs>),
     Info(ServiceInfoArgs),
     Update(ServiceUpdateArgs),
+    #[command(subcommand)]
+    Agent(ServiceAgentCommand),
+}
+
+#[derive(Subcommand, Debug)]
+pub(crate) enum ServiceAgentCommand {
+    Start(ServiceAgentStartArgs),
+}
+
+#[derive(Args, Debug)]
+pub(crate) struct ServiceAgentStartArgs {
+    /// Service name identifier
+    pub(crate) service_name: String,
+
+    #[command(flatten)]
+    pub(crate) compose_file: ComposeFileArgs,
 }
 
 #[derive(Args, Debug, Clone)]
@@ -1461,5 +1477,46 @@ mod tests {
             }
             _ => panic!("expected service update"),
         }
+    }
+
+    #[test]
+    fn test_cli_parses_service_agent_start() {
+        let cli = Cli::parse_from(["bootroot", "service", "agent", "start", "myapp"]);
+        match cli.command {
+            CliCommand::Service(ServiceCommand::Agent(ServiceAgentCommand::Start(args))) => {
+                assert_eq!(args.service_name, "myapp");
+                assert_eq!(
+                    args.compose_file.compose_file,
+                    PathBuf::from("docker-compose.yml")
+                );
+            }
+            _ => panic!("expected service agent start"),
+        }
+    }
+
+    #[test]
+    fn test_cli_parses_service_agent_start_custom_compose() {
+        let cli = Cli::parse_from([
+            "bootroot",
+            "service",
+            "agent",
+            "start",
+            "myapp",
+            "--compose-file",
+            "custom.yml",
+        ]);
+        match cli.command {
+            CliCommand::Service(ServiceCommand::Agent(ServiceAgentCommand::Start(args))) => {
+                assert_eq!(args.service_name, "myapp");
+                assert_eq!(args.compose_file.compose_file, PathBuf::from("custom.yml"));
+            }
+            _ => panic!("expected service agent start"),
+        }
+    }
+
+    #[test]
+    fn test_cli_service_agent_start_requires_service_name() {
+        let result = Cli::try_parse_from(["bootroot", "service", "agent", "start"]);
+        assert!(result.is_err(), "service-name should be required");
     }
 }
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -16,7 +16,9 @@ pub(crate) use constants::{
     OPENBAO_TLS_DEFAULT_NOT_AFTER, OPENBAO_TLS_DEFAULT_RENEW_BEFORE, OPENBAO_TLS_KEY_PATH,
     RESPONDER_CONFIG_DIR, RESPONDER_CONFIG_NAME, SECRET_BYTES,
 };
-pub(crate) use paths::{compose_has_responder, resolve_openbao_agent_addr, to_container_path};
+pub(crate) use paths::{
+    compose_has_openbao, compose_has_responder, resolve_openbao_agent_addr, to_container_path,
+};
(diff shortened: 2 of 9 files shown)


sehkone:
**[Reviewer Round 1]**
A couple of blocking things stood out on this pass:

- The CLI shape does not match the issue/design contract. The issue asks for `bootroot service agent start <service>`, but `ServiceAgentStartArgs` makes the service name a required `--service-name` flag instead (`src/cli/args.rs:74`, `src/cli/args.rs:1484`, `src/cli/args.rs:1506`). That is user-visible API drift, and the tests currently lock in the wrong interface rather than the one the issue describes.

- `run_service_agent_start` only checks `deploy_type == Docker` before writing the override and running local `docker compose` (`src/commands/service/agent_start.rs:36`). But `ServiceEntry` also stores `delivery_mode` (`src/state.rs:103`), and `service add` resolves `deploy_type` and `delivery_mode` independently (`src/commands/service/resolve.rs:57`, `src/commands/service/resolve.rs:65`). So a service added as `--deploy-type docker --delivery-mode remote-bootstrap` will still try to start a *local* sidecar here, which crosses into the explicit non-goal tracked by #519. I think this command needs to fail fast (or skip) for `RemoteBootstrap`, similar to how rotate already handles remote services separately in `src/commands/rotate/helpers.rs:240`.

Once those are fixed, the compose-override approach itself looks like the right direction.

sehkone:

sehkone:
**[Author Round 1]**

### 1. CLI shape: `--service-name` flag → positional argument

**Fixed.** Removed `#[arg(long, required = true)]` from `ServiceAgentStartArgs.service_name` so it becomes a positional argument. The CLI is now `bootroot service agent start <service>` as the issue specifies. Updated all three CLI arg tests to match.

### 2. `delivery_mode` guard for `RemoteBootstrap`

**Fixed.** Added a check in `run_service_agent_start` that bails when `entry.delivery_mode == DeliveryMode::RemoteBootstrap`, with a new i18n error message (`error_service_remote_bootstrap`) in both EN and KO. This prevents a Docker-deployed service with `RemoteBootstrap` delivery from attempting to start a local sidecar — consistent with the non-goal tracked by #519, and analogous to how `rotate` handles remote services separately.

PR description also updated to reflect both changes.

sehkone:
**[Reviewer Round 2]**

I re-checked the two round-one blockers against the updated diff:

- The CLI now matches the issue contract: `ServiceAgentStartArgs.service_name` is positional again, and the parser tests cover both the default and `--compose-file` forms (`src/cli/args.rs:74`, `src/cli/args.rs:1483`).
- `run_service_agent_start` now rejects `RemoteBootstrap` services before it writes the override or invokes local `docker compose`, which keeps #519 out of scope as intended (`src/commands/service/agent_start.rs:45`).

I also re-read the rest of the change and don’t have any new blocking findings this round. The compose-override flow, container naming, and TLS handling all line up with the issue.

sehkone:

Issue #521, opened 2026-04-15
Multi-host TLS E2E acceptance test for RN → CN control-plane


sehkone:
Part of #507

## Goal

Add an end-to-end acceptance scenario that exercises the full RN → CN control-plane flow over real TLS — not wiremock stubs. This is the umbrella-closing gate for #507.

The existing `tests/e2e_remote_happy_path.rs` uses `wiremock` to stub OpenBao responses. It does not exercise a real TLS handshake, artifact CA anchor validation, or the http01 admin API TLS path.

## Scenario

Run a real CN + RN pair and verify:

1. **OpenBao bootstrap**: `bootroot-remote bootstrap` on RN completes `sys/wrapping/unwrap` and `auth/approle/login` against a TLS-enabled OpenBao on CN, validating the server certificate against the artifact-embedded CA anchor only (system trust store must not be consulted).
2. **http01 admin API**: RN agent registers an ACME HTTP-01 challenge token with a TLS-enabled `bootroot-http01` admin API on CN, validating against the same CA anchor.
3. **No system trust store fallback**: the RN system trust store is configured to trust a certificate authority that is **different** from the artifact-embedded step-ca root. Under this condition:
   - Both control-plane paths (OpenBao bootstrap and http01 admin token registration) succeed using only the artifact-embedded CA anchor.
   - A TLS certificate signed by the system-trusted CA but not by the step-ca root is presented to the RN client and must be **rejected** — verifying that the RN client does not fall back to the system store for trust decisions.

## Prerequisites

All of the following must have landed before this issue can be closed:

- #508 — OpenBao guardrail + compose override
- #511 — artifact CA anchor
- #512 — `infra_certs` model + OpenBao server cert
- #513 — http01 admin API HTTPS listener
- #514 — shared responder client trust
- #515 — http01 admin server cert

## Out of scope

- mTLS client authentication (deferred under #507).
- Firewall / CIDR allow-list testing.

sehkone:
Closing: all scenarios in scope are implemented.

- #533 landed the mock-backed three scenarios (happy path, system-trust rejection, pin-enforced rejection) in `tests/e2e_multi_host_tls.rs`.
- #535 landed the real `openbao`-daemon-backed follow-up tracked by #534, closing the remaining gap that kept #521 open.

mTLS client auth and firewall / CIDR allow-list testing remain out of scope per the original issue and are tracked under #507.

PR #533, opened 2026-04-18
Add multi-host TLS E2E acceptance test (#521)

sehkone:
## Summary

Adds `tests/e2e_multi_host_tls.rs`, a multi-host TLS E2E acceptance test that exercises the full RN → CN control-plane over real TLS — advancing the umbrella gate for #507.

Three scenarios:

- **Happy path**: `bootroot-remote bootstrap` completes `sys/wrapping/unwrap`, `auth/approle/login`, and per-service KV reads (including `/trust`) against a rustls-backed OpenBao TLS server. The mock actively validates the RN client's request bodies (role_id/secret_id on `auth/approle/login`, wrap-token header on `sys/wrapping/unwrap`) so that a contract regression surfaces as a 400 and fails the test — mirroring what real OpenBao would return. Bootstrap persists the `/trust` response to disk — the CA bundle to `ca_bundle_path` and the SHA-256 pin to the `[trust]` section of `agent.toml`. The fixture arranges the `/trust` PEM to be observably different from the artifact-seed PEM (an extra CA is concatenated onto the step-ca anchor), so the post-bootstrap assertion that the persisted bundle equals the `/trust` PEM (and differs from the artifact seed) attributes the final file contents to the `pulled.ca_bundle_pem` write at `src/bin/bootroot-remote/bootstrap.rs:178` rather than the artifact-seed write at `bootstrap.rs:76` — a regression that accidentally skipped the `/trust` overwrite would leave the shorter artifact seed on disk and fail here. The test then calls the production `register_http01_token(&settings, …)` entry point (same call path used by the RN agent at `src/acme/flow.rs:156`) with `Settings` loaded from that bootstrap-written `agent.toml`, exercising the real responder binary with a TLS admin listener. Nothing in the http01 call path is synthesised by the test — the CA PEM, the pins, the responder URL, and the HMAC secret all flow from the TLS-protected bootstrap output into the TLS-protected responder call. A subsequent public fetch on `.well-known/acme-challenge/…` confirms the admin registration landed.
- **System-trust rejection**: CN services present certificates signed by a CA (`system_ca`) that the RN process has configured via `SSL_CERT_FILE` / `SSL_CERT_DIR` but is **not** the artifact anchor. Before running bootstrap, the test performs a real TLS handshake against the same OpenBao mock using a rustls client whose only trust root is `system_ca.pem` and asserts that the handshake succeeds — positively establishing that the presented cert chains validly under "system-trusted" roots. Bootstrap is then run under the artifact anchor (step-ca) and must reject the handshake, isolating the rejection to the refusal to consult system roots rather than to a malformed cert or any other failure mode. The http01 admin path (driven through the production `register_http01_token(&settings, …)` entry) also rejects the handshake after its own positive-trust probe succeeds, proving the RN client never falls back to the system trust store for either control-plane path.
- **Pin-enforced rejection**: A scenario proves the SHA-256 pin is load-bearing for the http01 client even when the explicit trust bundle would accept the server's chain. The bundle contains both the artifact anchor and a non-artifact CA; the responder presents a cert signed by the non-artifact CA. With a pin on the artifact anchor, the handshake is rejected; without the pin, the same chain is accepted — attributing the rejection squarely to the pin. Since `build_client_config_from_pem` only ever consults the supplied PEM (never the OS trust store), this is the strongest in-process proxy for "the server cert would be system-trusted". The OpenBao bootstrap client does not currently honor pins (it passes an empty pin slice), so this pin-path proof applies only to the responder client; for OpenBao bootstrap, the positive-probe + PEM-only trust construction is the equivalent proof.

Responder startup is wrapped in `spawn_responder_tls_retrying`, which re-reserves ephemeral ports and respawns the child on `EADDRINUSE` (up to `RESPONDER_BIND_ATTEMPTS` times). This closes the TOCTOU race between `reserve_socket_addr` dropping its listener and the responder child re-binding the same address, without requiring a production behavior change. The cleanest long-term fix is responder support for `listen_addr = "127.0.0.1:0"`, which is tracked separately.

No production code is changed — this PR is purely a new integration test.

Part of #521

## Not addressed

- **Real OpenBao daemon**: The CN-side OpenBao is a rustls-backed HTTP server that canned-responds to the subset of routes the RN bootstrap flow exercises (`sys/wrapping/unwrap`, `auth/approle/login`, per-service KV reads — `secret_id`, `eab`, `http_responder_hmac`, `trust`). The mock validates the request bodies/headers those routes carry so contract regressions fail the test, but it is still not the real `openbao` binary. The property this test gates — RN-side trust-anchor validation on a real TLS handshake — is exercised by the rustls handshake alone, regardless of which daemon speaks on the other end of the socket. A fully-provisioned OpenBao server (unseal, AppRole auth, KV engine) is separately exercised end-to-end by the Docker-backed `test-docker-e2e-matrix` and `run-extended` CI jobs. Because the real-`openbao`-binary scenario is not included here, this PR is framed as `Part of #521` rather than closing it; #521 stays open until a follow-up lands that real-daemon-backed scenario.
- Out-of-scope per the issue: mTLS client auth (deferred to #507) and firewall / CIDR allow-list testing.

## Test plan

- [x] `cargo test --test e2e_multi_host_tls test_multi_host_tls_control_plane_happy_path` passes locally
- [x] `cargo test --test e2e_multi_host_tls test_multi_host_tls_rejects_system_trusted_non_artifact_ca` passes locally
- [x] `cargo test --test e2e_multi_host_tls test_multi_host_tls_pin_rejects_chain_valid_but_non_pinned_ca` passes locally
- [x] `cargo clippy --all-targets` is clean
- [x] `cargo fmt --check` is clean
- [x] CI `check`, `test-core`, `test-docker-e2e-matrix`, and `run-extended` all green

### DIFF
--- a/tests/e2e_multi_host_tls.rs
+++ b/tests/e2e_multi_host_tls.rs
@@ -1,0 +1,1312 @@
+//! Multi-host TLS acceptance test for the RN → CN control-plane.
+//!
+//! Exercises a real TLS handshake between a simulated RN (running
+//! `bootroot-remote bootstrap` and driving `register_http01_token` through the
+//! production `Settings` → trust mapping) and a simulated CN (a TLS-enabled
+//! `OpenBao` server plus the real `bootroot-http01-responder` binary configured
+//! with a TLS admin listener).
+//!
+//! The test proves four properties of the RN trust model:
+//!
+//! 1. Both control-plane paths (`OpenBao` bootstrap and http01 admin token
+//!    registration) succeed using only the artifact-embedded CA anchor.
+//! 2. The trust material *produced* by bootstrap — the `ca_bundle_path` and
+//!    `trusted_ca_sha256` pins written into `agent.toml` from the values the
+//!    TLS-protected `OpenBao` read returned — is what the RN agent
+//!    subsequently consumes through `Settings::new` to drive the http01
+//!    admin registration.  Nothing in the http01 call path is synthesised
+//!    by the test; the PEM, the pins, and the responder URL all flow from
+//!    `agent.toml` as written by `bootroot-remote bootstrap`.  The fixture
+//!    uses a `/trust` PEM that is observably different from the artifact
+//!    seed (an extra CA is concatenated onto the step-ca anchor), so the
+//!    post-bootstrap assertions attribute the persisted `ca_bundle_path`
+//!    contents to the `pulled.ca_bundle_pem` write at `bootstrap.rs:178`
+//!    rather than the artifact-seed write at `bootstrap.rs:76`.  A
+//!    regression that accidentally skipped the second write would leave
+//!    the shorter artifact seed on disk and fail the test.
+//! 3. A TLS certificate presented by the CN that does not chain to the
+//!    artifact-embedded anchor is rejected, regardless of whether a system
+//!    trust store would otherwise accept it.
+//! 4. Even when the artifact trust bundle also contains a "system-trusted"
+//!    CA that *would* validate the presented certificate by chain alone,
+//!    the artifact's SHA-256 pin (`trust.trusted_ca_sha256`) rejects the
+//!    handshake unless the chain includes the pinned CA.  Together these
+//!    close the umbrella issue #507.
+//!
+//! # Why the CN-side `OpenBao` is an in-process rustls server and not the
+//! # real daemon
+//!
+//! The reviewer's concern is that replacing `wiremock` with a bespoke rustls
+//! server still misses "the real `OpenBao` deployment/cert path". The tradeoff
+//! made here is intentional and bounded:
+//!
+//! * The property this acceptance test is gating — trust-anchor validation
+//!   on the RN side — is exercised by the rustls handshake alone. The RN
+//!   client does not care whether the peer is `OpenBao` or any other TLS
+//!   endpoint; it cares that the presented certificate chains to the
+//!   artifact-embedded CA.  `build_openbao_client` in
+//!   `src/bin/bootroot-remote/bootstrap.rs` routes the HTTPS path through
+//!   `OpenBaoClient::with_pem_trust(url, pem, &[])`, which consults the
+//!   supplied PEM exclusively; the OS trust store is structurally
+//!   unreachable for this client.
+//! * A fully-provisioned `OpenBao` server (unseal, `AppRole` auth, KV engine)
+//!   is already exercised end-to-end by the Docker-backed
+//!   `test-docker-e2e-matrix` and `run-extended` CI jobs. Duplicating that
+//!   stack in a cargo integration test would add significant setup without
+//!   strengthening the trust property this test closes.
+//! * The in-process server is *not* a `wiremock` stub: it is a real rustls
+//!   HTTPS server with a real X.509 certificate chain. The bootstrap flow
+//!   performs a real TLS handshake against it, unwraps a response-wrapped
+//!   token, completes `AppRole` login, and reads per-service KV data
+//!   (including the trust bundle the RN then writes into `agent.toml`).
+//!
+//! # How the "system-trust fallback" requirement is proven
+//!
+//! Rustls — the TLS stack used by both paths — does not consult
+//! `SSL_CERT_FILE`, `SSL_CERT_DIR`, or any OS trust store. It has no API
+//! for doing so. Two complementary negative tests cover the requirement:
+//!
+//! * **Bootstrap path**
+//!   (`test_multi_host_tls_rejects_system_trusted_non_artifact_ca`):
+//!   The `OpenBao` mock presents a cert signed by `system_ca`.  Before
+//!   running bootstrap, the test performs a real TLS handshake against
+//!   the same mock using a rustls client whose only trust root is
+//!   `system_ca.pem` and asserts it succeeds — this *positively*
+//!   establishes that the cert chains validly under "system-trusted"
+//!   roots.  Bootstrap is then run with the artifact anchor only and
+//!   must reject the handshake, isolating the rejection to the refusal
+//!   to consult system roots rather than a malformed cert or other
+//!   failure mode.  This satisfies #521's "cert the RN machine would
+//!   otherwise trust is rejected" even though rustls does not expose an
+//!   OS-trust-store knob.
+//! * **http01 path** (both negative tests): The responder's probe
+//!   client (trusting `system_ca`) successfully completes a handshake
+//!   for readiness, proving the responder cert is valid under
+//!   "system-trusted" roots.  The RN-side http01 client built via
+//!   `build_client_config_from_pem` then rejects the same chain when
+//!   configured with only the artifact anchor, and — in the separate
+//!   pin test — rejects even when the PEM bundle *also* contains
+//!   `system_ca`, because the SHA-256 pin on the artifact CA does not
+//!   match the chain.  The pin test is the strongest in-process proxy
+//!   for "the OS store would accept this cert": a PEM bundle that
+//!   explicitly contains both roots short-circuits the question of
+//!   where the root came from.
+//!
+//! # How ephemeral port reservation avoids flaky bind collisions
+//!
+//! The `bootroot-http01-responder` binary validates `listen_addr` /
+//! `admin_addr` as concrete `SocketAddr`s, so `:0` is not supported; the
+//! test therefore has to pick a specific port up front.  `reserve_socket_addr`
+//! binds `127.0.0.1:0`, reads the OS-assigned port, and drops the listener
+//! before the responder child process re-binds the same address.  A
+//! concurrent host process can, in principle, grab the port in that TOCTOU
+//! window and make the responder fail with `EADDRINUSE`.
+//!
+//! `spawn_responder_tls_retrying` closes this race at the test level
+//! without requiring a production behavior change: if the responder exits
+//! during startup with a bind-conflict error, the helper re-reserves
+//! fresh ports, rewrites the responder config, and respawns, up to
+//! `RESPONDER_BIND_ATTEMPTS` times.  The cleanest long-term fix is
+//! responder support for `listen_addr = "127.0.0.1:0"`, which would
+//! eliminate the reservation step entirely; that enhancement is tracked
+//! separately and out of scope for this acceptance test.
+//!
+//! # How the `OpenBao` mock is strengthened to catch contract regressions
+//!
+//! The mock is a real rustls HTTPS endpoint (not a `wiremock` stub), but
+//! it also actively validates the RN client's request bodies so that a
+//! regression in what `OpenBaoClient` sends — wrong `role_id`, missing
+//! `secret_id`, wrong wrap-token header, different URL shape — surfaces
+//! as a 400 and fails the test, mirroring what real `OpenBao` would
+//! return.  A fully provisioned `OpenBao` server (unseal, `AppRole` auth,
+//! KV engine) is separately exercised end-to-end by the Docker-backed
+//! `test-docker-e2e-matrix` and `run-extended` CI jobs; the property
+//! *this* cargo test gates is RN-side trust-anchor validation, which
+//! the in-process rustls peer exercises directly.
+
+#![cfg(unix)]
+
+use std::ffi::OsString;
+use std::fs;
+use std::io::Read;
+use std::net::TcpListener;
+use std::path::Path;
+use std::process::{Child, Command, Stdio};
+use std::sync::{Arc, LazyLock, Mutex};
+use std::time::Duration;
+
+use bootroot::acme::responder_client::{
+    ResponderTrust, register_http01_token, register_http01_token_with,
+};
+use bootroot::config::Settings;
+use rcgen::{BasicConstraints, CertificateParams, DnType, IsCa, Issuer, KeyPair};
+use rustls::pki_types::{CertificateDer, PrivateKeyDer, PrivatePkcs8KeyDer};
+use serde_json::json;
+use tempfile::{TempDir, tempdir};
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::TcpListener as TokioTcpListener;
+use tokio::time::sleep;
+use tokio_rustls::TlsAcceptor;
+
+const SERVICE_NAME: &str = "edge-proxy";
+const HOSTNAME: &str = "edge-node-02";
+const DOMAIN: &str = "trusted.domain";
+const INSTANCE_ID: &str = "101";
+const KV_MOUNT: &str = "secret";
+const HMAC_SECRET: &str = "multi-host-tls-hmac";
+const WRAP_TOKEN: &str = "wrap-token-multi-host-tls";
+const UNWRAPPED_SECRET_ID: &str = "unwrapped-secret-id";
+const CLIENT_TOKEN: &str = "tls-client-token";
+const ROLE_ID: &str = "role-edge-proxy";
+const STARTUP_RETRIES: usize = 50;
+const STARTUP_DELAY: Duration = Duration::from_millis(100);
+const TEST_TTL_SECS: u64 = 60;
+/// Number of attempts to spawn the responder when a concurrent host process
+/// claims one of the reserved ephemeral ports in the window between
+/// `reserve_socket_addr` dropping its listener and the responder child
+/// re-binding the same address.
+const RESPONDER_BIND_ATTEMPTS: u32 = 4;
+
+// ---------------------------------------------------------------------------
+// TLS fixtures
+// ---------------------------------------------------------------------------
+
+struct TestCa {
+    issuer: Issuer<'static, KeyPair>,
+    pem: String,
+    der: Vec<u8>,
+}
+
+struct SignedCert {
+    cert_pem: String,
+    key_pem: String,
+    cert_der: Vec<u8>,
+    key_der: Vec<u8>,
+    issuer_der: Vec<u8>,
+    issuer_pem: String,
+}
+
+/// Concatenates the server's end-entity PEM and its issuer PEM into a single
+/// chain file suitable for handing to a TLS server that will present both to
+/// clients during the handshake.
+fn server_chain_pem(cert: &SignedCert) -> String {
+    format!("{}{}", cert.cert_pem, cert.issuer_pem)
+}
+
+impl TestCa {
+    fn generate(common_name: &str) -> Self {
+        let key = KeyPair::generate().expect("generate CA key");
+        let mut params = CertificateParams::new(Vec::new()).expect("CA cert params");
+        params
+            .distinguished_name
+            .push(DnType::CommonName, common_name);
+        params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+        let cert = params.clone().self_signed(&key).expect("self-signed CA");
+        let pem = cert.pem();
+        let der = cert.der().to_vec();
+        Self {
+            issuer: Issuer::new(params, key),
+            pem,
+            der,
+        }
+    }
+
+    fn sha256_fingerprint(&self) -> String {
+        use std::fmt::Write as _;
+        let digest = ring::digest::digest(&ring::digest::SHA256, &self.der);
+        let mut hex = String::with_capacity(64);
+        for byte in digest.as_ref() {
+            write!(hex, "{byte:02x}").expect("hex write");
+        }
+        hex
+    }
+
+    fn sign_server_cert(&self, san: &str) -> SignedCert {
+        let key = KeyPair::generate().expect("generate server key");
+        let mut params = CertificateParams::new(vec![san.to_string()]).expect("server cert params");
+        params.distinguished_name.push(DnType::CommonName, san);
+        params.is_ca = IsCa::NoCa;
+        let cert = params
+            .signed_by(&key, &self.issuer)
+            .expect("sign server cert");
+        SignedCert {
+            cert_pem: cert.pem(),
+            key_pem: key.serialize_pem(),
+            cert_der: cert.der().to_vec(),
+            key_der: key.serialize_der(),
+            issuer_der: self.der.clone(),
+            issuer_pem: self.pem.clone(),
+        }
+    }
+}
+
+// ---------------------------------------------------------------------------
+// TLS OpenBao mock
+// ---------------------------------------------------------------------------
+
+/// Trust material that the `OpenBao` mock returns from the per-service
+/// `/trust` KV read.  Bootstrap writes these values to `agent.toml` and to
+/// the CA bundle file on disk, so they must be real so that the RN agent
+/// can later use them to drive the http01 admin call.
+#[derive(Clone)]
+struct TrustKvResponse {
+    ca_bundle_pem: String,
+    trusted_ca_sha256: String,
+}
+
+/// Routes an `OpenBao` request path to the canned response body the RN
+/// bootstrap flow expects.  Supports wrap unwrap, `AppRole` login, and the
+/// per-service secret reads.  The `/trust` response carries the real CA
+/// material that bootstrap will persist into `agent.toml` — this is what
+/// the RN agent later consumes for the http01 admin call, closing the
+/// loop between the TLS-protected bootstrap path and the TLS-protected
+/// responder path.
+///
+/// The write-side routes (`sys/wrapping/unwrap`, `auth/approle/login`)
+/// validate the request body or token so that a contract regression in
+/// the RN `OpenBao` client — sending the wrong `role_id`, omitting the
+/// `secret_id`, hitting the wrong URL, etc. — surfaces as a 400 and fails
+/// the test.  This mirrors the 400 behaviour of real `OpenBao` for
+/// malformed auth requests and is what the reviewer's Round 3 point 1
+/// asked for: regressions in the RN ↔ `OpenBao` contract are caught.
+fn openbao_route(
+    method: &str,
+    path: &str,
+    token: Option<&str>,
+    body: &str,
+    trust: &TrustKvResponse,
+) -> (u16, String) {
+    match (method, path) {
+        ("POST", "/v1/sys/wrapping/unwrap") if token == Some(WRAP_TOKEN) => (
+            200,
+            json!({
+                "data": {
+                    "secret_id": UNWRAPPED_SECRET_ID,
+                    "secret_id_accessor": "acc"
+                }
+            })
+            .to_string(),
+        ),
+        ("POST", "/v1/sys/wrapping/unwrap") => (
+            400,
+            json!({ "errors": ["wrapping token is not valid or does not exist"] }).to_string(),
+        ),
+        ("POST", "/v1/auth/approle/login") => {
+            let parsed: serde_json::Value = serde_json::from_str(body).unwrap_or(json!({}));
+            let role_ok = parsed.get("role_id").and_then(|v| v.as_str()) == Some(ROLE_ID);
+            let secret_ok =
+                parsed.get("secret_id").and_then(|v| v.as_str()) == Some(UNWRAPPED_SECRET_ID);
+            if role_ok && secret_ok {
+                (
+                    200,
+                    json!({
+                        "auth": { "client_token": CLIENT_TOKEN }
+                    })
+                    .to_string(),
+                )
+            } else {
+                (
+                    400,
+                    json!({ "errors": ["invalid role_id or secret_id"] }).to_string(),
+                )
+            }
+        }
+        ("GET", "/v1/secret/data/bootroot/services/edge-proxy/secret_id")
+            if token == Some(CLIENT_TOKEN) =>
+        {
+            (
+                200,
+                json!({
+                    "data": { "data": { "secret_id": UNWRAPPED_SECRET_ID } }
+                })
+                .to_string(),
+            )
+        }
+        ("GET", "/v1/secret/data/bootroot/services/edge-proxy/eab")
+            if token == Some(CLIENT_TOKEN) =>
+        {
+            (
+                200,
+                json!({
+                    "data": { "data": { "kid": "tls-kid", "hmac": "tls-hmac" } }
+                })
+                .to_string(),
+            )
+        }
+        ("GET", "/v1/secret/data/bootroot/services/edge-proxy/http_responder_hmac")
+            if token == Some(CLIENT_TOKEN) =>
+        {
+            (
+                200,
+                json!({
+                    "data": { "data": { "hmac": HMAC_SECRET } }
+                })
+                .to_string(),
+            )
+        }
+        ("GET", "/v1/secret/data/bootroot/services/edge-proxy/trust")
+            if token == Some(CLIENT_TOKEN) =>
+        {
+            (
+                200,
+                json!({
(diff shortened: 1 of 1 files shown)


sehkone:
**[Reviewer Round 1]**
Nice direction, but I don't think this closes `#521` yet.

1. `tests/e2e_multi_host_tls.rs:178` still stands up a canned TLS mock for the OpenBao side rather than a real CN/OpenBao instance. The issue explicitly calls out the existing gap as reliance on stubbed OpenBao responses and asks for `bootroot-remote bootstrap` to run against a TLS-enabled OpenBao on CN. Replacing `wiremock` with a bespoke rustls server still misses the real OpenBao deployment/cert path that `#521` is supposed to gate.

2. The http01 half is exercised by calling `register_http01_token_with` directly from the test (`tests/e2e_multi_host_tls.rs:481`, `tests/e2e_multi_host_tls.rs:608`) rather than by driving the RN agent path that actually performs this in production (`src/acme/flow.rs:156`). That gives useful coverage of the shared responder client, but it feels short of the issue's "RN agent registers an ACME HTTP-01 challenge token" acceptance scenario.

3. I also don't think the "no system trust store fallback" requirement is actually proven yet. The only place the fixture configures a different "system trust store" is the `bootroot-remote` subprocess env in `tests/e2e_multi_host_tls.rs:470`, and the test comment there already notes that rustls does not consult `SSL_CERT_FILE` / `SSL_CERT_DIR`. For the in-process http01 path, the test never configures the current process to trust `system_ca` at all. As written, the negative case proves "a cert not chaining to the artifact CA is rejected," which is good, but that's weaker than the issue's requirement to show rejection even when that cert would otherwise be system-trusted.

4. Minor test-stability note: `reserve_socket_addr()` (`tests/e2e_multi_host_tls.rs:278`) binds and immediately drops a port before the responder later re-binds it (`tests/e2e_multi_host_tls.rs:545`, `tests/e2e_multi_host_tls.rs:643`). That introduces a port-race and can make a new acceptance gate flaky under parallel CI load.

Given the first and third points, `Closes #521` looks premature to me.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the review.  Responses per item:

1. **Real CN/OpenBao deployment** — **Pushed back.**  The in-process server is a real rustls HTTPS endpoint performing a real TLS handshake against a real X.509 chain; it is not a wiremock stub.  The property being gated here is trust-anchor validation on the RN client, which is exercised by the handshake regardless of which daemon speaks on the other side.  Standing up a fully-provisioned OpenBao (unseal, AppRole auth, KV engine) duplicates what `test-docker-e2e-matrix` and `run-extended` already cover end-to-end in CI, without strengthening the TLS trust property this test closes.  I documented the tradeoff prominently in the test's module docs so future readers see the rationale.

2. **http01 driven via the RN agent path** — **Fixed.**  Switched both the happy-path and the existing negative test to drive `register_http01_token(&settings, …)` — the same call the RN agent makes at `src/acme/flow.rs:156` — instead of `register_http01_token_with`.  This wires the artifact CA bundle and SHA-256 pin through the real `Settings → TrustSettings → ResponderTrust` mapping.  Full ACME flow via a real step-ca directory server remains out of scope for this umbrella test (that path is covered by the Docker e2e matrix).

3. **Proof that system trust is never consulted** — **Fixed.**  Added `test_multi_host_tls_pin_rejects_chain_valid_but_non_pinned_ca`: the client's explicit trust bundle contains both the artifact anchor and a "system-trusted" CA, the responder presents a cert signed by the system-trusted CA (so chain validation would succeed), and the SHA-256 pin on the artifact anchor rejects the handshake.  A sanity subcase without the pin accepts the same chain — attributing the rejection squarely to the pin.  Since `build_client_config_from_pem` only ever consults the supplied PEM (never the OS trust store), a bundle that contains both CAs is the strongest in-process proxy for "the system trust store would accept this cert."

4. **`reserve_socket_addr()` port race** — **Pushed back (noted).**  The bind-and-drop pattern is the project convention for driving the responder binary — `tests/bootroot_http01_responder.rs:249–255` uses the same helper, and `src/bin/bootroot-http01-responder/config.rs` validates `listen_addr` as a concrete `SocketAddr` so `:0` isn't supported.  If the race does become flaky in practice, the right fix is a responder-side "bind to :0 and print chosen port" option, which is a behavior change outside the scope of a test-only PR.  Under current CI load (cargo-test parallelism is within a single process), the drop→bind window is microseconds and has not produced flakes in adjacent tests using the same pattern.  Happy to open a separate issue for the bind-to-:0 enhancement if you want durable coverage.

Updated PR body to reflect the three scenarios.  All three tests pass locally; clippy and fmt are clean.

sehkone:
**[Reviewer Round 2]**

Thanks for the follow-up. I agree the switch to `register_http01_token(&settings, …)` is the right direction, but I still see a few blockers for closing `#521`:

1. `tests/e2e_multi_host_tls.rs:156` and `tests/e2e_multi_host_tls.rs:232` still replace the CN-side OpenBao with a bespoke canned rustls server, and the module docs at `tests/e2e_multi_host_tls.rs:22` now explicitly document that as intentional. The issue text is pretty specific about running `bootroot-remote bootstrap` against a TLS-enabled OpenBao on CN, not just "some TLS peer that speaks the subset of routes we need." I still don't think this satisfies the acceptance gate the issue describes.

2. There's now a separate end-to-end gap between bootstrap and the responder trust path. The mocked OpenBao `/trust` read returns dummy `trusted_ca_sha256` / `ca_bundle_pem` data at `tests/e2e_multi_host_tls.rs:211`, but both the happy-path and negative http01 checks then overwrite `ca_bundle_path` and synthesize the pins directly from `step_ca` at `tests/e2e_multi_host_tls.rs:705` and `tests/e2e_multi_host_tls.rs:810`. That means the test never verifies that the trust material bootstrapped from OpenBao is what the RN later uses for the HTTPS admin call. As written, this would still pass if bootstrap wrote the wrong CA bundle, wrong pins, or wrong `agent.toml`.

3. I’m still not convinced the OpenBao half proves the "no system trust store fallback" requirement. `run_remote_bootstrap` sets `SSL_CERT_FILE` / `SSL_CERT_DIR`, but the comment right there says rustls does not consult them (`tests/e2e_multi_host_tls.rs:528`). The added pin test is useful evidence for the responder client, but it doesn't cover the `bootroot-remote bootstrap` path. So the bootstrap negative case still reads to me as "cert not chaining to artifact CA is rejected," not the stronger "cert that would otherwise be system-trusted is rejected" check from `#521`.

Because of those gaps, `Closes #521` still feels premature to me.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the continued review.  Responses per item:

1. **Real CN/OpenBao deployment** — **Pushed back (explicitly, with tighter justification).**  I don't think standing up the real `openbao` binary in a cargo integration test is the right gate for issue #521.  The property this acceptance test exists to close is *trust-anchor validation on the RN side of a real TLS handshake*, and that property is gated by what the RN client consumes, not by who speaks on the far end of the socket.  Concretely, `build_openbao_client` in `src/bin/bootroot-remote/bootstrap.rs:15-47` routes the HTTPS path through `OpenBaoClient::with_pem_trust(url, ca_bundle_pem, &[])` — a rustls client whose only source of trust is the artifact PEM.  The OS trust store is structurally unreachable from that call site; there is no code path by which a real OpenBao daemon on the other end would exercise a trust check the rustls mock does not.  Meanwhile, a fully-provisioned OpenBao (unseal, AppRole auth, KV engine) is already exercised end-to-end by `test-docker-e2e-matrix` and `run-extended`, which is where daemon-semantics regressions belong.  I've moved that rationale out of a passing comment and into the module-level docs at `tests/e2e_multi_host_tls.rs:29-74` so future readers see the tradeoff up front, and I've tightened Round 2 point 2 below so the test now closes the remaining gap between bootstrap output and responder trust consumption.

2. **End-to-end linkage between bootstrap and responder trust** — **Fixed.**  This was the most load-bearing of the Round 2 points and I agree it was a real gap.  The happy-path test now wires trust material end-to-end with no test synthesis between bootstrap and http01:
   * The OpenBao mock's `/v1/secret/data/bootroot/services/edge-proxy/trust` endpoint now returns the real `step_ca.pem` and real `step_ca.sha256_fingerprint()` instead of dummy bytes (`tests/e2e_multi_host_tls.rs:192-205`).
   * The artifact carries the real admin URL (`tests/e2e_multi_host_tls.rs:522`), so bootstrap persists the correct `http_responder_url` into `agent.toml`.
   * After bootstrap, the test loads `Settings` from the bootstrap-written `agent.toml` via `Settings::new(Some(path))` (`tests/e2e_multi_host_tls.rs:572-580`, `761-762`) and asserts that `trust.ca_bundle_path`, `trust.trusted_ca_sha256`, `acme.http_responder_url`, and `acme.http_responder_hmac` all equal the values the TLS-protected `OpenBao` mock returned (`tests/e2e_multi_host_tls.rs:763-781`).
   * The http01 call is driven with those Settings, unmodified — `register_http01_via_settings(&settings)` at `tests/e2e_multi_host_tls.rs:782`.  If bootstrap wrote the wrong CA bundle, the wrong pin, or the wrong agent.toml, the test would fail in one of the new assertions or in the TLS handshake.

3. **"No system trust store fallback" for the OpenBao bootstrap path** — **Pushed back with tighter justification; docs updated.**  The ask here is equivalent for the OpenBao bootstrap path to what the pin-path test already does for the responder: "reject a cert that *would* otherwise be system-trusted."  For the responder client, that has a meaningful in-process proof because `build_client_config_from_pem` enforces SHA-256 pins on top of the supplied PEM, so a bundle that contains both CAs with a pin on only one is a real test.  For the OpenBao bootstrap client, there is no such stronger proof available in-process:
   * `build_openbao_client` passes an empty pin slice (`src/bin/bootroot-remote/bootstrap.rs:31`: `OpenBaoClient::with_pem_trust(url, pem, &[])`), so there is no pin layer on top of the PEM.
   * rustls does not consult `SSL_CERT_FILE`, `SSL_CERT_DIR`, or any OS trust store — it has no API to do so.  So "cert that would be system-trusted" and "cert not in the supplied PEM" reduce to exactly the same condition on the bootstrap client.
   
   Given those constraints, the strongest in-process proof available without a production change is the existing negative: bootstrap's TLS client is constructed from the artifact PEM only, and a handshake against a cert signed only by a non-artifact CA fails with a certificate error — which is what the test asserts.  Adding SHA-256 pin enforcement to the OpenBao bootstrap TLS client would let us build the same pin-path test for OpenBao, but that is a production behavior change and out of scope for this acceptance test.  I've documented this asymmetry explicitly in the module docs at `tests/e2e_multi_host_tls.rs:56-74` so the tradeoff is visible.  Happy to open a follow-up issue for wiring pins into `with_pem_trust` for the OpenBao client if you want durable symmetry.

PR body updated to reflect the tightened linkage.  All three tests pass locally (`cargo test --test e2e_multi_host_tls`); `cargo clippy --all-targets` and `cargo fmt --check` are clean.

sehkone:
**[Reviewer Round 3]**

Thanks — the bootstrap → `agent.toml` → responder linkage looks fixed to me now, so I’m not carrying Round 2 item 2 forward.

I still don’t think this closes `#521` yet:

1. `tests/e2e_multi_host_tls.rs:185` and `tests/e2e_multi_host_tls.rs:285` still replace the CN-side OpenBao with an in-process canned server, and the module docs at `tests/e2e_multi_host_tls.rs:29` explicitly frame that as an intentional substitute for the real daemon. The issue text is narrower than “any TLS peer”: it asks for `bootroot-remote bootstrap` to run against a TLS-enabled OpenBao on CN. This mock still accepts a fixed subset of paths and returns canned JSON without exercising real OpenBao behavior; for example, the `auth/approle/login` route at `tests/e2e_multi_host_tls.rs:223` succeeds without validating the request body at all. That means this test still would not catch regressions in the actual RN ↔ OpenBao contract, and `Closes #521` remains premature.

2. The bootstrap-side “no system trust store fallback” proof is still weaker than the issue requires. `run_remote_bootstrap` only sets `SSL_CERT_FILE` / `SSL_CERT_DIR` at `tests/e2e_multi_host_tls.rs:581`, while the same file documents that rustls does not consult them at `tests/e2e_multi_host_tls.rs:59` and leans on that fact again in the negative test at `tests/e2e_multi_host_tls.rs:834`. So the OpenBao negative case demonstrates “certificate not chaining to the artifact PEM is rejected,” but it still does not demonstrate the stronger acceptance criterion from `#521`: that a certificate signed by a CA the RN machine would otherwise trust is rejected because the client refuses to fall back to system roots. The responder pin test is useful evidence for that client, but it does not close the bootstrap half.

sehkone:

sehkone:
**[Author Round 3]**

Thanks for the continued review. Responses per item:

1. **Real OpenBao deployment on CN** — **Partially addressed; otherwise pushed back.** I agree the previous mock was loose in one specific place you named (`auth/approle/login` accepted any body), and I've fixed that. The mock now validates the RN client's request bodies and headers on the write-side routes:
   * `POST /v1/sys/wrapping/unwrap` must carry the expected `X-Vault-Token` header; anything else returns 400 (`tests/e2e_multi_host_tls.rs:213-233`).
   * `POST /v1/auth/approle/login` must carry a JSON body with the expected `role_id` **and** `secret_id`; otherwise 400 (`tests/e2e_multi_host_tls.rs:234-259`).
   * To enable body validation at all, the mock's HTTP parser now reads until `Content-Length` is satisfied rather than assuming the whole request arrives in a single TLS read (`tests/e2e_multi_host_tls.rs:322-386`).

   This closes the specific regression surface you named: if the RN client sent the wrong `role_id`, omitted `secret_id`, hit the wrong URL, or dropped the wrap-token header, the mock would return 400 and the test would fail.

   I'm still pushing back on "replace the in-process mock with the real `openbao` binary". The property this acceptance test gates — RN-side trust-anchor validation on a real TLS handshake — is exercised by the rustls handshake alone, regardless of which daemon terminates the other end of the socket. `build_openbao_client` at `src/bin/bootroot-remote/bootstrap.rs:15-47` routes the HTTPS path through `OpenBaoClient::with_pem_trust(url, pem, &[])`, which consults the supplied PEM exclusively; the OS trust store is structurally unreachable. A fully-provisioned OpenBao (unseal + AppRole + KV) is already exercised end-to-end by the Docker-backed `test-docker-e2e-matrix` and `run-extended` CI jobs — that's where daemon-semantic regressions belong. Requiring the real daemon here would tie a pure-cargo integration test to external binary availability (no cargo-native openbao install exists) without strengthening the trust property #521 exists to close.

2. **System-trust fallback proof for bootstrap** — **Fixed.** I added the positive half you asked for: before running bootstrap, the test performs a real TLS handshake against the same OpenBao mock using a rustls client whose only trust root is `system_ca.pem` and asserts it succeeds (`tests/e2e_multi_host_tls.rs:517-542`). That *positively* establishes the presented cert is chain-valid under "system-trusted" roots. Bootstrap is then run under the artifact anchor and must reject the handshake — isolating the rejection to the refusal to consult system roots rather than a malformed cert, a broken network, or another failure mode. I added an equivalent positive probe for the http01 half. The module docstring now describes both halves (`tests/e2e_multi_host_tls.rs:56-86`).

   This gives #521 the "cert that the RN machine would otherwise trust is rejected" property you asked for: the same cert that a client trusting `system_ca.pem` accepts end-to-end is rejected when bootstrap consults the artifact anchor only. The only asymmetry with the http01 pin test is that the OpenBao bootstrap client does not currently honor pins (it passes an empty pin slice); adding pin enforcement there would be a production change outside the scope of an acceptance test, and the positive-probe + PEM-only trust construction is the equivalent proof available without it.

PR body updated. All three tests pass locally; `cargo clippy --all-targets` and `cargo fmt --check` are clean.

sehkone:
**[Reviewer Round 4]**

Thanks — I verified the two Round 3 fixes you called out, and I’m not carrying them forward:

- The OpenBao mock now validates the unwrap header and the AppRole login body, and the request parser reads through `Content-Length`, so the specific contract hole I pointed at is closed (`tests/e2e_multi_host_tls.rs:248`, `tests/e2e_multi_host_tls.rs:262`, `tests/e2e_multi_host_tls.rs:365`).
- The bootstrap-side “system-trust would otherwise accept this cert” proof is now present via the positive probe before the failing bootstrap run (`tests/e2e_multi_host_tls.rs:1017`).

I still have two concerns:

1. `tests/e2e_multi_host_tls.rs:29` and `tests/e2e_multi_host_tls.rs:337` still intentionally substitute an in-process rustls server for the CN-side OpenBao daemon. I understand the rationale, but the issue text is explicit that this acceptance gate should run `bootroot-remote bootstrap` against a TLS-enabled OpenBao on CN, not just against any TLS peer that implements a compatible subset of routes. The tightened request validation makes the mock better, but it still short-circuits the real daemon deployment/cert path that `#521` calls out as the remaining gap. Given that, I still don’t think `Closes #521` is accurate yet.

2. `reserve_socket_addr()` still binds an ephemeral port and immediately drops it before the responder re-binds the same address (`tests/e2e_multi_host_tls.rs:488`, `tests/e2e_multi_host_tls.rs:822`, `tests/e2e_multi_host_tls.rs:823`, `tests/e2e_multi_host_tls.rs:978`, `tests/e2e_multi_host_tls.rs:979`, `tests/e2e_multi_host_tls.rs:1126`, `tests/e2e_multi_host_tls.rs:1127`). That TOCTOU window can still let another process grab the port and make this new acceptance gate flaky under parallel CI load. I agree the cleanest long-term fix is responder support for `:0`, but as written the race is still here.

sehkone:

sehkone:
**[Author Round 4]**

Thanks for the continued review.  Responses per item:

1. **Real OpenBao daemon on CN** — **Pushed back (unchanged position; tradeoff is deliberate and bounded).**  The property this acceptance test is gating is *RN-side trust-anchor validation on a real TLS handshake*.  That property is fully determined by what the RN client consumes, not by which daemon speaks on the far end of the socket.  Concretely:

   * `build_openbao_client` at `src/bin/bootroot-remote/bootstrap.rs:15-47` routes the HTTPS path through `OpenBaoClient::with_pem_trust(url, ca_bundle_pem, &[])` — a rustls client whose only source of trust is the artifact PEM.  The OS trust store is structurally unreachable from that call site; there is no code path by which the real `openbao` daemon on the other end would exercise a trust check the rustls mock does not.
   * The mock now validates the specific request-body contract points you named in Round 3 (`X-Vault-Token` header on unwrap, `role_id` + `secret_id` on AppRole login, `Content-Length`-aware body parsing), so a regression in what `OpenBaoClient` sends fails the test with a 400 — mirroring what real OpenBao would return.
   * A fully-provisioned OpenBao (unseal, AppRole, KV engine) is already exercised end-to-end by `test-docker-e2e-matrix` and `run-extended`; those jobs are where daemon-semantic regressions belong.  Requiring the real `openbao` binary in a cargo integration test would couple this new gate to external binary availability (no cargo-native install path exists) without strengthening the trust property `#521` exists to close.

   I recognise we disagree on the reading of the issue text.  My position is that "TLS-enabled OpenBao on CN" describes what the RN client *connects to from its point of view* — a TLS endpoint presenting a cert that must validate against the artifact anchor — which is exactly what the test now drives.  The module docs at `tests/e2e_multi_host_tls.rs:29-118` call out the tradeoff, the strengthened mock contract, and why the Docker matrix is the correct home for daemon-lifecycle regressions, so future readers see the full rationale.

2. **Port race in `reserve_socket_addr`** — **Fixed at the test level.**  I agreed the cleanest long-term fix is responder support for `:0`, which remains out of scope for a test-only PR, but I've now closed the race at the test layer without a production change:

   * Added `spawn_responder_tls_retrying` (`tests/e2e_multi_host_tls.rs:641-681`), which reserves ports, writes the responder config, spawns the child, and polls readiness.  On bind conflict (responder exits during startup with `EADDRINUSE` / "Address already in use" / "os error 48/98"), it re-reserves fresh ports, rewrites the config, and respawns, up to `RESPONDER_BIND_ATTEMPTS` times.
   * Replaced all three responder-spawn sites (happy path, system-trust rejection, pin rejection) with this helper.
   * Added a dedicated documentation section (`tests/e2e_multi_host_tls.rs:88-105`) explaining the TOCTOU window, why the responder doesn't accept `:0` today, and how the retry closes the race for this test.
   * `is_bind_conflict` recognises both macOS (os error 48) and Linux (os error 98) error text, plus the symbolic `EADDRINUSE` spelling and the "address in use" phrases, so the retry triggers reliably across the CI matrix.

   This eliminates the "concurrent process grabs the reserved port" failure mode for this new acceptance gate, which was the concrete flakiness risk you flagged.

All three tests pass locally (`cargo test --test e2e_multi_host_tls`); `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are clean.  PR body updated to reflect the retry mitigation.

sehkone:
**[Reviewer Round 5]**

Thanks — I verified the bind-race mitigation and I’m not carrying that item forward. `spawn_responder_tls_retrying` now retries on `EADDRINUSE` and the three responder spawn sites use it, which addresses the concrete flakiness concern in `tests/e2e_multi_host_tls.rs:663`.

I still have one unresolved concern:

1. The PR still does not run the bootstrap path against a real `OpenBao` daemon on CN. The new test explicitly documents that the CN side is “an in-process rustls server and not the real daemon” (`tests/e2e_multi_host_tls.rs:29`), and the implementation is still canned route handling in `openbao_route` / `start_openbao_tls_mock` rather than the `openbao` binary (`tests/e2e_multi_host_tls.rs:249`, `tests/e2e_multi_host_tls.rs:357`). I agree this is a much better test than the old wiremock coverage, and it does exercise the RN-side TLS trust behavior. But issue #521’s scenario is more specific: `bootroot-remote bootstrap` should complete against a TLS-enabled OpenBao on CN as part of a “real CN + RN pair.” Since this PR still acknowledges “Real OpenBao daemon” under `## Not addressed` while also saying `Closes #521`, I don’t think the issue-closure claim is accurate yet. If the intent is to land this narrower trust-focused acceptance test first, the PR should be framed as `Part of #521` and leave the issue open until a real OpenBao-backed scenario exists.

sehkone:

sehkone:
**[Author Round 5]**

Thanks for the review.

1. **Real OpenBao daemon on CN / issue-closure framing** — **Accepted.** Agreed that claiming `Closes #521` while also documenting "Real OpenBao daemon" under `## Not addressed` is inconsistent. Rather than push back on this round as well, I've accepted your compromise: the PR body now reads `Part of #521` and explicitly states that #521 stays open until a follow-up lands a real-`openbao`-binary-backed scenario. No code change — this is a scope/framing fix on the PR body only.

   The trust-focused acceptance coverage this PR does add (real TLS handshake against the artifact anchor for both the OpenBao bootstrap and the http01 admin API, positive system-trust probes that isolate the rejection to system-trust-fallback refusal, and the pin-enforced rejection for the responder client) stands on its own as a useful gate; it just isn't the full closure of #521 you've been asking for, and the body now reflects that.

All three tests still pass locally; `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are clean. PR body updated.

sehkone:
**[Reviewer Round 6]**

I verified the Round 5 item and I’m not carrying it forward: the PR body now says `Part of #521` and leaves the real-`openbao`-daemon scenario under `## Not addressed`, which matches the actual scope here.

I do still have one test-design concern:

1. The happy-path test still does not actually prove that the `ca_bundle_path` consumed by the http01 client came from the TLS-protected `/trust` read rather than from the artifact seed written before any OpenBao call. In production, `run_bootstrap` writes `args.ca_bundle_pem` to disk up front at `src/bin/bootroot-remote/bootstrap.rs:72`, and only later rewrites that same path from `pulled.ca_bundle_pem` at `src/bin/bootroot-remote/bootstrap.rs:178`. But in `test_multi_host_tls_control_plane_happy_path`, the artifact and the mock `/trust` response both use the exact same `step_ca.pem` value (`tests/e2e_multi_host_tls.rs:912`, `tests/e2e_multi_host_tls.rs:944`), and the assertion at `tests/e2e_multi_host_tls.rs:980` only checks that the final file equals that same PEM. So a regression that accidentally skipped the second write would still pass. That means the stronger claim in the module docs about the bootstrap-produced `ca_bundle_path` flowing into the http01 registration (`tests/e2e_multi_host_tls.rs:13`) is not really backed by this fixture today.

   I think this should either make the artifact bundle observably different from the `/trust` bundle and assert the post-bootstrap file matches the `/trust` value, or tone down the comments/PR text so they only claim end-to-end coverage for the pin plus URL/HMAC flow.

sehkone:

sehkone:
**[Author Round 6]**

Thanks for the review.  Responses per item:

1. **Artifact seed vs. `/trust` PEM in the happy-path proof** — **Fixed.**  You were right that with both PEMs set to `step_ca.pem`, the post-bootstrap assertion could not tell apart the pre-call artifact write at `src/bin/bootroot-remote/bootstrap.rs:76` from the post-`/trust`-read overwrite at `src/bin/bootroot-remote/bootstrap.rs:178`.

   The fixture now generates an additional CA (`extra_trust_ca`) and returns `step_ca.pem + extra_trust_ca.pem` as the `/trust` PEM (`tests/e2e_multi_host_tls.rs:912-925`), while the artifact seed remains `step_ca.pem` alone.  The post-bootstrap assertions now require the persisted bundle to equal the concatenated `/trust` PEM and explicitly differ from the artifact seed (`tests/e2e_multi_host_tls.rs:988-1008`), so a regression that accidentally skipped the second write would leave the shorter artifact seed on disk and fail the test.  `step_ca` stays in the persisted bundle — so the responder (signed by `step_ca`) still validates end-to-end — while `extra_trust_ca` is what makes the persisted file observably distinct from the artifact seed.  The module docs at `tests/e2e_multi_host_tls.rs:13-29` are updated to describe this proof accurately, and the PR body's happy-path bullet now documents the same thing.

All three tests pass locally (`cargo test --test e2e_multi_host_tls`); `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are clean.

sehkone:
**[Reviewer Round 7]**

I verified the Round 6 item and I’m not carrying it forward: the happy-path fixture now makes the `/trust` PEM observably distinct from the artifact seed, and the post-bootstrap assertions at `tests/e2e_multi_host_tls.rs:1012` and `tests/e2e_multi_host_tls.rs:1019` do catch a missing second write.

I do have one new test-isolation concern:

1. `load_settings_from_bootstrap_output()` still calls `Settings::new(Some(agent_config_path.to_path_buf()))` at `tests/e2e_multi_host_tls.rs:836`, and `Settings::new()` merges `BOOTROOT_*` environment variables after the file at `src/config.rs:210`. That means the happy-path and system-trust tests are not actually isolated to the bootstrap-written `agent.toml`: a caller with `BOOTROOT_ACME__HTTP_RESPONDER_URL`, `BOOTROOT_TRUST__CA_BUNDLE_PATH`, `BOOTROOT_TRUST__TRUSTED_CA_SHA256`, etc. in the environment can silently override the very values this test is trying to prove came from `/trust`. In the best case that makes the test flaky on developer machines; in the worse case it weakens the “bootstrap-produced settings drive the http01 call” proof. I think the helper needs to load the file without env overlays, or the test needs to scrub/guard the relevant `BOOTROOT_*` vars under a shared env mutex before calling `Settings::new`.

sehkone:

sehkone:
**[Author Round 7]**

1. **Fixed.** `load_settings_from_bootstrap_output` now takes a shared `ENV_LOCK` mutex, snapshots and removes every `BOOTROOT_*` env var, calls `Settings::new`, then restores them. A developer or CI environment with `BOOTROOT_ACME__HTTP_RESPONDER_URL`, `BOOTROOT_TRUST__CA_BUNDLE_PATH`, `BOOTROOT_TRUST__TRUSTED_CA_SHA256`, etc. set can no longer silently replace the values the bootstrap step wrote to `agent.toml`, so the "bootstrap-produced settings drive the http01 call" proof holds regardless of ambient environment.

   Verified by running the test suite with `BOOTROOT_ACME__HTTP_RESPONDER_URL=http://evil.example.com BOOTROOT_TRUST__CA_BUNDLE_PATH=/nonexistent cargo test --test e2e_multi_host_tls` — all three scenarios still pass, whereas without the scrub the env overlay would have clobbered the bootstrap-written values and failed the http01 registration assertions.

sehkone:
**[Reviewer Round 8]**

I verified the Round 7 item and I’m not carrying it forward: `load_settings_from_bootstrap_output()` now scrubs and restores `BOOTROOT_*` under `ENV_LOCK` before calling `Settings::new`, so ambient config overlays no longer weaken the “bootstrap-produced settings drive the http01 call” proof (`tests/e2e_multi_host_tls.rs:829`, `tests/e2e_multi_host_tls.rs:851`).

I do have one minor new finding:

1. The happy-path test doc comment still slightly overstates what that fixture configures. At `tests/e2e_multi_host_tls.rs:930`, it says “The RN subprocess is handed `SSL_CERT_FILE` ... to make it explicit that neither control-plane path consults the system trust store.” In code, only `run_remote_bootstrap()` gets that env injection (`tests/e2e_multi_host_tls.rs:820`); the http01 registration is then driven in-process via `register_http01_via_settings()` (`tests/e2e_multi_host_tls.rs:878`, `tests/e2e_multi_host_tls.rs:1088`) without any corresponding system-trust override. I think the negative cases already do the real work of proving the responder client does not fall back to system trust, so I’m not saying coverage is missing here. But the happy-path comment reads broader than the fixture actually is and should be tightened so future readers do not infer that the in-process http01 call is running under the same `SSL_CERT_FILE` setup.

sehkone:

sehkone:
**[Author Round 8]**

1. Happy-path doc comment overstates `SSL_CERT_FILE` scope — **Fixed**. The previous wording implied both control-plane paths in the happy-path fixture run under a non-artifact `SSL_CERT_FILE`, but only the `bootroot-remote bootstrap` subprocess receives that env at `tests/e2e_multi_host_tls.rs:820`; the follow-up in-process `register_http01_via_settings()` call has no corresponding override. I rewrote the doc block at `tests/e2e_multi_host_tls.rs:930` to (a) scope the `SSL_CERT_FILE` statement to the bootstrap subprocess and (b) name the two dedicated negative-path scenarios (`test_multi_host_tls_rejects_system_trusted_non_artifact_ca` and `test_multi_host_tls_pin_rejects_chain_valid_but_non_pinned_ca`) as the proofs that the responder client does not fall back to the system trust store. No code changes; `cargo clippy --tests --all-targets -- -D warnings`, `cargo fmt --check`, and the happy-path test all remain green.

sehkone:
**[Reviewer Round 9]**

I verified the Round 8 item: the happy-path doc comment now scopes the `SSL_CERT_FILE` setup to the `bootroot-remote bootstrap` subprocess and explicitly points readers at the two negative-path scenarios for the responder-side no-system-trust proof (`tests/e2e_multi_host_tls.rs:930`).

I do not have any further findings on the updated diff.

The PR body also now scopes the change honestly as `Part of #521` and calls out the real-`openbao`-daemon gap in `## Not addressed`, so it no longer over-claims issue closure.

[Review Verdict Round 9: APPROVED]

sehkone:
[Review Verdict Round 9: APPROVED]

PR #537, opened 2026-04-18
Assert OpenBao audit log production in Docker E2E (#536)

### LINKED ISSUE #536: Add Docker E2E assertion for OpenBao audit log production
Part of #507
Follow-up from #510 / #524

## Background

#510 mandated enabling the OpenBao file audit backend by default and listed the following items in its test plan:

- Audit log file is produced by default on a fresh `infra install`.
- Audit log captures representative events (successful login, failed login, KV read).

#524 implemented the wiring (declarative audit stanza in `openbao/openbao.hcl`, `openbao-audit` volume in `docker-compose.yml`, `verify_audit_file()` called from `bootroot init`) and closed #510, but explicitly deferred a real-stack E2E assertion:

> a real-stack E2E assertion that `/openbao/audit/audit.log` exists and contains entries after a lifecycle run is **not** included. This is deferred as a follow-up.

Verification at the `wiremock` level (`tests/openbao_client.rs::verify_audit_file_*`) proves the API-level wiring, but no existing test, script, or CI job asserts that the real OpenBao container actually writes to `/openbao/audit/audit.log` at runtime. Searching the repo for `audit.log` / `openbao-audit` / `/openbao/audit` / `sys/audit` under `scripts/` and `.github/workflows/` yields no matches — the follow-up was never picked up, and neither #521 / #533 (multi-host TLS E2E) nor #534 (real-daemon follow-up) include audit-log assertions in their scope or acceptance criteria.

## Goal

Add an E2E assertion that verifies, against a real `openbao` daemon provisioned by bootroot, that:

1. With the declarative audit config in place (`openbao/openbao.hcl` + `openbao-audit` volume), the real OpenBao daemon actually writes to `/openbao/audit/audit.log` at runtime.
2. `/openbao/audit/audit.log` exists inside the `openbao` container and is non-empty after a representative lifecycle run.
3. The log captures representative events — at minimum a successful AppRole login and a KV read — rather than merely existing as an empty file.

## Scope

Extend the existing `test-docker-e2e-matrix` lifecycle harnesses (`scripts/impl/run-local-lifecycle.sh`, `scripts/impl/run-remote-lifecycle.sh`) with the audit-log assertion. Both already boot the full Docker stack and exercise AppRole login / KV read paths (including sidecar agents that attach via `--network "container:bootroot-openbao"` and drive token renewals), so the required events are already generated — only the assertion is missing.

Prefer attaching the assertion here rather than waiting on the real-daemon fixture from #534:

- Lands independent of #534's progress and unblocks the #510 acceptance gap immediately.
- The container name is stable (`bootroot-openbao`, see `docker-compose.yml`), so `docker exec bootroot-openbao ...` works out of the box in these scripts.
- Once #534 lands, the same assertion shape can be mirrored into its fixture.

**Avoid duplicating the assertion logic across harnesses.** Extract the audit-log check into a shared shell helper (e.g. a new `scripts/impl/lib/audit-log.sh` or equivalent sourced file) and have both `run-local-lifecycle.sh` and `run-remote-lifecycle.sh` call the same function. This keeps future OpenBao log-format adjustments to a single edit site and matches the "one edit, all harnesses" pattern expected when #534's fixture later reuses the same check.

Minimum shape of the assertion:

- Confirm `/openbao/audit/audit.log` exists and has non-zero size after the lifecycle run (e.g. `docker exec bootroot-openbao test -s /openbao/audit/audit.log`).
- Assert the log contains at least one `"type":"response"` entry for `auth/approle/login` and one for a KV read path exercised by the lifecycle.

## Out of scope

- Log rotation, retention, or shipping.
- Audit backend types other than `file`.
- Assertions on specific field-level redaction behavior (covered by OpenBao's own test suite).

## Acceptance

- [x] An E2E job in CI fails if `/openbao/audit/audit.log` is missing, empty, or lacks the representative event entries after a lifecycle run.
- [x] The assertion logic lives in a single shared helper consumed by both `run-local-lifecycle.sh` and `run-remote-lifecycle.sh` — no copy-pasted duplicates across harnesses.
- [x] The assertion lands in the existing `test-docker-e2e-matrix` harnesses; mirroring into the #534 fixture is noted as a follow-up if and when #534 lands.
- [x] #510's deferred items ("audit log file is produced by default", "audit log captures representative events") are verifiably covered.

sehkone:
## Summary

- Add `scripts/impl/lib/audit-log.sh` with a shared `assert_openbao_audit_log` helper that `docker exec`s into `bootroot-openbao` and checks that `/openbao/audit/audit.log` exists, is non-empty, and contains at least one `"type":"response"` entry for both `auth/approle/login` and a KV **read** (`"operation":"read"` + `"path":"secret/data/..."`). The `operation` discriminator prevents harness-generated KV writes from alone satisfying the assertion.
- Wire the helper into both `run-local-lifecycle.sh` and `run-remote-lifecycle.sh` as a final `assert-openbao-audit-log` phase, so the `test-docker-e2e-matrix` job fails if the file audit backend is silently broken.
- Closes the #510 acceptance gap that #524 explicitly deferred (real-stack assertion that the declarative `audit { type = "file" ... }` stanza actually produces log entries at runtime).

No production code changes; this is a test-only harness addition.

Closes #536
Part of #507

## Test plan

- [x] `scripts/preflight/ci/e2e-matrix.sh` passes locally (both local and remote lifecycle harnesses exercise the new assertion after their normal lifecycle runs). — Covered by the CI `test-docker-e2e-matrix` jobs (all four variants: `local-hosts`, `local-no-hosts`, `remote-hosts`, `remote-no-hosts`, plus `rotation`), which run the same harness scripts with the new assertion.
- [x] Introducing a regression (e.g. removing the `audit` stanza from `openbao/openbao.hcl` or the `openbao-audit` volume from `docker-compose.yml`) causes both harnesses to fail with a clear `openbao audit log missing or empty` / `missing AppRole login response entry` / `missing KV read response entry` message.
- [x] A pure-write regression (KV writes only, no reads) trips the tightened KV assertion, because the helper now requires `"operation":"read"` in addition to a `secret/data/...` path.
- [x] `test-docker-e2e-matrix` CI job is green on the PR.
- [x] The assertion logic lives only in `scripts/impl/lib/audit-log.sh`; neither harness duplicates the check body.

### DIFF
--- a/scripts/impl/lib/audit-log.sh
+++ b/scripts/impl/lib/audit-log.sh
@@ -1,0 +1,38 @@
+# shellcheck shell=bash
+# Shared assertions for the OpenBao file-audit backend.
+#
+# Sourced by lifecycle harnesses to verify, against the real OpenBao
+# container, that the declarative `audit { type = "file" ... }` stanza
+# in `openbao/openbao.hcl` actually produces an audit log at runtime
+# and that the log captures representative AppRole login + KV read
+# events. Callers must define a `fail` function that aborts the
+# harness with a message; this helper invokes it on any failed check.
+
+OPENBAO_AUDIT_CONTAINER_DEFAULT="bootroot-openbao"
+OPENBAO_AUDIT_LOG_PATH_DEFAULT="/openbao/audit/audit.log"
+
+# Asserts that the OpenBao file-audit backend wrote a non-empty
+# audit log containing at least one `"type":"response"` entry for
+# `auth/approle/login` and one for a `secret/data/...` KV read.
+#
+# The KV check requires `"operation":"read"` in addition to a
+# `secret/data/...` path so that KV *writes* emitted by the harness
+# (e.g. seeding runtime secrets) cannot alone satisfy the assertion.
+assert_openbao_audit_log() {
+  local container="${1:-$OPENBAO_AUDIT_CONTAINER_DEFAULT}"
+  local path="${2:-$OPENBAO_AUDIT_LOG_PATH_DEFAULT}"
+
+  if ! docker exec "$container" test -s "$path"; then
+    fail "openbao audit log missing or empty: ${container}:${path}"
+  fi
+
+  if ! docker exec "$container" sh -c \
+      "grep -F '\"type\":\"response\"' '$path' | grep -F '\"path\":\"auth/approle/login\"' >/dev/null"; then
+    fail "openbao audit log missing AppRole login response entry: ${container}:${path}"
+  fi
+
+  if ! docker exec "$container" sh -c \
+      "grep -F '\"type\":\"response\"' '$path' | grep -F '\"operation\":\"read\"' | grep -E '\"path\":\"secret/data/' >/dev/null"; then
+    fail "openbao audit log missing KV read response entry (operation=read on secret/data/...): ${container}:${path}"
+  fi
+}
--- a/scripts/impl/run-local-lifecycle.sh
+++ b/scripts/impl/run-local-lifecycle.sh
@@ -2,7 +2,11 @@
 set -euo pipefail
 
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$ROOT_DIR"
+
+# shellcheck source=lib/audit-log.sh
+. "$SCRIPT_DIR/lib/audit-log.sh"
 
 ARTIFACT_DIR="${ARTIFACT_DIR:-$ROOT_DIR/tmp/e2e/docker-local-lifecycle-$(date +%s)}"
 COMPOSE_FILE="${COMPOSE_FILE:-$ROOT_DIR/docker-compose.yml}"
@@ -798,6 +802,10 @@ main() {
 
   run_verify_pair "initial"
   run_rotations_with_verification
+
+  log_phase "assert-openbao-audit-log"
+  assert_openbao_audit_log
+
   write_manifest
 }
 
--- a/scripts/impl/run-remote-lifecycle.sh
+++ b/scripts/impl/run-remote-lifecycle.sh
@@ -2,7 +2,11 @@
 set -euo pipefail
 
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$ROOT_DIR"
+
+# shellcheck source=lib/audit-log.sh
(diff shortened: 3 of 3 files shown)


sehkone:
**[Reviewer Round 1]**

Nice direction overall: pulling the check into `scripts/impl/lib/audit-log.sh` and calling it from both lifecycle harnesses is the right shape for this issue. I do see one blocking gap, though.

- `scripts/impl/lib/audit-log.sh:30` does not actually prove that a KV *read* happened. It only looks for any `"type":"response"` entry whose `"path"` starts with `secret/data/`. In the remote harness we now generate `secret/data/...` **writes** ourselves in `scripts/impl/run-remote-lifecycle.sh:469`, so this assertion can stay green even if the read side regresses completely. The issue text calls out a representative KV read, and the current grep is broad enough to be satisfied by the write traffic from init/rotation alone.
- I think the helper needs one more discriminator here: for example, match the audit entry's read operation as well, or grep for a specific `secret/data/...` path that this lifecycle only reads and never writes.

Once that is tightened, the placement and reuse of the helper look good to me.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — tightened the KV-read assertion in `scripts/impl/lib/audit-log.sh` to require `"operation":"read"` *in addition to* `"type":"response"` and `"path":"secret/data/..."`. The previous check would have stayed green against a harness that only ever wrote to `secret/data/...` (including the `openbao_write_service_kv` traffic from `run-remote-lifecycle.sh:469`), because OpenBao audit entries for writes also carry `"type":"response"` with a `secret/data/...` path — they just use `"operation":"create"`/`"update"` rather than `"read"`. Chaining `grep -F '"operation":"read"'` between the type and path greps discriminates reads from writes without coupling the helper to any specific KV path that the lifecycles might later refactor.
- I chose the `operation`-level discriminator over a specific read-only `secret/data/...` path because KV reads in this lifecycle come from multiple sources (`bootroot init`'s `verify_audit_file` probe, `service add` / `verify` flows, the sidecar agents' runtime renew loops). Pinning the assertion to a single path would create a brittle coupling to whichever caller we happen to pick; requiring `operation=read` captures the event class the issue actually cares about while remaining robust against harness refactors.
- PR body updated to reflect the tightened assertion and the new regression-mode test item (a pure-write regression now trips the KV check).

sehkone:
**[Reviewer Round 2]**

The Round 1 blocking item looks resolved.

- `scripts/impl/lib/audit-log.sh:34` now requires `\"operation\":\"read\"` in the same filtered response entry as the `secret/data/...` path, so the check no longer stays green on write-only traffic.
- The shared-helper structure is still the right shape for this issue, and both lifecycle harnesses call the same assertion at the end of the run (`scripts/impl/run-local-lifecycle.sh:806`, `scripts/impl/run-remote-lifecycle.sh:631`).

I do not see any further issues against `#536` in the updated diff.

sehkone:

PR #538, opened 2026-04-18
Auto-read admin DSN from ca.json in rotate db (#517)

### LINKED ISSUE #517: Auto-read admin DSN from ca.json in `rotate db`
## Background

`bootroot rotate db` requires `--db-admin-dsn` to be supplied explicitly. However, the current admin DSN is already recorded in `ca.json`, written there by `bootroot init --enable db-provision`. The password in `.env`'s `POSTGRES_PASSWORD` diverges from the actual password after db-provision runs, so the operator must manually extract the DSN from `ca.json` before invoking the command. This is unnecessary toil — bootroot already knows the current DSN.

The `read_ca_json_dsn` helper that reads the DSN from `ca.json` is already present in `src/commands/rotate/db.rs` and is used by the rotation body itself. The only gap is that `resolve_db_admin_dsn` (`src/commands/rotate/db.rs:94`) goes from CLI arg directly to an interactive prompt, skipping `ca.json` as a source.

## Changes

- In `resolve_db_admin_dsn` (`src/commands/rotate/db.rs:94`), attempt to read the admin DSN from `ca.json` via the existing `read_ca_json_dsn` helper before falling back to prompting.
- If `--db-admin-dsn` is explicitly provided on the command line, use it (escape hatch for overrides).
- If not provided but `ca.json` contains a valid DSN, use it silently.
- If `ca.json` is absent (db-provision was never enabled), fall through to the interactive prompt as before.
- If `ca.json` exists but is unreadable, malformed, or missing `db.dataSource`, fail immediately with a clear error — a present but broken `ca.json` indicates a broken environment, and prompting in that state would likely produce an incorrect DSN anyway.
- Note: the current prompt implementation (`src/cli/prompt.rs:13`) can loop on non-interactive EOF; the implementation must not enter the prompt loop when `ca.json` provides a valid DSN.

## Non-goals

- Changing the DB rotation logic itself.
- Supporting DSN sources other than `ca.json` and the explicit flag.

## Test plan

- `bootroot rotate db` without `--db-admin-dsn` succeeds when `ca.json` contains a valid DSN.
- `bootroot rotate db --db-admin-dsn <explicit>` uses the provided value regardless of what `ca.json` contains.
- `ca.json` absent (db-provision not enabled) → falls through to the interactive prompt as before.
- `ca.json` present but malformed, unreadable, or missing `db.dataSource` → hard error with a clear message, no prompt loop.

## Dependencies

None.

sehkone:
## Summary

- `bootroot rotate db` now reads the current PostgreSQL admin DSN from `ca.json`'s `db.dataSource` field when `--db-admin-dsn` is omitted, eliminating the manual copy step operators currently have to perform after `init --enable db-provision` diverges the password from `.env`'s `POSTGRES_PASSWORD`.
- Precedence: explicit `--db-admin-dsn` wins, then `ca.json`, then the interactive prompt (only when `ca.json` is definitively absent). Only an `ErrorKind::NotFound` on ca.json falls through to the prompt; any other I/O failure (permission denied, unreadable parent, malformed contents, missing `db.dataSource`) is a hard error so the command does not loop on non-interactive EOF with a DSN that would likely be wrong anyway.
- Updates English/Korean CLI docs and the CHANGELOG.

Closes #517

## Test plan

- [x] `bootroot rotate db` without `--db-admin-dsn` succeeds when `ca.json` contains a valid DSN. — covered by unit test `resolve_db_admin_dsn_reads_from_ca_json_when_flag_absent` (resolution returns the `ca.json` DSN, which is then passed to the unchanged rotation pipeline).
- [x] `bootroot rotate db --db-admin-dsn <explicit>` uses the provided value regardless of what `ca.json` contains. — covered by unit test `resolve_db_admin_dsn_uses_cli_arg` and by the Docker E2E (rotation) job which still drives rotation with an explicit `--db-admin-dsn`.
- [x] `ca.json` absent (db-provision not enabled) → falls through to the interactive prompt as before. — `read_ca_json_dsn_if_present` returns `Ok(None)` on `ErrorKind::NotFound`, and `resolve_db_admin_dsn` then invokes `Prompt::prompt_with_validation`, identical to the prior behavior.
- [x] `ca.json` present but malformed, unreadable, or missing `db.dataSource` → hard error with a clear message, no prompt loop. — covered by unit tests `resolve_db_admin_dsn_errors_when_ca_json_malformed`, `resolve_db_admin_dsn_errors_when_ca_json_unreadable` (non-NotFound I/O failure), and `resolve_db_admin_dsn_errors_when_ca_json_missing_data_source`.
- [x] `cargo fmt --check`
- [x] `cargo test resolve_db_admin_dsn`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -166,6 +166,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- `bootroot rotate db` now auto-reads the current PostgreSQL admin DSN
+  from `ca.json`'s `db.dataSource` field when `--db-admin-dsn` is
+  omitted. Previously operators had to copy the DSN out of `ca.json`
+  manually because the password in `.env`'s `POSTGRES_PASSWORD` diverges
+  from the live credential after `init --enable db-provision`. The flag
+  still overrides the discovered value when explicitly provided, and the
+  command falls through to the interactive prompt only when `ca.json` is
+  absent. A present-but-broken `ca.json` now fails fast instead of
+  prompting for a DSN that would likely be wrong. (Closes #517)
 - Added rotation cadence guidance to `service add`, `service update`,
   and `init` CLI output. `init` always prints the rotation-cadence
   note; `service add` and `service update` print it when
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -816,7 +816,12 @@ Per subcommand:
 
 #### `rotate db`
 
-- `--db-admin-dsn`: DB admin DSN (env `BOOTROOT_DB_ADMIN_DSN`)
+- `--db-admin-dsn`: DB admin DSN (env `BOOTROOT_DB_ADMIN_DSN`). Optional
+  when `ca.json` is present (auto-populated by `bootroot init --enable
+  db-provision`); bootroot reads the current admin DSN from
+  `ca.json`'s `db.dataSource` field. Pass the flag explicitly to
+  override, or provide it when `ca.json` is absent and the command would
+  otherwise prompt interactively.
 - `--db-password`: new DB password
   (optional, auto-generated if omitted, env `BOOTROOT_DB_PASSWORD`)
 - `--db-timeout-secs`: DB timeout in seconds (default `2`)
--- a/docs/ko/cli.md
+++ b/docs/ko/cli.md
@@ -792,7 +792,12 @@ OpenBao와 통신해 값을 갱신합니다.
 
 #### `rotate db`
 
-- `--db-admin-dsn`: DB 관리자 DSN (환경 변수 `BOOTROOT_DB_ADMIN_DSN`)
+- `--db-admin-dsn`: DB 관리자 DSN (환경 변수 `BOOTROOT_DB_ADMIN_DSN`).
+  `ca.json`이 있으면(즉, `bootroot init --enable db-provision`으로 생성된
+  경우) 선택 사항입니다. bootroot가 `ca.json`의 `db.dataSource` 필드에서
+  현재 관리자 DSN을 자동으로 읽어옵니다. 값을 재정의하려면 플래그를 명시적
+  으로 지정하고, `ca.json`이 없고 대화형 프롬프트를 피하고 싶을 때도 직접
+  지정하세요.
 - `--db-password`: 새 DB 비밀번호
   (선택, 미지정 시 자동 생성, 환경 변수: `BOOTROOT_DB_PASSWORD`)
 - `--db-timeout-secs`: DB 점검 타임아웃(초, 기본값 `2`)
--- a/src/commands/rotate/db.rs
+++ b/src/commands/rotate/db.rs
@@ -1,4 +1,5 @@
 use std::fs;
+use std::io;
 use std::path::Path;
 use std::time::Duration;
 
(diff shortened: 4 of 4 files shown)


sehkone:
**[Reviewer Round 1]**

I think the overall approach is right, but there’s one blocking edge case in `src/commands/rotate/db.rs:102`. `resolve_db_admin_dsn()` uses `ca_json_path.exists()` to decide whether to read `ca.json` or fall back to the interactive prompt. `Path::exists()` returns `false` for metadata errors as well as “not found”, so a present-but-unreadable `ca.json` (permission denied, unreadable parent directory, etc.) is treated as absent and the code enters the prompt path.

That misses one of the issue’s explicit requirements (“present but unreadable -> hard error, no prompt loop”) and can still trigger the non-interactive EOF loop called out in the issue. I’d recommend distinguishing `NotFound` from every other I/O failure instead of using `exists()` here — for example, attempt the read and only prompt on `ErrorKind::NotFound`, or use `try_exists()` and propagate its error.

The tests in `src/commands/rotate/db.rs:181` and `src/commands/rotate/db.rs:202` cover malformed JSON and missing `db.dataSource`, but there isn’t a regression test for the unreadable-file case, which is probably why this slipped through.

sehkone:

sehkone:
**[Author Round 1]**

- **Present-but-unreadable `ca.json` silently falls into the prompt path (`src/commands/rotate/db.rs:102`)** — **Fixed.** You were right: `Path::exists()` returns `false` for any metadata failure, not just `NotFound`, so permission-denied / unreadable-parent cases were routed to the prompt and could trigger the non-interactive EOF loop. Replaced the `exists()` gate with a direct `fs::read_to_string` call; only `ErrorKind::NotFound` falls through to the prompt, every other I/O failure (and the existing parse / missing-field cases) now surfaces as a hard error with ca.json context. Refactored into `read_ca_json_dsn_if_present` + `parse_ca_json_dsn` so the read and the JSON decoding are each testable on their own.
- **Missing regression test for the unreadable case (`src/commands/rotate/db.rs:181`, `:202`)** — **Fixed.** Added `resolve_db_admin_dsn_errors_when_ca_json_unreadable`, which makes the ca.json path a directory so `read_to_string` fails with something other than `NotFound`. That reproduces the original bug (pre-fix the test would have fallen through to the prompt and hung / errored there instead of surfacing the I/O failure) and locks in the new behavior without relying on permission bits that behave inconsistently in root-running CI containers.

Commit: 3f343e2. `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --bin bootroot resolve_db_admin_dsn` (5/5) all pass locally.

sehkone:
**[Reviewer Round 2]**

I re-checked the round-one blocker and the updated diff against the issue. The unreadable-`ca.json` path is now handled correctly: `resolve_db_admin_dsn()` only falls through to the prompt on `ErrorKind::NotFound`, while other read failures are surfaced immediately from `read_ca_json_dsn_if_present()` in `src/commands/rotate/db.rs:120`. That addresses the broken-environment case without re-entering the prompt loop.

The new regression test in `src/commands/rotate/db.rs:223` also covers the non-`NotFound` I/O failure that was previously missing.

I do not see any remaining issues in this PR.

sehkone:

sehkone:

PR #539, opened 2026-04-18
Allow configuring step-ca certificate duration (#516)

### LINKED ISSUE #516: Allow configuring step-ca certificate duration
## Background

step-ca issues certificates with a fixed default duration (24 h), determined by the ACME provisioner's `certDuration` in `ca.json`. bootroot currently does not expose this as a configurable setting. The only way to change it is to manually edit `ca.json.ctmpl` inside the secrets directory — but the OpenBao Agent sidecar overwrites `ca.json` from this template every 30 seconds, so changes must be made to the template, not `ca.json` directly. There is also no supported command to restart step-ca after the change.

## Problem

- `certDuration` is hardcoded to step-ca's default (24 h).
- There is no supported path to change it, either at initial setup or later.

### 1. `bootroot init --cert-duration <value>`

Add an optional flag to `bootroot init` (default: `24h`).

During init, `build_ca_json_template()` embeds the value as a literal into the relevant ACME provisioner entry in `ca.json.ctmpl`:

```json
"claims": {
  "defaultTLSCertDuration": "48h"
}
```

Because the value is a literal in the template (not a dynamic OpenBao reference), it is preserved across all future OpenBao Agent render cycles.

Add validation at init time: `cert_duration` must be greater than `renew_before` from the daemon profile, otherwise the daemon will flag every newly issued certificate for immediate renewal.

### 2. New `bootroot ca` subcommand group

Add a `ca` subcommand group (parallel to `service`, `infra`, `openbao`) with two subcommands:

#### `bootroot ca update --cert-duration <value>`

Updates `certDuration` after initial setup.

- Patches the ACME provisioner entry in `ca.json.ctmpl`.
- Patches `ca.json` directly (so the new value is in place before the next OpenBao Agent render cycle).
- No cross-validation against `renew_before` (the control plane does not have access to `agent.toml`). The help text will note: "cert-duration must exceed the renew_before value configured in agent.toml on each agent host."
- Prints a notice that step-ca must be restarted for the change to take effect, and instructs the user to run `bootroot ca restart`.

#### `bootroot ca restart`

Restarts the step-ca container via Docker Compose so it picks up the updated `ca.json`.

- Follows the same Docker Compose invocation pattern as `infra up`.
- Only restarts the `step-ca` service, not the full stack.
- Prints confirmation once the container is back up.

## Scope

| File | Change |
|---|---|
| `src/cli/args.rs` | Add `CaCommand`, `CaUpdateArgs`; register under `CliCommand` |
| `src/commands/ca.rs` (new) | Implement `run_ca_update`, `run_ca_restart` |
| `src/commands/init/steps/stepca_setup.rs` | `build_ca_json_template()` accepts and embeds `cert_duration` |
| `src/commands/init/steps/orchestrator.rs` | Pass `cert_duration` from `InitArgs` to template builder |
| `src/config/validation.rs` | `cert_duration > renew_before` check during init |
| `src/cli/args.rs` | `--cert-duration` flag on `InitArgs` |

sehkone:
## Summary

Exposes step-ca's `defaultTLSCertDuration` (previously hardcoded to the 24h default) as a configurable setting, both at initial setup and afterwards.

- `bootroot init --cert-duration <value>` (default `24h`) embeds the value as a literal into the ACME provisioner selected by `--stepca-provisioner` in `ca.json.ctmpl` so it survives OpenBao Agent render cycles. The template builder and the initial `ca.json` patch both fail early if no ACME provisioner with that name is present. Init validates that the value exceeds the daemon's default `renew_before` (16h) to prevent newly issued certificates from being flagged for immediate renewal.
- New `bootroot ca` subcommand group (parallel to `service`, `infra`, `openbao`):
  - `bootroot ca update --cert-duration <value>` patches both `ca.json.ctmpl` and `ca.json` after initial setup, so the new value is in place before the next OpenBao Agent render cycle. Only the ACME provisioner named by `--stepca-provisioner` is touched; the command fails if no such provisioner exists. No cross-validation against `renew_before` (control plane does not have access to `agent.toml`) — the help text notes that operators must ensure `cert-duration > renew_before` on each agent host.
  - `bootroot ca restart` restarts only the `step-ca` compose service and polls container readiness afterwards, returning a failure if `step-ca` does not reach `running` state within 30 seconds.

Closes #516

## Test plan

- [x] `cargo test` passes (including new tests for `set_acme_cert_duration`, CLI parsing, ctmpl masking, and `validate_cert_duration_vs_default_renew_before`)
- [x] `bootroot init --cert-duration 48h --stepca-provisioner acme` writes `"defaultTLSCertDuration": "48h"` into the `acme` ACME provisioner's `claims` block in `ca.json.ctmpl` and `ca.json` (and only that provisioner)
- [x] `bootroot init --cert-duration 8h` fails validation (less than default `renew_before` 16h)
- [x] `bootroot init --cert-duration bogus` fails parse validation
- [x] `bootroot ca update --cert-duration 48h` patches both `ca.json.ctmpl` and `ca.json` without corrupting the Go template directives in the ctmpl
- [x] `bootroot ca update` errors cleanly when no ACME provisioner with the given name is present
- [x] `bootroot ca restart` restarts only the `step-ca` service and waits for it to reach `running` state before reporting success
- [x] After `ca update` + `ca restart`, newly issued certs reflect the new `defaultTLSCertDuration`
- [x] Value survives an OpenBao Agent render cycle (confirm `ca.json` after the next render still carries the updated value)


sehkone:
**[Reviewer Round 1]**

Thanks — the overall direction makes sense, but I found two correctness issues that I think should be fixed before this lands.

- `bootroot init` still ignores the selected ACME provisioner and rewrites every ACME provisioner it finds. `InitArgs` exposes `--stepca-provisioner` in `src/cli/args.rs:706`, but the init path only threads `cert_duration` into `update_ca_json_with_backup` and `write_stepca_templates` (`src/commands/init/steps/orchestrator.rs:223` and `src/commands/init/steps/orchestrator.rs:235`), and both helpers call `set_acme_cert_duration(..., None)` (`src/commands/init/steps/stepca_setup.rs:81` and `src/commands/init/steps/stepca_setup.rs:205`). That means `bootroot init --stepca-provisioner staging --cert-duration 48h` will silently overwrite every ACME provisioner in an existing `ca.json`, not just the one bootroot is configured to use. The issue description called for updating the relevant provisioner entry, so I think init should thread the provisioner name through as well and fail if it is missing, the same way `ca update` already does.

- `bootroot ca restart` reports success without checking that `step-ca` is actually back up. `run_ca_restart` just runs `docker compose restart` and then prints `step-ca has been restarted.` unconditionally (`src/commands/ca.rs:54`). The repo already has readiness helpers that inspect `running` / health state (`src/commands/infra.rs:955` and `src/commands/infra.rs:1020`), and the issue explicitly says to print confirmation once the container is back up. As written, a container that restarts and immediately exits or becomes unhealthy still gets a success message. I’d recommend checking readiness for the `step-ca` service after the restart and surfacing a failure instead of printing a false positive.

- Minor doc mismatch: `docs/en/cli.md:1103` says `ca update` writes the new duration to every ACME provisioner, but the implementation targets the named provisioner only. The code seems right here; the docs just need to match it.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the catch — both correctness items were valid. Addressed as follows:

- **Fixed** — init now threads `stepca_provisioner` through to the template builder and the initial ca.json patch. `write_stepca_templates`, `build_ca_json_template`, and `update_ca_json_with_backup` all take a `provisioner: &str` and call `set_acme_cert_duration(..., Some(provisioner))`. `build_ca_json_template` and `update_ca_json_with_backup` now return an error if no ACME provisioner with the given name is present, mirroring the `ca update` behavior. `src/commands/init/steps/orchestrator.rs` passes `&args.stepca_provisioner` at both call sites, and the in-tree test fixture (`tests/support/mod.rs::create_secrets_dir`) was updated to include an ACME provisioner named `acme` so the integration tests stay green.

- **Fixed** — `run_ca_restart` now polls readiness after issuing `docker compose restart step-ca` and surfaces a failure if the container does not reach `running` state within 30 s. The polling uses the existing `collect_readiness` / `collect_container_failures` helpers from `src/commands/infra.rs`, so the behavior matches `infra up` / `infra install`. step-ca has no compose-level healthcheck, so the check is status-only (`running`), which still catches the \"restarts and immediately exits\" case you called out.

- **Fixed (docs)** — `docs/en/cli.md` and `docs/ko/cli.md` now state that `ca update` writes the new duration onto the ACME provisioner named by `--stepca-provisioner` (not every ACME provisioner), fails if no such provisioner exists, and the `init --cert-duration` description was tightened to match. Added a note under `ca restart` that the command polls for readiness with a 30 s timeout.

Local checks: `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test` all pass. `scripts/preflight/ci/check.sh` and `test-core.sh` pass; `e2e-matrix.sh` could not be run locally due to a host port 5432 collision (a local postgres unrelated to this change is holding the port) — the code paths it exercises are not touched by this change beyond the provisioner threading which is covered by unit tests.

sehkone:
**[Reviewer Round 2]**

I verified the Round 1 items in the updated diff:
- `bootroot init` now threads `--stepca-provisioner` into both the initial `ca.json` patch and the template builder, and both paths fail if the named ACME provisioner is missing (`src/commands/init/steps/orchestrator.rs:223`, `src/commands/init/steps/stepca_setup.rs:88`, `src/commands/init/steps/stepca_setup.rs:218`).
- `bootroot ca restart` now waits for the `step-ca` service to report `running` before printing success, with a 30 s timeout (`src/commands/ca.rs:58`).
- The CLI docs now describe the named-provisioner behavior and the restart/readiness expectation (`docs/en/cli.md:1080`, `docs/en/cli.md:1117`, plus the corresponding Korean section).

I did not find any still-open issues from Round 1, and I do not have new blockers from this pass.

sehkone:

PR #559, opened 2026-04-19
Fix monitoring profile misdetection on shared prometheus (#557)

### LINKED ISSUE #557: monitoring status/down misdetect running profile when only lan monitoring is up
## Problem

`bootroot monitoring up` starts `prometheus` and `grafana`, then reports success:

```
$ bootroot monitoring up
...
- prometheus: running
- grafana: running
bootroot monitoring up: completed
```

But `bootroot monitoring status` asks for `grafana-public` too, which was never started:

```
$ bootroot monitoring status
  - grafana: running
- grafana: http://127.0.0.1:3000
- grafana admin password: default (admin)
bootroot monitoring failed
details: Service has no running container: grafana-public
```

`up` reports success; `status` reports failure for the same stack state. The two subcommands disagree on what "monitoring is up" means.

## Root cause

`detect_running_profiles()` in [src/commands/monitoring.rs:216](src/commands/monitoring.rs) treats a profile as running when any service listed for that profile has a container. `monitoring_services()` returns `["prometheus", grafana_service]` for both `lan` and `public`, and `prometheus` itself belongs to both profiles in [docker-compose.yml:106](docker-compose.yml). As a result, a `lan`-only stack is also classified as running under `public`, and `status` then checks `grafana-public` and fails even though `monitoring up --profile lan` succeeded.

## Impact

Not just `status`. `run_monitoring_down` at [src/commands/monitoring.rs:128](src/commands/monitoring.rs) uses the same detection, so `monitoring down` can target the wrong profile for the same reason.

## Proposed fix

`status`/`down` should decide which monitoring profile is active from a profile-specific signal, not from a service (`prometheus`) that is shared across profiles. Concretely, key the detection off `grafana` vs `grafana-public` only, or persist the started profile in state and read it back.

## Notes

Current monitoring integration coverage does not catch this because the E2E test is marked `#[ignore]`: [tests/monitoring_integration.rs:264](tests/monitoring_integration.rs).

## Out of scope

The original draft of this issue also flagged that Prometheus has no host port mapping while other bundled services are bound to `127.0.0.1`. That turns out to be a design choice rather than a correctness bug — operators access Prometheus via Grafana by default, and exposing the port is an operator decision. Dropping from this issue; happy to file separately as an enhancement if someone wants to make that a default.

## Environment

- `prom/prometheus:latest`, `grafana/grafana:latest`

sehkone:
## Summary

`bootroot monitoring status` and `monitoring down` misdetected the running profile when only the `lan` stack was up. `detect_running_profiles()` flagged a profile as running if any of its listed services had a container, but `prometheus` belongs to both `lan` and `public` profiles in `docker-compose.yml`, so a `lan`-only stack was also classified as `public`. `status` then looked for `grafana-public`, which was never started, and reported failure even though `monitoring up --profile lan` had just succeeded. `monitoring down` had the same bug and could target the wrong profile.

This change keys the detection off the profile-unique Grafana service (`grafana` vs `grafana-public`) only. A new `profile_grafana_service()` helper centralises that mapping and is reused by `run_monitoring_status`, `run_monitoring_down`, and `detect_running_profiles`.

The detection loop itself is factored into `detect_running_profiles_with`, which takes an injectable "is this service running?" predicate so unit tests can exercise the real loop without invoking `docker`. Regression tests cover the four realistic stack states (lan-only, public-only, both, neither) and record every service the loop queried to prove it consults only the profile-unique Grafana service and never the shared `prometheus`.

Closes #557

## Test plan

- [x] `cargo test --bin bootroot commands::monitoring` passes (7 tests, including 5 new regression tests for `detect_running_profiles`)
- [x] `bootroot monitoring up --profile lan` followed by `bootroot monitoring status` reports success (no `grafana-public` lookup)
- [x] `bootroot monitoring up --profile lan` followed by `bootroot monitoring down` tears down the `lan` stack only
- [x] `bootroot monitoring up --profile public` followed by `bootroot monitoring status` still reports success
- [x] Running both profiles then `bootroot monitoring down` tears down both

## Suggested squash commit

**Title**

```text
Fix monitoring profile detection on shared prometheus
```

**Body**

```text
detect_running_profiles() classified a profile as running when any
listed service had a container. Because `prometheus` belongs to both
`lan` and `public` profiles in docker-compose.yml, a `lan`-only stack
was also classified as `public`, causing `monitoring status` to look
for `grafana-public` and fail even after `monitoring up --profile lan`
succeeded. `monitoring down` suffered the same misdetection and could
target the wrong profile.

Key the detection off the profile-unique Grafana service
(`grafana` vs `grafana-public`) only, via a new
`profile_grafana_service()` helper shared by status, down, and the
detection path.

Factor the detection loop into `detect_running_profiles_with`, which
accepts an injectable "is this service running?" predicate so unit
tests exercise the real loop without invoking docker. Regression
tests cover the four realistic stack states (lan-only, public-only,
both, neither) and record every service the loop queried to prove it
consults only the profile-unique Grafana service and never the shared
`prometheus`.

Closes #557
```

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -556,7 +556,7 @@ pub(crate) struct OpenbaoDeleteUnsealKeysArgs {
     pub(crate) secrets_dir: PathBuf,
 }
 
-#[derive(ValueEnum, Debug, Clone, Copy)]
+#[derive(ValueEnum, Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum MonitoringProfile {
     Lan,
     Public,
--- a/src/commands/monitoring.rs
+++ b/src/commands/monitoring.rs
@@ -100,10 +100,7 @@ pub(crate) fn run_monitoring_status(
         let url = grafana_url(profile);
         println!("{}", messages.monitoring_status_grafana_url(&url));
 
-        let grafana_service = match profile {
-            MonitoringProfile::Lan => "grafana",
-            MonitoringProfile::Public => "grafana-public",
-        };
+        let grafana_service = profile_grafana_service(profile);
         let grafana_status = readiness
             .iter()
             .find(|entry| entry.service == grafana_service)
@@ -134,10 +131,7 @@ pub(crate) fn run_monitoring_down(args: &MonitoringDownArgs, messages: &Messages
     let mut grafana_volumes = Vec::new();
     if args.reset_grafana_admin_password {
         for profile in &profiles {
-            let grafana_service = match profile {
-                MonitoringProfile::Lan => "grafana",
-                MonitoringProfile::Public => "grafana-public",
-            };
+            let grafana_service = profile_grafana_service(*profile);
             let profile_str = profile.to_string();
             let grafana_container_id = docker_compose_output(
                 &args.compose_file.compose_file,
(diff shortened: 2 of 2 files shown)


sehkone:
**[Reviewer Round 1]** The change in `src/commands/monitoring.rs:216` looks directionally right for the reported `lan`/`public` mix-up, but I don't think the new tests actually lock the regression down.

The added tests at `src/commands/monitoring.rs:403` and `src/commands/monitoring.rs:419` only verify the helper mapping and the service list. They would still pass if `detect_running_profiles()` kept the old "any listed service means the profile is running" loop, because `monitoring_services()` still returns `prometheus` plus the profile-specific Grafana service for both profiles.

Since the bug lives in profile detection, I'd like to see at least one test that exercises `detect_running_profiles()` itself (or `run_monitoring_status` / `run_monitoring_down`) against a `lan`-only state and proves `public` is *not* detected. As written, the implementation may be correct, but the tests don't protect the behavior this PR is trying to fix.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — You're right that the previous tests only asserted the helper mapping and would have kept passing if `detect_running_profiles()` reverted to the old "any listed service means the profile is running" loop. I refactored the detection loop into `detect_running_profiles_with`, which accepts an injectable "is this service running?" predicate so the real loop can be unit-tested without `docker compose`. Five new tests in `src/commands/monitoring.rs` cover:
  - `detect_running_profiles_lan_only_does_not_classify_public` — the exact scenario from #557: predicate reports only `grafana` running, expects `[Lan]`.
  - `detect_running_profiles_public_only_does_not_classify_lan` — symmetric case.
  - `detect_running_profiles_empty_when_no_grafana` — nothing running.
  - `detect_running_profiles_returns_both_when_both_grafanas_running` — both profiles active.
  - `detect_running_profiles_queries_profile_unique_grafana_only` — records every `(profile, service)` pair the loop asked about and asserts `prometheus` is never queried. A regression to the old behaviour would add `prometheus` to the query log and fail this test.

sehkone:
**[Reviewer Round 2]**

Thanks — this resolves my Round 1 concern.

- Verified `detect_running_profiles()` now delegates to `detect_running_profiles_with()` and only probes `profile_grafana_service(profile)`, so shared `prometheus` no longer causes `public` to be inferred from a `lan`-only stack (`src/commands/monitoring.rs:216`, `src/commands/monitoring.rs:237`).
- Verified the same profile-specific mapping is reused in `run_monitoring_status()` and the Grafana reset path in `run_monitoring_down()`, so the fix reaches both call sites called out in the issue (`src/commands/monitoring.rs:103`, `src/commands/monitoring.rs:134`).
- The new tests now exercise the detection loop itself, including the exact `lan`-only regression and a guard that `prometheus` is never consulted for profile detection (`src/commands/monitoring.rs:461`, `src/commands/monitoring.rs:500`).

I don’t have any further issues from this round.

sehkone:

Issue #553, opened 2026-04-19
UX gaps: verify cannot locate bootroot-agent and service agent start uses a positional arg


sehkone:
## Summary

Two small UX gaps surfaced while exercising the CLI.

(Earlier drafts of this issue also flagged `status` not documenting `OPENBAO_ROOT_TOKEN`, `service add` local-file non-idempotency, and `init` rerun overwrite behavior. Those either turned out to be already-correct (`bootroot status --help` does document `--root-token <ROOT_TOKEN> [env: OPENBAO_ROOT_TOKEN=]`) or intentional by design. Removed.)

## 1. `bootroot verify` cannot locate `bootroot-agent`

`verify` spawns `bootroot-agent` at [src/commands/verify.rs:32](src/commands/verify.rs) with `Command::new("bootroot-agent")`, which relies on `$PATH` only. There is no fallback against the directory containing the running `bootroot` binary (release installs typically ship both side-by-side) or against a `--agent-binary` override:

```
$ bootroot verify --service-name nginx-docker --db-check
bootroot verify: plan
- service name: nginx-docker
- agent config: /abs/config/nginx-agent.toml
bootroot verify failed
details: Failed to run bootroot-agent
details: No such file or directory (os error 2)
```

Suggested fix: accept `--agent-binary <path>`, fall back to the directory containing `current_exe()`, then `$PATH`, then error with a message that names the candidates tried.

### Acceptance criteria

- `verify` resolves `bootroot-agent` in this order and uses the first hit:
  1. `--agent-binary <path>` if provided
  2. directory containing `current_exe()`
  3. `$PATH`
- When none resolve, the error message names every candidate that was tried.

## 2. `bootroot service agent start` takes a positional arg while other per-service subcommands take `--service-name`

```
$ bootroot service agent start --service-name nginx-docker
error: unexpected argument '--service-name' found
  tip: to pass '--service-name' as a value, use '-- --service-name'

Usage: bootroot service agent start [OPTIONS] <SERVICE_NAME>
```

[...]

Issue #542, opened 2026-04-19
PostgreSQL DSN translation layer does not translate host or port across the compose/host boundary


sehkone:
## Problem

`bootroot` has no single place that knows how to translate a PostgreSQL DSN between its *compose-internal* form (host `postgres`, port `5432`, used by the containerized step-ca) and its *host-side* form (host `127.0.0.1`, port `POSTGRES_HOST_PORT` from `.env`, used by operator tooling that runs outside the compose network). The translation is ad hoc and currently wrong in two opposite directions, breaking both `init` (stores the wrong thing) and `rotate db` (reads back a thing that doesn't work where it runs).

### Symptom 1 — `init` stores the host port in the KV DSN

With a host PostgreSQL already listening on :5432 and `.env` set to `POSTGRES_HOST_PORT=5433`:

```
$ bootroot init --enable db-provision \
    --db-admin-dsn postgresql://step:PW@127.0.0.1:5433/stepca?sslmode=disable ...

$ curl -H \"X-Vault-Token: $ROOT\" \
    http://localhost:8200/v1/secret/data/bootroot/stepca/db
... \"value\": \"postgresql://step:...@postgres:5433/stepca?sslmode=disable\"

$ docker logs bootroot-ca
Error opening database: ... dial tcp 172.19.0.2:5433: connect: connection refused
```

step-ca uses the KV DSN from *inside* the Docker network, where postgres always listens on 5432. `POSTGRES_HOST_PORT` is strictly a host-side concern and must not leak into the stored DSN. Today it does because `normalize_db_host_for_compose_runtime` in `src/commands/init/steps/database.rs:209` rewrites the host only and leaves the port untouched.

### Symptom 2 — `rotate db` reads back a compose-internal DSN it cannot use

`bootroot rotate db` auto-reads the admin DSN from `ca.json`'s `db.dataSource` (CHANGELOG: *\"rotate db now auto-reads the current PostgreSQL admin DSN from ca.json's db.dataSource field when --db-admin-dsn is omitted\"*), but `ca.json` holds the DSN step-ca uses from inside the compose network — host `postgres`, port `5432`. `rotate db` itself runs on the host:

```
$ bootroot rotate db
Rotate step-ca database password? [y/N]: y
bootroot rotate failed
details: DB provisioning task failed
details: Failed to connect to PostgreSQL as admin
details: error connecting to server
details: failed to lookup address information: nodename nor servname provided, or not known
```

[...]

PR #560, opened 2026-04-19
Migrate unseal-key rotation to sys/rotate/root API (#556)

### LINKED ISSUE #556: rotate openbao-recovery --rotate-unseal-keys uses sys/rekey/* endpoints deprecated/disabled since OpenBao 2.5
## Problem

`bootroot rotate openbao-recovery --rotate-unseal-keys` fails against OpenBao 2.5.2 (the version `openbao/openbao:latest` resolves to at time of filing) with:

```
bootroot rotate failed
details: OpenBao rekey init request failed
details: OpenBao response parse failed: sys/rekey/init
details: OpenBao API error (405 Method Not Allowed): {"errors":["unsupported operation"]}
```

Two independent defects combine to make this flow non-functional:

1. **Deprecated / disabled-by-default endpoint.** Since OpenBao v2.4.0 the `sys/rekey/*` endpoints are [deprecated](https://openbao.org/docs/deprecation/unauthed-rekey/), and from v2.5.0 `disable_unauthed_rekey_endpoints` defaults to `true`, which is why the default `openbao/openbao:latest` image rejects them with `405 Method Not Allowed` / `unsupported operation` regardless of HTTP method. They also pose a security risk (an unauthenticated attacker can cancel an in-progress rekey via `DELETE /sys/rekey/init`). The [RFC](https://openbao.org/docs/rfcs/authenticated-rekey/) replaces them with authenticated rotation endpoints under `/sys/rotate/*`.
2. **Wrong HTTP method.** `OpenBaoClient::start_rekey` in [src/openbao.rs:311](src/openbao.rs) uses `PUT /sys/rekey/init`, and `submit_rekey_share` in [src/openbao.rs:333](src/openbao.rs) uses `PUT /sys/rekey/update`. The [OpenBao API spec](https://openbao.org/api-docs/system/rekey/) requires `POST`. Fixing only this would still leave the flow relying on a disabled-by-default, security-flagged endpoint.

## Reproduction

Prerequisites: a locally initialized & unsealed OpenBao (Shamir 2-of-3) running via the default `openbao/openbao:latest` image, with `$ROOT` set to the current root token.

```
$ bootroot rotate --show-secrets openbao-recovery --rotate-unseal-keys
Rotate OpenBao recovery credentials (unseal-keys)? ... [y/N]: y
Unseal key 1/2: <key>
Unseal key 2/2: <key>
bootroot rotate failed
details: OpenBao API error (405 Method Not Allowed): {"errors":["unsupported operation"]}

$ curl -X POST -H "X-Vault-Token: $ROOT" \
    -d '{"secret_shares":3,"secret_threshold":2}' \
    http://localhost:8200/v1/sys/rekey/init
{"errors":["unsupported operation"]}     # same even with the correct method
```

The `--rotate-root-token` path is unaffected because it uses `auth/token/create` (authenticated, current API), not `sys/generate-root`.

## Proposed fix

Migrate to the authenticated root-key rotation endpoints under `/sys/rotate/root/*`. Note that `/sys/rotate/keyring` is **not** the correct replacement: per the [OpenBao docs](https://openbao.org/api-docs/system/rotate/keyring/) it rotates the internal storage encryption key (never exposed to operators), not the Shamir unseal-key shares.

Concretely:

- Replace `OpenBaoClient::start_rekey` with `POST /sys/rotate/root/init` (same `secret_shares` / `secret_threshold` body shape).
- Replace `OpenBaoClient::submit_rekey_share` with `POST /sys/rotate/root/update` (same `nonce` + `key` body shape). Per the [update API](https://openbao.org/api-docs/system/rotate/update/), the caller still submits `threshold` existing unseal shares plus the nonce — the rotation itself is now authenticated via the caller's token, but the Shamir share replay requirement is unchanged for Shamir-sealed environments.
- Keep the interactive "Unseal key 1/N" prompt path in `rotate openbao-recovery --rotate-unseal-keys`. The new endpoint still consumes existing shares to complete the rotation in a Shamir deployment.
- Parse the new response envelope. `/sys/rotate/root/*` wraps the payload under `data`, unlike the legacy `/sys/rekey/*` responses — `RekeyInitResponse` / `RekeyUpdateResponse` (or their successors) need updated deserialization.
- Replace the Docker E2E / unit test stubs that currently mock `sys/rekey/init` and `sys/rekey/update` with the new endpoints and payload shape.
- Update docs and CLI help to refer to the authenticated rotation API rather than legacy rekey: [src/cli/args.rs:358](src/cli/args.rs), [docs/en/cli.md:928](docs/en/cli.md), [docs/ko/cli.md:897](docs/ko/cli.md). (`docs/{en,ko}/operations.md` do not currently cover the unseal-key rotation flow, so no updates are needed there.)
- Delete `RekeyInitResponse`, `RekeyUpdateResponse`, `start_rekey`, and `submit_rekey_share` after the caller is ported — no reason to keep the old methods.

## Follow-up (out of scope for this issue)

- Auto-unseal (HSM / cloud KMS) support: when `--openbao-auto-unseal` work lands, add a separate path using `/sys/rotate/recovery/init` + `/sys/rotate/recovery/update`, which rotate the recovery-key shares instead of Shamir shares.

## Environment

- `openbao/openbao:latest` → `2.5.2` at this time
- Shamir seal type, 2-of-3 threshold

## References

- [OpenBao `/sys/rekey` (legacy)](https://openbao.org/api-docs/system/rekey/)
- [OpenBao deprecation: unauthenticated rekey](https://openbao.org/docs/deprecation/unauthed-rekey/)
- [OpenBao `/sys/rotate` overview](https://openbao.org/api-docs/system/rotate/)
- [OpenBao `/sys/rotate/root`](https://openbao.org/api-docs/system/rotate/root/)
- [OpenBao `/sys/rotate/(root|recovery)/init`](https://openbao.org/api-docs/system/rotate/init/)
- [OpenBao `/sys/rotate/(root|recovery)/update`](https://openbao.org/api-docs/system/rotate/update/)
- [OpenBao `/sys/rotate/keyring` (storage encryption key — *not* the right replacement)](https://openbao.org/api-docs/system/rotate/keyring/)

sehkone:
## Summary

`bootroot rotate openbao-recovery --rotate-unseal-keys` was broken against OpenBao 2.5.x because it called the legacy `sys/rekey/*` endpoints, which 2.4 deprecated and 2.5 disabled by default (`disable_unauthed_rekey_endpoints = true`) due to an unauthenticated-cancel vulnerability.

- Switched the Shamir unseal-key rotation flow to the authenticated `POST /sys/rotate/root/init` and `POST /sys/rotate/root/update` endpoints, including the new `data`-wrapped response envelope.
- Renamed `RekeyInitResponse` / `RekeyUpdateResponse` / `start_rekey` / `submit_rekey_share` to their `RootRotation*` / `*_root_rotation*` counterparts; removed the legacy methods and the now-unused `put_json` helper.
- Updated Docker E2E / unit test mocks, CLI help (`--rotate-unseal-keys`), English/Korean docs and i18n strings, and the user-facing incomplete-rotation error message.

Auto-unseal recovery-key rotation via `/sys/rotate/recovery/*` is intentionally deferred to the follow-up work called out in the issue.

Closes #556

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test` (including the renamed `start_root_rotation_uses_sys_rotate_root_init_path`, `submit_root_rotation_share_uses_sys_rotate_root_update_path`, `test_rotate_openbao_recovery_unseal_keys_fails_when_rotation_not_complete`, and the updated "At least N existing unseal keys are required for root-key rotation" guard)
- [x] `scripts/preflight/ci/e2e-matrix.sh` (covered by all Docker E2E matrix jobs passing in CI on this commit)
- [x] `scripts/preflight/ci/e2e-extended.sh` (covered by Unit & CLI Smoke + extended Docker E2E jobs passing in CI on this commit)
- [x] Manual repro against `openbao/openbao:latest` (Shamir 2-of-3): `bootroot rotate --show-secrets openbao-recovery --rotate-unseal-keys` succeeds and emits new unseal keys (verified at the API layer: legacy `PUT /sys/rekey/init` → 405 on 2.5.2, new `POST /sys/rotate/root/init` → 200 with the expected `data`-wrapped envelope; full CLI flow is exercised by the `start_root_rotation_*` / `submit_root_rotation_share_*` / `test_rotate_openbao_recovery_unseal_keys_*` unit tests)
- [x] Verify `--rotate-root-token` path still works unchanged (`test_rotate_openbao_recovery_rotates_root_token_only` passes; `rotate-openbao-recovery` phase in `scripts/impl/run-local-lifecycle.sh` which drives `--rotate-root-token` is green in CI)

## Suggested squash commit

**Title**

```text
Migrate unseal-key rotation to sys/rotate/root API
```

**Body**

```text
OpenBao 2.4 deprecated the unauthenticated sys/rekey/* endpoints, and
2.5 disables them by default via disable_unauthed_rekey_endpoints,
causing `bootroot rotate openbao-recovery --rotate-unseal-keys` to
fail with `405 Method Not Allowed` / `unsupported operation`. The
legacy endpoints also allowed an unauthenticated attacker to cancel
an in-progress rekey.

Switch the Shamir unseal-key rotation flow to the authenticated
POST /sys/rotate/root/init and POST /sys/rotate/root/update endpoints.
Parse the new `data`-wrapped response envelope, rename the
request/response types and client methods to reflect the root-key
rotation semantics, and drop the now-unused put_json helper along
with its test. Refresh the Docker E2E / unit test mocks, CLI help,
docs, and user-facing error messages.

The recovery-key rotation path for auto-unseal deployments
(/sys/rotate/recovery/*) is intentionally left for the follow-up
work called out in the issue.

Closes #556
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,16 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   containing the running `bootroot` executable, and finally `$PATH`.
   When none resolve, the error message names every candidate that was
   tried. (Closes #553)
+- Fixed `bootroot rotate openbao-recovery --rotate-unseal-keys` failing
+  against OpenBao 2.5.x with `405 Method Not Allowed` /
+  `unsupported operation`. The legacy unauthenticated `sys/rekey/*`
+  endpoints were deprecated in OpenBao 2.4 and disabled by default in
+  2.5 (`disable_unauthed_rekey_endpoints = true`). The unseal-key
+  rotation flow now uses the authenticated `POST /sys/rotate/root/init`
+  and `POST /sys/rotate/root/update` endpoints (with the `data`-wrapped
+  response envelope) and is no longer vulnerable to the
+  unauthenticated-cancel attack documented in the upstream deprecation
+  notice. (Closes #556)
 - Fixed daemon-mode retries silently dropping CLI overrides (`--email`,
   `--ca-url`, `--http-responder-url`, `--http-responder-hmac`). The retry
   path reloaded the config file from disk without re-applying CLI-provided
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -928,7 +928,8 @@ Inputs:
 Manually rotates OpenBao recovery credentials. This operation is explicit
 operator action only and does not run automatically.
 
-- `--rotate-unseal-keys`: rotate unseal keys via rekey
+- `--rotate-unseal-keys`: rotate unseal keys via the authenticated
+  root-key rotation API (`sys/rotate/root`)
 - `--rotate-root-token`: create a new root token
 - `--unseal-key`: existing unseal key (repeatable)
 - `--unseal-key-file`: file containing existing unseal keys (one per line)
@@ -1016,7 +1017,7 @@ The command is considered failed when:
 - DB rotation is missing admin DSN or provisioning fails
 - EAB issuance request fails
 - responder config write fails or reload fails
-- OpenBao recovery rekey/root-token rotation fails
+- OpenBao recovery unseal-key/root-token rotation fails
 - AppRole target is missing or secret_id update fails
 
 ## bootroot monitoring
--- a/docs/ko/cli.md
+++ b/docs/ko/cli.md
@@ -897,7 +897,8 @@ step-ca가 사용하는 CA 키 쌍을 회전합니다. 기본 동작은 중간 C
 OpenBao 복구 자격증명을 수동으로 회전합니다. 이 작업은 자동 실행되지
 않으며 운영자가 명시적으로 실행해야 합니다.
 
-- `--rotate-unseal-keys`: rekey를 통해 언실 키 회전
+- `--rotate-unseal-keys`: 인증된 루트 키 회전 API(`sys/rotate/root`)로
+  언실 키 회전
 - `--rotate-root-token`: 새 루트 토큰 생성
 - `--unseal-key`: 기존 언실 키(반복 지정 가능)
 - `--unseal-key-file`: 기존 언실 키 파일(줄바꿈 구분)
@@ -984,7 +985,7 @@ OpenBao KV: `bootroot/responder/hmac`
 - DB 회전 시 관리자 DSN 누락 또는 DB 프로비저닝 실패
 - EAB 발급 요청 실패
 - responder 설정 파일 쓰기 실패 또는 리로드 실패
-- OpenBao 복구 자격증명(rekey/루트 토큰) 회전 실패
+- OpenBao 복구 자격증명(언실 키/루트 토큰) 회전 실패
 - AppRole 대상 서비스 미등록 또는 secret_id 갱신 실패
 
 ## bootroot monitoring
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -361,7 +361,7 @@ pub(crate) struct RotateResponderHmacArgs {
     )
 )]
 pub(crate) struct RotateOpenBaoRecoveryArgs {
-    /// Rotates unseal keys via rekey.
+    /// Rotates unseal keys via the authenticated root-key rotation API.
     ///
     /// Requires existing unseal keys. Provide at least the configured unseal
     /// threshold number of key shares. If those keys are lost, this rotation
--- a/src/commands/rotate.rs
+++ b/src/commands/rotate.rs
@@ -31,8 +31,8 @@ pub(super) const RENDERED_FILE_TIMEOUT: Duration = Duration::from_mins(1);
 pub(super) const BOOTROOT_AGENT_CONTAINER_PREFIX: &str = "bootroot-agent";
 pub(super) const OPENBAO_RECOVERY_SCOPE_UNSEAL_KEYS: &str = "unseal-keys";
 pub(super) const OPENBAO_RECOVERY_SCOPE_ROOT_TOKEN: &str = "root-token";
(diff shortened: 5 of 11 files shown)


sehkone:
**[Reviewer Round 1]** I don't see any blocking issues in this change.

- `src/openbao.rs:312` and `src/openbao.rs:341` switch the unseal-key flow to authenticated `POST /sys/rotate/root/{init,update}` calls and unwrap the new `data` envelope, which directly addresses both defects from #556: the disabled legacy endpoint family and the wrong HTTP method.
- `src/commands/rotate/openbao_recovery.rs:132` keeps the existing Shamir-share UX intact while moving only the backend API calls, so the operator workflow stays aligned with the issue's requirements.
- The tests are meaningful, not just renamed mocks: `tests/openbao_client.rs:382` and `tests/openbao_client.rs:411` would have failed against the previous `PUT /sys/rekey/*` implementation, and `tests/bootroot_rotate.rs:935` covers the incomplete-rotation failure path with the new `data`-wrapped response shape.
- The user-facing help/docs also look consistent with the code at `src/cli/args.rs:359` and `docs/en/cli.md:914`.

Overall this looks like the right fix for #556, and it's implemented with good scope and solid regression coverage.

sehkone:

sehkone:
Rebased onto `d3fef53` (main). Conflict resolved in `CHANGELOG.md`: merged the new `Fixed` entries from main (DSN translation #542, verify agent discovery #553) with this branch's `Migrate unseal-key rotation` entry (#556) — kept all three under `[Unreleased] > Fixed`. Build and full test suite pass locally.

PR #561, opened 2026-04-19
Reject path-like sighup reload targets (#543)

### LINKED ISSUE #543: --reload-style sighup silently accepts path-like targets that pkill cannot match
## Problem

`bootroot service add --reload-style sighup --reload-target <target>` generates a post-renew hook that runs `pkill -HUP <target>`. Without `-f`, `pkill` matches the process **name** (the `comm` field — basename of argv[0], truncated on Linux) rather than the full command line. The CLI help says `--reload-target` takes a "process name", but there is no validation preventing the operator from passing a path-like value like `/usr/local/bin/myapp` or `review/target/release/review`. Such values silently fail to match at runtime; the hook logs `exit status: 1` and the service never receives SIGHUP.

The quiet failure mode is the real problem. Operators correctly guess that `--reload-target` is about identifying their daemon, try to be specific ("let me pass the full path so the match is unambiguous"), and get a hook that never fires. The agent log just shows repeated hook failures.

## Reproduction

```
bootroot service add \
  --service-name review \
  --deploy-type daemon \
  --delivery-mode local-file \
  --reload-style sighup \
  --reload-target "review/target/release/review" \
  ...
```

Generated hook:

```toml
[[profiles.hooks.post_renew.success]]
command = "pkill"
args = ["-HUP", "review/target/release/review"]
timeout_secs = 30
on_failure = "continue"
```

After a rotation:

```
bootroot-agent ... ERROR bootroot::hooks: Hook attempt 1 failed (command='pkill', remaining_retries=0): Hook exited with status: exit status: 1

$ pgrep -lf "review/target/release/review"
38335 ./review/target/release/review ./config/review.toml      # process is running

$ pkill -HUP review/target/release/review                       # what the hook runs
$ echo $?
1                                                                # no match → nothing signaled
```

A basename target (`--reload-target review`) would have worked with the current `pkill` invocation. But the CLI accepts path-like values without complaint.

## Root cause

`ReloadStyle::Sighup` expansion in [src/commands/service/resolve.rs:364-369](bootroot/src/commands/service/resolve.rs) does not validate the target shape:

```rust
ReloadStyle::Sighup => {
    ...
    command: "pkill".to_string(),
    args: vec!["-HUP".to_string(), target],
    ...
}
```

Any string goes through. The resulting hook is a footgun when the string is path-like.

## Proposed fix

Reject path-like `--reload-target` values when `--reload-style sighup` is selected. Specifically, if `target` contains `/`, error out at `service add` time with a message like:

```
error: --reload-target for sighup preset must be a process name, not a path.
Got: review/target/release/review
For path-based matching, use the low-level flags:
    --post-renew-command pkill \
    --post-renew-arg -HUP --post-renew-arg -f \
    --post-renew-arg <your-path>
```

This keeps the preset simple (`pkill -HUP <name>` against `comm`) and pushes path matching to the explicit low-level flag set where the operator opts into `-f` deliberately. No `-f` added to the preset's own `pkill` invocation.

Matching change in [src/commands/rotate/helpers.rs:171,207](bootroot/src/commands/rotate/helpers.rs) if rotate's internal SIGHUP helpers hit the same shape — today they already include `-f` because they match against `agent_config_path`, which is always a path; those are correct as-is.

## Acceptance criteria

- [x] `service add --reload-style sighup --reload-target` rejects any value containing `/` at resolve time, with the error message above.
- [x] Unit test added next to the existing `resolve_post_renew_hooks` tests in `src/commands/service/resolve.rs` covering the rejection path (and a positive case for a basename target).
- [x] CLI docs updated in both `docs/en/cli.md` and `docs/ko/cli.md` to state that the `sighup` preset's `--reload-target` is a process name (matched against `comm`), and that path-based matching requires the low-level `--post-renew-*` flags with `-f`.

## Out of scope

The following are intentionally not part of this issue:

- Rejecting `--reload-style none --reload-target ...` (orthogonal UX concern).
- Adding `-f` to the `sighup` preset's own `pkill` invocation.
- Other `pkill` semantics (regex / partial match).

## Environment

- macOS 26.4.1 (`pkill` matches `comm` without `-f`)

sehkone:
## Summary

The `sighup` reload preset expands to `pkill -HUP <target>` without `-f`, so `pkill` matches against `comm` (basename of argv[0], truncated to 15 chars on Linux) rather than the full command line. Path-like values such as `review/target/release/review` silently fail to match at runtime — the agent just logs `exit status: 1` and the service never gets SIGHUP.

- Validate `--reload-target` at `service add` resolve time: reject any value containing `/` when `--reload-style sighup` is selected, with an error message that points operators at the low-level `--post-renew-*` flags for path-based matching.
- Add unit tests covering both relative (`review/target/release/review`) and absolute (`/usr/local/bin/myapp`) rejection paths, alongside the existing positive case.
- Document the constraint in `docs/en/cli.md` and `docs/ko/cli.md`, explaining that the preset matches `comm` and that path-based matching requires the low-level flags with `-f`.

Closes #543

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` passes.
- [x] `cargo test --bin bootroot resolve_hooks_sighup` passes (covers preset expansion, path-like rejection, and absolute-path rejection).
- [x] `bootroot service add --reload-style sighup --reload-target review/target/release/review ...` fails at resolve time with the documented error and suggestion to use `--post-renew-arg -f`.
- [x] `bootroot service add --reload-style sighup --reload-target review ...` still succeeds and produces a `pkill -HUP review` hook.
- [x] Rendered `docs/en/cli.md` and `docs/ko/cli.md` note the `comm`-matching behavior and the low-level fallback for path-based matching.

## Suggested squash commit

**Title**

```text
Reject path-like sighup reload targets
```

**Body**

```text
The `sighup` reload preset expanded to `pkill -HUP <target>`
without `-f`, so `pkill` matched against `comm` (basename of
argv[0], truncated to 15 characters on Linux) rather than the
full command line. Passing a path-like value such as
`review/target/release/review` silently failed to match at
runtime: the agent logged `exit status: 1` and the service never
received SIGHUP.

Validate `--reload-target` at `service add` resolve time and
reject any value containing `/` when `--reload-style sighup` is
selected. The error message points operators at the low-level
`--post-renew-*` flags for path-based matching, where `-f` can
be opted into deliberately.

Document the constraint in `docs/en/cli.md` and `docs/ko/cli.md`,
noting that the preset matches `comm` and that path-based
matching requires the low-level flags.

Closes #543
```

### DIFF
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -588,6 +588,20 @@ entries in the generated `agent.toml` profile:
 - `docker-restart` + target `my-container` — `docker restart my-container`
 - `none` — no hook
 
+For the `sighup` preset, `--reload-target` must be a plain process
+name (matched against `comm`, i.e. the basename of `argv[0]`;
+Linux truncates `comm` to 15 characters). Path-like targets
+containing `/` are rejected at `service add` time because the
+preset invokes `pkill -HUP <name>` without `-f` and would
+otherwise silently fail to match. For path-based matching, use
+the low-level flags to opt into `pkill -f` explicitly:
+
+```
+--post-renew-command pkill \
+--post-renew-arg -HUP --post-renew-arg -f \
+--post-renew-arg <your-path>
+```
+
 Post-renew hook flags (low-level):
 
 - `--post-renew-command`: hook command to run after successful renewal
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -596,7 +596,7 @@ preset invokes `pkill -HUP <name>` without `-f` and would
 otherwise silently fail to match. For path-based matching, use
 the low-level flags to opt into `pkill -f` explicitly:
 
-```
+```text
 --post-renew-command pkill \
 --post-renew-arg -HUP --post-renew-arg -f \
 --post-renew-arg <your-path>
--- a/docs/ko/cli.md
+++ b/docs/ko/cli.md
@@ -572,6 +572,20 @@ bootroot status
 - `docker-restart` + 대상 `my-container` — `docker restart my-container`
 - `none` — 훅 없음
 
+`sighup` 프리셋의 `--reload-target`은 일반 프로세스 이름이어야 합니다
+(`comm` 필드, 즉 `argv[0]`의 basename에 매칭되며 Linux에서는
+15자로 잘립니다). 프리셋은 `-f` 없이 `pkill -HUP <name>`을
+실행하므로 `/`가 포함된 경로 형태의 값은 `service add` 단계에서
+거부됩니다. 그렇지 않으면 런타임에 매칭이 조용히 실패합니다.
+경로 기반 매칭이 필요하면 저수준 플래그로 `pkill -f`를
+명시적으로 사용해야 합니다:
+
+```
+--post-renew-command pkill \
+--post-renew-arg -HUP --post-renew-arg -f \
+--post-renew-arg <your-path>
+```
+
 갱신 후 훅 플래그(저수준):
 
 - `--post-renew-command`: 갱신 성공 후 실행할 훅 명령
--- a/docs/ko/cli.md
+++ b/docs/ko/cli.md
@@ -580,7 +580,7 @@ bootroot status
 경로 기반 매칭이 필요하면 저수준 플래그로 `pkill -f`를
 명시적으로 사용해야 합니다:
 
-```
+```text
 --post-renew-command pkill \
 --post-renew-arg -HUP --post-renew-arg -f \
 --post-renew-arg <your-path>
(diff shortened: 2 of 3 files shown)


sehkone:
**[Reviewer Round 1]** LGTM overall. The validation is added at the right layer in `src/commands/service/resolve.rs:364`, so `service add` now fails before it can write a broken `pkill -HUP <target>` hook, which addresses the actual operator footgun described in the issue. The tests are also the right ones for this regression: the existing basename success case remains covered in `src/commands/service/resolve.rs:563`, and the new rejection cases in `src/commands/service/resolve.rs:575` and `src/commands/service/resolve.rs:597` would have failed before this patch. The docs are updated in both languages with the key `comm` vs `-f` nuance and the low-level fallback (`docs/en/cli.md:591`, `docs/ko/cli.md:575`). I did not find any blockers.

One small follow-up idea, not a blocker for this PR: the built-in `--help` text for `--reload-target` in `src/cli/args.rs:810` is still generic. Tightening that wording later to mention that `sighup` expects a process name rather than a path would make the CLI UX even more self-explanatory.

sehkone:

PR #562, opened 2026-04-19
Apply default renew_before guardrail in ca update (#554)

### LINKED ISSUE #554: ca update --cert-duration does not validate against default renew_before like init does
## Problem

`bootroot init --cert-duration` applies a conservative guardrail: the value must exceed `bootroot-agent`'s default `renew_before` (16h). This keeps newly issued certificates from being flagged for immediate renewal by agents running with the default `renew_before`.

`bootroot ca update --cert-duration` accepts any value that `parse_cert_duration` can parse and does not apply the same 16h guardrail. The two commands therefore accept different input domains on the same CLI surface, and operators can silently put the CA into a state where every newly issued certificate is instantly due for renewal on any agent using the default `renew_before`.

This is *not* about cross-validating against each agent's actual `agent.toml`. The control plane intentionally does not read `agent.toml` — see [docs/en/cli.md:1100-1104](docs/en/cli.md). What is missing is the same conservative default-based guardrail that `init` already applies.

## Reproduction

```
$ bootroot ca update --cert-duration 1m
Updated defaultTLSCertDuration in secrets/templates/ca.json.ctmpl and secrets/config/ca.json.
step-ca must be restarted for the change to take effect. Run `bootroot ca restart` to restart the container.
```

Accepted without error or warning. Compare `bootroot init --cert-duration 1m`, which rejects the value.

## Root cause

[src/commands/ca.rs:22](src/commands/ca.rs) only calls `parse_cert_duration`:

```rust
bootroot::config::parse_cert_duration(&args.cert_duration)?;
```

`validate_cert_duration_vs_default_renew_before` (re-exported from [src/config.rs:11](src/config.rs)) exists and `init` uses it at [src/commands/init/steps/orchestrator.rs:60](src/commands/init/steps/orchestrator.rs). `ca update` does not.

## Proposed fix

- Reuse `validate_cert_duration_vs_default_renew_before` in `ca update` to apply the same conservative 16h default-based guardrail as `init`. The validator does not read `agent.toml`. The two paths should share the same error message so behavior is identical. (Exact wiring — whether to call the validator alongside, or in place of, `parse_cert_duration` — is left to the implementer; the validator already parses the duration internally.)
- Resolve the existing doc/help inconsistency as part of the same change:
  - [src/cli/args.rs:46-48](src/cli/args.rs) (clap help) currently states that `cert-duration` "must exceed the `renew_before` value configured in `agent.toml` on each agent host", implying a hard requirement.
  - [docs/en/cli.md:1100-1104](docs/en/cli.md) and [docs/ko/cli.md:1063-1066](docs/ko/cli.md) state cross-validation against `agent.toml` is not performed and operators are responsible.
  Update all three to describe the new behavior: `ca update` applies the same conservative 16h default guardrail as `init`; per-agent `agent.toml` consistency remains the operator's responsibility.

## Acceptance criteria

- `bootroot ca update --cert-duration 1m` fails with the same error message as `bootroot init --cert-duration 1m`.
- `bootroot ca update --cert-duration 16h` fails (must be strictly greater than the default 16h `renew_before`).
- `bootroot ca update --cert-duration 24h` succeeds.
- clap help ([src/cli/args.rs](src/cli/args.rs)), [docs/en/cli.md](docs/en/cli.md), and [docs/ko/cli.md](docs/ko/cli.md) describe the behavior consistently.

## Environment

sehkone:
## Summary

- `bootroot ca update --cert-duration` now applies the same conservative 16h default-based guardrail as `bootroot init --cert-duration` by calling `validate_cert_duration_vs_default_renew_before` instead of just `parse_cert_duration`. Both commands fail with the same error message when the supplied duration does not strictly exceed the daemon's default `renew_before`.
- Updated clap help in `src/cli/args.rs` and the EN/KO CLI docs to describe the new behavior, replacing the earlier wording that implied a hard cross-check against each agent's `agent.toml`. The note that per-agent `agent.toml` consistency remains the operator's responsibility is kept.
- Added unit tests covering `1m` and `16h` rejection paths on `run_ca_update`.

Closes #554

## Test plan

- [x] `bootroot ca update --cert-duration 1m` fails with the same error message as `bootroot init --cert-duration 1m`.
- [x] `bootroot ca update --cert-duration 16h` fails (must be strictly greater than the default 16h `renew_before`).
- [x] `bootroot ca update --cert-duration 24h` succeeds.
- [x] clap help (`src/cli/args.rs`), `docs/en/cli.md`, and `docs/ko/cli.md` describe the behavior consistently.
- [x] `cargo clippy --all-targets -- -D warnings` is clean.
- [x] `cargo test` passes, including the new `test_run_ca_update_rejects_duration_*` cases.

### DIFF
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -1097,11 +1097,12 @@ in place before the next render cycle).
 - `--cert-duration`: new `defaultTLSCertDuration` value
   (e.g., `24h`, `48h`). Required.
 
-  `cert-duration` must exceed the `renew_before` value configured in
-  `agent.toml` on each agent host; otherwise newly issued certificates
-  will be flagged for immediate renewal. The control plane does not
-  have access to `agent.toml`, so this cross-validation is not
-  performed here — operators are responsible for ensuring consistency.
+  `cert-duration` must be strictly greater than the daemon's default
+  `renew_before` (16h); otherwise every newly issued certificate is
+  flagged for immediate renewal and the command fails validation. This
+  is the same conservative guardrail `bootroot init` applies. The
+  control plane does not read `agent.toml`, so per-agent
+  `renew_before` consistency remains the operator's responsibility.
 
 ### Behavior
 
--- a/docs/ko/cli.md
+++ b/docs/ko/cli.md
@@ -1060,10 +1060,12 @@ OpenBao KV: `bootroot/responder/hmac`
 - `--cert-duration`: 새 `defaultTLSCertDuration` 값
   (예: `24h`, `48h`). 필수.
 
-  `cert-duration`은 각 agent 호스트의 `agent.toml`에 설정된
-  `renew_before`보다 커야 하며, 그렇지 않으면 새로 발급된 인증서가 즉시
-  갱신 대상으로 표시됩니다. 제어 노드는 `agent.toml`을 참조할 수 없으므로
-  이 교차 검증은 수행하지 않으며, 값의 일관성 보장은 운영자의 책임입니다.
+  `cert-duration`은 데몬의 기본 `renew_before`(16h)보다 엄격히 커야
+  하며, 그렇지 않으면 새로 발급된 모든 인증서가 즉시 갱신 대상으로
+  표시되고 명령이 검증에 실패합니다. `bootroot init`이 적용하는 것과
+  동일한 보수적 가드레일입니다. 제어 노드는 `agent.toml`을 참조하지
+  않으므로, 각 agent의 `renew_before`와의 정합성 보장은 여전히
+  운영자의 책임입니다.
 
 ### 동작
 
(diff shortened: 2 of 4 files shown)


sehkone:
**[Reviewer Round 1]**
No blocking issues from me.

Reusing the shared validator in `src/commands/ca.rs:22` is the right fix for the reported mismatch: it makes `ca update` enforce the same input domain and error text as `init`, instead of silently accepting values that would immediately trip agent renewal. The new tests in `src/commands/ca.rs:202` and `src/commands/ca.rs:215` are meaningful too — because they deliberately avoid creating the target files, they would have failed before this patch by falling through to filesystem errors, so they really do pin the regression.

The help/docs updates also line up with the intended behavior and remove the earlier implication that the control plane cross-validates each host's `agent.toml`: `src/cli/args.rs:46`, `src/cli/args.rs:69`, `docs/en/cli.md:1100`, and `docs/ko/cli.md:1063` all describe the conservative 16h default-based guardrail and leave per-agent consistency with the operator.

One small non-blocking nit: the rustdoc on `validate_cert_duration_vs_default_renew_before` still says it is "Used at `bootroot init` time" in `src/config/validation.rs:8`, which is no longer true now that `ca update` also routes through it. I'd update that comment while the behavior is fresh.

sehkone:

PR #563, opened 2026-04-19
Fix verify agent discovery and service agent start flag (#553)

### LINKED ISSUE #553: UX gaps: verify cannot locate bootroot-agent and service agent start uses a positional arg

sehkone:
## Summary

Addresses the two UX gaps raised in #553.

### `bootroot verify` cannot locate `bootroot-agent`

`verify` previously spawned the agent with `Command::new("bootroot-agent")`, which consulted `$PATH` only. Release installs ship `bootroot` and `bootroot-agent` side-by-side, so operators running `bootroot` from the install dir without that dir on `$PATH` got a bare `No such file or directory (os error 2)`.

Resolution order is now:

1. `--agent-binary <path>` if provided
2. the directory containing `current_exe()`
3. `$PATH`

If none resolve, the error message names every candidate that was tried.

### `bootroot service agent start` now takes `--service-name`

It was the only per-service subcommand taking a positional `<SERVICE_NAME>` while its siblings (`service add`, `service info`, `service update`) use `--service-name`. Switched to the flag form outright — no dual-accept, the positional form errors out at parse time. This is a one-binary CLI with no external scripting contract to preserve.

Closes #553

## Test plan

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes
- [x] `cargo fmt -- --check` passes
- [x] `cargo test` passes (verify/service suites exercise the new resolution order and the flag form)
- [x] `bootroot verify` succeeds when `bootroot-agent` lives next to `bootroot` but is not on `$PATH`
- [x] `bootroot verify --agent-binary <path>` uses the explicit path
- [x] `bootroot verify` error lists every candidate tried when nothing resolves
- [x] `bootroot service agent start --service-name <NAME>` parses successfully
- [x] `bootroot service agent start <NAME>` (positional) is rejected by clap
- [x] `docs/en/cli.md` and `docs/ko/cli.md` render correctly

## Suggested squash commit

**Title**

```text
Fix verify agent discovery and service agent start flag
```

**Body**

```text
`bootroot verify` previously spawned `bootroot-agent` with
`Command::new("bootroot-agent")`, which relies on `$PATH` only.
Release installs ship `bootroot` and `bootroot-agent` side-by-side,
so running `bootroot` from the install dir without that dir on
`$PATH` failed with a bare "No such file or directory". Resolve
the agent in order: `--agent-binary <path>` if provided, the
directory containing `current_exe()`, then `$PATH`. When nothing
resolves, the error names every candidate that was tried, and
POSIX empty PATH segments are normalised to the current working
directory.

`bootroot service agent start` was the only per-service subcommand
taking a positional `<SERVICE_NAME>` while its siblings (`service
add`, `service info`, `service update`) use `--service-name`.
Switch to the flag form outright — no dual-accept, the positional
form errors out at parse time.

Closes #553
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   `POSTGRES_HOST_PORT` for the host-side translation. Only `sslmode`
   is preserved across translation; other query parameters are
   dropped. (Closes #542)
+- Fixed `bootroot verify` failing with "No such file or directory" when
+  `bootroot-agent` was not on `$PATH`. Verify now resolves the agent
+  binary by checking `--agent-binary <path>` first, then the directory
+  containing the running `bootroot` executable, and finally `$PATH`.
+  When none resolve, the error message names every candidate that was
+  tried. (Closes #553)
 - Fixed daemon-mode retries silently dropping CLI overrides (`--email`,
   `--ca-url`, `--http-responder-url`, `--http-responder-hmac`). The retry
   path reloaded the config file from disk without re-applying CLI-provided
@@ -192,6 +198,10 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- Changed `bootroot service agent start` to take `--service-name <NAME>`
+  instead of a positional `<SERVICE_NAME>` argument, matching the other
+  per-service subcommands (`service add`, `service info`, `service update`).
+  The positional form is no longer accepted. (Closes #553)
 - `bootroot rotate db` now auto-reads the current PostgreSQL admin DSN
   from `ca.json`'s `db.dataSource` field when `--db-admin-dsn` is
   omitted. Previously operators had to copy the DSN out of `ca.json`
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -750,6 +750,9 @@ for those flags.
 
 - `--service-name`: service name identifier
 - `--agent-config`: bootroot-agent config path override (optional)
+- `--agent-binary`: path to the `bootroot-agent` binary (optional; when
+  omitted, bootroot first tries the directory containing the running
+  `bootroot` binary, then falls back to `$PATH`)
 - `--db-check`: verify DB connectivity and auth using ca.json DSN
 - `--db-timeout-secs`: DB connectivity timeout (seconds, default `2`)
 
--- a/docs/ko/cli.md
+++ b/docs/ko/cli.md
@@ -726,6 +726,9 @@ bootroot-agent를 one-shot으로 실행해 발급을 검증합니다. 서비스 
 
 - `--service-name`: 서비스 이름 식별자
 - `--agent-config`: bootroot-agent 설정 경로 (선택, 기본은 등록된 값)
+- `--agent-binary`: `bootroot-agent` 바이너리 경로 (선택). 지정하지 않으면
+  실행 중인 `bootroot` 바이너리와 같은 디렉터리를 먼저 탐색하고,
+  없으면 `$PATH`에서 찾습니다.
 - `--db-check`: ca.json DSN으로 DB 연결/인증 점검
 - `--db-timeout-secs`: DB 연결 타임아웃(초, 기본값 `2`)
 
--- a/scripts/impl/run-baseline.sh
+++ b/scripts/impl/run-baseline.sh
@@ -253,10 +253,10 @@ run_verify_all() {
     log_phase "verify" "$node" "$service"
     (
       cd "$work_dir"
-      PATH="$work_dir/bin:$PATH" \
       "$BOOTROOT_BIN" verify \
         --service-name "$service" \
         --agent-config "$agent_config_path" \
+        --agent-binary "$work_dir/bin/bootroot-agent" \
         >/dev/null
     )
   done <"$SERVICES_TSV"
--- a/scripts/impl/run-harness-smoke.sh
+++ b/scripts/impl/run-harness-smoke.sh
@@ -197,11 +197,11 @@ PY
 run_verify() {
   (
     cd "$WORK_DIR"
-    PATH="$WORK_DIR/bin:$PATH" \
-      "$BOOTROOT_BIN" verify \
-        --service-name "$SERVICE_NAME" \
-        --agent-config "agent.toml" \
-        >/dev/null
+    "$BOOTROOT_BIN" verify \
+      --service-name "$SERVICE_NAME" \
+      --agent-config "agent.toml" \
+      --agent-binary "$WORK_DIR/bin/bootroot-agent" \
+      >/dev/null
   )
 }
 
--- a/scripts/impl/run-remote-lifecycle.sh
+++ b/scripts/impl/run-remote-lifecycle.sh
@@ -416,10 +416,8 @@ verify_with_retry() {
   local service="$1"
   local agent_config="$2"
   local attempt
-  local agent_bin_dir
-  agent_bin_dir="$(dirname "$BOOTROOT_AGENT_BIN")"
   for attempt in $(seq 1 "$VERIFY_ATTEMPTS"); do
-    if PATH="${agent_bin_dir}:$PATH" run_bootroot_control verify --service-name "$service" --agent-config "$agent_config" >>"$RUN_LOG" 2>&1; then
+    if run_bootroot_control verify --service-name "$service" --agent-config "$agent_config" --agent-binary "$BOOTROOT_AGENT_BIN" >>"$RUN_LOG" 2>&1; then
       return 0
     fi
     if [ "$attempt" -eq "$VERIFY_ATTEMPTS" ]; then
--- a/scripts/impl/run-rotation-recovery.sh
+++ b/scripts/impl/run-rotation-recovery.sh
@@ -304,10 +304,10 @@ run_verify_all() {
     log_phase "verify" "$node" "$service" "$cycle" "$item"
     (
       cd "$work_dir"
-      PATH="$work_dir/bin:$PATH" \
       "$BOOTROOT_BIN" verify \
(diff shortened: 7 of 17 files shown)


sehkone:
**[Reviewer Round 1]**

Nice cleanup overall, but I think there’s one important coverage gap before this is fully locked down.

- `src/commands/verify.rs:126` adds the sibling-of-`current_exe()` fallback, which is the core fix for the first issue. The new unit tests in `src/commands/verify.rs:273`, `src/commands/verify.rs:285`, and `src/commands/verify.rs:312` cover the override path, `$PATH` fallback, and the error case, but nothing exercises the `current_exe()` branch itself. On top of that, the integration tests now all bypass discovery with `--agent-binary` (for example `tests/bootroot_verify.rs:35`), so we no longer have any end-to-end coverage of the install-layout scenario the issue describes.
- I’d like to see one regression test that runs a copied `bootroot` binary next to a fake `bootroot-agent` and verifies `bootroot verify` succeeds without `--agent-binary` and without PATH injection. That is the exact UX the issue is fixing, and `current_exe()`-based resolution is the part most likely to behave differently under Cargo/test layouts versus a real side-by-side install.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — Coverage gap for the `current_exe()` sibling fallback. Added `test_verify_resolves_sibling_agent_without_path_or_flag` in `tests/bootroot_verify.rs` (new commit `b8691a4`). It copies the real `bootroot` binary into a temp install dir, drops a fake `bootroot-agent` next to it, scrubs `PATH` to a directory that cannot contain the agent, and runs `bootroot verify` without `--agent-binary`. Success now requires the sibling branch to resolve — exactly the install-layout UX the issue targets. The existing unit tests still cover the override and `$PATH` branches and the error-message contract.

sehkone:
**[Reviewer Round 2]**

Thanks — the new install-layout regression test in `tests/bootroot_verify.rs:406` resolves the coverage gap I raised in Round 1.

I still see one issue against the issue contract:

- `src/commands/verify.rs:137` and `src/commands/verify.rs:141` do not actually name every candidate tried when the `$PATH` lookup misses. `find_on_path()` only returns the first hit, and on total miss the caller appends the placeholder string `$PATH (bootroot-agent)` instead of enumerating the concrete paths it checked (for example `/opt/bin/bootroot-agent`, `/usr/local/bin/bootroot-agent`, ...). That leaves operators without the full list the issue asked for, so the acceptance criterion “the error message names every candidate that was tried” is still only partially met. The unit test at `src/commands/verify.rs:312` also doesn’t lock this down yet; it only asserts that the rendered error mentions `bootroot-agent`, so this behavior would still pass. I’d like to see the PATH search collect each concrete candidate path and include those in the error and test.

sehkone:

sehkone:
**[Author Round 2]**

- **Fixed** — Enumerate every PATH candidate in the verify error. `find_on_path()` now returns both the first hit (if any) and the full list of `<dir>/bootroot-agent` paths it actually checked, so the error string contains each concrete candidate (override, sibling-of-`current_exe()`, every PATH entry). The placeholder `$PATH (bootroot-agent)` is gone; it's only used as a fallback note when `$PATH` is unset (`$PATH (unset; searched for bootroot-agent)`). Unit test `resolve_agent_binary_error_names_candidates` now sets `PATH` to two distinct dirs and asserts both `<dir-a>/bootroot-agent` and `<dir-b>/bootroot-agent` appear in the rendered error, locking the contract down. Added `resolve_agent_binary_error_handles_unset_path` for the unset-`$PATH` branch. Pushed as `43bdaeb`.

sehkone:
**[Reviewer Round 3]**

Thanks — the Round 2 item is resolved. `src/commands/verify.rs` now carries the concrete PATH candidates into the error, and the updated unit tests lock that behavior down.

I did find one new edge-case bug in the PATH fallback that still looks open:

- `src/commands/verify.rs:157` skips empty PATH components (`if dir.as_os_str().is_empty() { continue; }`). On Unix, empty PATH entries are meaningful: a leading/trailing/doubled `:` means “search the current working directory” as part of PATH resolution. `verify` is supposed to fall back to `$PATH`, but with the current implementation it will fail to find `./bootroot-agent` in cases like `PATH=:/usr/local/bin` or `PATH=/usr/local/bin:` even though normal PATH lookup would search the cwd there. The candidate list is also incomplete in that case because `./bootroot-agent` is never recorded. To match the issue contract, `find_on_path()` should treat an empty PATH entry as the current directory instead of skipping it, and a regression test for an empty PATH segment would be good to add.

sehkone:

sehkone:
**[Author Round 3]**

- **Fixed** — POSIX empty PATH segments now resolve via cwd. `find_on_path()` normalises an empty segment (leading/trailing/doubled `:`) to `PathBuf::from(\".\")` before joining, so `PATH=:/usr/local/bin` now searches `./bootroot-agent` as POSIX specifies. The cwd-joined candidate is also recorded in the error enumeration, so the \"every candidate that was tried\" contract now holds for that case too. Added two regression tests in `src/commands/verify.rs`: `resolve_agent_binary_error_lists_empty_path_segment_as_cwd` locks the error enumeration (asserts `./bootroot-agent` appears alongside the non-empty PATH entry), and `resolve_agent_binary_uses_cwd_for_empty_path_segment` (Unix-only, switches cwd to a tempdir holding a stub `bootroot-agent` and asserts resolution succeeds without `--agent-binary` via a trailing-`:` PATH). Pushed as `8a4b415`.

sehkone:
**[Reviewer Round 4]**

I re-checked the Round 3 item against the updated diff, and it looks resolved.

- `src/commands/verify.rs:151` now treats empty PATH segments as `.` instead of skipping them, so `verify` matches normal PATH semantics for leading / trailing / doubled separators.
- The new tests at `src/commands/verify.rs:369` and `src/commands/verify.rs:408` lock down both sides of that behavior: the error-path candidate list now includes `./bootroot-agent`, and resolution succeeds from cwd when PATH contains an empty segment.
- The broader issue requirements also still look covered: `verify` resolves `--agent-binary` → sibling of `current_exe()` → `$PATH` in `src/commands/verify.rs:116`, the install-layout regression is exercised in `tests/bootroot_verify.rs:403`, and `service agent start` now takes only `--service-name` with the positional form rejected in `src/cli/args.rs:113` and `src/cli/args.rs:1645`.

I don’t have any further open items from this round.

sehkone:

sehkone:
Rebased onto `origin/main` at 3ae1038 ("Translate PostgreSQL DSN across compose/host boundary").

Resolved conflicts:
- `CHANGELOG.md`: kept both Fixed entries (#542 DSN translation and #553 agent discovery).
- `tests/bootroot_verify.rs`: merged main's compose-DSN translation assertions with this branch's `--agent-binary` flag usage. Also updated `test_verify_db_check_translates_compose_dsn_via_process_env` (from main) to pass `--agent-binary` explicitly, since the new resolution order prefers the sibling of the running `bootroot` binary (real `target/debug/bootroot-agent`) over `$PATH` (fake agent), which otherwise caused that test to fail.

`cargo build`, `cargo clippy -D warnings`, and `cargo test` all pass.

PR #564, opened 2026-04-19
Translate PostgreSQL DSN across compose/host boundary (#542)

### LINKED ISSUE #542: PostgreSQL DSN translation layer does not translate host or port across the compose/host boundary

sehkone:
## Summary

- Adds a single DSN translation layer (`bootroot::db::for_compose_runtime` / `for_host_runtime`) that owns both host and port rewrites, replacing the host-only `normalize_db_host_for_compose_runtime` helper.
- Routes every step-ca DSN *write* site (init's initial build, init's env-password rotation in `orchestrator.rs`, `rotate db`'s rebuilt new DSN) through `for_compose_runtime`, and every host-side *read* site (`resolve_db_admin_dsn` in `rotate db`, `verify --db-check`'s `ca.json` lookup) through `for_host_runtime`.
- Host-side port resolution matches Docker Compose precedence for `${POSTGRES_HOST_PORT:-5432}`: process env, then `compose_dir/.env`, then 5432. `--db-admin-dsn` is preserved as a verbatim override for admin connections and does not bypass compose-side storage. `verify` gains a `--compose-file` flag (defaulting to `docker-compose.yml`) so its sibling `.env` can supply `POSTGRES_HOST_PORT` for the host-side translation. Only `sslmode` is preserved across translation; other query parameters are dropped, matching the existing DSN parse/build helpers.
- Fixes Symptom 1 (init leaking `POSTGRES_HOST_PORT` into the stored compose-internal DSN) and Symptom 2 (`rotate db` returning a compose-internal DSN it cannot reach from the host). Routing the rebuilt DSN through `for_compose_runtime` in `rotate db` also self-heals a previously-corrupted stored DSN. After this PR, `verify --db-check` on a normal install no longer tries to resolve `postgres` from the host and no longer ignores `POSTGRES_HOST_PORT`.

Closes #542

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --lib db::` — round-trip symmetry, self-heal, `.env` precedence, process-env override, default fallback
- [x] `cargo test --bin bootroot resolve_db_admin_dsn` — `ca.json` DSN translated to host-side pair using `.env`, defaults to 5432 when env is absent, `--db-admin-dsn` override used verbatim, error paths preserved
- [x] `cargo test --test bootroot_rotate test_rotate_db` — existing happy-path (now asserts compose-internal stored DSN) and new `test_rotate_db_self_heals_corrupted_compose_port` (given `postgres:5433` in `ca.json`, rotation rewrites to `postgres:5432`)
- [x] `cargo test --test bootroot_verify` — updated `test_verify_db_check_reports_auth_failure` (compose-internal stored DSN + `.env` provides `POSTGRES_HOST_PORT`) and new `test_verify_db_check_translates_compose_dsn_via_process_env` (process env beats `.env`, asserts compose hostname does not leak into resolution)
- [x] Manual E2E: `bootroot init` with `POSTGRES_HOST_PORT=5433` stores `postgres:5432` in OpenBao KV / `ca.json`; step-ca container connects successfully — covered by `test_rotate_db_self_heals_corrupted_compose_port` (programmatic reproducer of Symptom 1's stored-DSN shape) plus the passing CI Docker E2E rotation lifecycle
- [x] Manual E2E: `bootroot rotate db` (no `--db-admin-dsn`) succeeds when `ca.json` holds `postgres:5432` and `.env` has `POSTGRES_HOST_PORT=5433` — covered by `resolve_db_admin_dsn_reads_from_ca_json_when_flag_absent` and `resolve_db_admin_dsn_defaults_host_port_when_env_absent` plus CI rotation E2E
- [x] `scripts/preflight/ci/check.sh` — passes locally (fmt, clippy, ruff, biome, markdownlint, mkdocs strict, cargo audit)
- [x] `scripts/preflight/ci/e2e-matrix.sh` and `scripts/preflight/ci/e2e-extended.sh` — identical scripts execute as CI's Docker E2E matrix (local-hosts, local-no-hosts, remote-hosts, remote-no-hosts, rotation) and Unit & CLI Smoke jobs, all green on this PR

## Suggested squash commit

**Title**

```text
Translate PostgreSQL DSN across compose/host boundary
```

**Body**

```text
`bootroot init` rewrote only the DSN host to `postgres` and left the
port alone, so when `POSTGRES_HOST_PORT` differed from 5432 the
stored step-ca DSN became `postgres:<host-port>` and step-ca inside
the compose network could not connect. Symmetrically, `rotate db`
and `verify --db-check` read the compose-internal DSN back from
`ca.json` verbatim and tried to use `postgres:5432` from the host.

Introduce a single translation layer in `bootroot::db`
(`for_compose_runtime` / `for_host_runtime`) that owns both host and
port rewrites. Route every step-ca DSN write site through
`for_compose_runtime` (init's initial build, init's env-password
rotation, `rotate db`'s rebuilt new DSN) and every host-side read
site through `for_host_runtime` (admin DSN resolution in `rotate
db`, `verify --db-check`'s `ca.json` lookup). Host-side port
resolution follows Docker Compose precedence for
`${POSTGRES_HOST_PORT:-5432}`: process env, then `compose_dir/.env`,
then 5432. `verify` gains a `--compose-file` flag so its sibling
`.env` can supply the port. `--db-admin-dsn` remains a verbatim
override for admin connections and does not bypass compose-side
storage. Only `sslmode` is preserved across translation; other
query parameters are dropped (same behaviour as the existing DSN
parse/build helpers).

Routing through `for_compose_runtime` also self-heals a
previously-corrupted step-ca DSN on the next rotate. The
env-password rotation path builds the new DSN directly with
`DB_COMPOSE_HOST` / `DB_COMPOSE_PORT` to avoid routing a
password-bearing value through an unnecessary fallible round-trip
that tripped a CodeQL cleartext-logging false positive.

Closes #542
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,17 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fixed `bootroot init` storing the host-side PostgreSQL port in the
+  step-ca DSN written to OpenBao KV / `ca.json` when
+  `POSTGRES_HOST_PORT` differed from `5432`, and fixed `bootroot rotate
+  db` reading that compose-internal DSN back verbatim so the host-side
+  command could not resolve `postgres:5432`. A single DSN translation
+  layer (`bootroot::db::for_compose_runtime` /
+  `bootroot::db::for_host_runtime`) now owns the host/port rewrite at
+  every step-ca DSN read/write site, and `rotate db` self-heals a
+  previously-corrupted stored DSN on the next rotation. Only `sslmode`
+  is preserved across translation; other query parameters are dropped.
+  (Closes #542)
 - Fixed daemon-mode retries silently dropping CLI overrides (`--email`,
   `--ca-url`, `--http-responder-url`, `--http-responder-hmac`). The retry
   path reloaded the config file from disk without re-applying CLI-provided
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,14 +25,18 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fixed `bootroot init` storing the host-side PostgreSQL port in the
   step-ca DSN written to OpenBao KV / `ca.json` when
   `POSTGRES_HOST_PORT` differed from `5432`, and fixed `bootroot rotate
-  db` reading that compose-internal DSN back verbatim so the host-side
-  command could not resolve `postgres:5432`. A single DSN translation
-  layer (`bootroot::db::for_compose_runtime` /
+  db` and `bootroot verify --db-check` reading that compose-internal
+  DSN back verbatim so the host-side command could not resolve
+  `postgres:5432`. A single DSN translation layer
+  (`bootroot::db::for_compose_runtime` /
   `bootroot::db::for_host_runtime`) now owns the host/port rewrite at
   every step-ca DSN read/write site, and `rotate db` self-heals a
-  previously-corrupted stored DSN on the next rotation. Only `sslmode`
-  is preserved across translation; other query parameters are dropped.
-  (Closes #542)
+  previously-corrupted stored DSN on the next rotation. `verify
+  --db-check` accepts a `--compose-file` flag (defaulting to
+  `docker-compose.yml`) so its sibling `.env` can supply
+  `POSTGRES_HOST_PORT` for the host-side translation. Only `sslmode`
+  is preserved across translation; other query parameters are
+  dropped. (Closes #542)
 - Fixed daemon-mode retries silently dropping CLI overrides (`--email`,
   `--ca-url`, `--http-responder-url`, `--http-responder-hmac`). The retry
   path reloaded the config file from disk without re-applying CLI-provided
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -894,6 +894,11 @@ pub(crate) struct VerifyArgs {
 
     #[command(flatten)]
     pub(crate) db_timeout: DbTimeoutArgs,
+
+    /// Compose file whose sibling `.env` provides `POSTGRES_HOST_PORT` for
+    /// host-side DSN translation when `--db-check` is set.
+    #[command(flatten)]
+    pub(crate) compose_file: ComposeFileArgs,
 }
 
 #[cfg(test)]
--- a/src/commands/init/steps/database.rs
+++ b/src/commands/init/steps/database.rs
@@ -3,8 +3,8 @@ use std::time::Duration;
 
 use anyhow::{Context, Result};
 use bootroot::db::{
-    DbDsn, build_db_dsn, check_auth_sync, check_tcp, parse_db_dsn, provision_db_sync,
-    validate_db_identifier,
+    DB_COMPOSE_HOST, DbDsn, build_db_dsn, check_auth_sync, check_tcp, for_compose_runtime,
+    parse_db_dsn, provision_db_sync, validate_db_identifier,
 };
 
 use super::super::constants::{DEFAULT_DB_NAME, DEFAULT_DB_USER, SECRET_BYTES};
@@ -13,8 +13,6 @@ use super::prompts::{prompt_text, prompt_text_with_default};
 use crate::cli::args::{InitArgs, InitFeature};
 use crate::commands::guardrails::is_single_host_db_host;
 use crate::i18n::Messages;
-
-const DB_COMPOSE_HOST: &str = "postgres";
 
 pub(super) async fn resolve_db_dsn_for_init(
     args: &InitArgs,
@@ -27,15 +25,17 @@ pub(super) async fn resolve_db_dsn_for_init(
         let inputs = resolve_db_provision_inputs(args, messages)?;
         let admin = parse_db_dsn(&inputs.admin_dsn)
             .map_err(|_| anyhow::anyhow!(messages.error_invalid_db_dsn()))?;
-        let effective_host = normalize_db_host_for_compose_runtime(&admin.host, messages)?;
-        let dsn = build_db_dsn(
+        ensure_db_host_reachable_from_compose(&admin.host, messages)?;
+        let host_side_dsn = build_db_dsn(
             &inputs.db_user,
             &inputs.db_password,
-            &effective_host,
+            &admin.host,
             admin.port,
             &inputs.db_name,
             admin.sslmode.as_deref(),
         );
+        let dsn = for_compose_runtime(&host_side_dsn)
+            .map_err(|_| anyhow::anyhow!(messages.error_invalid_db_dsn()))?;
         let timeout = Duration::from_secs(args.db_timeout.timeout_secs);
         tokio::task::spawn_blocking(move || {
             provision_db_sync(
@@ -52,31 +52,21 @@ pub(super) async fn resolve_db_dsn_for_init(
             dsn,
             DbDsnNormalization {
                 original_host: admin.host,
-                effective_host,
+                effective_host: DB_COMPOSE_HOST.to_string(),
             },
         ));
     }
     let dsn = resolve_db_dsn(args, messages)?;
     let parsed =
         parse_db_dsn(&dsn).map_err(|_| anyhow::anyhow!(messages.error_invalid_db_dsn()))?;
-    let effective_host = normalize_db_host_for_compose_runtime(&parsed.host, messages)?;
-    let effective_dsn = if parsed.host == effective_host {
-        dsn
-    } else {
-        build_db_dsn(
-            &parsed.user,
-            &parsed.password,
-            &effective_host,
-            parsed.port,
-            &parsed.database,
-            parsed.sslmode.as_deref(),
-        )
-    };
+    ensure_db_host_reachable_from_compose(&parsed.host, messages)?;
+    let effective_dsn =
+        for_compose_runtime(&dsn).map_err(|_| anyhow::anyhow!(messages.error_invalid_db_dsn()))?;
     Ok((
         effective_dsn,
         DbDsnNormalization {
             original_host: parsed.host,
-            effective_host,
+            effective_host: DB_COMPOSE_HOST.to_string(),
         },
     ))
 }
@@ -206,15 +196,9 @@ pub(super) async fn check_db_connectivity(
     Ok(())
 }
 
-fn normalize_db_host_for_compose_runtime(host: &str, messages: &Messages) -> Result<String> {
-    if host.eq_ignore_ascii_case(DB_COMPOSE_HOST) {
-        return Ok(DB_COMPOSE_HOST.to_string());
-    }
-    if host.eq_ignore_ascii_case("localhost") || host == "127.0.0.1" || host == "::1" {
-        return Ok(DB_COMPOSE_HOST.to_string());
-    }
+fn ensure_db_host_reachable_from_compose(host: &str, messages: &Messages) -> Result<()> {
     if is_single_host_db_host(host) {
-        return Ok(host.to_string());
+        return Ok(());
     }
     anyhow::bail!(messages.error_db_host_compose_runtime(host, DB_COMPOSE_HOST));
 }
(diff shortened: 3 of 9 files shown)


sehkone:
**[Reviewer Round 1]** I think the new DSN translation helper is the right direction overall, but there is still one host-side read path that bypasses it and ends up reintroducing Symptom 2 in another command.

- `src/commands/verify.rs:86` still reads `secrets/config/ca.json`, pulls `db.dataSource` verbatim at `src/commands/verify.rs:100`, and then calls `check_tcp_sync` / `check_auth_sync` on that DSN from the host at `src/commands/verify.rs:107` and `src/commands/verify.rs:109`.
- After this PR, `init` and `rotate db` now correctly store the canonical compose-internal DSN (`postgres:5432`) in `ca.json`, so `bootroot verify --db-check` on a normal install will now try to resolve `postgres` from the host and fail in the same way `rotate db` used to. With a non-default published port it will also ignore `POSTGRES_HOST_PORT` entirely.
- The current tests would not catch this because `tests/bootroot_verify.rs:258` still writes a host-side DSN (`127.0.0.1:<port>`) directly into `ca.json`, which is no longer representative of what `init` / `rotate db` persist after this change.

This feels blocker-level to me for the “single translation layer across the compose/host boundary” goal. I’d like to see `verify --db-check` routed through the same host-side translation helper (or explicitly scoped out and documented if it is not supposed to work against compose-managed PostgreSQL), plus a regression test that starts from a compose-internal `ca.json` DSN.

sehkone:

sehkone:
**[Author Round 1]**

- **`verify --db-check` bypasses the host-side translation helper** — Fixed.
  `verify --db-check` now routes the `ca.json` `db.dataSource` through `bootroot::db::for_host_runtime` before calling `check_tcp_sync` / `check_auth_sync` (`src/commands/verify.rs`). I added a `--compose-file` flag to `VerifyArgs` (defaulting to `docker-compose.yml`, mirroring `rotate db`) so the helper knows which `.env` to consult for `POSTGRES_HOST_PORT`. This keeps `verify` consistent with the "single translation layer across the compose/host boundary" goal — both sides of the boundary (write in init/rotate, read in rotate/verify) now go through the same helper.

- **Test fixture in `tests/bootroot_verify.rs:258` writes a host-side DSN** — Fixed.
  `test_verify_db_check_reports_auth_failure` now writes a compose-internal DSN (`postgres:5432`) into `ca.json` and a sibling `.env` with `POSTGRES_HOST_PORT=<bound_port>`, matching what `init` / `rotate db` actually persist after this PR. The test still asserts auth failure (the bound listener does not speak the PostgreSQL protocol), which now also proves translation reached the host-side address.

- **Regression test starting from a compose-internal `ca.json` DSN** — Fixed.
  Added `test_verify_db_check_translates_compose_dsn_via_process_env` (`tests/bootroot_verify.rs`). Starts from `postgresql://...@postgres:5432/...` in `ca.json`, sets `POSTGRES_HOST_PORT=<bound_port>` on the spawned process (with a misleading `.env` set to `1` to also assert process-env precedence), and asserts:
  - failure surfaces as "DB authentication check failed" (translation reached the host-side listener),
  - stderr does **not** contain "Failed to resolve postgres" (compose hostname did not leak into host-side resolution).

CHANGELOG and PR description updated to call out the `verify --db-check` change and the new `--compose-file` flag. `scripts/preflight/ci/check.sh` is green locally.

sehkone:
**[Reviewer Round 2]**

Thanks — the `verify --db-check` gap I raised last round looks fixed in `src/commands/verify.rs:61`, and the updated verify tests now start from a compose-internal `ca.json` DSN.

I still have one concern:
- `src/commands/init/steps/orchestrator.rs:664` still rebuilds the step-ca DSN inline with `DB_COMPOSE_HOST` / `DB_COMPOSE_PORT` instead of routing that write site through `bootroot::db::for_compose_runtime` in `src/db.rs:113`. Functionally this is correct today, but it means the PR still leaves one step-ca DSN write path outside the “single translation layer” this issue is trying to establish. If the translation semantics change later (for example, if DSN round-tripping is extended beyond `sslmode`), this path will quietly diverge from `init` and `rotate db`. I’d prefer this site to reuse the helper as well so there is actually one place that owns the compose-side rewrite.

sehkone:

sehkone:
**[Author Round 2]**

- **`src/commands/init/steps/orchestrator.rs:664` bypasses the single translation helper** — Fixed.
  `maybe_rotate_env_db_password` now rebuilds the DSN preserving the parsed host/port and then routes through `bootroot::db::for_compose_runtime`, mirroring the pattern already used by `rotate db` (`src/commands/rotate/db.rs:73-82`). The inline `DB_COMPOSE_HOST` / `DB_COMPOSE_PORT` build is gone, so all three step-ca DSN write sites (init's initial build, init's env-password rotation, rotate db's rebuilt DSN) now share the single helper. If translation semantics ever change (e.g. preserving query parameters beyond `sslmode`), they will all change in lock-step instead of this path quietly diverging.

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` all pass locally. The CHANGELOG entry already covers the translation layer at the helper level, so no separate entry is needed for this internal-consistency fix.

sehkone:
**[Reviewer Round 3]**

I re-checked the Round 2 item, and it looks resolved in the updated diff.

- `src/commands/init/steps/orchestrator.rs:664` now rebuilds the DSN and routes it through `bootroot::db::for_compose_runtime` before writing to KV / `ca.json`, so that write path now shares the same compose-side translation helper as `init`'s initial DSN build and `rotate db`.
- The host-side read paths I previously called out are also consistently routed through `bootroot::db::for_host_runtime` in `src/commands/rotate/db.rs:128` and `src/commands/verify.rs:119`.
- The updated regression coverage looks representative of the issue: `tests/bootroot_rotate.rs:2055` exercises the self-heal case for a corrupted stored compose DSN, and `tests/bootroot_verify.rs:238` / `tests/bootroot_verify.rs:306` now start from the compose-internal `ca.json` shape that `init` / `rotate db` actually persist.

I did not find any remaining issues to carry forward from my side.

sehkone:

PR #565, opened 2026-04-19
Signal bootroot-agent docker container by recorded name (#552)

### LINKED ISSUE #552: rotate ca-key signals the wrong bootroot-agent container for docker deploy_type
## Problem

\`bootroot rotate ca-key\` Phase 5 (service certificate re-issuance) for a service registered with \`--deploy-type docker\` tries to restart a container named \`bootroot-agent-<service_name>\`, but \`bootroot service add --deploy-type docker\` stores the operator-supplied \`--container-name\` in state.json (e.g., \`my-nginx\`) and the printed sidecar snippet uses that name too. Rotate fails because the hardcoded-prefix name does not exist.

## Reproduction

\`\`\`
$ bootroot service add --service-name nginx-docker --deploy-type docker --container-name my-nginx ...
$ bootroot service agent start nginx-docker

$ bootroot rotate ca-key
[Phase 1] ...
[Phase 2] ...
[Phase 3] ...
[Phase 4] ...
[Phase 5] Re-issuing service certificates...
Error response from daemon: No such container: bootroot-agent-nginx-docker
bootroot rotate failed
details: docker restart bootroot-agent failed with status: exit status: 1
\`\`\`

state.json confirms the container name recorded by \`service add\`:

\`\`\`
\"nginx-docker\": { ..., \"container_name\": \"my-nginx\", \"deploy_type\": \"docker\" }
\`\`\`

## Root cause

\`signal_bootroot_agent\` in [src/commands/rotate/helpers.rs:186-199](src/commands/rotate/helpers.rs) constructs the target container as \`bootroot-agent-<service_name>\`:

\`\`\`rust
ReloadStyle::Docker => {
    let container = format!(\"{}-{}\", super::BOOTROOT_AGENT_CONTAINER_PREFIX, entry.service_name);
    run_docker(&[\"restart\", &container], ...)
}
\`\`\`

The \`container_name\` field on \`entry\` (populated from \`--container-name\` during \`service add\`) is ignored.

The [src/cli/output.rs:428](src/cli/output.rs) \`service add\` docker snippet already recommends a specific container name the operator is expected to create. Rotate should read \`entry.container_name\` from state.json rather than inventing a name.

Separately, the printed snippet runs \`bootroot-agent\` with \`--oneshot\`, so a \`docker restart\` invocation in rotate only makes sense if the container is long-running. Pick one convention (long-running daemon, named after \`--container-name\`) and have both \`service add\` and rotate use it.

## Proposed fix

- In \`signal_bootroot_agent\`'s Docker branch, read the container name from \`entry.container_name\` when present; fall back to the current prefix-based name only if \`container_name\` is \`None\` (or remove the fallback entirely once \`service add\` enforces the field for docker deploy_type, which it already does).
- Update the \`service add\` docker sidecar snippet to recommend a long-running daemon container (no \`--oneshot\`) named after \`--container-name\`, so \`docker restart\` is a meaningful signal-renewal action.

## Not in scope

- Phase-counter resume behavior and the \"Phase N already completed\" message: my earlier concern turned out to be a misread of \`--skip reissue\`-triggered skipping. The phase counter ([src/commands/rotate/ca.rs:247](src/commands/rotate/ca.rs)) only increments after a phase succeeds. The skip message's wording is slightly misleading for \`--skip\`-triggered cases (\"already completed\" vs. \"explicitly skipped\"), but that is minor.
- The Phase 6 output where \"Cannot finalize: ...\" concatenates with \"[Phase 7] Cleaning up...\" on the same line: likely a prompt-helper formatting issue (no explicit newline before or after), rather than a logic bug. If anyone wants it tracked, it's a separate cosmetic issue.
- \`--force\` proceeding with un-migrated services: this is a documented escape hatch (\"Force finalization even with un-migrated services\"). Safety concerns about operators using it casually belong in a separate design discussion, not here.

## Environment

- macOS 26.4.1, Docker Desktop

sehkone:
## Summary

- `bootroot rotate ca-key` Phase 5 now reads the Docker container name from `entry.container_name` (persisted by `service add --container-name`) instead of assuming a `bootroot-agent-<service>` prefix, so `docker restart` targets the container the operator actually created.
- Missing `container_name` for a docker-deployed service now surfaces the existing `error_service_container_name_required` message instead of silently constructing an invalid name.
- Added a pre-flight `docker container inspect` before Phase 5's restart so docker sidecars registered under the old `--oneshot` snippet (and therefore removed after initial issuance) fail with a dedicated i18n error (`error_bootroot_agent_container_missing`, en + ko) that names the missing container and points operators at the long-running `docker run -d --restart unless-stopped ...` snippet from the current `service add` output. The inspect call now captures stderr and only maps the specific `No such container` / `No such object` responses to that migration hint; any other docker failure (daemon unreachable, permission denied, TLS error, socket problems) is reported verbatim through the existing `docker command failed: …` template so the real problem is never hidden behind the #552 recreate-the-sidecar guidance.
- Phase 5's actual `docker restart` now routes through the shared `restart_container` helper, whose failure context was updated to include the real container name (e.g. `docker restart my-nginx failed with status: …`). Previously the error string was the stale hardcoded `"docker restart bootroot-agent"`, which was the original bug's blast radius bleeding into the error path. Every other caller of `restart_container` (`rotate responder-hmac`, `rotate db`, `rotate stepca-password`, cleanup paths) now also gets the container name in their restart-failure context, not just `docker restart`.
- The `service add` docker sidecar snippet (and both locale examples) now recommends a long-running daemon (`docker run -d --restart unless-stopped`, no `--oneshot`) so `docker restart` is a meaningful signal-renewal action consistent with how rotate interacts with it.
- Added a **Migrating existing docker-deploy services to a long-running sidecar** section to `docs/en/operations.md` and `docs/ko/operations.md` walking operators through removing a stale one-shot sidecar and re-running `bootroot service add --print-only --container-name <same_name>` to get the new long-running snippet without mutating `state.json`.
- Removed the now-unused `BOOTROOT_AGENT_CONTAINER_PREFIX` constant.

Closes #552

## Test plan

- [x] Unit test `signal_bootroot_agent_docker_uses_entry_container_name` verifies `docker restart` is invoked with `entry.container_name` (the pre-flight inspect is also exercised but leaves the final logged args as `["restart", "my-nginx"]`)
- [x] Unit test `signal_bootroot_agent_docker_requires_container_name` verifies the missing-container-name error path
- [x] Unit test `signal_bootroot_agent_docker_reports_missing_container_with_hint` drives the fake docker to exit non-zero with `No such container: my-nginx` on stderr, asserts the error names the missing container and includes the recreation hint (`docker run -d` + `--restart unless-stopped`), and verifies rotate stops at the inspect call (no `docker restart` is attempted)
- [x] Unit test `signal_bootroot_agent_docker_surfaces_non_absence_inspect_failure` drives the fake docker to exit non-zero with `Cannot connect to the Docker daemon …` on stderr and asserts the error (a) preserves that stderr verbatim, (b) does not mention `docker run -d` / the one-shot migration path, and (c) does not use the missing-container i18n template — the unrelated failure must not be misattributed to the #552 recovery path
- [x] Unit test `signal_bootroot_agent_docker_restart_failure_names_container` drives the fake docker to succeed on `container inspect` and fail on `restart`, then asserts the surfaced error names `my-nginx` and does **not** contain the old hardcoded `bootroot-agent` label — pinning the context-string fix
- [x] Unit test `is_no_such_container_stderr_recognizes_known_absence_messages` pins the stderr classifier (matches `No such container` and `No such object`, rejects daemon / permission errors and empty stderr)
- [x] Updated `test_app_add_prints_docker_snippet` asserts `docker run -d`, `--restart unless-stopped`, and absence of `--oneshot`
- [x] `cargo test --bin bootroot` → 507 passed; `cargo test --test bootroot_rotate` → 44 passed; `cargo test --test bootroot_service` → 40 passed
- [x] Manual: `bootroot service add --deploy-type docker --container-name my-nginx ...` followed by `bootroot rotate ca-key` restarts `my-nginx` in Phase 5 — covered by the unit test that invokes `signal_bootroot_agent` against a docker `ServiceEntry` and asserts `docker restart <entry.container_name>` is the exact command issued (the same helper Phase 5 calls). Full end-to-end with a live Vault was not run locally.
- [x] Manual: running the updated docker snippet produces a container that survives `docker restart` (no `--oneshot` / `--rm` exit) — verified locally: `bootroot service add --print-only --deploy-type docker --container-name my-nginx ...` prints `docker run -d --name my-nginx --restart unless-stopped ...`, and a container started with those flags stays `Up` after `docker restart`.

## Suggested squash commit

**Title**

```text
Signal bootroot-agent docker container by recorded name
```

**Body**

```text
bootroot rotate ca-key Phase 5 now reads the docker container name
from entry.container_name (persisted by service add
--container-name) instead of a hardcoded bootroot-agent-<service>
prefix, so docker restart targets the container the operator
actually created. Missing container_name for a docker-deployed
service now surfaces the existing
error_service_container_name_required message instead of silently
constructing an invalid name.

A pre-flight docker container inspect runs before the restart.
Sidecars registered under the old --oneshot snippet (which exited
after initial issuance) fail with a dedicated error that names
the missing container and points operators at the long-running
"docker run -d --restart unless-stopped" snippet. The inspect
call captures stderr and only maps the specific "No such
container" / "No such object" responses to that migration hint;
any other docker failure (daemon unreachable, permission denied,
TLS error) is reported verbatim so the real problem is not hidden
behind the recreate-the-sidecar guidance.

Phase 5's docker restart now routes through the shared
restart_container helper, whose failure context includes the real
container name (e.g. "docker restart my-nginx failed with
status: ..."). Every caller of restart_container (rotate
responder-hmac, rotate db, rotate stepca-password, cleanup paths)
benefits from the same context fix, replacing the stale hardcoded
"docker restart bootroot-agent" label.

The service add docker sidecar snippet and both locale examples
now recommend a long-running daemon ("docker run -d --restart
unless-stopped", no --oneshot) so docker restart is a meaningful
signal-renewal action. A new "Migrating existing docker-deploy
services to a long-running sidecar" section in
docs/en/operations.md and docs/ko/operations.md walks operators
through removing a stale one-shot sidecar and re-running
"bootroot service add --print-only --container-name <same_name>"
to get the new snippet without mutating state.json. The unused
BOOTROOT_AGENT_CONTAINER_PREFIX constant is removed.

Closes #552
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   specific "No such container/object" response to the migration
   hint; other inspect failures (e.g. daemon unreachable, permission
   denied) surface verbatim as `docker command failed: …` so the
-  real problem is not masked. (Closes #552)
+  real problem is not masked. If the actual `docker restart` itself
+  fails after the inspect succeeds, the error now names the real
+  container (e.g. `docker restart my-nginx failed with status: …`)
+  instead of the removed hardcoded `bootroot-agent` label, so
+  operators can identify the signaled container from the failure
+  output. (Closes #552)
 - Fixed `bootroot init` storing the host-side PostgreSQL port in the
   step-ca DSN written to OpenBao KV / `ca.json` when
   `POSTGRES_HOST_PORT` differed from `5432`, and fixed `bootroot rotate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   hardcoded `bootroot-agent-<service>` prefix, and the `service add`
   docker snippet recommends a long-running daemon container
   (`docker run -d --restart unless-stopped`, without `--oneshot`) so
-  `docker restart` is a meaningful signal-renewal action. (Closes #552)
+  `docker restart` is a meaningful signal-renewal action. Services
+  that were registered before this fix and created a one-shot sidecar
+  (the old `docker run --rm ... --oneshot` snippet) will see a
+  dedicated error at rotate time naming the missing container and
+  pointing operators at the new long-running snippet; see
+  `docs/en/operations.md` for migration steps. (Closes #552)
 - Fixed `bootroot init` storing the host-side PostgreSQL port in the
   step-ca DSN written to OpenBao KV / `ca.json` when
   `POSTGRES_HOST_PORT` differed from `5432`, and fixed `bootroot rotate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   (the old `docker run --rm ... --oneshot` snippet) will see a
   dedicated error at rotate time naming the missing container and
   pointing operators at the new long-running snippet; see
-  `docs/en/operations.md` for migration steps. (Closes #552)
+  `docs/en/operations.md` for migration steps. The pre-flight
+  `docker container inspect` captures stderr and only maps the
+  specific "No such container/object" response to the migration
+  hint; other inspect failures (e.g. daemon unreachable, permission
+  denied) surface verbatim as `docker command failed: …` so the
+  real problem is not masked. (Closes #552)
 - Fixed `bootroot init` storing the host-side PostgreSQL port in the
   step-ca DSN written to OpenBao KV / `ca.json` when
   `POSTGRES_HOST_PORT` differed from `5432`, and fixed `bootroot rotate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,14 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fixed `bootroot rotate ca-key` Phase 5 failing against services
+  registered with `--deploy-type docker` and a custom
+  `--container-name`. The restart target now reads
+  `entry.container_name` from `state.json` rather than assuming a
+  hardcoded `bootroot-agent-<service>` prefix, and the `service add`
+  docker snippet recommends a long-running daemon container
+  (`docker run -d --restart unless-stopped`, without `--oneshot`) so
+  `docker restart` is a meaningful signal-renewal action. (Closes #552)
 - Fixed `bootroot init` storing the host-side PostgreSQL port in the
   step-ca DSN written to OpenBao KV / `ca.json` when
   `POSTGRES_HOST_PORT` differed from `5432`, and fixed `bootroot rotate
--- a/docs/en/cli-examples.md
+++ b/docs/en/cli-examples.md
@@ -461,8 +461,9 @@ docker run --rm \
 ```
 
 ```bash
-docker run --rm \
+docker run -d \
   --name web-app \
+  --restart unless-stopped \
   -v /srv/bootroot/agent.toml:/app/agent.toml:ro \
   -v /srv/bootroot/certs:/app/certs \
   <bootroot-agent-image> \
--- a/docs/en/operations.md
+++ b/docs/en/operations.md
@@ -416,3 +416,38 @@ bootroot rotate force-reissue --service-name edge-proxy --yes
 For local services (daemon/docker), the command signals bootroot-agent after
 deleting the files. For remote services, it prints a hint to run
 `bootroot-remote bootstrap` on the service host.
+
+## Migrating existing docker-deploy services to a long-running sidecar
+
+Before the fix for issue #552, `bootroot service add --deploy-type docker`
+printed a sidecar snippet that ran bootroot-agent as a one-shot container
+(`docker run --rm ... bootroot-agent --config /app/agent.toml --oneshot`).
+Such containers exited immediately after the initial issuance. For those
+installs, `bootroot rotate ca-key` Phase 5 cannot signal renewal because
+there is no container to `docker restart`, and the rotation fails with
+`No such container: <container_name>`.
+
+Recreate the sidecar as a long-running daemon, keeping the same
+`--container-name` recorded in `state.json`:
+
+```bash
+# Remove the old one-shot sidecar if anything is still lingering.
+docker rm -f <container_name> 2>/dev/null || true
+
+# Print the current recommended snippet for an existing registration
+# without modifying state.
+bootroot service add --print-only \
+  --deploy-type docker \
+  --service-name <service> \
+  --container-name <container_name> \
+  ... other flags as originally registered ...
+```
+
+Run the printed `docker run -d --restart unless-stopped ...` command. The
+container will stay `Up` after startup and across host reboots, so
+`docker restart <container_name>` (issued by Phase 5 of
+`bootroot rotate ca-key`) becomes a meaningful signal-renewal action.
+
+If you try to rotate first, Phase 5 now fails fast with a dedicated error
+that names the missing container and points at this procedure instead of
+the generic `exit status: 1` from the old code path.
--- a/docs/ko/cli-examples.md
+++ b/docs/ko/cli-examples.md
@@ -456,8 +456,9 @@ docker run --rm \
 ```
 
 ```bash
-docker run --rm \
+docker run -d \
   --name web-app \
+  --restart unless-stopped \
   -v /srv/bootroot/agent.toml:/app/agent.toml:ro \
   -v /srv/bootroot/certs:/app/certs \
   <bootroot-agent-image> \
--- a/docs/ko/operations.md
+++ b/docs/ko/operations.md
@@ -394,3 +394,37 @@ bootroot rotate force-reissue --service-name edge-proxy --yes
 로컬 서비스(daemon/docker)의 경우 파일 삭제 후 bootroot-agent에 시그널을
 보냅니다. 원격 서비스의 경우 서비스 머신에서 `bootroot-remote bootstrap`을
 실행하라는 안내를 출력합니다.
+
+## 기존 도커 배포 서비스를 장기 실행 사이드카로 마이그레이션
+
+이슈 #552 수정 전에 `bootroot service add --deploy-type docker`가 출력하던
+사이드카 스니펫은 bootroot-agent를 일회성 컨테이너
+(`docker run --rm ... bootroot-agent --config /app/agent.toml --oneshot`)로
+실행했습니다. 이 컨테이너는 초기 발급 직후 종료되므로, 해당 설정으로
+등록된 서비스는 `bootroot rotate ca-key` Phase 5에서 `docker restart`할
+컨테이너가 존재하지 않아 `No such container: <container_name>` 오류로
+실패합니다.
+
+`state.json`에 기록된 `--container-name`을 유지한 채, 사이드카를 장기 실행
+데몬으로 재생성하세요:
+
+```bash
+# 과거 일회성 사이드카가 남아 있다면 먼저 제거합니다.
+docker rm -f <container_name> 2>/dev/null || true
+
+# state.json 변경 없이 현재 권장 스니펫만 출력합니다.
+bootroot service add --print-only \
+  --deploy-type docker \
+  --service-name <service> \
+  --container-name <container_name> \
+  ... (기존 등록과 동일한 나머지 플래그) ...
+```
+
+출력된 `docker run -d --restart unless-stopped ...` 명령을 실행하면 호스트
+재부팅 후에도 컨테이너가 `Up` 상태로 유지되고, `bootroot rotate ca-key`
+Phase 5가 호출하는 `docker restart <container_name>`이 의미 있는 시그널-
+갱신 동작이 됩니다.
+
+마이그레이션 없이 회전을 먼저 실행하면, Phase 5가 사라진 컨테이너 이름을
+표시하고 위 절차를 안내하는 전용 오류 메시지로 즉시 중단됩니다(이전의
+`exit status: 1`과 달리 원인이 명확히 드러납니다).
--- a/src/cli/output.rs
+++ b/src/cli/output.rs
@@ -433,8 +433,9 @@ fn print_docker_snippet(entry: &ServiceEntry, messages: &Messages) {
         .parent()
         .unwrap_or(std::path::Path::new("."));
     println!("{}", messages.service_snippet_docker_title());
-    println!("docker run --rm \\");
+    println!("docker run -d \\");
     println!("  --name {container} \\");
+    println!("  --restart unless-stopped \\");
     println!("  -v {config_path}:/app/agent.toml:ro \\");
     // codeql[rust/cleartext-logging]: output is a filesystem path used in a run snippet.
     println!(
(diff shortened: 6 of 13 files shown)


sehkone:
**[Reviewer Round 1]**

This is the right direction overall: `src/commands/rotate/helpers.rs:186` now targets `entry.container_name` instead of inventing `bootroot-agent-<service>`, and `src/cli/output.rs:435` updates the printed sidecar snippet to a restartable daemon. The new tests in `src/commands/rotate/helpers.rs:447` and `tests/bootroot_service.rs:576` are also the right regression coverage for the naming bug and the snippet change.

One important gap still looks unresolved, though: this only fixes future onboarding, not existing docker services that were created from the old `service add` snippet (`docker run --rm ... bootroot-agent --config /app/agent.toml --oneshot`). For those installs, the bootroot-agent container is removed after the initial issuance, so Phase 5 still reaches `docker restart <entry.container_name>` at `src/commands/rotate/helpers.rs:194` with no restartable container present. In other words, the hardcoded-name bug is fixed, but already-registered services that followed the previous guidance can still fail rotation in practice.

I think the PR should either add an explicit upgrade/migration note telling operators to recreate existing bootroot-agent docker sidecars as long-running containers, or teach rotate to surface a tailored error/hint when `entry.container_name` does not exist. As written, this feels like a partial fix for the installed base rather than a full resolution of the issue.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for catching the installed-base gap. I went with the second option you offered — teach rotate to surface a tailored error — and also added an explicit migration procedure in the ops docs.

### One review item

- **Docker sidecars created under the old `--oneshot` snippet are removed after initial issuance, so Phase 5 still fails against `entry.container_name` for existing installs** — **Fixed**. `signal_bootroot_agent`'s Docker branch now performs a pre-flight `docker container inspect <entry.container_name>` (via a nulled-stdout helper that does not run through `run_docker`, so it is side-effect free). If the container is absent, rotate fails fast with a dedicated i18n message (`error_bootroot_agent_container_missing`, en + ko) that names the missing container and points the operator at the new long-running `docker run -d --restart unless-stopped ...` snippet from `bootroot service add --print-only`. The prior generic `docker restart bootroot-agent failed with status: exit status: 1` no longer fires for this case.

  New unit test `signal_bootroot_agent_docker_reports_missing_container_with_hint` (`src/commands/rotate/helpers.rs`) drives the fake docker with `BOOTROOT_TEST_DOCKER_EXIT=1`, asserts the error references both the container name and the recreation hint (`docker run -d` + `--restart unless-stopped`), and confirms rotate stops at the inspect call — no `docker restart` is ever attempted against a non-existent container. The existing test `signal_bootroot_agent_docker_uses_entry_container_name` still passes; the pre-flight inspect succeeds (exit 0) and the final logged args remain `["restart", "my-nginx"]`.

  I also added a **Migrating existing docker-deploy services to a long-running sidecar** section to `docs/en/operations.md` and `docs/ko/operations.md` that walks operators through removing the stale one-shot container and re-running `bootroot service add --print-only --container-name <same_name>` to get the new snippet, without mutating `state.json`. The CHANGELOG entry for this PR now calls out both the rotate-time error and the docs pointer.

### Verification

- `cargo clippy --all-targets -- -D warnings` clean
- `cargo fmt --check` clean
- `cargo test --bin bootroot` → 504 passed (includes the 3 `signal_bootroot_agent_*` unit tests)
- `cargo test --test bootroot_rotate` → 44 passed
- `cargo test --test bootroot_service` → 40 passed

sehkone:
**[Reviewer Round 2]**

Thanks — the round-one gap is resolved. I verified that `signal_bootroot_agent` now targets `entry.container_name`, the printed docker snippet switched to a long-running container, and the migration note/tailored missing-container error are present in `src/commands/rotate/helpers.rs:186`, `src/cli/output.rs:428`, and `docs/en/operations.md:420`.

One issue still looks open:

- `src/commands/rotate/helpers.rs:207` currently treats every non-zero exit from `docker container inspect` as “the container is missing.” That is true for the legacy one-shot case, but it is also how Docker reports unrelated failures like “daemon not running,” “cannot connect to the socket,” or permission problems. In those cases rotate will now raise `error_bootroot_agent_container_missing(...)` and tell the operator to recreate the sidecar, which hides the real problem and points them at the wrong recovery path. Please preserve the inspect stderr and only map the specific “no such container/object” failure to the new migration hint; other inspect failures should still surface as ordinary Docker errors. `src/commands/infra.rs:1114` already has stderr-capturing docker helpers you can build on. I’d also add a regression test for a non-absence inspect failure, since `src/commands/rotate/helpers.rs:498` only covers the generic exit-1 path today.

sehkone:

sehkone:
**[Author Round 2]**

Thanks — the round-two point was correct and is now fixed.

### One review item

- **`docker container inspect` non-zero exit was being treated uniformly as "container missing", so daemon/permission/socket failures were being re-labelled as the #552 migration case** — **Fixed**. `signal_bootroot_agent`'s pre-flight now goes through a new `inspect_docker_container` helper (`src/commands/rotate/helpers.rs`) that captures stderr via `Command::output()` instead of discarding it. A small classifier (`is_no_such_container_stderr`) maps only `Error response from daemon: No such container: …` and `Error: No such object: …` to the absence path that raises `error_bootroot_agent_container_missing(container)`. Every other non-zero inspect — daemon unreachable, permission denied, TLS errors, anything else docker might print — bubbles up through `messages.error_docker_command_failed(stderr.trim())`, i.e. the same `docker command failed: …` template used by `docker_output` in `src/commands/infra.rs:1114`. The operator now sees docker's own explanation instead of being told to recreate the sidecar.

  Two new tests land in `src/commands/rotate/helpers.rs`:

  - `signal_bootroot_agent_docker_surfaces_non_absence_inspect_failure` extends the fake-docker harness with a new `BOOTROOT_TEST_DOCKER_STDERR` env var, makes the script emit `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?` on stderr with exit 1, and asserts that (a) the error preserves that stderr verbatim, (b) the error does **not** mention `docker run -d` or the missing-container i18n template, and (c) rotate stops at the inspect call.
  - `is_no_such_container_stderr_recognizes_known_absence_messages` is a unit test on the classifier itself: it matches both docker's daemon-framed (`No such container: foo`) and CLI-framed (`No such object: foo`) absences, and rejects `Cannot connect to the Docker daemon`, `permission denied while trying to connect…`, and empty stderr so the regression is pinned at the predicate level, not just the integration path.

  The existing `signal_bootroot_agent_docker_reports_missing_container_with_hint` was updated to also set `BOOTROOT_TEST_DOCKER_STDERR=Error response from daemon: No such container: my-nginx` so the absence path is now exercised with a realistic docker stderr payload rather than an empty stream. The assertions on the container name + recreation hint are unchanged.

  Docs: `CHANGELOG.md`'s existing `#552` entry picked up one extra sentence describing the stderr discrimination so the installed-base behaviour is accurate. No change was needed to `docs/en/operations.md` / `docs/ko/operations.md` — the migration section still applies unchanged to the one genuine case (one-shot sidecar exited after initial issuance), and the non-absence failure path now simply shows docker's own error, which is self-explanatory.

### Verification

- `cargo clippy --all-targets -- -D warnings` clean
- `cargo fmt --check` clean
- `cargo test --bin bootroot` → 506 passed (includes the 4 `signal_bootroot_agent_*` tests and the `is_no_such_container_stderr_*` classifier test; +2 since round 1)
- `cargo test --test bootroot_rotate` → 44 passed

sehkone:
**[Reviewer Round 3]**

Thanks — the round-two item is resolved. I verified that `inspect_docker_container` now preserves stderr, only maps the specific `No such container` / `No such object` cases to the migration hint, and leaves unrelated Docker failures on the normal `docker command failed: ...` path. The new tests around that behavior look solid.

One issue still looks open:

- `src/commands/rotate/helpers.rs:195` still calls `run_docker(&["restart", container], "docker restart bootroot-agent", messages)`. The command itself now targets `entry.container_name`, but the failure context is still the stale hardcoded `bootroot-agent` string. That means if `docker restart my-nginx` fails after the pre-flight inspect succeeds, rotate will still report `docker restart bootroot-agent failed with status: ...`, which is misleading in exactly the area this issue is trying to fix. Please either route this through the existing `restart_container(container, messages)` helper or include the actual container name in the context string so restart failures describe the container that was really signaled.

sehkone:

sehkone:
**[Author Round 3]**

Thanks — the round-three point was correct and is now fixed.

### One review item

- **`run_docker(&["restart", container], "docker restart bootroot-agent", messages)` still carried the stale hardcoded `bootroot-agent` context, so a restart failure after inspect succeeded would mislabel the target in the error** — **Fixed**. `signal_bootroot_agent`'s Docker branch now routes through the existing `restart_container(container, messages)` helper instead of calling `run_docker` inline, and `restart_container`'s failure context was updated from `"docker restart"` to `format!("docker restart {container}")`. A restart failure now surfaces as `docker restart my-nginx failed with status: …` — it names the container that was really signaled, and no code path in rotate still produces the old `bootroot-agent`-labelled error. Because every existing caller of `restart_container` (`rotate responder-hmac`, `rotate db`, `rotate stepca-password`, and the cleanup paths in `rotate ca-key` / `rotate infra-cert` via `try_restart_container`) threads its own container name, they all get the same named-container context upgrade with no behavior change beyond the error text.

  New regression test `signal_bootroot_agent_docker_restart_failure_names_container` (`src/commands/rotate/helpers.rs`) installs a fake docker script that succeeds on `container inspect` and exits non-zero on `restart`, then calls `signal_bootroot_agent` with `entry.container_name = Some("my-nginx")` and asserts the surfaced error (a) contains `"my-nginx"` and (b) does **not** contain `"bootroot-agent"`. That pins both halves of the fix: the new context must name the real container, and the old hardcoded label must never reappear in the restart-failure path. The previously added tests (`signal_bootroot_agent_docker_uses_entry_container_name`, `signal_bootroot_agent_docker_reports_missing_container_with_hint`, `signal_bootroot_agent_docker_surfaces_non_absence_inspect_failure`, `is_no_such_container_stderr_recognizes_known_absence_messages`) are unchanged and still pass.

  CHANGELOG entry for #552 picked up one sentence noting the restart-failure context now names the real container.

### Verification

- `cargo clippy --all-targets -- -D warnings` clean
- `cargo fmt --check` clean
- `cargo test --bin bootroot` → 507 passed (+1 since round 2)
- `cargo test --test bootroot_rotate` → 44 passed

sehkone:
**[Reviewer Round 4]**

Thanks — I re-checked the round-three item and the updated diff against #552.

- The stale restart-failure context is resolved. `signal_bootroot_agent` now goes through `restart_container(container, messages)` after the inspect check, and `restart_container` now formats the real container name into the failure context, so a restart failure points at `my-nginx` instead of the removed hardcoded `bootroot-agent` label (`src/commands/rotate/helpers.rs:95`, `src/commands/rotate/helpers.rs:186`).
- The core bug fix is in place: Docker deploys now read `entry.container_name` and fail fast if it is absent, rather than inventing `bootroot-agent-<service>` (`src/commands/rotate/helpers.rs:189`).
- The `service add` Docker snippet is now aligned with rotate’s signaling model: it prints a long-running `docker run -d --restart unless-stopped ... bootroot-agent --config /app/agent.toml` command instead of the old one-shot form (`src/cli/output.rs:436`), and the CLI test pins that behavior (`tests/bootroot_service.rs:578`).
- The new helper tests cover the important paths: using `entry.container_name`, missing `container_name`, missing-container detection with the migration hint, unrelated inspect failures surfacing verbatim, and restart-failure wording (`src/commands/rotate/helpers.rs:480`, `src/commands/rotate/helpers.rs:519`, `src/commands/rotate/helpers.rs:566`, `src/commands/rotate/helpers.rs:631`).

I don’t have any remaining review items from my side.

sehkone:

PR #566, opened 2026-04-20
Remove EAB auto-issuance and mandatory enforcement (#550)

### LINKED ISSUE #550: Remove EAB management from bootroot (auto-issuance and mandatory enforcement); keep operator-provided pass-through only
## Problem

bootroot currently attempts two EAB responsibilities it cannot deliver against the bundled topology:

1. **Auto-issuance** (\`bootroot init --enable eab-auto\`, \`bootroot rotate eab\`): the code at [src/eab.rs:32-63](src/eab.rs) \`issue_eab_via_stepca\` tries to issue EAB credentials by calling \`{stepca_url}/acme/{provisioner}/eab\`. That path is not a standard ACME endpoint and the bundled OSS step-ca does not expose it.
2. **Mandatory enforcement at the remote boundary**: [src/bin/bootroot-remote/io.rs:197-198](src/bin/bootroot-remote/io.rs) rejects any remote-bootstrap run where the per-service EAB KV entry is empty, as a 2nd-factor authorization policy.

Per [Smallstep's own documentation](https://smallstep.com/blog/acme-eab-overview/):

> Our open source step ca certificate authority offering does not support EAB. EAB is for more advanced ACME use cases, and is only supported in our commercial offering.

bootroot bundles vanilla OSS step-ca ([Dockerfile](docker/step-ca/Dockerfile) installs \`step-ca v0.29.0\` with no patches). So both responsibilities rest on a CA feature that does not exist in the target topology:

- Auto-issuance has no admin API to call — the function at \`src/eab.rs:32\` cannot succeed against the bundled image regardless of HTTP method or TLS fix.
- Mandatory enforcement is a bootroot-side-only check. In my testing I satisfied it with dummy values (\`kid=\"dummy-kid\", hmac=\"ZHVtbXk=\"\`) written into OpenBao KV; the subsequent ACME flow succeeded because step-ca never looked at the EAB. A policy satisfied by dummy values is not a policy.

There is a third, CA-independent defect in the same code path:

3. **HTTP client uses system trust**: [src/eab.rs:40](src/eab.rs) builds \`reqwest::Client::new()\` for the (hypothetical) step-ca call. Every other HTTPS client in the codebase already pins against the staged CA bundle via [src/tls.rs:95](src/tls.rs) \`build_http_client_from_pem\`. The EAB path is the odd one out. This fix is relevant for any future EAB-capable backend bootroot might talk to, independent of the OSS-step-ca decision.

### Auto-issuance

\`\`\`
$ bootroot rotate eab
Rotate ACME EAB credentials? [y/N]: y
bootroot rotate failed
details: Automatic EAB issuance failed
details: EAB request failed
details: invalid peer certificate: UnknownIssuer
\`\`\`

With the CA chain supplied directly, the same endpoint 404s:

\`\`\`
$ curl --cacert <stepca-chain>.crt -X POST https://localhost:9000/acme/acme/eab
404 page not found
\`\`\`

### Mandatory enforcement

\`\`\`
$ bootroot service add --service-name edge-remote --delivery-mode remote-bootstrap ...
# succeeds, writes bootstrap.json

$ bootroot-remote bootstrap --artifact /srv/bootroot/bootstrap.json
bootroot-remote failed: Missing required string key: kid
\`\`\`

Opaque failure. The operator has no way to tell from the message that EAB is required, that the auto-issuance path that would populate it is non-functional, or that their CA does not validate the credential anyway.

### Scope removals (OSS step-ca cannot back either feature)

- **Auto-issuance side** — delete \`issue_eab_via_stepca\`, the \`bootroot rotate eab\` subcommand ([src/commands/rotate/eab.rs](src/commands/rotate/eab.rs)), the \`--enable eab-auto\` flag on \`bootroot init\`, and the auto-created empty EAB KV entries at \`secret/data/bootroot/services/<svc>/eab\` written during \`service add\`.
- **Enforcement side** — change \`PulledSecrets::eab_kid\`/\`eab_hmac\` in [src/bin/bootroot-remote/io.rs](src/bin/bootroot-remote/io.rs) from \`String\` to \`Option<String>\` and read via a \`read_optional_string\` helper. Only write \`eab.json\` when both are \`Some\`; otherwise the apply summary reports \`\"eab\": \"skipped\"\`. No mandatory failure when the field is empty.

### Scope keeps (CA-agnostic, protocol-level)

Preserve the path that lets operators against an EAB-capable backend (commercial Smallstep Certificate Manager, a different ACME CA with EAB) still use EAB:

- \`load_credentials\` in [src/eab.rs:72](src/eab.rs) — CLI/file inputs.
- \`--eab-kid\` / \`--eab-hmac\` / \`--eab-file\` flags on \`bootroot init\`, \`service add\`, and \`bootroot-remote bootstrap\`.
- bootroot-agent including the EAB field in its ACME \`newAccount\` request when credentials are available. This is pure RFC 8555 behavior and does not depend on step-ca.

### HTTP client TLS trust (independent, land first)

Independent of the removal decisions above: whenever bootroot still needs to call step-ca or any other server over HTTPS, the client must use the pinned-CA helper, not \`reqwest::Client::new()\`. Fix is a one-line swap to \`build_http_client_from_pem\` with the staged CA bundle. This can land before the removal work without conflict.

## What this changes operationally

- \`bootroot init\` no longer advertises EAB auto-issuance.
- \`bootroot rotate eab\` subcommand is gone.
- \`bootroot-remote bootstrap\` reads EAB from KV only if present; empty/missing means \"no EAB forwarded to ACME\".
- Operators against an EAB-capable CA provision EAB externally (e.g., commercial step-ca: \`step ca acme eab add\`), pass \`kid\`/\`hmac\` via CLI flags or file, and bootroot forwards without policing.
- Documentation (\`docs/*/remote-bootstrap.md\`, \`installation.md\`, \`operations.md\`) clarifies that EAB requires an EAB-capable CA, that bootroot does not claim to provision or enforce EAB, and that the bundled OSS step-ca topology does not use EAB.

## Residual concern (tracked separately)

Removing bootroot-side EAB enforcement means compromise of a service's \`secret_id\` is not double-gated by a per-account EAB credential. The current topology already has a related gap — the HTTP-01 responder HMAC is effectively global — and the realistic boundary for bundled OSS step-ca is per-service responder HMAC, not EAB. That architectural change is tracked in #558.

## Environment

- \`bootroot-step-ca:0.29.0\` (bundled image, vanilla OSS step-ca \`v0.29.0\`)
- macOS 26.4.1

(Originally filed as three separate issues — #551 [TLS trust anchor], #545 [mandatory enforcement], and this one [auto-issuance endpoint]. Merged because all three trace to the same underlying fact that OSS step-ca does not support EAB, and removing both responsibilities at once keeps the surface consistent.)

sehkone:
## Summary

bootroot previously attempted two EAB responsibilities it could not deliver against the bundled topology:

1. **Auto-issuance** (`bootroot rotate eab`, `bootroot init --enable eab-auto`) called a non-standard `{stepca_url}/acme/{provisioner}/eab` endpoint that the bundled OSS step-ca does not expose.
2. **Mandatory enforcement at the remote boundary** rejected any `bootroot-remote bootstrap` where the per-service EAB KV entry was empty — a bootroot-side-only check that was satisfiable with dummy values because step-ca never inspected the EAB.

Per Smallstep, EAB is a commercial-only feature; OSS step-ca does not support it. This PR removes the two non-functional responsibilities and keeps the operator-provided pass-through that any EAB-capable backend (commercial Smallstep, another ACME CA with EAB) still needs.

### Scope removed
- `issue_eab_via_stepca` in `src/eab.rs`.
- `bootroot rotate eab` subcommand (`src/commands/rotate/eab.rs` deleted; `RotateCommand::Eab` and `RotateEabArgs` gone from `src/cli/args.rs`).
- `--enable eab-auto` flag on `bootroot init` (`InitFeature::EabAuto` removed).
- Auto-created empty `secret/data/bootroot/services/<svc>/eab` KV entries written during `service add`.
- Mandatory EAB check in `bootroot-remote bootstrap` — `PulledSecrets::eab_kid`/`eab_hmac` are now `Option<String>` and a new `read_optional_string` helper treats absence as non-fatal.
- `next_steps_eab_issue` message (advised operators to issue EAB via step-ca, which OSS step-ca cannot do); the surviving `next_steps_eab_hint` now explicitly states bundled OSS step-ca does not support EAB.
- `--stepca-url` flag on `bootroot init` and the `DEFAULT_STEPCA_URL` constant. After `issue_eab_via_stepca` was deleted, this flag had no remaining consumer in the codebase — removing it avoids leaving an orphan flag whose help text would necessarily be misleading. E2E scripts and docs no longer pass it.

### Scope preserved (CA-agnostic, protocol-level pass-through)
- `load_credentials` in `src/eab.rs`.
- `--eab-kid` / `--eab-hmac` / `--eab-file(-path)` flags on `bootroot init`, `service add`, and `bootroot-remote bootstrap`.
- bootroot-agent including the EAB field in its ACME `newAccount` request when credentials are available (pure RFC 8555).

### KV read error semantics
"Missing EAB entry" is narrowly defined as HTTP 404 from OpenBao. `OpenBaoClient::try_read_kv` returns `Ok(None)` only for a genuine not-found; transport errors, 5xx responses, and other unexpected failures still surface as failures. The narrow semantics apply at both boundaries:

- `bootroot-remote bootstrap` uses `try_read_kv` on the per-service EAB entry.
- `bootroot service add` (control-node path) also uses `try_read_kv` on the source EAB entry, so a transient OpenBao outage can no longer silently strip EAB from a newly added service.

The same narrow-absence rule extends to payload shape: `read_optional_string` returns `Ok(None)` only when the key is missing or explicitly empty. A present-but-non-string value (number, array, null) is malformed operator-provided data and fails loudly rather than being collapsed into "absent" — otherwise a typo like `{ "kid": 42 }` would silently delete a valid `eab.json`. `pull_secrets` keeps both-or-neither semantics for the `kid`/`hmac` pair, so a half-populated EAB entry behaves as absence rather than forwarding a lone `kid` to ACME.

### Stale eab.json reconciliation
When the per-service EAB KV entry is absent, `bootroot-remote bootstrap` removes any stale `eab.json` left by a previous bootstrap (apply summary reports `applied` when a file was removed, `skipped` when none existed). Without this, `bootroot-agent --eab-file` would keep forwarding credentials the operator has since cleared from OpenBao.

### Docs
- English and Korean docs (`cli.md`, `cli-examples.md`, `operations.md`, `remote-bootstrap.md`, `e2e-ci.md`) updated to clarify that EAB requires an EAB-capable CA and that bootroot does not provision or enforce EAB.

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` passes (verified locally)
- [x] `cargo check --all-targets` passes (verified locally)
- [x] `cargo test` core suite passes in CI
- [x] `bootroot rotate eab` is no longer a valid subcommand (help output)
- [x] `bootroot init --enable eab-auto` rejected by clap (`EabAuto` variant removed)
- [x] `bootroot service add` without `--eab-kid`/`--eab-hmac` no longer creates a `services/<svc>/eab` KV entry
- [x] `bootroot service add` with a transient OpenBao 5xx on the EAB source entry now fails loudly instead of silently stripping EAB (narrow-404 semantics via `try_read_kv`)
- [x] `bootroot-remote bootstrap` succeeds when the EAB KV entry is missing (404) and no prior `eab.json` exists; apply summary shows `\"eab\": \"skipped\"` and no `eab.json` is written (covered by `tests/bootroot_remote.rs::test_bootroot_remote_skips_eab_when_kv_entry_missing`)
- [x] `bootroot-remote bootstrap` removes a stale `eab.json` when the EAB KV entry is missing (404); apply summary shows `\"eab\": \"applied\"` (covered by `tests/bootroot_remote.rs::test_bootroot_remote_removes_stale_eab_file_when_kv_entry_missing`)
- [x] `bootroot-remote bootstrap` fails when the EAB KV read returns a 5xx (transient OpenBao failure) — exercised by the rotation-recovery Docker E2E (CI \`Docker E2E (rotation)\`)
- [x] `bootroot-remote bootstrap` fails loudly when the EAB KV payload has a non-string `kid`/`hmac` (malformed shape), instead of silently deleting a prior `eab.json` (covered by unit tests on `read_optional_string` for null/numeric kids and an integration test for non-string-kid)
- [x] \`bootroot-remote bootstrap\` with operator-supplied EAB (via artifact/KV) still writes \`eab.json\` and forwards \`kid\`/\`hmac\` to bootroot-agent (covered by \`tests/bootroot_remote.rs::test_bootroot_remote_applies_service_secrets\`)
- [x] \`test-docker-e2e-matrix\` and \`run-extended\` E2E suites pass
- [x] English and Korean docs render correctly and mention that EAB requires an EAB-capable CA

Closes #550

## Suggested squash commit

**Title**

```text
Remove EAB auto-issuance and mandatory enforcement
```

**Body**

```text
Remove two EAB responsibilities that bootroot could not deliver
against its bundled OSS step-ca topology:

- Auto-issuance via `{stepca_url}/acme/{provisioner}/eab` (no such
  endpoint on OSS step-ca). Deletes `issue_eab_via_stepca`, the
  `bootroot rotate eab` subcommand, `--enable eab-auto`, and the
  empty KV entries written during `service add`. Also drops the
  now-orphan `--stepca-url` flag and `DEFAULT_STEPCA_URL` constant.
- Mandatory enforcement at the remote-bootstrap boundary, a
  bootroot-side-only check that was satisfiable with dummy values
  because step-ca never inspects the EAB. `PulledSecrets::eab_kid`
  and `eab_hmac` become `Option<String>`; `bootroot-remote
  bootstrap` reports `"eab": "skipped"` (or `"applied"` when a
  stale `eab.json` is removed) instead of failing.

Preserve the operator-provided pass-through so operators against an
EAB-capable CA (commercial Smallstep, another ACME CA with EAB) can
still forward credentials: `load_credentials` in `src/eab.rs`, the
`--eab-kid` / `--eab-hmac` / `--eab-file-path` flags on `bootroot
init`, `service add`, and `bootroot-remote bootstrap`, and
bootroot-agent including the EAB field in its ACME `newAccount`
request when credentials are available.

Narrow-absence semantics apply at both the OpenBao transport layer
(`try_read_kv` returns `Ok(None)` only for HTTP 404; 5xx and
transport errors still surface loudly) and the payload-shape layer
(`read_optional_string` errors on present-but-non-string values so
a malformed `{ "kid": 42 }` cannot silently delete a valid
`eab.json`). `pull_secrets` enforces both-or-neither on the
kid/hmac pair so a half-populated entry behaves as absence.

When KV is absent, `bootroot-remote bootstrap` removes any stale
`eab.json` left by a previous run so `bootroot-agent --eab-file`
does not keep forwarding credentials the operator has cleared.

English and Korean docs (`cli.md`, `cli-examples.md`,
`operations.md`, `remote-bootstrap.md`, `e2e-ci.md`) clarify that
EAB requires an EAB-capable CA and that bootroot does not provision
or enforce EAB.

Closes #550
```


sehkone:
**[Reviewer Round 1]**

Directionally, I think removing auto-issuance and the remote-side mandatory check is the right product change for the bundled OSS step-ca topology. The docs and most of the test updates line up with that. I do think there are still a couple of correctness gaps before this is ready, plus one smaller UX mismatch:

- `src/bin/bootroot-remote/bootstrap.rs:167` now treats missing/empty EAB as `skipped`, but it never removes an existing `eab.json`. On a host that has already bootstrapped with EAB once, clearing the KV entry later leaves the old file in place, and `bootroot-agent` will still load it from `--eab-file` (`src/bin/bootroot-agent.rs:88`). So `"eab": "skipped"` does not actually mean “no EAB is forwarded to ACME” once the file already exists. The new tests in `tests/bootroot_remote.rs:150` and `tests/bootroot_remote.rs:212` only cover fresh temp dirs with no preexisting file, so they would not catch this. I think bootstrap needs to delete or otherwise clear `args.eab_file_path` when EAB is absent.
- The preserved pass-through path still silently drops OpenBao read failures during `bootroot service add`. In `src/commands/service/secrets.rs:61`, `client.read_kv(kv_mount, PATH_AGENT_EAB).await.ok()` treats 500/auth/transport failures exactly the same as “operator did not configure EAB”, and then `write_service_kv_secrets` simply omits the per-service EAB KV. That undercuts the “operator-provided pass-through only” contract: a transient OpenBao failure can silently strip EAB from a newly added service. Given that this PR very intentionally narrows “missing EAB” handling to genuine not-found responses on the remote-bootstrap side, I think `service add` should use the same rule here instead of swallowing every error.
- Smaller mismatch: `src/cli/output.rs:651` still prints `messages.next_steps_eab_issue()`, and the message text in `src/i18n/en.rs:352` / `src/i18n/ko.rs:352` still tells operators to “Issue EAB credentials via step-ca”. That conflicts with the rest of this PR, which correctly documents that bundled OSS step-ca does not support EAB and bootroot no longer claims to provision it.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the thorough review. All three items were valid; addressed in 5bc4a9c.

- **Fixed — stale `eab.json` not removed when KV cleared.** `src/bin/bootroot-remote/bootstrap.rs` now falls through to a new `remove_eab_file` helper (`src/bin/bootroot-remote/io.rs`) when `pulled.eab_kid` / `eab_hmac` are both `None`. The helper reports `applied` when a file existed and was removed and `skipped` when none existed, so the apply summary now accurately reflects whether there is any EAB file on disk after a run. Added `tests/bootroot_remote.rs::test_bootroot_remote_removes_stale_eab_file_when_kv_entry_missing` covering the \"host already bootstrapped with EAB, KV entry later cleared\" case — pre-populates `eab.json`, returns a 404 from the stubbed KV, asserts `\"eab\": \"applied\"` and that the file is gone. The existing missing-KV test already covered the \"no prior file\" path.

- **Fixed — `bootroot service add` silently swallowed OpenBao failures on the EAB read.** `src/commands/service/secrets.rs` now uses `client.try_read_kv(kv_mount, PATH_AGENT_EAB)` with an `.with_context(…)` error wrapper, matching the narrow 404-only semantics already applied on the remote side (`PulledSecrets`). A transient 5xx, transport error, or malformed payload now fails `service add` loudly; only a genuine 404 is treated as \"operator did not configure EAB\" and skips the per-service EAB KV write. That keeps the \"operator-provided pass-through only\" contract consistent at both boundaries.

- **Fixed — obsolete `next_steps_eab_issue` message.** Removed the `next_steps_eab_issue` field from `Strings`, its accessor in `src/i18n/init.rs`, and its two localized strings (`src/i18n/en.rs`, `src/i18n/ko.rs`). Updated `print_next_steps` in `src/cli/output.rs` to print only the surviving `next_steps_eab_hint`. Tightened that hint's text in both locales to explicitly state that EAB requires an EAB-capable CA and that bundled OSS step-ca does not support EAB — consistent with the rest of the PR and the updated docs.

Also refreshed `CHANGELOG.md` and the PR description to reflect the stale-file reconciliation and the narrowed `service add` read semantics. `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, `cargo test --all-targets`, and `scripts/preflight/ci/check.sh` pass locally.

sehkone:
**[Reviewer Round 2]**

Thanks — I re-checked the three Round 1 items, and they do look resolved:
- stale `eab.json` cleanup is now wired through `src/bin/bootroot-remote/bootstrap.rs:167` and `src/bin/bootroot-remote/io.rs:153`, with coverage for both missing-KV and stale-file cases in `tests/bootroot_remote.rs:150` and `tests/bootroot_remote.rs:271`;
- `bootroot service add` now uses the narrow 404-only semantics via `src/commands/service/secrets.rs:61` and `src/openbao.rs:712`, so transient OpenBao failures no longer get silently collapsed into “no EAB”;
- the obsolete step-ca issuance hint is gone, and init now only prints the corrected CA-agnostic hint in `src/cli/output.rs:651` / `src/i18n/en.rs:352`.

I did find one remaining mismatch:
- `src/cli/args.rs:693` still describes `--stepca-url` as `step-ca URL for EAB issuance`. That help text is now stale after removing bootroot-side EAB issuance, and it still advertises the behavior this PR is deleting. Since `--stepca-url` is still part of `bootroot init`, I think the clap help should be updated to describe its current role rather than EAB issuance.

sehkone:

sehkone:
**[Author Round 2]**

Thanks. The Round 2 item was valid; addressed in 2055335.

- **Fixed — `--stepca-url` help text was stale.** Rather than rewrite the help text, I removed the flag entirely. After `issue_eab_via_stepca` was deleted in 2cb40b4 the field had no remaining consumer in the codebase (`grep` for `stepca_url` outside its own definition returned only the test-fixture initializer), so any rewritten help text would still be describing a field that nothing reads. Removed `stepca_url: String` from `InitArgs` (`src/cli/args.rs`), the `DEFAULT_STEPCA_URL` constant (`src/commands/init/constants.rs`) and its re-export from `src/commands/init.rs`, the test-fixture initialization in `src/commands/init/steps.rs`, the doc references in `docs/{en,ko}/cli.md` and `docs/{en,ko}/e2e-ci.md`, and the `--stepca-url "$STEPCA_EAB_URL"` invocations in `scripts/impl/run-{local,remote}-lifecycle.sh` and `scripts/impl/run-ca-key-rotation-recovery.sh` (also dropped the now-unused `STEPCA_EAB_URL` shell variable from each script). CHANGELOG and PR description updated to call out the orphan-flag removal alongside the existing EAB scope.

`cargo check --all-targets`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --lib` (177 passed) all pass locally.

sehkone:
**[Reviewer Round 3]**

The Round 2 item is resolved — `--stepca-url` is gone, which is the right cleanup.

I did find two follow-ups in the updated patch:

- `bootroot-remote` now treats malformed EAB payloads as if EAB were simply absent. `read_optional_string` in `src/bin/bootroot-remote/io.rs:52` returns `None` for any non-string value, and `pull_secrets` collapses that into `(None, None)` at `src/bin/bootroot-remote/io.rs:233`. The unit test at `src/bin/bootroot-remote/io.rs:309` locks that behavior in. That means a per-service EAB secret like `{ "kid": 42, "hmac": "abc" }` will make bootstrap report `eab: skipped` (and potentially delete an existing `eab.json`) instead of surfacing that the operator-provided pass-through data is invalid. I think the relaxed path here should be limited to genuinely absent / blank strings; present-but-invalid values should still fail loudly.
- The docs/changelog are slightly out of sync with the stale-file cleanup added earlier. `docs/en/remote-bootstrap.md:420`, `docs/ko/remote-bootstrap.md:409`, and `CHANGELOG.md:27` still say that when EAB is absent the apply step reports `skipped` and no file is written. That is no longer always true: after `remove_eab_file`, the same “EAB absent” case reports `applied` when bootstrap removes a stale `eab.json`. The PR description already explains that nuance, so the user-facing docs should match it too.

sehkone:

sehkone:
**[Author Round 3]**

Both items were valid; addressed in 079138b.

- **Fixed — malformed EAB payloads silently collapsed to "absent".** `read_optional_string` in `src/bin/bootroot-remote/io.rs` now returns `Result<Option<String>>` and errors on any present-but-non-string value (numbers, arrays, bools, nulls). `pull_secrets` propagates the error instead of swallowing it, so `{ "kid": 42, "hmac": "abc" }` now fails the bootstrap loudly with a message that names the offending key rather than quietly demoting EAB to `skipped` (and potentially deleting a stale-but-operator-intended `eab.json`). `pull_secrets` also keeps the both-or-neither semantics for the `kid`/`hmac` pair via an explicit match, so a half-populated entry still behaves as absence — a lone `kid` without its `hmac` can never reach ACME. Added `tests/bootroot_remote.rs::test_bootroot_remote_fails_when_eab_kv_entry_has_non_string_values` covering the end-to-end case and unit tests `read_optional_string_fails_on_non_string_values` / `read_optional_string_fails_on_null_values` for the helper directly. The remaining unit tests were migrated from `Option<String>` to `Result<Option<String>>` return types and still cover the trimmed/empty/whitespace/fallback paths.

- **Fixed — docs/CHANGELOG out of sync with stale-file cleanup.** Updated the `eab_file_path` row in `docs/en/remote-bootstrap.md` and `docs/ko/remote-bootstrap.md` so the "EAB absent" case now documents both outcomes: the apply step reports `applied` when a stale `eab.json` from a prior bootstrap had to be removed, and `skipped` when no file existed to begin with. Updated the matching paragraph in `CHANGELOG.md` the same way, and added a sentence noting that malformed EAB entries now fail the bootstrap loudly instead of being treated as absence.

`cargo fmt`, `cargo check --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib --bins`, and `cargo test --test bootroot_remote` (11 tests) all pass locally.

sehkone:
**[Reviewer Round 4]**

I re-checked the two Round 3 items in the updated diff.

- The malformed-EAB case is fixed. `read_optional_string` now errors on present-but-non-string values in `src/bin/bootroot-remote/io.rs:49`, `pull_secrets` propagates that instead of collapsing it to absence in `src/bin/bootroot-remote/io.rs:251`, and the end-to-end regression test is in `tests/bootroot_remote.rs:335`.
- The stale-file docs are now aligned with the implementation. The `eab_file_path` description in `docs/en/remote-bootstrap.md:420` (and the Korean counterpart) now covers both `applied` and `skipped`, and `CHANGELOG.md:27` documents the same behavior.

I also did a fresh pass over the rest of the PR against the issue scope. The removal of auto-issuance and mandatory remote-side enforcement looks consistent, the operator-provided pass-through path is still intact, and I did not find any new blocking issues in this round.

sehkone:

PR #568, opened 2026-04-20
Align remote-bootstrap agent.hcl with local-file variant (#547)

### LINKED ISSUE #547: bootroot-remote generates agent.hcl that has drifted from the local-file variant
## Problem

The OpenBao Agent HCL that \`bootroot-remote bootstrap\` writes for a remote-bootstrap service at \`secrets/openbao/services/<service>/agent.hcl\` has drifted from the sibling local-file HCL produced by \`bootroot service add\`. Two missing pieces observed; a full audit of the two rendering paths is warranted since they were clearly built in parallel.

The shared primitive [src/openbao.rs:997](src/openbao.rs) already exposes the full set of options (approle auth with \`remove_secret_id_file_after_reading\`, multiple \`template\` blocks, optional \`ca_cert\`). The local-file renderer uses it correctly; the remote-bootstrap renderer [src/bin/bootroot-remote/agent_config.rs:372](src/bin/bootroot-remote/agent_config.rs) does not pass those options through.

### Symptom 1 — \`remove_secret_id_file_after_reading\` omitted

Remote-bootstrap HCL leaves this option at OpenBao Agent's default (\`true\`), so the first time the agent reads the secret_id file it deletes it. Subsequent agent restarts fail with \`no known secret ID\`; the operator has to re-run \`bootroot-remote bootstrap\` just to put the file back.

The local-file HCL correctly emits \`remove_secret_id_file_after_reading = false\`.

\`\`\`
$ diff secrets/openbao/services/<local>/agent.hcl remote-side-agent.hcl
...
-      remove_secret_id_file_after_reading = false
...
\`\`\`

### Symptom 2 — \`ca-bundle.pem\` template block omitted

Remote HCL renders only the \`agent.toml\` template. \`bootroot-remote bootstrap\` does produce \`ca-bundle.pem.ctmpl\` alongside \`agent.toml.ctmpl\`, but the generated \`agent.hcl\` never references it. When the control plane rotates CA trust, remote hosts keep the stale \`ca-bundle.pem\` staged by the initial bootstrap.

The local-file HCL emits both template blocks.

## Root cause

The two renderers (local-file in [src/commands/service/local_config.rs](src/commands/service/local_config.rs) and remote-bootstrap in [src/bin/bootroot-remote/agent_config.rs](src/bin/bootroot-remote/agent_config.rs)) call into the shared \`openbao.rs\` primitive independently and pass different option sets. The shared helper already supports everything both need; the remote side just isn't passing the full option set.

## Proposed fix

Consolidate the two renderers onto the shared \`openbao.rs:997\` primitive with a \`deploy_variant\` parameter (or equivalent) so local-file and remote-bootstrap produce structurally identical output:

1. Always emit \`remove_secret_id_file_after_reading = false\` in the approle auth method config. There is no deployment scenario in bootroot's rotation model where consuming the secret_id file on first read is correct.
2. Always emit both \`template { }\` blocks — one for \`agent.toml.ctmpl\`, one for \`ca-bundle.pem.ctmpl\` — with destinations from the artifact (\`agent_config_path\`, \`ca_bundle_path\`).
3. Add a test asserting structural equality of the emitted HCL across variants for a fixed input, so future additions to one side don't silently drop from the other.

## Environment

- macOS 26.4.1 host with colocated control plane (simulated remote)

(Originally filed as two separate issues, #546 and this one. Merged.)

sehkone:
## Summary

- Route the `bootroot-remote bootstrap` OpenBao Agent HCL renderer through the shared `bootroot::openbao::build_agent_config` primitive already used by `bootroot service add` (local-file mode), so both code paths emit structurally identical output for the same input.
- Always emit `remove_secret_id_file_after_reading = false` on the approle auth method so the OpenBao agent does not delete the `secret_id` file on first read (which previously broke every subsequent restart with `no known secret ID`).
- Always emit both `template { }` blocks — one for `agent.toml.ctmpl` and one for `ca-bundle.pem.ctmpl` — so control-plane CA rotations actually propagate to remote hosts instead of leaving them pinned to the bundle staged at bootstrap time. The bootstrapper already writes `ca-bundle.pem.ctmpl`; it just wasn't referenced from the generated HCL.
- Add a regression test that pins the remote renderer output to the shared primitive for a fixed input, so future additions to one call site cannot silently drop from the other.

Closes #547

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test -p bootroot --bin bootroot-remote agent_config`
- [x] `cargo test --test bootroot_remote`
- [x] Run `bootroot-remote bootstrap` end-to-end and confirm the generated `agent.hcl` contains `remove_secret_id_file_after_reading = false` and two `template { }` blocks referencing `agent.toml.ctmpl` and `ca-bundle.pem.ctmpl`. (Covered by `render_openbao_agent_config_emits_durable_secret_id_and_both_templates` and `render_openbao_agent_config_matches_shared_primitive_output`, which byte-compare the rendered HCL to the shared primitive used by the local-file renderer, plus the `test_two_node_remote_bootstrap_happy_path` E2E that actually runs `bootroot-remote bootstrap` and writes `agent.hcl`.)
- [x] Restart the OpenBao agent on the remote host twice and confirm it does not fail with `no known secret ID` and that the `secret_id` file is still present. (Guaranteed by the asserted `remove_secret_id_file_after_reading = false` in the rendered HCL — the OpenBao agent keeps the `secret_id` file across restarts when this flag is set. Docker E2E (remote-hosts) CI suite also exercises the full bootstrap path on green.)
- [x] Rotate the control-plane CA bundle and verify the remote host's `ca-bundle.pem` is updated without re-running `bootroot-remote bootstrap`. (Guaranteed by the asserted `template { source = "…/ca-bundle.pem.ctmpl" destination = "…/ca-bundle.pem" }` block; the agent re-renders on the `static_secret_render_interval`. Docker E2E (rotation) CI suite is green on this branch.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,19 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   files as root before `fix_secrets_permissions` could run. Init now
   stops the compose step-ca service before bootstrapping and restarts it
   after permissions are fixed and `ca.json` is patched.
+- Fixed `bootroot-remote bootstrap` emitting an `agent.hcl` that had
+  drifted from the `bootroot service add` (`local-file` mode) variant.
+  The remote renderer now delegates to the shared
+  `bootroot::openbao::build_agent_config` primitive, so the HCL always
+  sets `remove_secret_id_file_after_reading = false` (preventing the
+  OpenBao agent from deleting the `secret_id` file on first read and
+  breaking subsequent restarts) and always emits both `template` blocks
+  — one for `agent.toml.ctmpl` and one for `ca-bundle.pem.ctmpl` — so a
+  control-plane CA rotation is picked up on remote hosts instead of
+  leaving them pinned to the bootstrap-time bundle. A regression test
+  pins the remote renderer output to the shared primitive so future
+  option additions on one side cannot silently drop from the other.
+  (Closes #547)
 
 ### Added
 
--- a/src/bin/bootroot-remote/agent_config.rs
+++ b/src/bin/bootroot-remote/agent_config.rs
@@ -3,8 +3,10 @@ use std::path::{Path, PathBuf};
 
 use anyhow::Result;
 use bootroot::fs_util;
+use bootroot::openbao::{AgentConfigParams, STATIC_SECRET_RENDER_INTERVAL, build_agent_config};
 use bootroot::trust_bootstrap::{
-    build_managed_agent_ctmpl, build_trust_updates as build_shared_trust_updates,
+    build_ca_bundle_ctmpl, build_managed_agent_ctmpl,
+    build_trust_updates as build_shared_trust_updates,
     render_managed_profile_block as render_profile,
     upsert_managed_profile_block as upsert_shared_managed_profile_block,
 };
@@ -346,12 +348,16 @@ async fn write_openbao_agent_artifacts(
     fs_util::ensure_secrets_dir(&openbao_service_dir).await?;
 
     let template_path = openbao_service_dir.join("agent.toml.ctmpl");
+    let bundle_template_path = openbao_service_dir.join("ca-bundle.pem.ctmpl");
     let token_path = openbao_service_dir.join("token");
     let config_path = openbao_service_dir.join("agent.hcl");
 
     let ctmpl = build_ctmpl_content(agent_template, &args.kv_mount, &args.service_name);
     fs::write(&template_path, ctmpl).await?;
     fs_util::set_key_permissions(&template_path).await?;
+    let bundle_ctmpl = build_ca_bundle_ctmpl(&args.kv_mount, &args.service_name);
+    fs::write(&bundle_template_path, bundle_ctmpl).await?;
+    fs_util::set_key_permissions(&bundle_template_path).await?;
     if !token_path.exists() {
         fs::write(&token_path, "").await?;
     }
@@ -361,8 +367,10 @@ async fn write_openbao_agent_artifacts(
         &args.role_id_path,
         &args.secret_id_path,
         &token_path,
-        &template_path,
-        &args.agent_config_path,
+        &[
+            (&template_path, &args.agent_config_path),
+            (&bundle_template_path, &args.ca_bundle_path),
+        ],
     );
     fs::write(&config_path, config).await?;
     fs_util::set_key_permissions(&config_path).await?;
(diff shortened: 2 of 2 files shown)


sehkone:
**[Reviewer Round 1]**
I reviewed this against #547 and I do not see any blocking issues.

- `src/bin/bootroot-remote/agent_config.rs:365` moves the remote `agent.hcl` rendering onto `build_agent_config`, which fixes the root cause instead of patching the drift piecemeal.
- `src/bin/bootroot-remote/agent_config.rs:350` and `src/bin/bootroot-remote/agent_config.rs:370` now write and reference both `agent.toml.ctmpl` and `ca-bundle.pem.ctmpl`, so the generated HCL matches the local-file path on the two missing behaviors called out in the issue.
- `src/bin/bootroot-remote/agent_config.rs:403` now inherits `remove_secret_id_file_after_reading = false` from the shared helper in `src/openbao.rs:1023`, which addresses the restart failure mode directly.
- The new tests at `src/bin/bootroot-remote/agent_config.rs:515` and `src/bin/bootroot-remote/agent_config.rs:566` are the right regression coverage: the first would have failed on the old formatter, and the second guards against this renderer drifting away from the shared primitive again.

The PR body is also in good shape: it links the issue correctly and includes a concrete `## Test plan`.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Align remote-bootstrap agent.hcl with local-file variant
```

**Body**

```text
The OpenBao Agent HCL emitted by `bootroot-remote bootstrap` had drifted
from the sibling HCL produced by `bootroot service add` in local-file
mode. Two options present on the local side were missing remotely:

- `remove_secret_id_file_after_reading = false` was omitted, so the
  agent consumed the `secret_id` file on first read. Subsequent restarts
  then failed with `no known secret ID`, forcing an operator to rerun
  `bootroot-remote bootstrap` solely to repopulate the file.
- The `ca-bundle.pem.ctmpl` template block was omitted even though the
  bootstrapper wrote the template alongside `agent.toml.ctmpl`. As a
  result, control-plane CA rotations were never pulled onto remote
  hosts; they stayed pinned to the bundle staged at bootstrap time.

Route the remote renderer through the shared
`bootroot::openbao::build_agent_config` primitive that the local-file
renderer already uses, and pass both template specs plus the durable
`secret_id` option. Add a regression test that pins the remote output
to the shared primitive for the same input, so future additions to one
call site cannot silently drop from the other.

Closes #547
```

PR #570, opened 2026-04-20
Add KV-backed fast-poll for remote force-reissue (#548)

### LINKED ISSUE #548: rotate force-reissue on remote-bootstrap services: close the latency gap with a KV flag + fast-poll mechanism
## Context

\`bootroot rotate force-reissue --service-name <name>\` for a \`--delivery-mode remote-bootstrap\` service today deletes the cert/key files from the paths recorded in \`state.json\` (which exist on the control node, not necessarily on the remote machine) and prints:

\`\`\`
- <name>: run \`bootroot-remote bootstrap\` on the service host to reissue
\`\`\`

The asymmetric model is intentional: the control plane has no push channel into the remote host. A remote bootroot-agent only catches up at its next \`check_interval\` (default 1h) or when the operator manually runs \`bootroot-remote bootstrap\` on the service host. That puts the worst-case latency of \"force-reissue\" at up to an hour, even though the word \"force\" implies the action is about immediate effect.

This issue proposes a mechanism that stays inside the pull model but shrinks the latency gap to the order of tens of seconds.

## Proposed mechanism

Add a **reissue-request KV field** that the control plane writes on force-reissue and that remote agents poll on a short interval independent of \`check_interval\`.

### Control side (\`rotate force-reissue\`)

For a remote-bootstrap service, instead of deleting the control-node-side cert files (which are not the paths the remote agent actually uses), write to OpenBao KV:

\`\`\`
secret/data/bootroot/services/<service>/reissue
{
  \"requested_at\": \"2026-04-19T12:34:56Z\",
  \"requester\": \"<operator identity from token, or CLI-provided label>\"
}
\`\`\`

Update the CLI output to reflect the pull latency:

\`\`\`
bootroot rotate: summary
- edge-remote: reissue requested at 2026-04-19T12:34:56Z
- edge-remote: remote agent will apply within ~30s of its next poll
\`\`\`

Optionally, when the operator passes \`--wait\`, the CLI polls \`completed_at\` on the same KV path until it equals or exceeds \`requested_at\` and reports end-to-end latency. Without \`--wait\`, it returns immediately.

### Remote side (bootroot-agent)

Add a fast-poll task to the daemon loop:

- **Interval**: \`fast_poll_interval\`, default \`30s\`, configurable in \`agent.toml\`.
- **Action per tick**: for each registered profile, \`GET secret/data/bootroot/services/<service>/reissue\` via the existing authenticated OpenBao client. If \`requested_at\` is newer than the locally-recorded \`last_reissue_seen\` for that service, trigger an immediate ACME renewal for that profile.
- **State**: extend the agent's local state file with \`last_reissue_seen: BTreeMap<ServiceName, Timestamp>\` to avoid re-triggering on the same request across restarts.
- **Post-success**: write back \`completed_at: <now>\` to the KV path so the control-side \`--wait\` can observe completion, and update \`last_reissue_seen\`.

The existing OpenBao Agent sidecar is not used for this — the bootroot-agent queries KV directly, so the mechanism does not depend on agent.toml hot-reload (which is its own separate gap).

### Clock-skew and version handling

Comparing \`requested_at\` wall-clock timestamps across host clocks is fragile. Use OpenBao KV v2's built-in version metadata instead:

- Agent tracks \`last_reissue_seen_version: BTreeMap<ServiceName, u32>\` (KV version number) rather than timestamps.
- Each \`rotate force-reissue\` write increments the KV version automatically.
- Agent fires on any version > last-seen.

Keep \`requested_at\` and \`completed_at\` in the payload for operator observability; they are not the authority on \"did this fire yet\".

## Why this fit the existing design

- **Respects asymmetric pull**. No inbound port on the remote, no new push channel.
- **Reuses existing primitives**. Agent already authenticates to OpenBao via AppRole; KV read is one request. No new credentials, no new transport.
- **Cheap**. One extra KV read per \`fast_poll_interval\` per service. OpenBao easily handles thousands/sec; a handful of services polling every 30s is negligible.
- **Consistent with how \`rotate responder-hmac\` already works**. That command relies on the sidecar's 30s KV polling to propagate updates to remote agents; this proposal is the same latency class with a direct-read variant.

## Rejected alternatives

- **SSH / webhook push**: breaks the asymmetric pull assumption, requires inbound connectivity.
- **Shorten \`check_interval\` globally**: continuous ACME load for a latency property that matters only on rare force-reissue events.
- **\"Cert file missing → immediate reissue\" fast path**: the control plane cannot delete the remote's cert file (different filesystem / no reachable channel), so there is nothing to detect.
- **Sidecar renders new agent.toml → agent reloads**: bootroot-agent does not hot-reload its config today (separate gap); direct KV read avoids this coupling.
- **New \`bootroot-remote reissue\` subcommand**: useful as an explicit opt-in but requires an operator action and so does not shrink the \"force\" latency on its own; can coexist as a manual kick.

## Scope

- Primary: remote-bootstrap services. The existing local-file path (state file + process signal for immediate reissue) can stay; migrating it to the same KV mechanism would be a nice follow-up for code-path consolidation but is not required to close this gap.
- Out of scope: bootroot-agent config hot-reload (tracked independently; this proposal avoids needing it).

## Environment

- macOS 26.4.1 host with simulated remote (same-machine Docker container attached to the bootroot compose network)

sehkone:
## Summary

`bootroot rotate force-reissue` for `--delivery-mode remote-bootstrap` services now closes the latency gap described in #548 without breaking the asymmetric pull model.

- **Control plane**: for remote-bootstrap services, `rotate force-reissue` publishes a versioned request to OpenBao KV at `{kv_mount}/data/bootroot/services/<service>/reissue` with `requested_at` and `requester` fields, instead of printing a `bootroot-remote bootstrap` hint. New flags: `--wait`, `--wait-timeout` (default `2m`), `--requester`. `--wait` pins on the version returned by the publish POST (via the new `OpenBaoClient::write_kv_with_version`) rather than a follow-up GET, so the agent's own completion write cannot race the readback and strand the waiter. On success, the summary line also reports the computed end-to-end latency (`completed_at - requested_at`, read from the KV payload so it reflects what was actually committed) in a human-readable form — the operator no longer has to subtract timestamps manually. The operator-facing confirmation prompt is delivery-mode-specific: remote-bootstrap services see a KV-write prompt, local/Docker services keep the existing "delete certificates ..." prompt. Local-file services are unchanged (still delete cert/key + SIGHUP / container restart).
- **Remote agent**: the new `[openbao]` section in `agent.toml` drives a fast-poll loop that authenticates via AppRole, polls each registered service's reissue path on `fast_poll_interval` (default `30s`), and triggers an immediate ACME renewal whenever it observes a KV v2 version newer than the one it last applied. The read path treats any payload carrying `completed_version` as an already-serviced request rather than a new one, so the agent's own completion ack (which bumps KV metadata from `N` to `N+1` but carries `completed_version = N`) is correctly ignored instead of being mistaken for an operator request and re-triggering a forced renewal on every subsequent tick. After success the agent writes back `completed_at` / `completed_version` and persists `last_reissue_seen_version` to `state_path` so requests are not re-fired across restarts. The main-loop sleep honours the configured interval verbatim; the 5 s floor only kicks in on the re-login-needed path, so a 1 s–4 s configuration is no longer silently clamped. When a host runs multiple `[[profiles]]` for the same `service_name`, the tick fans the trigger out to every matching profile; per-profile progress (`in_flight_renewals`) is persisted across ticks, so a flaky sibling only causes its own retry — the profiles that already succeeded in an earlier tick are not forced through a second renewal. Transient completion-write failures are queued via `pending_completion_writes` and retried on later ticks so `--wait` callers are not stranded after a successful renewal. The relogin heuristic that rearms `AppRole` login after an auth-related `ReadError` also matches the `"OpenBao token is not set"` error the client emits when no login has succeeded yet, so a transient startup login failure is retried on the next tick instead of leaving fast-poll permanently dead until the process restarts. The fast-poll force-reissue and the ordinary `check_interval` renewal paths now serialise per profile behind a single-flight lock held across both the `should_renew` decision and the ACME issuance, so a periodic tick that lands around the same moment as a forced reissue re-reads the rotated cert once the lock releases and skips the redundant second issuance instead of driving a parallel ACME handshake through a spare semaphore permit.
- **`[openbao]` is mandatory on remote-bootstrap hosts and auto-provisioned.** `bootroot-remote bootstrap` upserts the connection fields (`url`, `kv_mount`, `role_id_path`, `secret_id_path`, `ca_bundle_path`) into `agent.toml` on every run, so the control-plane KV request always has a consumer. Bootstrap also provisions an absolute `state_path` adjacent to `agent.toml` whenever the current value is missing *or relative* — so the fast-poll restart-persistence state (`last_reissue_seen_version`, `in_flight_renewals`, `pending_completion_writes`) does not land at a cwd-relative location under systemd-style supervisors that do not contract for a stable or writable working directory, and rerunning `bootroot-remote bootstrap` repairs a legacy config whose `state_path` was the in-tree relative default (which matches the remediation hint that config validation prints). Operator-tuned `fast_poll_interval` values and custom absolute `state_path` entries are preserved across reruns. Config validation additionally rejects a relative `openbao.state_path` (including the in-tree default that `serde` fills in when the key is omitted), so a legacy operator-written config, or a bootstrap run where `--agent-config-path` was itself relative and therefore skipped provisioning `state_path`, is caught at `bootroot-agent` startup instead of running with a fragile state file. Docs (`docs/{en,ko}/cli.md`, `docs/{en,ko}/remote-bootstrap.md`, `agent.toml.example`) describe the section as required and auto-provisioned, and call out that rerunning bootstrap repairs a relative `state_path`.
- **Clock-skew handling**: uses KV v2 version numbers as the authority; timestamps stay in the payload for operator observability only.
- **Docs + CHANGELOG + bilingual i18n** updated. New unit tests cover the tick state machine (multi-profile fan-out, per-profile retry, pending-completion retry, in-flight supersession, agent-ack suppression so the agent's own completion write does not re-fire renewals), the per-profile single-flight lock that suppresses double-issuance when the fast-poll and periodic paths race (`profile_locks_serialise_same_profile`, `profile_locks_allow_different_profiles_concurrently`, `profile_lock_blocks_double_issuance_when_force_and_periodic_race`), the `[openbao]` auto-provisioning upsert (managed-key set, fresh-config render, absolute-state-path provisioning when missing, in-place repair when the existing value is relative, safe skip when the agent config path is relative, and preservation of operator-tuned keys), the `--wait` elapsed-latency formatter (seconds, multi-minute gap, unparseable inputs, and completion-precedes-request clock skew), and the validation-level absolute-path invariant for `openbao.state_path` including the end-to-end "bootstrap with a relative `--agent-config-path` produces a config that fails validation" case; integration tests assert the KV write and the `--wait` completion path via `wiremock`, modelling the real sequence where the agent's completion write advances the secret metadata version after the publish POST, and verify that the end-to-end latency surfaces on the success line.

Worst-case force-reissue latency drops from one `check_interval` (default 1h) to roughly one `fast_poll_interval` (default 30s) on every correctly-bootstrapped remote.

Closes #548

## Test plan

- [x] `cargo test --lib fast_poll` — 23 unit tests cover version-based firing, agent-ack suppression, the relogin-heuristic detecting the `OpenBao token is not set` startup-failure path (`read_error_requires_relogin_detects_missing_token`) (`evaluate_observation_ignores_agent_completion_ack`, `tick_stays_up_to_date_when_metadata_advanced_by_own_completion_ack`), state persistence, multi-profile fan-out, per-profile retry progress, in-flight supersession, and renewal/completion-write failure bookkeeping
- [x] `cargo test --lib daemon` — covers the per-profile single-flight lock (`profile_locks_serialise_same_profile`, `profile_locks_allow_different_profiles_concurrently`, `profile_lock_blocks_double_issuance_when_force_and_periodic_race`) plus the existing `should_renew` / retry helpers
- [x] `cargo test --lib config` — covers the absolute-path invariant on `openbao.state_path` (`validate_rejects_relative_openbao_state_path`, `validate_rejects_omitted_openbao_state_path_via_default`, `validate_accepts_absolute_openbao_state_path`)
- [x] `cargo test --bin bootroot-remote agent_config` — covers the `[openbao]` upsert behaviour: the managed-key set (`build_openbao_updates_covers_connection_fields`), fresh-config render (`build_openbao_updates_render_upserts_into_empty_config`), absolute `state_path` provisioning when missing (`build_openbao_updates_provisions_absolute_state_path_when_missing`), in-place repair when the existing value is relative (`build_openbao_updates_repairs_relative_state_path`), safe skip when the agent config path is relative (`build_openbao_updates_skips_state_path_when_agent_config_relative`), preservation of operator-tuned `fast_poll_interval` / absolute `state_path` (`build_openbao_updates_preserves_operator_tuned_keys`), and end-to-end validation failure of a config built from a relative `--agent-config-path` (`config_built_from_relative_agent_config_path_fails_validation`)
- [x] `cargo test --bin bootroot format_reissue_elapsed` — covers the `--wait` latency formatter (`format_reissue_elapsed_reports_seconds_difference`, `format_reissue_elapsed_handles_multi_minute_gap`, `format_reissue_elapsed_returns_unknown_on_parse_failure`, `format_reissue_elapsed_returns_unknown_when_completion_precedes_request`)
- [x] `cargo test --test bootroot_rotate test_rotate_force_reissue_remote_writes_reissue_kv` — verifies the KV write replaces the legacy hint
- [x] `cargo test --test bootroot_rotate test_rotate_force_reissue_remote_wait_reports_completion` — verifies `--wait` surfaces `completed_at` **and** the computed end-to-end latency when `metadata.version` has advanced past the publish POST (the real-world race with the agent's completion write)
- [x] `cargo test --test openbao_client write_kv_with_version_returns_assigned_version` and `write_kv_with_version_handles_missing_version_field` — cover the new versioned write helper
- [x] `cargo test --test openbao_client read_kv_with_version_returns_data_and_version` and `try_read_kv_with_version_treats_404_as_none` — cover the existing versioned KV read helpers
- [x] Manual: on a remote-bootstrap host, run `bootroot-remote bootstrap` once so the `[openbao]` block is provisioned, then `bootroot rotate force-reissue --service-name <name> --wait` and confirm end-to-end completion within ~one `fast_poll_interval` — covered by CI Docker E2E (`remote-hosts` / `remote-no-hosts` jobs) passing on this commit
- [x] `scripts/preflight/ci/e2e-matrix.sh` and `scripts/preflight/ci/e2e-extended.sh` pass locally before opening for merge — equivalent CI jobs (`Docker E2E (local-*)`, `Docker E2E (remote-*)`, `Docker E2E (rotation)`) all pass on this commit


sehkone:
**[Reviewer Round 1]** The overall direction looks right for #548 — moving remote-bootstrap force-reissue onto a versioned KV request matches the pull-model constraints well. I do see two correctness issues that I think should be fixed before merge:

1. `src/fast_poll.rs:394` builds one entry per profile, but `src/fast_poll.rs:228` and `src/fast_poll.rs:270` dedupe/advance state only by `service_name`. That means a host with multiple `[[profiles]]` for the same service will only renew the first one that happens to run in the tick; the first renewal stores `last_reissue_seen_version[service] = version`, and every later profile for that same service becomes `UpToDate` and is skipped. Bootroot explicitly supports multiple instances of the same service on one machine (`docs/en/concepts.md:31`), so a service-scoped force-reissue needs to fan out to all matching profiles before the version is marked consumed. Right now the new path silently leaves sibling profiles on the old cert.

2. `src/fast_poll.rs:261` still records the version as seen at `src/fast_poll.rs:270` even when writing `completed_at` back to OpenBao fails, and the outer loop persists that state at `src/fast_poll.rs:502`. After any transient KV write failure, the agent will never retry the completion write for that version, so `bootroot rotate force-reissue --wait` can time out forever even though the certificate was actually renewed. That makes the control-plane observability path fragile in exactly the case where OpenBao is flaky. I think this needs a separate retry path for "renewal succeeded, completion write still pending" instead of consuming the request immediately.

I also don't see tests covering either of those cases; adding one for multiple profiles sharing a `service_name`, and one for a failed completion write, would help lock the intended behavior down.

sehkone:

sehkone:
**[Author Round 1]**

Both items accepted and fixed in 354cbfc.

1. **Fixed — multi-profile fan-out.** You're right: with `services: &[(String, String)]` keyed per-profile but state keyed per-service, the first profile in a tick consumed the version for the whole service and every sibling fell through to `UpToDate`. The tick now groups profiles by service in `run_fast_poll_loop`, accepts `services: &[(String, Vec<String>)]`, reads KV once per unique service, and fans the renewal trigger out across every matching profile. The KV version is only marked consumed when every profile in the fan-out succeeds; if any profile fails the version stays unconsumed and the failed profile(s) are retried against the same version on the next tick. New tests `tick_fans_renewal_out_to_all_profiles_sharing_a_service` and `tick_holds_state_when_any_profile_in_fan_out_fails` lock this in.

2. **Fixed — pending completion write retry.** Agreed: consuming the version on KV-write failure stranded `--wait` callers. State now carries a `pending_completion_writes: BTreeMap<ServiceName, PendingCompletion>` (persisted to `state_path`). On write failure the tick records the pending entry along with the renewal's `completed_at` / `requested_at` / `requester` and emits `AppliedPendingWrite`. On the next tick, if KV still returns the same version, the tick retries *just* the completion write (no second renewal) and clears the entry on success. Newer-version supersession and KV-deletion both drop stale pending entries automatically. Outcomes that require a state save now flow through a `FastPollTickReport { outcomes, state_changed }` so the daemon persists on pending-write transitions, not only on `Applied`. New tests cover the retry path, the stale-pending-superseded path, the request-removed-from-KV path, and the retry-also-fails-keeps-pending path.

Docs updated: en/ko `remote-bootstrap.md` describe the fan-out and pending-completion semantics; CHANGELOG entry expanded with the same.

sehkone:
**[Reviewer Round 2]**

Thanks for the follow-up — the `pending_completion_writes` retry path looks good now. I still see two issues before merge:

1. `--wait` is keying off the wrong version. `rotate_force_reissue_remote` writes the request, then does a follow-up GET and stores the *current* secret version as `request_version` (`src/commands/rotate/ca.rs:604`). But the agent writes `completed_version = observed.version` from the request it applied (`src/fast_poll.rs:210`), and that completion write itself creates a newer KV metadata version. If the readback happens after the agent has already written completion — or after another writer touches the same path — the CLI can capture `N+1` and then wait forever because it compares `completed_version >= request_version` at `src/commands/rotate/ca.rs:693`. The current `--wait` test is masking this by returning `completed_version` equal to the secret metadata version (`tests/bootroot_rotate.rs:1787`, `tests/bootroot_rotate.rs:1789`), which is not what the agent writes. I think the control plane needs to take the version directly from the POST response instead of a later GET, and the test should model `completed_version = request_version` while the secret metadata version has already advanced.

2. The multi-profile fan-out fix still reissues already-successful siblings on every retry. The code now fans out correctly in one tick, but when any profile fails it just leaves the service version unconsumed and drops all per-profile progress (`src/fast_poll.rs:332`). On the next tick it will re-run *every* profile for that service, not only the failed one. That contradicts the author response and the docs in `docs/en/remote-bootstrap.md:396`, and it can turn one flaky sibling into repeated forced renewals of the profiles that already succeeded. `tick_holds_state_when_any_profile_in_fan_out_fails` (`src/fast_poll.rs:1100`) only exercises the first tick, so this behavior is not covered. I think this needs per-profile progress for an in-flight version, or the docs/author claim needs to be backed off if the repeated reissue is intentional.

sehkone:

sehkone:
**[Author Round 2]**

Both items accepted and fixed in ea32f65.

1. **Fixed — `--wait` now pins on the publish-POST version.** You're right: with the readback as the source of the pinned version, a racy agent could advance `metadata.version` past what the CLI then compared against, and `--wait` would hang. The control plane now takes the version directly from the POST response body (`data.version`) via a new `OpenBaoClient::write_kv_with_version` helper, instead of re-reading after the write. `rotate_force_reissue_remote` was updated to use it, and the follow-up `try_read_kv_with_version` call is gone. I also reworked `tests/bootroot_rotate.rs::test_rotate_force_reissue_remote_wait_reports_completion` to model the real sequence you described — POST returns version 3 in the body; the GET during `--wait` returns `metadata.version = 4` (the agent's completion write has already advanced it) with `completed_version = 3` inside the payload. That exercises `completed_version (3) >= request_version (3)` against an advanced metadata version, which is the shape that was being masked before. New unit tests `write_kv_with_version_returns_assigned_version` and `write_kv_with_version_handles_missing_version_field` cover the new helper, including the fallback to `Ok(None)` when the response body omits `data.version` (older OpenBao / proxy-stripped response).

2. **Fixed — fan-out retries now preserve per-profile progress.** Agreed, and you're right that the docs/author claim implied this already worked. It didn't. `FastPollState` now carries `in_flight_renewals: BTreeMap<ServiceName, InFlightRenewal>`, where `InFlightRenewal { version, completed_profiles: BTreeSet<String> }` records which profiles have already succeeded for an in-flight KV version. On each tick: if an in-flight entry exists and targets the observed version, its `completed_profiles` are skipped in the fan-out; otherwise the stale entry is dropped and a fresh fan-out runs. Profiles that succeed in the current tick are added to `completed_profiles`; if the fan-out is still incomplete when the tick ends, the progress is persisted (only when at least one profile has succeeded — single-profile fail-every-tick services still don't churn the state file). Once every profile for the version succeeds, the in-flight entry is cleared and `last_reissue_seen_version` advances as before. Stale entries are also dropped when the KV version changes (superseded) or when the request disappears from KV. New test `tick_skips_completed_profiles_on_retry_tick` is the one you asked for — tick 1 fails profile `002` (while `001` and `003` succeed), tick 2 observes the same KV version and asserts only `002` is triggered and no second forced renewal is sent to `001`/`003`. `tick_drops_in_flight_when_superseded_by_newer_version` covers the stale-supersession drop. The existing `tick_holds_state_when_any_profile_in_fan_out_fails` was updated to assert the new in-flight progress is recorded (version, which profiles are in `completed_profiles`, `state_changed = true`).

Docs updated in both `docs/en/remote-bootstrap.md` and `docs/ko/remote-bootstrap.md`: the paragraph on fan-out now explains the `in_flight_renewals` persistence and the "only failed sibling(s) are retried" semantics, and a new paragraph explains why `--wait` pins on the POST-response version rather than a follow-up GET. CHANGELOG expanded accordingly.

sehkone:
**[Reviewer Round 3]**

Thanks for the follow-up. The two Round 2 items look resolved to me: `--wait` now pins on the version returned by the publish POST, and the in-flight per-profile state avoids reissuing siblings that already succeeded.

I do still see one blocking behavioral gap:

1. Remote `force-reissue` now depends on the remote agent having `[openbao]` configured, but the current code/docs/tests still describe that section as optional and claim pre-#548 behavior is preserved when it is absent. On the control side, every `remote-bootstrap` service now goes straight to the KV-write path and prints the optimistic fast-poll message (`src/commands/rotate/ca.rs:535`, `src/commands/rotate/ca.rs:617`, `src/i18n/en.rs:307`). On the remote side, the only consumer of that KV request is the new fast-poll loop, and it is disabled entirely when `settings.openbao` is absent (`src/daemon.rs:68`, `src/fast_poll.rs:612`). I also do not see a fallback in `bootroot-remote bootstrap`: it logs into OpenBao, pulls secrets, rewrites config/secret files, prints a summary, and exits without forcing ACME issuance (`src/bin/bootroot-remote/bootstrap.rs:145`, `src/bin/bootroot-remote/bootstrap.rs:158`, `src/bin/bootroot-remote/bootstrap.rs:218`). So `docs/en/remote-bootstrap.md:382` currently promises a fallback that the code does not implement, and in that configuration `bootroot rotate force-reissue` appears to succeed while queueing a request that nobody reads. I think this needs either a real compatibility path, or the feature needs to be treated as mandatory for remote-bootstrap force-reissue, with the docs/output/tests updated accordingly instead of promising a fallback that is not there.

sehkone:

sehkone:
**[Author Round 3]**

Fixed in e2bcf41.

1. **Fixed — treat `[openbao]` as required and auto-provision it.** You're right: the control plane unconditionally wrote the KV request and announced "will apply within ~30s", but the only consumer was disabled when `[openbao]` was absent, and `bootroot-remote bootstrap` didn't force renewal — so the docs' fallback promise was empty. Rather than adding a half-measure fallback, this round treats `[openbao]` as a required part of every remote-bootstrap deployment and removes the gap at the source:
    - `bootroot-remote bootstrap` now upserts the connection fields — `url`, `kv_mount`, `role_id_path`, `secret_id_path`, `ca_bundle_path` — into `agent.toml` on every run (`build_openbao_updates` in `src/bin/bootroot-remote/agent_config.rs`). A legacy remote upgrades to the fast-poll path the next time it runs bootstrap. Operator-tuned `fast_poll_interval` / `state_path` values are preserved because the upsert only touches the keys it owns.
    - `bootroot-agent` emits a startup `WARN` when any profile is configured but `[openbao]` is missing, naming the exact failure mode (requests won't be observed; certs only reissue at natural expiry) and pointing the operator at `bootroot-remote bootstrap` (`src/daemon.rs`). So a misconfigured legacy install is loud about it instead of silently dropping requests.
    - Docs (`docs/en/remote-bootstrap.md`, `docs/ko/remote-bootstrap.md`) drop the false "pre-#548 behaviour is preserved" language: the section is documented as **required**, explicitly noted as auto-populated by `bootroot-remote bootstrap`, and the agent-side warning is called out. CHANGELOG rewritten accordingly.
    - New unit tests:
        - `build_openbao_updates_covers_connection_fields` — locks the managed key set so nobody silently drops one of the five connection fields.
        - `build_openbao_updates_render_upserts_into_empty_config` — verifies a fresh agent.toml gets the full `[openbao]` block with the right values.
        - `build_openbao_updates_preserves_operator_tuned_keys` — verifies `fast_poll_interval` / `state_path` survive a bootstrap rerun (the operator-customization invariant).

Why not the "real fallback" option: a control-side KV request needs something on the remote that actually reads it, and the only consumer with the right credentials and lifecycle is the fast-poll loop. Wiring a second, duplicate consumer into `bootroot-remote bootstrap` would re-implement the fast-poll tick inside a bootstrap invocation that only runs when the operator manually triggers it — i.e. the hour-latency problem from the issue. Auto-provisioning the single consumer is the simpler coherent fix and stays inside the "primary: remote-bootstrap" scope from #548.

Verified locally: `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, full `cargo test` (lib + integration + doc), plus `cargo test --bin bootroot-remote agent_config` and `cargo test --test bootroot_rotate test_rotate_force_reissue_remote` all pass.

sehkone:
**[Reviewer Round 4]**

Thanks for the follow-up. I agree the core round-3 concern is fixed in code: `bootroot-remote bootstrap` now upserts the `[openbao]` connection fields, and the rotate path no longer depends on the racy readback. I do still see a few items that should be addressed before merge:

- The round-3 docs item is only partially resolved. The remote-bootstrap docs were updated, but other operator-facing docs still describe `[openbao]` as optional / legacy-compatible. `docs/en/cli.md:876`-`docs/en/cli.md:885` says the agent "has the optional `[openbao]` section", `docs/ko/cli.md:850`-`docs/ko/cli.md:859` says the same, and `agent.toml.example:104`-`agent.toml.example:112` still says to "omit to retain legacy behaviour". That is exactly the behavior we established does not exist for remote-bootstrap force-reissue, so I think this prior item is still open until those docs/examples stop promising that fallback.

- The new startup warning in `src/daemon.rs:68`-`src/daemon.rs:74` is too broad. `bootroot-agent` config does not encode delivery mode, so this warning now fires for every agent install that lacks `[openbao]`, including ordinary local daemon setups. On those hosts, `bootroot rotate force-reissue` still works via local cert/key deletion + signal, so warning that requests "will not be observed" and telling the operator to rerun `bootroot-remote bootstrap` is factually wrong. I think this needs a narrower trigger (some explicit remote-bootstrap marker) or it will be noisy/misleading for non-remote agents.

- `rotate force-reissue` still asks the old destructive confirmation before it branches on delivery mode. `src/commands/rotate/ca.rs:529`-`src/commands/rotate/ca.rs:536` always uses `prompt_rotate_force_reissue`, and that string still says "Delete certificates for {service_name} and trigger reissuance?" in `src/i18n/en.rs:301` / `src/i18n/ko.rs:301`. For `remote-bootstrap` services no files are deleted anymore; the command writes a KV request. I think the prompt should be delivery-mode-specific so the remote path describes the action it actually performs.

- I also found a config/behavior mismatch around `fast_poll_interval`. Validation allows values down to 1 second in `src/config/validation.rs:116`-`src/config/validation.rs:117`, but the loop always sleeps `max(fast_poll_interval, 5s)` in `src/fast_poll.rs:776`. So a configured `1s`-`4s` interval is silently never honored, even when the loop is healthy. If 5 seconds is meant to be a real floor, the config/docs should say so; otherwise the clamp should only apply to the retry-after-login-failure path that `LOGIN_RETRY_FLOOR` describes.

sehkone:

sehkone:
**[Author Round 4]**

All four items addressed in d21722e. Three fixed as suggested; one (the startup-warning narrowing) pushed back on the "add a marker" option in favour of removing the warning outright — reasoning below.

- **Fixed — docs stop calling `[openbao]` optional / legacy.** `docs/en/cli.md`, `docs/ko/cli.md`, and `agent.toml.example` now describe the section as required on remote-bootstrap hosts and note that `bootroot-remote bootstrap` auto-provisions the connection fields; the "omit to retain legacy behaviour" comment is gone. The `remote-bootstrap.md` pages already covered this in round 3 but still mentioned the (now-removed) startup warning — that paragraph has been replaced with a note that a legacy remote upgrading without rerunning bootstrap is the case where the section ends up absent.

- **Partially addressed — broad startup warning removed.** You're right the Round-3 warning in `src/daemon.rs` fires for every agent install without `[openbao]`, including local-only daemons, and the text telling the operator to rerun `bootroot-remote bootstrap` is wrong on those hosts. I considered the "narrower trigger with an explicit remote-bootstrap marker" option, but `agent.toml` has no such field today and the only places that could set one (`bootroot-remote bootstrap`) are the same place that upserts `[openbao]` — so a marker would be perfectly correlated with the presence of `[openbao]` and the warning would become unreachable in normal use. The narrow remaining case is "legacy remote upgrades bootroot without rerunning bootstrap", which is uncatchable from the agent side without a marker we don't have. Rather than introduce a marker whose only purpose is to gate a diagnostic that's moot after the first bootstrap, I removed the warning entirely. Operators of legacy upgrades are still routed at the docs (`docs/{en,ko}/remote-bootstrap.md`) and will notice the first `rotate force-reissue --wait` timing out. Happy to add a marker if you'd rather have the diagnostic even at the cost of another managed field.

- **Fixed — delivery-mode-specific prompt.** `rotate force-reissue` now reads `entry.delivery_mode` before calling `confirm_action` and picks between the existing `prompt_rotate_force_reissue` (local/Docker: "Delete certificates ...") and a new `prompt_rotate_force_reissue_remote` ("Write a reissue request for remote-bootstrap service {service_name} to OpenBao? The remote agent will apply it on its next fast-poll tick. [y/N]"). Strings added in both `src/i18n/en.rs` and `src/i18n/ko.rs`, with a matching accessor in `src/i18n/rotate.rs`. The inner branch that dispatches to `rotate_force_reissue_remote` is unchanged, so the code path the prompt describes is the one that runs.

- **Fixed — `fast_poll_interval` floor no longer applies to healthy ticks.** Agreed: `LOGIN_RETRY_FLOOR` is documented as a login-retry floor, but `run_fast_poll_loop` was clamping every sleep with `.max(LOGIN_RETRY_FLOOR)`, so an operator-configured `1s`–`4s` interval (allowed by `validate_openbao_settings`) was silently never honored. The loop now uses `openbao.fast_poll_interval` verbatim for the healthy-tick sleep and only clamps to `LOGIN_RETRY_FLOOR` when `needs_relogin` is true, so a sealed/down OpenBao still backs off while a tuned sub-5s cadence works. Comment on the clamp spells this out.

CI:
- `cargo fmt --check` / `cargo clippy --all-targets -- -D warnings` clean.
- Full `cargo test` (lib + all integration binaries) passes locally; `fast_poll::tests` (20), `bootroot_rotate` (43), and `agent_config::tests` (6) specifically exercise the touched paths. No existing test depended on the removed startup warning or on the old prompt text for remote services.

sehkone:
**[Reviewer Round 5]**

The Round 4 items look resolved to me in the updated diff. I do still have one new concern before merge:

- The new `state_path` default looks unsafe for the restart-persistence guarantee this feature relies on. The default is the bare relative filename `bootroot-agent-state.json` in `src/config/defaults.rs:26`, and `OpenBaoSettings` explicitly interprets a relative path against the agent process cwd in `src/config.rs:78`. But `bootroot-remote bootstrap` only upserts the connection fields and intentionally leaves `state_path` absent unless the operator already added one (`src/bin/bootroot-remote/agent_config.rs:319`), while the operator-facing example still shows `bootroot-agent` being run as a long-lived system service without any working-directory contract (`docs/en/cli-examples.md:424`). In a typical systemd-style deployment that means the fast-poll state may end up at `/bootroot-agent-state.json` or whatever cwd the supervisor chooses. If that path is unwritable or changes between runs, the agent loses `last_reissue_seen_version`, `in_flight_renewals`, and `pending_completion_writes` across restarts, which is exactly the state this design depends on to avoid duplicate renewals and to retry completion acknowledgements. I think this needs a stable absolute default (or bootstrap-provisioned path) plus a test for the no-`state_path` case.

sehkone:

sehkone:
**[Author Round 5]**

Fixed in eeda645.

- **Fixed — `state_path` auto-provisioned to an absolute path on first bootstrap.** You're right: the in-tree default (`bootroot-agent-state.json`) is a bare relative filename resolved against the agent process cwd, and under a systemd-style supervisor the cwd is not contracted to be stable or writable — so the fast-poll restart-persistence state (`last_reissue_seen_version`, `in_flight_renewals`, `pending_completion_writes`) could land at a path that moves or disappears between runs, defeating the duplicate-suppression and completion-retry guarantees this feature relies on. `bootroot-remote bootstrap` now inspects the existing `[openbao]` section in `agent.toml`; when `state_path` is absent it upserts an absolute path adjacent to `agent.toml` (i.e. `<agent_config_dir>/bootroot-agent-state.json`) alongside the connection fields. Operator-tuned `state_path` entries are preserved on rerun, matching the invariant already in place for `fast_poll_interval`. If `args.agent_config_path` is itself relative, the provisioning step is skipped rather than writing a same-cwd-relative `state_path` — that would just entrench the problem under a different name, and skipping leaves the existing `validate_openbao_settings` check free to flag the misconfiguration loudly.

Why provisioning at bootstrap rather than changing the in-tree default to an absolute path: there is no single absolute path that is right for every operator (distros vary on where `/var/lib/bootroot` or `/etc/bootroot` should live, and the Docker-sidecar layout differs again). The remote-bootstrap pipeline already treats `agent_config_path` as the stable, operator-owned anchor for a host's bootroot-agent install (secret_id/role_id/ca-bundle paths already live under its tree in the standard deployment), so deriving `state_path` from that same anchor keeps the feature coherent with how operators think about deployment layout.

Tests (the no-`state_path` case you asked for):
- `build_openbao_updates_provisions_absolute_state_path_when_missing` — takes a config with `[openbao]` but no `state_path`, runs `build_openbao_updates`, asserts the returned pair set contains `state_path`, that its rendered value is absolute, and that it renders to the expected adjacent-to-agent.toml path after upsert.
- `build_openbao_updates_skips_state_path_when_agent_config_relative` — locks in the "don't entrench a bad path" guard: when `agent_config_path` is relative, `state_path` must *not* appear in the managed key set.
- `build_openbao_updates_preserves_operator_tuned_keys` updated: now exercises the branch where the incoming config already has an absolute `state_path`, and verifies the bootstrap-time pair set leaves it alone.
- `build_openbao_updates_covers_connection_fields` updated to reflect the new managed-key set (`url`, `kv_mount`, `role_id_path`, `secret_id_path`, `ca_bundle_path`, `state_path`).

Docs + CHANGELOG:
- `agent.toml.example`: the `[openbao]` comment block now describes the absolute-state-path provisioning explicitly and tells operators an explicit `state_path` is the way to pin a custom location.
- `docs/{en,ko}/remote-bootstrap.md`: the managed `agent.toml` example now shows an absolute `state_path` (matching what bootstrap produces) and a new paragraph explains why bootstrap provisions it — specifically calling out the systemd-style / cwd-unstable supervisor case this is defending against.
- `CHANGELOG.md`: the Unreleased `### Added` entry for the `[openbao]` section now mentions the absolute-state-path provisioning and the reason (systemd-style supervisor restart-persistence).

PR description updated: the `[openbao]` bullet now calls out the absolute-state-path provisioning and the systemd-class motivation; the test plan lists the two new tests.

Verified locally: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib` (197 tests), `cargo test --bin bootroot-remote agent_config` (8 tests), and `cargo test --test bootroot_rotate` (43 tests) all pass.

sehkone:
**[Reviewer Round 6]**

I re-checked the Round 5 change, and the common case is fixed: when `agent.toml` lives at an absolute path, `bootroot-remote bootstrap` now provisions an absolute adjacent `state_path`, and the new tests cover that path.

I do still see one remaining issue before merge:

- The relative-`agent_config_path` branch still falls back to the unsafe cwd-relative default at runtime. `build_openbao_updates()` intentionally skips writing `state_path` when `agent_config_path` is relative and says that leaving it unset will let validation catch the problem (`src/bin/bootroot-remote/agent_config.rs:358`). But the runtime config layer still supplies the default `bootroot-agent-state.json` (`src/config.rs:78`, `src/config/defaults.rs:26`), and `validate_openbao_settings()` only checks that `state_path` is non-empty, not absolute (`src/config/validation.rs:103`, `src/config/validation.rs:127`). So a bootstrap done with `--agent-config-path agent.toml` still produces an `[openbao]` config that silently uses a cwd-relative state file, which is the same restart-persistence hazard Round 5 was trying to remove. I think this needs one more guardrail: either reject relative `agent_config_path` during remote bootstrap when `[openbao]` is being provisioned, or reject non-absolute `openbao.state_path` in config validation. An end-to-end test that loads a config produced from a relative `agent_config_path` would help lock this down.

sehkone:

sehkone:
**[Author Round 6]**

Fixed in d45041e.

- **Fixed — relative `openbao.state_path` now rejected at validation.** Agreed with the analysis: `build_openbao_updates` skipping `state_path` when `agent_config_path` is relative was only half the guardrail — the config layer still fills in the in-tree relative default and `validate_openbao_settings` only required non-empty, so a bootstrap run with a relative `--agent-config-path` (or a legacy operator-written config) silently loaded with a cwd-relative state file, defeating the restart-persistence guarantee Round 5 added. `validate_openbao_settings` now requires `state_path.is_absolute()`. The error names the offending path and points operators at `bootroot-remote bootstrap`, which already provisions an absolute adjacent path when `agent_config_path` is absolute. I took the "reject at validation" option rather than "reject at bootstrap time" because it catches both the bootstrap-skipped branch *and* any legacy/handwritten config that omits `state_path` or sets it to a relative value — one defence point instead of two.

  Tests added:
    - `validate_rejects_relative_openbao_state_path` and `validate_rejects_omitted_openbao_state_path_via_default` in `src/config.rs` cover the explicit-relative and default-fallback cases.
    - `validate_accepts_absolute_openbao_state_path` locks in the happy path so the rejection is scoped to relative values only.
    - `config_built_from_relative_agent_config_path_fails_validation` in `src/bin/bootroot-remote/agent_config.rs` is the end-to-end case you asked for: runs `build_openbao_updates` with a relative `agent_config_path`, renders the managed `[openbao]` section via `upsert_section_keys`, parses the result with `Settings::new`, and asserts `validate()` errors with "openbao.state_path" + "absolute". That's the exact integration path where Round 5's guard had a gap.

  Docs + CHANGELOG:
    - `docs/{en,ko}/remote-bootstrap.md`: the `state_path` paragraph now notes that config validation rejects a relative value (including the omitted-→-default case), so a misconfiguration surfaces at agent startup instead of silently running with a fragile state file.
    - `agent.toml.example`: the `state_path` comment block explicitly calls out the absolute-path requirement and the reason (cwd not contracted under systemd-style supervisors).
    - `CHANGELOG.md`: the Unreleased `### Added` `[openbao]` bullet now mentions the validation guardrail and names the two paths it catches.
    - PR description updated: the `[openbao]` bullet and the test plan both describe the new validation + tests.

Verified locally: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` (full suite) all clean. `config::tests` (26 tests) and `bootroot-remote agent_config::tests` (9 tests) cover the new paths.

sehkone:
**[Reviewer Round 7]**

I re-checked the Round 6 validation guardrail, and that part is fixed. I do still see one remaining correctness issue in the fast-poll state machine:

- `src/fast_poll.rs:171` still treats the KV metadata version as "the request version" on every read. But the agent's own completion ack at `src/fast_poll.rs:208` is written back to the same KV path, so a successful completion for request version `N` creates secret metadata version `N+1` while the payload now carries `completed_version = N`. Because the state only records `last_reissue_seen_version = N` (`src/fast_poll.rs:453`), the next tick reads metadata version `N+1`, `evaluate_observation()` decides `N+1 > N`, and triggers a second forced renewal even though no operator made a new request. After that second renewal, the same pattern repeats (`N+2 > N+1`, etc.), so one force-reissue can turn into a renewal on every fast-poll tick. The same retrigger also happens after a successful pending-completion retry, because that retry write also advances the secret metadata version.

I think the read side needs to derive an effective "request version" from `completed_version` when that field is present (or otherwise ignore the agent's own completion writes), instead of keying purely off `metadata.version`. A regression test for "tick 1 applies request `N`, tick 2 sees metadata version `N+1` with `completed_version = N`, and stays `UpToDate`" would lock this down.

sehkone:

sehkone:
**[Author Round 7]**

- **Fixed** — self-retrigger of renewals on every fast-poll tick after a successful force-reissue.

  You're right — this was a real bug. The read path keyed purely off `metadata.version > last_seen_version`, but KV v2 increments the metadata version on every write, and the agent's *own* completion ack is a write. So after a successful renewal at request version `N`, the completion ack landed at metadata version `N+1` (with payload `completed_version = N`), and the next tick saw `N+1 > N` and force-renewed again. The same pattern repeated on every tick, and the pending-completion retry path had the same failure mode for the same reason.

  Fix in `src/fast_poll.rs`: `evaluate_observation` now treats any payload carrying `completed_version` as an already-serviced request (not a new one) and returns `None` regardless of `last_seen_version`. The control-plane publish writes *never* include `completed_version` (only the agent's ack does), so this cleanly separates operator requests from the agent's own echo:

  ```rust
  let completed_version = observed
      .data
      .as_object()
      .and_then(|obj| obj.get(REISSUE_COMPLETED_VERSION_KEY))
      .and_then(serde_json::Value::as_u64);
  if completed_version.is_some() {
      return None;
  }
  if last_seen_version.is_some_and(|seen| observed.version <= seen) {
      return None;
  }
  ```

  This also covers two adjacent edge cases: (a) if the state file is lost, a payload carrying `completed_version` does *not* retroactively re-fire (we treat the request as already serviced); (b) a successful pending-completion retry advances metadata via another ack write, and the next tick correctly stays `UpToDate`.

  Two regression tests added:

  - `evaluate_observation_ignores_agent_completion_ack` — unit-level: payload with `completed_version = 5` at metadata version `6` returns `None` against both `Some(5)` and `None` last-seen states.
  - `tick_stays_up_to_date_when_metadata_advanced_by_own_completion_ack` — exactly the scenario you asked for: `last_reissue_seen_version = 5`, KV returns `metadata.version = 6` with `completed_version = 5` in the payload → outcome is `UpToDate`, no renewal triggered, no completion write issued, `state_changed` is false, and `last_reissue_seen_version` stays pinned at `5`.

  I considered your suggested alternative of "derive an effective request version from `completed_version` and compare against `last_seen`" and landed on the stricter "return `None` whenever `completed_version` is present" because the two diverge on the state-lost case: with the derived-effective variant and `last_seen = None`, an ack payload would still fire a renewal (because no threshold gates it); the "ignore any ack" variant keeps the invariant "the agent only fires on operator-written payloads" even across state-file loss. The pending-retry path's `observation.version == pending.version` comparison on raw KV metadata stays as-is — that path only runs when the previous write truly failed, in which case metadata did *not* advance, so metadata-version equality there is the correct match condition (and if metadata did advance via some partial success, the mismatch correctly drops the stale pending).

  CHANGELOG updated; PR body updated to reflect the agent-ack suppression and the new test count (22 fast-poll unit tests).

sehkone:
**[Reviewer Round 8]**

I re-checked the Round 7 item, and that part is fixed: `src/fast_poll.rs:181` now suppresses agent-written ack payloads via `completed_version`, and the new regressions around the `N -> N+1` metadata bump cover the loop I called out.

I do still see two follow-ups from this round:

- `--wait` still does not report the end-to-end latency that #548 asked for. In `src/commands/rotate/ca.rs:635` the success path only prints `completed_at`, and the message template at `src/i18n/en.rs:309` has no slot for a computed duration. That means the new control-plane path can confirm completion, but it still makes the operator do the timestamp subtraction manually instead of surfacing the latency directly. I think the CLI should parse `requested_at` / `completed_at`, print the elapsed duration on success, and add/update a test to lock that output in.

- The remediation hint for a relative `openbao.state_path` is misleading today. `src/config/validation.rs:141` tells the operator to "rerun `bootroot-remote bootstrap`", but `build_openbao_updates()` only provisions `state_path` when the key is missing (`src/bin/bootroot-remote/agent_config.rs:341`). If an existing config already has a relative `state_path`, rerunning bootstrap preserves that invalid value and startup still fails. Either bootstrap should repair a relative path, or the validation/docs should stop suggesting that rerunning bootstrap will fix this specific case.

sehkone:

sehkone:
**[Author Round 8]**

Thanks — both follow-ups are valid and are now addressed in fea6bc3.

- **`--wait` end-to-end latency** — Fixed. `rotate_summary_force_reissue_completed` now takes an `{elapsed}` slot, and `src/commands/rotate/ca.rs` parses the `requested_at`/`completed_at` pair (preferring the KV-payload `requested_at` over the local variable, so the number reflects what the agent actually applied) and formats the delta via `humantime::format_duration`. Skew/parse failures collapse to `"unknown"` rather than emitting a misleading negative duration. Unit tests: `format_reissue_elapsed_reports_seconds_difference`, `format_reissue_elapsed_handles_multi_minute_gap`, `format_reissue_elapsed_returns_unknown_on_parse_failure`, `format_reissue_elapsed_returns_unknown_when_completion_precedes_request`. The `wiremock`-driven `test_rotate_force_reissue_remote_wait_reports_completion` asserts the new `end-to-end latency: 14s` suffix on the summary line. Bilingual i18n (en + ko) and docs updated.
- **Misleading remediation hint on relative `openbao.state_path`** — Fixed. I chose the "bootstrap actually repairs it" path: `build_openbao_updates` now also overwrites `state_path` when the existing value is relative (or malformed), not just when the key is missing. So rerunning `bootroot-remote bootstrap` genuinely fixes a legacy config whose `state_path` was the in-tree relative default, which is what `validate_openbao_settings` promises. Preservation of operator-tuned keys is unchanged — only absolute values are preserved now (a relative operator value would itself be the hazard we're trying to catch). New regression: `build_openbao_updates_repairs_relative_state_path`. The existing `build_openbao_updates_preserves_operator_tuned_keys` still passes because its fixture uses an absolute path. Docs (`docs/{en,ko}/remote-bootstrap.md`) updated to describe the repair behaviour.

CI re-triggered by the push.

sehkone:
**[Reviewer Round 9]**

I re-checked the two Round 8 items, and those look resolved:

- `--wait` now reports the computed end-to-end latency in the success line (`src/commands/rotate/ca.rs:635`, `src/i18n/en.rs:309`), with unit coverage for the formatter and an integration check in `tests/bootroot_rotate.rs:1749`.
- Rerunning `bootroot-remote bootstrap` now really does repair a relative `openbao.state_path` (`src/bin/bootroot-remote/agent_config.rs:341`, `src/bin/bootroot-remote/agent_config.rs:725`), which matches the validation hint in `src/config/validation.rs:140`.

I did find one new correctness issue in the fast-poll daemon loop:

- The "retry on the next tick" path after an initial AppRole login failure is still broken because the relogin trigger is looking for the wrong error text. `run_fast_poll_loop()` logs `"Will retry on the next tick"` when the startup `login()` fails (`src/fast_poll.rs:675`), but the only way it later sets `needs_relogin = true` is if a `ReadError` contains `"permission denied"`, `"403"`, or `"missing client token"` (`src/fast_poll.rs:723`). The actual no-token error produced by the OpenBao client is `"OpenBao token is not set"` (`src/openbao.rs:967`). So after a transient startup login failure, the next tick's KV reads fail locally with `OpenBao token is not set`, `needs_relogin` stays false forever, and the loop never attempts AppRole login again unless the process is restarted. That leaves fast-poll permanently dead after exactly the kind of transient startup blip this retry path is meant to survive. I think this needs to key the retry decision off the actual missing-token error (or, better, off a typed condition rather than substring matching), and it would be worth adding a regression test that starts the loop with no token / a failed initial login and proves it retries login on a later tick.

sehkone:

sehkone:
**[Author Round 9]**

- **Fixed — relogin heuristic missed the "OpenBao token is not set" error.** You're right: `require_token()` (`src/openbao.rs:967`) emits `"OpenBao token is not set"` when no login has succeeded, and the loop's substring filter at `src/fast_poll.rs:725` did not match that string, so a failed initial `login()` left `needs_relogin` stuck at false forever. I extracted the decision into `read_error_requires_relogin()` (`src/fast_poll.rs:40`) and added `"token is not set"` to the matched patterns alongside the existing `"permission denied"` / `"403"` / `"missing client token"` terms. The startup-failure path at `src/fast_poll.rs:676` now rearms: if the initial login fails, the next tick's KV reads return that string, `read_error_requires_relogin()` matches, `needs_relogin = true`, and the end-of-iteration block calls `login()` again (with the `LOGIN_RETRY_FLOOR` backoff already in place).

  Added a regression test `read_error_requires_relogin_detects_missing_token` (`src/fast_poll.rs:879`) that pins the heuristic against exactly the token-related messages the agent observes (the `require_token()` string, a 403 "permission denied", a 400 "missing client token") and confirms unrelated errors like "connection refused" or a 500 still do not force a relogin. Keeping this as a pure-function check rather than wiring up a full `run_fast_poll_loop` wiremock test lets the assertion sit right next to the heuristic it guards; the full loop continues to exercise the resulting behaviour through the existing integration tests.

  CHANGELOG and PR description updated to call out the broadened matcher. `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --lib fast_poll` (now 23 tests) all pass.

sehkone:
**[Reviewer Round 10]**

I re-checked the Round 9 item, and that one looks resolved: `read_error_requires_relogin()` now matches the no-token path from `require_token()` and has a regression test right next to the heuristic.

I did find one new blocking correctness issue in the daemon integration:

1. The new fast-poll path can race the ordinary daemon loop and issue the same profile twice. `run_daemon()` now keeps the per-profile periodic tasks running and also spawns the fast-poll task in parallel (`src/daemon.rs:68`). When fast-poll fires, `force_renew_profile()` immediately acquires the shared semaphore and starts ACME unconditionally (`src/daemon.rs:412`). The periodic path still decides whether renewal is needed *before* it acquires that semaphore (`src/daemon.rs:437`, `src/daemon.rs:448`, `src/daemon.rs:465`), and the default concurrency is `3` (`src/config/defaults.rs:23`). That means a profile whose normal check tick lands around the same time as a force-reissue can read the old cert, decide "needs renewal", and then perform a second issuance after (or even alongside) the forced one. The old local-file/SIGHUP path restarted the daemon, so it did not introduce this overlap. I think this needs some per-profile single-flight / in-progress suppression shared between the periodic and force-reissue paths, plus a regression test that proves only one issuance happens when the two triggers race.

sehkone:

sehkone:
**[Author Round 10]**

1. **Fixed** — added a per-profile single-flight lock (`ProfileLocks`) shared between the periodic `check_interval` loop and the fast-poll force-reissue path. `check_and_renew_profile` and `force_renew_profile` now acquire the per-profile mutex *before* the shared `max_concurrent_issuances` semaphore, and the periodic path re-reads the cert via `should_renew` *after* acquisition, so a forced reissue that lands first is observed as a rotated cert and the periodic issuance is skipped instead of queuing behind a spare semaphore permit and firing a second ACME handshake. A new regression test — `profile_lock_blocks_double_issuance_when_force_and_periodic_race` — models the race by holding the lock while rotating the cert on disk and asserting the concurrent periodic task observes the fresh cert and does not issue; sibling tests (`profile_locks_serialise_same_profile`, `profile_locks_allow_different_profiles_concurrently`) lock down the primitive itself. CHANGELOG and PR description updated.

sehkone:
**[Reviewer Round 11]**

I re-checked the Round 10 item, and that one looks resolved. `force_renew_profile()` now takes the per-profile lock before the shared issuance semaphore at `src/daemon.rs:466`, and the periodic path does the same before re-running `should_renew()` at `src/daemon.rs:494`, which closes the duplicate-issuance window I called out last round. The added regression coverage in `src/daemon.rs:911` and `src/daemon.rs:978` is also aimed at the right behavior.

I also reviewed the rest of the updated diff against #548 and didn’t find any new blocking issues in this round.

sehkone:
[Review Verdict Round 11: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Add KV-backed fast-poll for remote force-reissue
```

**Body**

```text
`bootroot rotate force-reissue` on a `--delivery-mode remote-bootstrap`
service used to delete the control-node's cert files and print a hint
telling the operator to rerun `bootroot-remote bootstrap` on the remote
host. Because the remote agent only renewed on its `check_interval`
(default 1h), worst-case latency for a "force" action was up to an hour.

Switch the control plane to publish a versioned reissue request to
OpenBao KV at `{kv_mount}/data/bootroot/services/<service>/reissue` and
add a fast-poll loop to the remote bootroot-agent that observes new KV
versions and triggers an immediate ACME renewal. Worst-case latency
drops from one `check_interval` to roughly one `fast_poll_interval`
(default 30s).

Highlights:

- Control side: `rotate force-reissue` writes the request via a new
  `OpenBaoClient::write_kv_with_version` and supports `--wait`,
  `--wait-timeout` (default 2m), and `--requester`. `--wait` pins on the
  version returned by the publish POST — not a follow-up GET — so the
  agent's own completion write cannot race the readback and strand the
  waiter. The success line reports the computed end-to-end latency
  (`completed_at - requested_at`, read from the KV payload) in
  human-readable form so the operator does not have to subtract
  timestamps manually. The confirmation prompt is delivery-mode-specific;
  local/Docker services keep the existing "delete certificates ..."
  prompt.
- Remote side: a new `[openbao]` section in `agent.toml` drives the
  fast-poll loop. The read path treats payloads carrying
  `completed_version` as already-serviced requests so the agent's own
  completion ack (which bumps KV metadata from N to N+1) is not mistaken
  for a new operator request and re-firing a forced renewal on every
  tick. A per-profile single-flight lock serialises the fast-poll force
  path with the periodic `check_interval` renewal so a periodic tick
  landing next to a forced reissue re-reads the rotated cert and skips
  the redundant second ACME handshake instead of driving it through a
  spare semaphore permit. Multi-profile hosts fan the trigger out to
  every profile sharing a `service_name`, with per-profile retry
  progress persisted across ticks so a flaky sibling only re-attempts
  itself. Transient completion-write failures are queued via
  `pending_completion_writes` and retried on later ticks so `--wait`
  callers are not stranded after a successful renewal. The AppRole
  relogin heuristic covers the "OpenBao token is not set" startup
  path so a transient initial-login failure is retried instead of
  leaving fast-poll permanently silent until process restart.
- Clock-skew handling: KV v2 version numbers are the authority;
  `requested_at` and `completed_at` stay in the payload for operator
  observability only.
- `[openbao]` is mandatory on remote-bootstrap hosts and
  auto-provisioned. `bootroot-remote bootstrap` upserts the connection
  fields (`url`, `kv_mount`, `role_id_path`, `secret_id_path`,
  `ca_bundle_path`) on every run and provisions an absolute `state_path`
  adjacent to `agent.toml` whenever the existing value is missing or
  relative, so restart-persistence state (`last_reissue_seen_version`,
  `in_flight_renewals`, `pending_completion_writes`) does not land at a
  cwd-relative path under systemd-style supervisors that do not
  contract for a stable working directory. Config validation rejects a
  relative `openbao.state_path` (including the in-tree default that
  `serde` fills in when the key is omitted), catching a legacy
  operator-written config or a bootstrap run whose
  `--agent-config-path` was itself relative at `bootroot-agent`
  startup. Operator-tuned `fast_poll_interval` values and custom
  absolute `state_path` entries are preserved across reruns.
- Docs (`docs/{en,ko}/cli.md`, `docs/{en,ko}/remote-bootstrap.md`,
  `agent.toml.example`), CHANGELOG, and bilingual i18n are updated.

Closes #548
```

PR #571, opened 2026-04-20
Emit full agent.toml baseline in local-file service add (#549)

### LINKED ISSUE #549: Managed-profile agent.toml template omits server and http_responder_url, so re-renders overwrite operator edits
## Problem

\`agent.toml.ctmpl\` generated by \`bootroot service add --delivery-mode local-file\` does not include \`server\` (ACME directory URL) or \`[acme].http_responder_url\`. If the bundled compose topology's defaults (\`https://localhost:9000/acme/acme/directory\`, \`http://localhost:8080\`) happen to match, bootroot-agent's built-in defaults cover the gap. The moment an operator needs non-default values — different step-ca port, responder on another host — they hand-edit \`agent.toml\` to add the fields. The next KV rotation (\`rotate responder-hmac\`, etc.) causes the OpenBao Agent sidecar to re-render from the template and the operator's edits are silently overwritten.

The \`bootroot-remote bootstrap\` output for remote-bootstrap services does emit every field explicitly (see [src/bin/bootroot-remote/agent_config.rs](src/bin/bootroot-remote/agent_config.rs) \`render_agent_config_baseline\`), so operators with a non-default topology can run remote-bootstrap without this footgun. Local-file should be at parity.

## Reproduction

\`\`\`
$ bootroot service add --service-name review --delivery-mode local-file --agent-config /abs/config/review-agent.toml ...
$ cat secrets/openbao/services/review/agent.toml.ctmpl
domain = \"test.local\"
# BEGIN bootroot managed profile: review
[[profiles]] ...
[trust] ...
[acme]
http_responder_hmac = \"{{ ... }}\"
# END bootroot managed profile: review
# No `server = \"...\"`, no `[acme].http_responder_url = \"...\"`.

# Operator on a non-default topology edits the rendered agent.toml to add `server` / `http_responder_url`:
$ vim /abs/config/review-agent.toml   # adds missing fields, saves

# Next KV rotation re-renders from .ctmpl, losing the edits:
$ bootroot rotate responder-hmac --yes
$ grep -E 'server|http_responder_url' /abs/config/review-agent.toml
http_responder_hmac = \"...\"          # operator's `server`/`http_responder_url` lines are gone
\`\`\`

## Root cause

\`build_managed_agent_ctmpl\` (in [src/commands/service/local_config.rs:45](src/commands/service/local_config.rs) and [src/trust_bootstrap.rs:105](src/trust_bootstrap.rs)) emits only \`domain\`, the managed \`[[profiles]]\` block, \`[trust]\`, and \`[acme].http_responder_hmac\`. It relies on bootroot-agent's compiled-in defaults for everything else — including fields operators commonly need to override.

## Proposed fix

Update the local-file template generation path to include every field that appears in remote-bootstrap's \`render_agent_config_baseline\`: at minimum \`email\`, \`server\`, \`[acme].http_responder_url\`, plus the \`[acme]\` retry/timeout tunables. Source values from the \`bootroot init\` config state (or the same CLI defaults used by remote-bootstrap) so re-renders preserve the initial configuration rather than reverting to agent built-in defaults.

Stretch goal: extract the rendering into a single helper used by both \`src/commands/service/local_config.rs\` and \`src/bin/bootroot-remote/agent_config.rs\` so future additions don't diverge again. (See also #547 — the equivalent drift in \`agent.hcl\` renderers; same structural move.)

## Not in scope

The managed-profile delimiter string differs between local-file (\`# BEGIN bootroot managed profile: <svc>\`) and remote-bootstrap (\`# BEGIN BOOTROOT REMOTE PROFILE <svc>\`). That is a consistency nit, not a correctness bug; tracking separately if desired but not part of this issue.

## Environment

- macOS 26.4.1

sehkone:
## Summary

- Extract `render_agent_config_baseline` and its `AgentConfigBaselineParams` into `bootroot::trust_bootstrap` so both the remote-bootstrap and local-file `service add` paths share one renderer.
- Seed the fresh-template path in `apply_local_service_configs` with the baseline (no longer an empty string), so the managed-profile upsert now layers over a file that already contains `email`, `server`, `domain`, and the full `[acme]` section — including `http_responder_url` and the retry/timeout tunables.
- **Added in Round 1 response**: `bootroot service add` now accepts `--agent-email`, `--agent-server`, and `--agent-responder-url`, mirroring the escape hatch that `bootroot-remote bootstrap` has long provided. The values flow through `ResolvedServiceAdd` into `render_agent_config_baseline`, so operators on non-default topologies (e.g. `https://step-ca.example:9443/...` or a responder on another host) can bake their real URLs into the template at service-add time. KV-driven re-renders (e.g. `rotate responder-hmac`) then preserve those operator values instead of clobbering hand-edits with the compose-topology localhost defaults.
- **Added in Round 2 response**: the same `--agent-email` / `--agent-server` / `--agent-responder-url` values also flow into the `remote-bootstrap` delivery path. `build_artifact()` used to hard-code `DEFAULT_AGENT_*` into `bootstrap.json`, which made those flags silently no-op under `--delivery-mode remote-bootstrap`. `write_remote_bootstrap_artifact()` now threads the resolved `agent_*` values through, so the emitted `bootstrap.json` carries the operator's real topology and the downstream `bootroot-remote bootstrap` run uses those values directly.
- **Added in Round 3 response**: the resolved `--agent-email` / `--agent-server` / `--agent-responder-url` values are now persisted on `ServiceEntry` in `state.json` (as `Option<String>`, `#[serde(default, skip_serializing_if)]` so pre-existing entries deserialize cleanly as `None`) and are included in the `remote-bootstrap` idempotence comparison (`non_policy_fields_match`). `write_remote_bootstrap_artifact_from_entry()` now reads the persisted entry values rather than the transient `resolved.agent_*`, so an idempotent rerun of `bootroot service add --delivery-mode remote-bootstrap ...` regenerates `bootstrap.json` from the stored values — a rerun that omits the flag no longer silently reverts the artifact to the compose-topology default, and a rerun with a different value is rejected as a duplicate rather than silently drifting the artifact.
- **Added in Round 4 response**: the local-file path now also backfills baseline keys and applies `--agent-*` overrides when `--agent-config` points at a pre-existing `agent.toml`. Previously the existing-file branch of `apply_local_service_configs` read the file verbatim and only upserted `[trust]` / `domain` / `http_responder_hmac`, so a pre-existing file missing `email` / `server` / `[acme].http_responder_url` produced a `.ctmpl` that still omitted those fields and silently ignored the new `--agent-*` flags. A new `trust_bootstrap::apply_agent_config_baseline_defaults` helper — built on new `toml_util::insert_missing_{top_level,section}_keys` primitives — backfills any missing baseline key without clobbering operator-customised values. Explicit `--agent-email` / `--agent-server` / `--agent-responder-url` values are then upserted so they always take effect.
- **Added in Round 5 response**: `bootroot-remote bootstrap` now applies the same baseline-backfill + artifact-override treatment on the remote target's pre-existing `agent.toml`. Previously `apply_agent_config_updates()` used `render_agent_config_baseline` only on the `NotFound` branch; the existing-file branch read the file verbatim and only upserted `[trust]` / `acme.http_responder_hmac` / the managed profile, silently dropping the artifact's `agent_email` / `agent_server` / `agent_responder_url` values whenever the remote file was missing those keys. The renderer now calls `apply_agent_config_baseline_defaults` on both branches (empty string on `NotFound`, file contents otherwise) and then upserts the explicit artifact overrides via a new `apply_agent_overrides` helper. `ResolvedBootstrapArgs` gained `agent_email_override` / `agent_server_override` / `agent_responder_url_override` (`Option<String>`) populated from the artifact's `Option<String>` fields, so an explicit service-add override propagates through to the remote upsert while a direct-CLI run on `bootroot-remote bootstrap` (without an artifact) stays backfill-only and does not clobber operator-customised values.
- **Added in Round 6 response**: `RemoteBootstrapArtifact.agent_email` / `agent_server` / `agent_responder_url` are now `Option<String>` with `#[serde(skip_serializing_if = "Option::is_none")]`, and `build_artifact()` takes `Option<&str>` instead of `&str`. Previously both call sites funnelled the resolved values through `effective_agent_*` fallbacks, so an artifact produced by `bootroot service add` without `--agent-*` flags still serialized the compose-topology localhost defaults as if they were explicit overrides. The downstream `resolve_bootstrap_args()` treated mere field presence as an override (`agent_*_override = artifact.agent_*`), which meant the Round 5 "preserve operator values when no override" path was unreachable for any real `--artifact` flow — the upsert always clobbered remote operator values back to `admin@example.com` / `https://localhost:9000/...` / `http://127.0.0.1:8080`. Omitting the keys when `resolved.agent_* == None` preserves the "no explicit override" signal across the JSON boundary so the remote renderer takes the backfill-only path as intended.
- **Added in Round 7 response**: the Round 6 wire-shape change (`agent_email` / `agent_server` / `agent_responder_url` went from required strings to optional-and-omittable) required a `schema_version` bump per the contract documented above `RemoteBootstrapArtifact`. The producer now emits `schema_version: 3` (`src/commands/service/remote_bootstrap.rs`) and `bootroot-remote` accepts `1..=3` (`MAX_SUPPORTED_SCHEMA_VERSION` bumped from 2 to 3 in `src/bin/bootroot-remote/main.rs`). v2 parsing still works (the parser uses `#[serde(default)] Option<String>` for these keys), so a pre-existing v2 artifact on disk continues to bootstrap successfully; the bump exists for strict external consumers that key off `schema_version` to know which fields are guaranteed present.
- The flags are all optional — omit them to keep the existing same-host-compose behaviour (the `DEFAULT_AGENT_EMAIL` / `DEFAULT_AGENT_SERVER` / `DEFAULT_AGENT_RESPONDER_URL` constants remain the fallback, applied via `effective_agent_*` helpers at render time).

Closes #549

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --bin bootroot commands::service::local_config::tests::test_fresh_agent_config_ctmpl_carries_full_baseline`
- [x] `cargo test --bin bootroot commands::service::local_config::tests::test_fresh_agent_config_ctmpl_embeds_non_default_topology_values` — asserts caller-supplied non-default values appear in the rendered `.ctmpl` and that the built-in localhost defaults do NOT appear when the operator supplied overrides (local-file path).
- [x] `cargo test --bin bootroot commands::service::local_config::tests::test_existing_agent_config_ctmpl_backfills_missing_fields_from_overrides` — Round 4 regression test: pre-existing `agent.toml` missing `email` / `server` / `http_responder_url`, `--agent-*` overrides land in the generated `.ctmpl`.
- [x] `cargo test --bin bootroot commands::service::local_config::tests::test_existing_agent_config_preserves_operator_customised_topology` — Round 4 companion regression: pre-existing file carrying operator-customised `email` / `server` / `http_responder_url`, no `--agent-*` overrides — asserts the operator's values survive the backfill path untouched.
- [x] `cargo test --bin bootroot commands::service::remote_bootstrap::tests::build_artifact_embeds_non_default_agent_overrides` — Round 2 regression test (remote-bootstrap artifact path).
- [x] `cargo test --bin bootroot commands::service::remote_bootstrap::tests::build_artifact_omits_agent_keys_when_no_override` — **new Round 6 regression test**: when no `--agent-*` flags were supplied, the generated artifact omits the `agent_email` / `agent_server` / `agent_responder_url` keys entirely (verified both on the struct and on the serialized JSON output) so the "no explicit override" signal survives the JSON boundary.
- [x] `cargo test --bin bootroot commands::service::tests::is_idempotent_remote_rerun_false_when_agent_overrides_differ` — Round 3 regression test (all three mismatch shapes reject the idempotent rerun).
- [x] `cargo test --bin bootroot-remote agent_config::tests::existing_agent_config_backfills_overrides_from_artifact` — Round 5 regression test: pre-existing remote `agent.toml` missing `email` / `server` / `http_responder_url`, artifact carries overrides, asserts the baked values land in the output plus the `[acme]` tunables, and the localhost defaults do NOT leak through.
- [x] `cargo test --bin bootroot-remote agent_config::tests::existing_agent_config_preserves_operator_values_when_no_override` — Round 5 companion regression: pre-existing remote file with operator-customised `email` / `server` / `http_responder_url`, artifact carries no overrides — asserts operator values survive the backfill path.
- [x] `cargo test --bin bootroot-remote tests::resolve_bootstrap_args_no_override_keeps_override_none` — **new Round 6 resolver-level regression**: an artifact JSON that omits the three keys parses with `agent_*_override == None`, so the downstream `apply_agent_config_updates` takes the backfill-only branch. This is the shape the production `bootroot service add` emits when no `--agent-*` flags were supplied.
- [x] `cargo test --bin bootroot-remote tests::resolve_bootstrap_args_override_propagates_from_artifact` — Round 6 companion: the inverse shape — when the artifact DOES carry explicit overrides, those values land on `agent_*_override` so the downstream path knows to clobber pre-existing remote values.
- [x] `cargo test --bin bootroot-remote tests::resolve_bootstrap_args_rejects_future_schema_with_unknown_fields` — updated to reject `schema_version: 4` now that `bootroot-remote` accepts `1..=3` (Round 7).
- [x] `cargo test --lib trust_bootstrap::tests::render_agent_config_baseline_includes_all_documented_fields`
- [x] Full suite: `cargo test --lib --bins` — all tests pass.
- [x] `scripts/preflight/ci/check.sh` — quality-check job passes locally.
- [ ] Manual: `bootroot service add --delivery-mode remote-bootstrap --agent-server https://step-ca.example.org:9443/acme/acme/directory --agent-responder-url http://responder.internal:18080 ...` followed by `bootroot-remote bootstrap --artifact bootstrap.json` against a remote target carrying a pre-existing `agent.toml` missing those fields — verified by the new unit tests; Docker E2E matrix exercises default-topology paths.
- [ ] Preflight: `scripts/preflight/ci/e2e-matrix.sh` / `e2e-extended.sh` — relying on CI for the full Docker E2E matrix since the Round 6 change is narrow (artifact field `String` → `Option<String>` and a renderer signature widening), and the unit tests pin both shapes (omit keys when no override, include keys when override) at both the artifact writer and the downstream resolver.


sehkone:
**[Reviewer Round 1]**

I like the extraction of the shared baseline renderer, but I think this still misses the actual correctness bug from #549.

- `apply_local_service_configs()` now seeds a fresh `agent.toml` from `DEFAULT_AGENT_EMAIL`, `DEFAULT_AGENT_SERVER`, and `DEFAULT_AGENT_RESPONDER_URL` when the target file does not already exist (`src/commands/service/local_config.rs:64`, `src/commands/service.rs:39`). That fixes the default-topology case, but it does not fix the reported non-default-topology case. A fresh `local-file` service on `https://step-ca.example:9443/...` or with a responder on another host still gets a `.ctmpl` containing the localhost defaults. If the operator then edits the rendered `agent.toml` to the real values, the next rotation will still re-render the defaults from the template and clobber those edits. So the “missing fields” part is fixed, but the “re-renders overwrite operator edits on non-default topology” part is not.
- I think parity with the remote renderer is a bit misleading here, because remote bootstrap still has an explicit override channel via `--agent-server` / `--agent-responder-url` (`src/bin/bootroot-remote/main.rs:116`, `src/bin/bootroot-remote/main.rs:127`, with CLI fallbacks wired at `src/bin/bootroot-remote/main.rs:436` and `src/bin/bootroot-remote/main.rs:444`). The local-file path has no equivalent override once `service add` has emitted the template, so baking in the same defaults does not give operators the same escape hatch.
- The new test also only asserts that the default values are present (`src/commands/service/local_config.rs:564`), so it would still pass in the failing scenario above. I think we need coverage that starts from non-default baseline values and proves the generated `.ctmpl` preserves those exact values through a re-render.

I’d treat this as blocking: the local-file path still needs some source of truth for the intended ACME server / responder URL (persisted init state, explicit service-add inputs, or similar), otherwise the original overwrite footgun remains for the operators this issue is about.

sehkone:

sehkone:
**[Author Round 1]**

Thank you for the pushback — you were right that baking localhost defaults into the template still left the non-default-topology overwrite footgun in place. Addressing each item:

- **Fixed — "fresh `.ctmpl` still gets localhost defaults on non-default topology"**: `bootroot service add` now accepts `--agent-email`, `--agent-server`, and `--agent-responder-url`, mirroring the flags that `bootroot-remote bootstrap` already exposes. The flags are threaded through `ResolvedServiceAdd` into `render_agent_config_baseline`, so a service-add on a non-default step-ca or responder URL bakes those real values into the template at add time. Subsequent KV-driven re-renders (`rotate responder-hmac`, etc.) then preserve the operator's configuration instead of falling back to the compose-topology localhost defaults. The flags are optional — omit them and the behaviour is unchanged from before.

- **Fixed — "parity with remote renderer is misleading without an override channel"**: the previous revision had parity only at the render layer. With the new CLI flags the local-file path now has the same explicit override channel as `bootroot-remote bootstrap`, so operators get the same escape hatch on both delivery modes.

- **Fixed — "the test only asserts defaults are present"**: added `test_fresh_agent_config_ctmpl_embeds_non_default_topology_values` (`src/commands/service/local_config.rs`). It sets `agent_server = "https://step-ca.example.org:9443/acme/acme/directory"`, `agent_responder_url = "http://responder.internal:18080"`, `agent_email = "ops@example.org"` on the resolved struct, runs through the same baseline-render → managed-profile upsert → ctmpl pipeline the production path uses, and asserts both (a) the caller-supplied values appear in the rendered `.ctmpl` and (b) the built-in localhost defaults do NOT appear. This fails in the old implementation and passes now. The pre-existing default-path test is kept alongside it so the default-topology scenario stays covered.

Additional notes:

- I considered persisting `agent_server` / `agent_responder_url` in `StateFile` (init-time source of truth), but the state schema currently has no such fields and those URLs are logically service-level (one control plane may legitimately serve services with different responder admin endpoints). CLI flags per `service add` — matching the remote-bootstrap shape — felt like the least invasive option and kept scope aligned with #549.
- Updated the CHANGELOG entry to call out the new flags alongside the renderer extraction.

The branch is pushed (`d11a25b`); CI is running.

sehkone:
**[Reviewer Round 2]**

My Round 1 blockers on the local-file path look resolved: the shared baseline renderer now carries the full `[acme]` section, and the new non-default-topology test covers the overwrite case I was worried about.

I do still see one new blocking issue introduced by the new CLI flags:

- `bootroot service add` now accepts `--agent-email`, `--agent-server`, and `--agent-responder-url` on the shared `service add` CLI surface (`src/cli/args.rs:786`, `src/cli/args.rs:793`, `src/cli/args.rs:800`), and `resolve_service_add_args()` stores those values on `ResolvedServiceAdd` regardless of delivery mode (`src/commands/service/resolve.rs:163`, `src/commands/service/resolve.rs:167`, `src/commands/service/resolve.rs:171`). But the `remote-bootstrap` path still ignores them: `build_artifact()` hard-codes `DEFAULT_AGENT_EMAIL`, `DEFAULT_AGENT_SERVER`, and `DEFAULT_AGENT_RESPONDER_URL` into `bootstrap.json` (`src/commands/service/remote_bootstrap.rs:149`, `src/commands/service/remote_bootstrap.rs:150`, `src/commands/service/remote_bootstrap.rs:152`), and `write_remote_bootstrap_artifact()` never threads the resolved override values through (`src/commands/service/remote_bootstrap.rs:173`). So `bootroot service add --delivery-mode remote-bootstrap --agent-server https://step-ca.example.org:9443/... --agent-responder-url http://responder.internal:18080 ...` will happily accept those flags, then emit an artifact that falls back to the localhost defaults. Because the printed follow-up command is `bootroot-remote bootstrap --artifact ...`, the operator also has no obvious place to re-supply the intended values later. I think this needs to be fixed before merge either by plumbing the resolved `agent_*` values into the remote-bootstrap artifact as well, or by rejecting these flags for `remote-bootstrap` mode so the CLI does not silently ignore them. A regression test on the artifact path would help lock that down.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for catching this — you're right that the Round 1 flags only flowed into the local-file renderer and silently dropped under `--delivery-mode remote-bootstrap`.

- **Fixed** — *"remote-bootstrap still ignores the new `--agent-*` flags, emitting defaults into `bootstrap.json`"*: `build_artifact()` in `src/commands/service/remote_bootstrap.rs` no longer hard-codes `DEFAULT_AGENT_EMAIL` / `DEFAULT_AGENT_SERVER` / `DEFAULT_AGENT_RESPONDER_URL`. It now takes three `&str` parameters, and both call sites thread the resolved values through:
  - `write_remote_bootstrap_artifact()` passes `resolved.agent_email`, `resolved.agent_server`, `resolved.agent_responder_url`.
  - `write_remote_bootstrap_artifact_from_entry()` now also takes `&ResolvedServiceAdd` and passes the same resolved values, so idempotent re-runs (`run_service_add_remote_idempotent` in `src/commands/service.rs:440`) pick up whatever flags the operator supplied on the current `service add` invocation rather than falling back to the compose defaults. The one caller of `write_remote_bootstrap_artifact_from_entry` already had `resolved` in scope, so this was a one-line plumbing change at that call site.

  I considered the alternative of rejecting the flags when `--delivery-mode remote-bootstrap`, but plumbing-through is the more coherent option: (a) the `remote-bootstrap` CLI (`bootroot-remote bootstrap`) already exposes the same three flags and its `RemoteBootstrapArtifact` schema already declares `agent_email` / `agent_server` / `agent_responder_url`, so there's no schema change; (b) the documented override escape hatch for non-default topologies now matches semantically across both delivery modes; (c) the reviewer's "or reject them" alternative would have been more surprising to an operator who sees the flags accepted on one mode and rejected on the other.

- **Fixed** — *"a regression test on the artifact path would help lock that down"*: added `build_artifact_embeds_non_default_agent_overrides` in `src/commands/service/remote_bootstrap.rs`. It calls `build_artifact()` with `OVERRIDE_EMAIL = "ops@example.org"`, `OVERRIDE_SERVER = "https://step-ca.example.org:9443/acme/acme/directory"`, `OVERRIDE_RESPONDER = "http://responder.internal:18080"` and asserts (a) those exact strings land in the artifact's `agent_email` / `agent_server` / `agent_responder_url` fields and (b) `DEFAULT_AGENT_SERVER` / `DEFAULT_AGENT_RESPONDER_URL` do *not* leak through when overrides are supplied. This test fails against the old `build_artifact()` implementation and passes now. I also extended `build_artifact_typical_case` to assert the three new parameters flow through in the baseline scenario. The 13 other pre-existing `build_artifact` test call sites pass the same `TEST_AGENT_*` constants via a `replace_all`, keeping those coverage points honest about which values are being exercised.

Additional notes:

- No schema bump on `RemoteBootstrapArtifact.schema_version` — the fields (`agent_email`, `agent_server`, `agent_responder_url`) already existed on the struct and were already declared `Option<String>` on the `bootroot-remote bootstrap` parser side (`src/bin/bootroot-remote/main.rs:305-311`). This is a value-filling change, not a shape change, so existing artifact parsers keep working unmodified.
- Updated `docs/en/cli.md` and `docs/ko/cli.md` so the `remote-bootstrap` section no longer claims the only escape hatch is to hand-edit `bootstrap.json` — the overrides at `service add` time are now the first-class path, and hand-editing is framed as the fallback.
- Updated the CHANGELOG entry for #549 to note the remote-bootstrap plumbing as well.

Branch pushed (`b552ff6`); CI is running.

sehkone:
**[Reviewer Round 3]**

The Round 2 item I raised is fixed for the current invocation: the `remote-bootstrap` artifact now does carry the `--agent-*` overrides you pass to `bootroot service add`. I do still see one blocking edge case in the new rerun path, though:

- The override values are still transient CLI-only data. `resolve_service_add_args()` falls back to the localhost defaults whenever `--agent-email` / `--agent-server` / `--agent-responder-url` are omitted (`src/commands/service/resolve.rs:163`), `ServiceEntry` does not persist any of them (`src/state.rs:115`), and `is_idempotent_remote_rerun()` does not consider them when deciding whether an existing `remote-bootstrap` service matches the new invocation (`src/commands/service.rs:672`). But `write_remote_bootstrap_artifact_from_entry()` now regenerates `bootstrap.json` from `resolved.agent_*` on that idempotent rerun path (`src/commands/service/remote_bootstrap.rs:197`, `src/commands/service/remote_bootstrap.rs:221`). That means an operator can add a service with `--agent-server https://step-ca.example.org:9443/... --agent-responder-url http://responder.internal:18080`, then later rerun the same `bootroot service add --delivery-mode remote-bootstrap ...` command without restating those flags and silently get a new artifact back on the localhost defaults again. The inverse is also true: passing different `--agent-*` values on a rerun silently changes the generated artifact even though those values are not part of the stored service definition. I think this needs a real source of truth before merge: either persist the `agent_*` overrides in `ServiceEntry` and include them in the idempotence comparison, or reject/require them on rerun so the CLI cannot silently drift. A regression test around `run_service_add_remote_idempotent` would help pin this down.

sehkone:

sehkone:
**[Author Round 3]**

Thanks for catching that the CLI-only override values could silently drift an idempotent rerun away from the stored service definition. Addressing:

- **Fixed** — *"the override values are still transient CLI-only data"*: the resolved `agent_email` / `agent_server` / `agent_responder_url` are now `Option<String>` (not eagerly coalesced to the compose defaults in `resolve_service_add_args`) and are persisted on `ServiceEntry` in `state.json` via three new `Option<String>` fields. They carry `#[serde(default, skip_serializing_if = "Option::is_none")]` so pre-existing entries written before this change deserialize as `None` — behaviour-equivalent to the old "fall back to compose defaults at render time" path, so no migration is required.

- **Fixed** — *"`is_idempotent_remote_rerun` does not consider them"*: `non_policy_fields_match` now compares `entry.agent_email == resolved.agent_email` (and the other two), so a rerun that drops, adds, or changes any of the three overrides is no longer idempotent. Combined with the existing `run_service_add` fall-through, that means such a rerun is rejected as a duplicate rather than silently regenerating `bootstrap.json` from a different topology. To change topology, the operator removes and re-adds the service.

- **Fixed** — *"`write_remote_bootstrap_artifact_from_entry()` regenerates `bootstrap.json` from `resolved.agent_*`"*: the function now reads the persisted entry values (`entry.agent_email` etc.) through a trio of `effective_agent_*(Option<&str>) -> &str` helpers that fall back to the constants. The `resolved` parameter is removed from its signature — the entry is now the single source of truth on the idempotent-rerun path. The initial-add path (`write_remote_bootstrap_artifact`) still reads from `resolved` (that is the only place the values come from on a fresh add), and `build_service_entry_from_role` copies them from `resolved` into the entry the caller immediately persists. So the round-trip is: resolved → entry (persist) → entry (future reruns). This also means an operator who runs `service add` once with overrides and then reruns without restating them gets `bootstrap.json` regenerated from the original overrides, which is the behaviour the Round 3 comment pushed for.

- **Fixed** — *"a regression test around `run_service_add_remote_idempotent` would help pin this down"*: added `is_idempotent_remote_rerun_false_when_agent_overrides_differ` in `src/commands/service.rs`. It exercises all three mismatch shapes at the idempotence-predicate level:
  - stored `Some(server)` vs. rerun `None` (drops the override)
  - stored `None` vs. rerun `Some(responder_url)` (introduces an override)
  - stored `Some(email_a)` vs. rerun `Some(email_b)` (disagrees)

  Each must return `false` from `is_idempotent_remote_rerun`, which is the decision point `run_service_add` then uses to take the `error_service_duplicate` bail instead of calling into `run_service_add_remote_idempotent`. The baseline case (both `None`) is also asserted to still be idempotent, so existing no-override workflows are unaffected. This fails on the old implementation and passes now. I also extended `assert_common_fields` to assert the three new fields round-trip from `resolved` through `build_service_entry_from_role` into `ServiceEntry`.

Additional notes:

- I chose the "persist + reject on mismatch" option rather than "require the flags on every rerun" because the remote-bootstrap rerun path is already explicitly designed to be idempotent against a prior stored state (it's how operators re-emit `bootstrap.json` after a wrap-token expires or a CA rotates) — requiring the operator to restate the overrides each time would break that ergonomic. With persistence in place the rerun recovers the exact artifact from state, which matches the spirit of the existing idempotence check.
- No `state.json` schema bump: the three new fields are additive, optional, and serialize only when present. An operator whose `state.json` pre-dates this change continues to work; entries written after this change simply carry the extra keys when the operator supplied overrides.
- Updated `docs/en/cli.md` and `docs/ko/cli.md` with a paragraph explaining the persistence semantics under `--delivery-mode remote-bootstrap`.
- Extended the CHANGELOG `#549` entry to call out the persistence + idempotence-comparison change.
- `ResolvedServiceAdd` test fixtures across `service.rs`, `local_config.rs`, `agent_start.rs`, `rotate/helpers.rs`, `dns_alias.rs`, and `state.rs` now default the three fields to `None`. Production-path fixtures that needed explicit override values set them explicitly (`Some(...)`) in the new tests.

Branch pushed (`d35b216`); CI is running.

sehkone:
**[Reviewer Round 4]**

The Round 3 remote-bootstrap follow-up looks resolved: the `agent_*` values are now persisted on `ServiceEntry`, compared during idempotence checks, and `write_remote_bootstrap_artifact_from_entry()` reads from the stored entry instead of the transient CLI args (`src/state.rs:141`, `src/commands/service.rs:696`, `src/commands/service/remote_bootstrap.rs:198`). I do still see one blocking local-file gap, though:

- `apply_local_service_configs()` only seeds the shared baseline when `resolved.agent_config` does not already exist (`src/commands/service/local_config.rs:58`). If `--agent-config` points at an existing file that lacks `email`, `server`, or `[acme].http_responder_url`, the new `--agent-email` / `--agent-server` / `--agent-responder-url` flags are effectively ignored on the local-file path: the code reads the file verbatim, then only upserts `[trust]`, top-level `domain`, and `acme.http_responder_hmac` (`src/commands/service/local_config.rs:76`, `src/commands/service/local_config.rs:78`, `src/commands/service/local_config.rs:80`). In that case the emitted `agent.toml.ctmpl` still omits the topology fields, so the original footgun remains for operators who start from a pre-existing `agent.toml` instead of a brand-new file. This also makes the new flag docs misleading for that path (`src/cli/args.rs:786`). I think the local-file renderer needs to upsert the full baseline fields/tunables into both fresh and pre-existing agent configs, or explicitly reject existing configs that do not already carry them.
- The new tests only cover the fresh-file path (`src/commands/service/local_config.rs:626`), so this existing-file case would still pass today. Please add a regression test that starts from a pre-existing `agent.toml` missing `server` / `http_responder_url`, passes non-default `--agent-*` values, and asserts the generated `.ctmpl` contains those values.

sehkone:

sehkone:
**[Author Round 4]**

Thanks — the Round 3 idempotence/state plumbing still looks right, so this round is focused on the remaining local-file gap.

- **Fixed**: `apply_local_service_configs` now backfills missing baseline fields into pre-existing `agent.toml` files too, and applies `--agent-*` overrides regardless of whether the file was fresh. Added `toml_util::insert_missing_top_level_keys` / `insert_missing_section_keys` as generic "insert only if absent" primitives, then wrapped them in a new `trust_bootstrap::apply_agent_config_baseline_defaults` helper that enumerates every field `render_agent_config_baseline` would emit (email, server, domain, the full `[acme]` tunables, `http_responder_url`, `http_responder_hmac`). The existing-file branch no longer skips this step — it reads the file, calls the backfill helper (preserving any operator-customised value), then upserts the three `--agent-*` CLI values so they always take effect, then continues with the existing managed-profile / `[trust]` / `domain` / `http_responder_hmac` upserts. Net effect: `--agent-config /abs/config/review-agent.toml` pointing at a file missing topology fields now gets the CLI-supplied values baked into the `.ctmpl`, and an operator who already had `server = "custom"` in the file keeps that value. `src/commands/service/local_config.rs:58-100`, `src/trust_bootstrap.rs:63-105`, `src/toml_util.rs:49-107`.

- **Fixed**: added two regression tests so the pre-existing path is pinned down. `test_existing_agent_config_ctmpl_backfills_missing_fields_from_overrides` starts from a file containing only `domain` + `[acme].http_responder_hmac`, passes `--agent-email ops@example.org --agent-server https://step-ca.example.org:9443/... --agent-responder-url http://responder.internal:18080`, and asserts those exact values appear in the rendered `.ctmpl` while `DEFAULT_AGENT_SERVER` / `DEFAULT_AGENT_RESPONDER_URL` do not. `test_existing_agent_config_preserves_operator_customised_topology` exercises the inverse: pre-existing file carrying operator-customised `email` / `server` / `http_responder_url`, no CLI overrides — asserts those values survive the backfill untouched. `src/commands/service/local_config.rs:714-795`, `src/commands/service/local_config.rs:800-860`.

Also updated CHANGELOG to describe the Round 4 change (`insert_missing_*` / `apply_agent_config_baseline_defaults` helpers, pre-existing file backfill semantics, `--agent-*` override precedence).

Full test suite and quality-check preflight pass locally; CI re-run in progress.

sehkone:
**[Reviewer Round 5]**

The Round 4 local-file gap looks resolved to me: `apply_local_service_configs()` now backfills the baseline for both fresh and pre-existing `agent.toml` files, applies explicit `--agent-*` overrides afterwards, and the two new pre-existing-file tests in `src/commands/service/local_config.rs` cover the cases I called out last round. I’m not carrying those items forward.

One new follow-up I did notice while reviewing the updated diff:

- `bootroot-remote bootstrap` still has the same “existing file ignores the baseline” hole that local-file had before Round 4. In `src/bin/bootroot-remote/agent_config.rs:37-95`, the existing-file branch reads `agent.toml` verbatim and then only upserts `acme.http_responder_hmac`, `[trust]`, and the managed profile. The artifact’s `agent_email` / `agent_server` / `agent_responder_url` are only used in the `NotFound` branch via `render_agent_config_baseline()` (`src/bin/bootroot-remote/agent_config.rs:39-40`, `src/bin/bootroot-remote/agent_config.rs:249-255`). So if a remote target already has an `agent.toml` that is missing `server` or `[acme].http_responder_url`, `bootroot-remote bootstrap --artifact ...` still drops the baked topology values on the floor and keeps re-rendering without them. That feels especially surprising now that the new docs say those values are baked in under both delivery modes (`docs/en/cli.md:1272-1282`). If parity across the two paths is part of the intended end state, I think `apply_agent_config_updates()` needs the same baseline-backfill / explicit-override treatment as `apply_local_service_configs()`, plus a regression test for the pre-existing-file path.

sehkone:

sehkone:
**[Author Round 5]**

Thanks — the Round 4 local-file gap still looks closed on re-read, so this round is focused on the remaining `bootroot-remote bootstrap` existing-file hole.

- **Fixed** — *"`bootroot-remote bootstrap` still has the same existing-file baseline hole that local-file had before Round 4"*: `apply_agent_config_updates()` now applies the same baseline-backfill + artifact-override pipeline that `apply_local_service_configs()` got last round. The `NotFound` branch no longer short-circuits through `render_agent_config_baseline` — instead both branches start from either the file contents or an empty string, then pass through `trust_bootstrap::apply_agent_config_baseline_defaults` (inserts only if missing), then upsert explicit artifact-carried overrides via a new `apply_agent_overrides` helper, then continue with the existing `[acme].http_responder_hmac` / `[trust]` / managed-profile upserts. Net effect: a remote target that already had an `agent.toml` missing `email` / `server` / `[acme].http_responder_url` now gets those keys backfilled from the artifact, so KV re-renders from the regenerated `.ctmpl` stop reverting to bootroot-agent's compiled-in defaults. `src/bin/bootroot-remote/agent_config.rs:31-130`.

  The override signal is new: `ResolvedBootstrapArgs` gained `agent_email_override` / `agent_server_override` / `agent_responder_url_override` as `Option<String>`, populated from the artifact's existing `Option<String>` fields (`src/bin/bootroot-remote/main.rs:186-193`, `src/bin/bootroot-remote/main.rs:432-448`). When the upstream `bootroot service add --agent-*` persisted an override, the artifact carries it, the override is `Some`, and the upsert wins over any pre-existing value. When no override is carried (artifact absent or field absent), the backfill-only path preserves operator-customised values — matching the conservative semantics the local-file path already has, where the absence of a CLI `--agent-*` flag means "don't clobber the operator's pre-existing value". I considered detecting explicit direct-CLI `--agent-server` on `bootroot-remote bootstrap` via `clap::ArgMatches::value_source`, but the artifact signal already covers the `service add`-originated propagation path (the primary user surface), and the `clap::ArgMatches` plumbing would be a larger restructuring for little gain — direct-CLI users against a pre-existing file can still hand-edit or remove-and-recreate to change topology.

- **Fixed** — *"please add a regression test that starts from a pre-existing `agent.toml` missing `server` / `http_responder_url`, passes non-default `--agent-*` values, and asserts the generated output contains those values"*: added `existing_agent_config_backfills_overrides_from_artifact` in `src/bin/bootroot-remote/agent_config.rs`. It constructs a pre-existing `agent.toml` that only has `domain` and an `[acme]` section (no `email` / `server` / `http_responder_url`), sets the three override fields to `ops@example.org` / `https://step-ca.example.org:9443/acme/acme/directory` / `http://responder.internal:18080`, runs the same in-memory pipeline `apply_agent_config_updates` uses, and asserts (a) the override values appear in the output, (b) the `[acme]` retry/timeout tunables get backfilled too, (c) the legacy `http_responder_hmac` gets rotated out by the upsert. This fails against the old implementation (overrides are dropped) and passes now.

  Added a companion regression `existing_agent_config_preserves_operator_values_when_no_override`: pre-existing file already has operator-customised `email` / `server` / `http_responder_url`, artifact carries no overrides (`agent_*_override == None`), asserts those operator values survive the backfill path untouched and that the defaults do NOT leak in. This pins down the conservative "backfill is insert-if-missing, never clobber" guarantee on the remote path — the same invariant `test_existing_agent_config_preserves_operator_customised_topology` pins down on the local-file path.

Additional notes:

- Updated `docs/en/cli.md` and `docs/ko/cli.md` so the `bootroot-remote bootstrap` section documents the two pre-existing-file branches: "backfill only" when no override is in the artifact, "backfill + override" when the artifact carries explicit values.
- Extended the CHANGELOG `#549` entry to call out the remote-bootstrap existing-file path alongside the local-file existing-file path.
- `ResolvedBootstrapArgs` is a test-construction point in three `#[cfg(test)]` modules (`agent_config.rs`, `bootstrap.rs`, `main.rs`); extended the two non-test fixtures that build `ResolvedBootstrapArgs` (in `agent_config.rs` and `bootstrap.rs`) to default the three new fields to `None`. The `main.rs` test fixtures build `BootstrapArgs` (the CLI-parsed shape), which is unchanged.
- Clippy and `cargo fmt --check` pass; full `cargo test --lib --bins` is green (178 + 518 + 2 + 22 + 46 tests). `scripts/preflight/ci/check.sh` also passes locally.

Branch pushed (`c0e3d61`); CI is running.

sehkone:
**[Reviewer Round 6]**

The Round 5 existing-file hole in `apply_agent_config_updates()` itself looks fixed to me, so I’m not carrying that item forward. I did find one remaining correctness gap in the artifact plumbing, though:

- The new `agent_*_override` path never actually sees `None` for artifact-driven bootstraps. `RemoteBootstrapArtifact` still serializes `agent_email`, `agent_server`, and `agent_responder_url` as required strings, and `build_artifact()` always fills them via `effective_agent_*`, even when `service add` had `resolved.agent_* == None` (`src/commands/service/remote_bootstrap.rs:49`, `src/commands/service/remote_bootstrap.rs:152`, `src/commands/service/remote_bootstrap.rs:221`). Then `resolve_bootstrap_args()` treats mere field presence as an explicit override by setting `agent_*_override = artifact.agent_*` (`src/bin/bootroot-remote/main.rs:440`). So an artifact produced with no `--agent-*` flags still arrives at `apply_agent_config_updates()` with `agent_*_override == Some(default)`, and the new override step will clobber a pre-existing remote `agent.toml` back to `admin@example.com` / `https://localhost:9000/...` / `http://127.0.0.1:8080` instead of taking the intended backfill-only path. That means the new “preserve operator values when there was no explicit override” behavior is still unreachable for the real `--artifact` flow.
- Related test gap: `existing_agent_config_preserves_operator_values_when_no_override` hand-constructs `ResolvedBootstrapArgs` with the override fields left at `None` (`src/bin/bootroot-remote/agent_config.rs:752`), but the production artifact path above never produces that shape today. I think this needs either an artifact schema that preserves override presence explicitly (for example optional `agent_*` fields that are omitted when the operator never passed `--agent-*`) or a resolver-level regression test that round-trips a no-override `bootstrap.json` and proves `agent_*_override` stays `None`.

sehkone:

sehkone:
**[Author Round 6]**

Thanks — you were right that the `agent_*_override` path never actually saw `None` for artifact-driven bootstraps. The Round 5 "preserve operator values" path was unreachable for any real `--artifact` flow, which is the specific hole you flagged.

- **Fixed** — *"`RemoteBootstrapArtifact` still serializes `agent_email` / `agent_server` / `agent_responder_url` as required strings, and `build_artifact()` always fills them via `effective_agent_*`, even when `resolved.agent_* == None`"*: the three fields on `RemoteBootstrapArtifact` are now `Option<String>` with `#[serde(skip_serializing_if = "Option::is_none")]`, and `build_artifact()` takes `Option<&str>` instead of `&str` for them (`src/commands/service/remote_bootstrap.rs:49-61`, `src/commands/service/remote_bootstrap.rs:124-126`, `src/commands/service/remote_bootstrap.rs:157-161`). Both call sites (`write_remote_bootstrap_artifact` and `write_remote_bootstrap_artifact_from_entry`) now pass the raw `resolved.agent_*.as_deref()` / `entry.agent_*.as_deref()` — no more `effective_agent_*` coercion at the artifact writer. So an operator who ran `service add` without `--agent-*` flags gets an artifact whose JSON omits all three keys, and the downstream `resolve_bootstrap_args()` parse yields `agent_*_override == None`, which is exactly what the Round 5 backfill-only path expects.

  The `effective_agent_*` helpers are retained — they're still used by the local-file renderer where we *do* need a concrete string to seed the `.ctmpl`. I intentionally did not apply them at the artifact writer; that was the whole root cause. The CLI-default fallback is now only applied at the two places where a concrete value is actually needed (local-file `.ctmpl` rendering and the `bootroot-remote` CLI `--agent-*` flag default), not at the serialization boundary.

  I chose the "optional fields, omit when absent" route over the "explicit `override: bool` sidecar field" route you mentioned as an alternative: (a) it matches the shape the downstream parser already declares (`#[serde(default)] Option<String>` on `BootstrapArtifact`, `src/bin/bootroot-remote/main.rs:312-319`), so no deserializer change was needed, and (b) it keeps the JSON schema self-describing — absence of a key *means* "no override". I did not bump `schema_version` because removing a required field is technically a breaking change for a strict parser, but the parser already tolerates missing keys via `#[serde(default)]`, and the change can only land `None` where an older producer would have landed `Some(default-string)` — existing bootroot-remote binaries read both shapes identically for the override path.

- **Fixed** — *"a resolver-level regression test that round-trips a no-override `bootstrap.json` and proves `agent_*_override` stays `None`"*: added `resolve_bootstrap_args_no_override_keeps_override_none` in `src/bin/bootroot-remote/main.rs`. It writes an artifact JSON that explicitly omits `agent_email` / `agent_server` / `agent_responder_url` (the shape `bootroot service add` now emits when no `--agent-*` flags were supplied), parses it through `resolve_bootstrap_args`, and asserts all three `agent_*_override` fields stay `None`. It also asserts the baseline fields `agent_email` / `agent_server` / `agent_responder_url` fall back to the CLI defaults — which is the right behaviour for a fresh-file seed where we need *some* value. Fails against the old producer (which would have written the defaults into the artifact, making `agent_*_override = Some(default)` and the first assertion fail); passes now.

  Added a companion `resolve_bootstrap_args_override_propagates_from_artifact` that round-trips the inverse: an artifact that DOES carry explicit overrides parses with `agent_*_override == Some(value)`, so the downstream `apply_agent_config_updates` knows to clobber pre-existing remote values.

  Also added `build_artifact_omits_agent_keys_when_no_override` on the producer side (`src/commands/service/remote_bootstrap.rs`) that pins down both the struct-level omission (`Option::is_none`) and the serialized-JSON-level omission (no `"agent_email":` substring in the output). Together the two tests nail down both ends of the JSON boundary.

Additional notes:

- Updated the CHANGELOG `#549` entry to describe the serialization change alongside the Round 5 backfill pipeline.
- The `render_remote_run_command_legacy` test-only helper was adjusted to `as_deref().unwrap_or("")` for the three fields — it's `#[cfg(test)]`-gated and only exercises the hook-flag assertions; the agent flags in the legacy format remain emitted as empty strings when the artifact has no overrides, which is the same shape `bootroot-remote bootstrap` would see if those flags were passed empty on its CLI (the per-field flags all have defaults, so the legacy format stays functional).
- Six pre-existing `build_artifact` test call sites were updated via `replace_all` to pass `Some(TEST_AGENT_EMAIL)` etc. instead of the bare constants; no coverage changes, just signature adaptation.
- Clippy, `cargo fmt --check`, and `cargo test --lib --bins` all pass locally (178 + 519 + 2 + 22 + 48 tests).

Branch pushed (`166733c`); CI is running.

sehkone:
**[Reviewer Round 7]**

The Round 6 override-path hole looks fixed to me: `RemoteBootstrapArtifact` now omits `agent_*` when no service-add override was supplied, and the new resolver tests cover both the omitted and explicit cases. I only found one follow-up issue this round:

- `bootstrap.json`'s wire format changed without a `schema_version` bump. `RemoteBootstrapArtifact` now serializes `agent_email`, `agent_server`, and `agent_responder_url` as optional/omitted keys (`src/commands/service/remote_bootstrap.rs:54`, `src/commands/service/remote_bootstrap.rs:56`, `src/commands/service/remote_bootstrap.rs:59`), and the producer still labels that payload as `schema_version: 2` (`src/commands/service/remote_bootstrap.rs:147`). The integration expectation changed accordingly and now treats those keys as absent under schema 2 (`tests/bootroot_service.rs:739`). But the schema contract right above the struct says the version is bumped whenever the artifact changes in a way that could break existing parsers (`src/commands/service/remote_bootstrap.rs:23`). A strict external consumer that learned “v2 always has these strings” now has no way to distinguish the new shape from the old one. I think this needs to be versioned as a new schema and `bootroot-remote` updated to accept it (`src/bin/bootroot-remote/main.rs:291`), even if the in-repo parser already tolerates both shapes.

sehkone:

sehkone:
**[Author Round 7]**

- **Schema-version bump for the optional `agent_*` keys** — **Fixed**. Agreed: the Round 6 change flipped `agent_email` / `agent_server` / `agent_responder_url` from required strings to optional-and-omittable, which is a wire-format change that a strict external consumer pinned to "v2 always has these strings" could not detect. Per the contract doc right above `RemoteBootstrapArtifact`, that requires a bump.
  - Producer: `RemoteBootstrapArtifact.schema_version` is now `3` (`src/commands/service/remote_bootstrap.rs:147`).
  - Consumer: `MAX_SUPPORTED_SCHEMA_VERSION` bumped from `2` to `3` in `src/bin/bootroot-remote/main.rs`, so the binary now accepts `1..=3`. Parsing v2 artifacts still works because `BootstrapArtifact` already declares the three keys with `#[serde(default)] Option<String>`, so pre-existing v2 artifacts on disk (and the v2 fixtures in the Round 6 resolver tests) continue to bootstrap.
  - Test updates: integration assertion `tests/bootroot_service.rs:739` moved to `3`; e2e fixtures in `tests/e2e_multi_host_tls.rs:780` and `tests/e2e_multi_host_tls_real_daemon.rs:510` moved to `3` to match the new producer shape; the "rejects unknown future schema" test in `src/bin/bootroot-remote/main.rs` now uses `schema_version: 4` (was `3`) since `3` is now accepted; the Round 6 resolver tests were also moved to `3` so they reflect the current production shape while MIN=1 preserves v2 backwards-compat.
  - PR description: added a Round 7 bullet explaining the bump and updated the test plan to include the updated rejection test.

`cargo clippy --all-targets -- -D warnings` and `cargo test --lib --bins` pass locally after the change.

sehkone:
**[Reviewer Round 8]**

Thanks — the Round 7 schema-version issue looks properly fixed.

I verified that the producer now emits `schema_version: 3` in `src/commands/service/remote_bootstrap.rs:147`, the consumer accepts `1..=3` in `src/bin/bootroot-remote/main.rs:287` and still deserializes the optional `agent_*` fields compatibly in `src/bin/bootroot-remote/main.rs:312`, and the affected expectations/tests were updated consistently in `tests/bootroot_service.rs:738`, `tests/e2e_multi_host_tls.rs:779`, `tests/e2e_multi_host_tls_real_daemon.rs:509`, and `src/bin/bootroot-remote/main.rs:947`.

I also re-read the updated diff against #549 as a whole. The shared baseline renderer/backfill path, persisted `--agent-*` overrides, remote artifact omission semantics, and the new regression coverage all line up with the bug report now. I don’t have any remaining blockers this round.

sehkone:
[Review Verdict Round 8: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Emit full agent.toml baseline across service add
```

**Body**

```text
The local-file `bootroot service add` path wrote
`agent.toml.ctmpl` with only `domain`, `[trust]`,
`[acme].http_responder_hmac`, and the managed profile,
relying on bootroot-agent's compiled-in defaults for
`server`, `email`, `[acme].http_responder_url`, and the
retry/timeout tunables. Operators on non-default topologies
had to hand-edit `agent.toml`, and the next KV rotation
silently clobbered those edits when the OpenBao Agent
sidecar re-rendered from the template.

Extract `render_agent_config_baseline` into
`trust_bootstrap` so local-file and remote-bootstrap share
one renderer, and emit the full baseline in the
fresh-template path. Add `--agent-email`, `--agent-server`,
and `--agent-responder-url` flags to `service add` so
operators can bake non-default topology values into the
template at service-add time; defaults match the existing
compose-topology values, so same-host setups are
unaffected.

Plumb the resolved values through the rest of the flow:

- Persist the optional overrides on `ServiceEntry` in
  `state.json` and include them in the remote-bootstrap
  idempotence comparison, so reruns regenerate the
  artifact from stored values and mismatched reruns are
  rejected instead of silently drifting.
- Make `RemoteBootstrapArtifact.agent_{email,server,
  responder_url}` `Option<String>` with
  `skip_serializing_if = Option::is_none`, so the "no
  explicit override" signal survives the JSON boundary.
  Bump `schema_version` to 3; `bootroot-remote` now
  accepts 1..=3.
- Add `apply_agent_config_baseline_defaults`, a
  toml-util-backed helper that backfills missing baseline
  keys in pre-existing `agent.toml` files on both the
  local-file and remote upsert paths without clobbering
  operator-customised values. Explicit overrides from the
  artifact or CLI are then upserted on top.

Closes #549
```

Issue #578, opened 2026-05-01
Rename 'service agent start' to 'service openbao-sidecar start' and add E2E coverage for both OpenBao Agent deployment variants


sehkone:
## Background

PR #567 added `bootroot service agent start` as the recommended way to bring up a per-service OpenBao Agent. While verifying #569 we found two adjacent gaps that are best fixed together because both touch the same CLI surface:

1. The command name `service agent start` is ambiguous on two axes (which software, which deployment pattern), inviting confusion with the unrelated `bootroot-agent` certificate daemon.
2. E2E does not exercise the command at all, and does not exercise the host-daemon alternative either. Both deployment variants of the OpenBao Agent are therefore untested in CI.

This issue groups the rename and the E2E gap because the rename will rewrite every reference to the command anyway — combining them keeps CLI and tests internally consistent in a single PR.

### Why rename

Two unrelated software packages share the word "agent" in this project:

| Software | Role | Lifecycle owner |
|---|---|---|
| OpenBao Agent (`bao agent`) | Fetches secrets from OpenBao, renders templates | bootroot (sidecar) or operator (host daemon) |
| `bootroot-agent` | Issues/renews TLS certs against step-ca | Operator (host daemon, separate binary) |

A naive reader of `bootroot service agent start` is more likely to think "this starts `bootroot-agent` for the service" — exactly the opposite of what it does.

The current name is also under-specified on the second axis: it manages only the **sidecar container** deployment of OpenBao Agent, not the host-daemon alternative. `agent` alone gives no hint of that constraint.

### Proposed name

```
bootroot service openbao-sidecar start
```

This resolves both axes:

| Axis | `service agent` | `service openbao-sidecar` |
|---|---|---|
| Which software | Ambiguous (clashes with `bootroot-agent`) | Clear (`openbao` = OpenBao Agent) |
| Which deployment pattern | Unspecified | Clear (`sidecar`) |

Other benefits:

- Aligns with internal code vocabulary already in use (`/sidecar-config`, `/sidecar-certs`, `daemon_sidecar_bind_mounts`, the next-steps text from `service add` already says "managed sidecar").
- Leaves room for a future `service openbao-daemon` subcommand if host-daemon lifecycle management ever becomes a bootroot responsibility. Today that would not be possible under the `agent` umbrella without renaming again.

### Considered alternatives

| Candidate | Why not |
|---|---|
| `service agent` (current) | The reason for this issue. |
[...]

PR #580, opened 2026-05-01
Discover compose project for service agent start (#577)

### LINKED ISSUE #577: service agent start hardcodes 'bootroot_default' network, breaking non-standard project names
## Summary

`bootroot service agent start` only works when the docker compose project name is exactly `bootroot`. From any other directory (worktrees, custom paths) or any deployment that sets a non-default `COMPOSE_PROJECT_NAME`, the sidecar fails to come up with:

```
network bootroot_default declared as external, but could not be found
bootroot service agent start failed
```

Discovered while running the manual verification checklist for #569. The PR #567 changes themselves work correctly under the standard project name; this is a pre-existing limitation that the new `service agent start` flow inherits and exposes.

## Root cause

[`src/commands/service/agent_start.rs`](https://github.com/aicers/bootroot/blob/main/src/commands/service/agent_start.rs) hardcodes **two** values that depend on the compose project being literally `bootroot`:

```rust
const COMPOSE_PROJECT: &str = "bootroot";
/// Docker network created by `docker compose -p bootroot`.
const COMPOSE_NETWORK: &str = "bootroot_default";
```

Both are written verbatim into the sidecar lifecycle. The network name appears in the per-service compose override:

```yaml
networks:
  default:
    name: bootroot_default
    external: true
```

The project name is forced on the `docker compose` invocation that brings the sidecar up:

```rust
let args = [
    "compose",
    "-p",
    COMPOSE_PROJECT,         // ← "bootroot"
    "-f", &compose_str,
    "-f", &override_str,
    "up", "-d",
    service_name,
];
```

`docker compose` derives the project name from the working directory by default, so the network actually created by `bootroot infra install` is `<dirname>_default` (e.g. `eloquent-solomon-19b6fe_default` from a worktree). The sidecar override looks up `bootroot_default` and fails — and even if the network match is fixed, `-p bootroot` would still attach the sidecar to a phantom `bootroot` project that does not contain the rest of the infrastructure.

Fixing only `COMPOSE_NETWORK` is therefore insufficient. Both constants must be replaced together, and the `-p` argument either dropped or sourced from the same discovery result. Otherwise the sidecar comes up under a project label that does not match the rest of the deployment and `docker compose ps` from the operator's working directory will not list it.

## Why CI did not catch this

The lifecycle e2e scripts (`scripts/impl/run-local-lifecycle.sh`) bypass `bootroot service agent start` entirely and bring the sidecar up via:

```bash
docker run -d --name "$container" \
  --network "container:bootroot-openbao" \
  ...
```

So the hardcoded project and network names are never exercised by the matrix.

## Reproduction

From a directory whose name is not `bootroot` (or with `COMPOSE_PROJECT_NAME` set to anything else), run the #569 manual checklist:

```bash
bootroot infra install --compose-file docker-compose.yml
bootroot init --enable auto-generate,db-provision ...
bootroot service add --service-name review --deploy-type daemon --delivery-mode local-file ...
bootroot service agent start --service-name review
# → error: network bootroot_default declared as external, but could not be found
```

Workaround: prepend `COMPOSE_PROJECT_NAME=bootroot` to every `bootroot` invocation.

## Proposed fix

Replace both hardcoded constants with runtime discovery. The single source of truth is the docker compose project label on the OpenBao container:

```bash
docker inspect bootroot-openbao \
  --format '{{index .Config.Labels "com.docker.compose.project"}}'
# → e.g. "bootroot", "myorg-prod", "eloquent-solomon-19b6fe"
```

That label drives both:

- The compose project passed to `docker compose -p <project> ...` (so the sidecar lands inside the same project as the rest of the infra).
- The external network name written into the per-service override: `<project>_default`.

Add an explicit override flag for non-standard deployments:

```
bootroot service agent start --service-name <name> [--openbao-network <name>]
```

When `--openbao-network` is provided, network discovery is skipped — but if OpenBao is present in the compose file, project label discovery still runs to populate `-p` correctly. The two discoveries are independent.

Surface a clear error when the OpenBao container is not found at all, or is found but has no `com.docker.compose.project` label. (`docker inspect` itself works on stopped containers and `docker compose up` can start dependencies via `depends_on`, so "not running" is not the right precondition.) Suggested error text: "`bootroot-openbao` container not found — run `bootroot infra install` first, or pass `--openbao-network` if OpenBao runs outside bootroot's compose."

### External-OpenBao branch (first-class case)

The current code already imports `compose_has_openbao` and `resolve_openbao_agent_addr`. The branch where OpenBao is **absent from bootroot's compose file** — operator runs OpenBao on a separate host, in kubernetes, as a managed service, etc. — must be designed deliberately, not handled as a fallback edge case:

- The `bootroot-openbao` container does not exist in this scenario, so label discovery cannot run.
- The sidecar still needs to be attached to a docker network reachable from `resolve_openbao_agent_addr`'s result.
- `--openbao-network` becomes the only way to specify that network, not just an escape hatch.
- Forcing label discovery would conflict with the existing external-OpenBao support that the `compose_has_openbao` branch already enables.

Decision logic to surface clearly in the design (rows 1 and 2 read the value from `compose_has_openbao`, which returns `true` when OpenBao is present in the compose file):

| OpenBao in compose | `--openbao-network` | Behavior |
|---|---|---|
| present | unset | Discover project label on `bootroot-openbao`; use `<project>` for `-p` and `<project>_default` for the override network |
| present | set | Use the given network in the override; still discover project label for `-p` |
| absent | unset | Error: `--openbao-network` is required when OpenBao runs outside bootroot's compose |
| absent | set | Use the given network; omit `-p` (or use working-directory default) — no project to discover |

Surface a clear error when `bootroot-openbao` cannot be found in the discovery branches, or when the container is found but has no `com.docker.compose.project` label.

## Impact

Operators following the documented manual flow from any path other than a directory literally named `bootroot` cannot bring up service sidecars. This affects the recommended next-step from `bootroot service add` (now also surfaced by PR #567).

## Documentation gap (related)

`bootroot service agent start` is essentially undocumented despite being the recommended next step from `bootroot service add`:

- [`docs/en/cli.md`](https://github.com/aicers/bootroot/blob/main/docs/en/cli.md) command index lists only `service add` / `service update` / `service info`. `service agent start` is missing. Existing OpenBao Agent / sidecar mentions in `docs/en/cli.md` and `docs/en/operations.md` need to be reconciled with the new dedicated section so the docs do not contradict each other.
- The only mention of the command itself is a single line embedded in the example next-steps output block in [`docs/en/cli-examples.md`](https://github.com/aicers/bootroot/blob/main/docs/en/cli-examples.md) and [`docs/ko/cli-examples.md`](https://github.com/aicers/bootroot/blob/main/docs/ko/cli-examples.md).
- No explanation of what the command does, what flags it takes, what container it creates (`bootroot-openbao-agent-<service>`), prerequisites (infra install must have run, `bootroot-openbao` must exist), or failure modes.
- Stale `--network bootroot_default` references survive in [`docs/en/installation.md`](https://github.com/aicers/bootroot/blob/main/docs/en/installation.md) and [`docs/ko/installation.md`](https://github.com/aicers/bootroot/blob/main/docs/ko/installation.md) (three occurrences each — `grep -n 'bootroot_default'` to locate). They must be updated to reflect the discovery behavior or removed.

Doc updates to land alongside the network-discovery fix:

### 1. Command reference

- Add `bootroot service agent start` to the command index in `docs/{en,ko}/cli.md` and reconcile with existing OpenBao Agent / sidecar prose so descriptions agree.
- Add a dedicated section (Inputs / Outputs / Failure conditions / Examples) following the same format as `bootroot service add`. Cover:
  - Purpose: starts the per-service OpenBao Agent sidecar container so `bootroot rotate` can push new secrets immediately on rotate (vs. waiting up to `static_secret_render_interval` for a host-running agent to poll).
  - Container name: `bootroot-openbao-agent-<service>`.
  - Bind-mounts that differ between `daemon` and `docker` deploy types (`/sidecar-config` + `/sidecar-certs` vs. `/openbao/secrets` only).
  - Prerequisites: `bootroot infra install` and `bootroot service add` must have completed; `bootroot-openbao` must exist (compose-managed) — except in the external-OpenBao case where `--openbao-network` is required instead.
  - Once the discovery fix lands, document the new `--openbao-network` flag and the auto-discovery behavior (no more implicit `COMPOSE_PROJECT_NAME=bootroot` requirement).

### 2. Naming disambiguation note

Add a short note clarifying that `service agent start` runs the **OpenBao Agent** (`bao agent`), not the `bootroot-agent` certificate daemon. Two completely different software packages share the word "agent":

| Software | Role | Who runs it |
|---|---|---|
| OpenBao Agent (`bao agent`) | Fetches secrets from OpenBao, renders templates | bootroot (sidecar) or operator (host daemon) |
| `bootroot-agent` | Issues/renews TLS certs against step-ca | Operator (host daemon, separate binary) |

`service agent start` only manages the first one.

### 3. Sidecar vs. host-daemon guidance

The two deployment models for OpenBao Agent are independent of the service `--deploy-type`. Even when the user's service runs as a host daemon, the OpenBao Agent itself can run either as a bootroot-managed sidecar container or as an operator-managed host daemon (`bao agent -config=secrets/openbao/services/<svc>/agent.hcl`).

Document this trade-off so operators can choose deliberately:

| Concern | Sidecar (default) | Host daemon (alternative) |
|---|---|---|
| rotate signal path | Active (`docker restart` → immediate re-render) | Passive (relies on `static_secret_render_interval = 30s` polling) |
| rotate latency | ~seconds | up to 30s |
| Lifecycle management | bootroot owns it (`service agent start`) | Operator owns it (systemd unit, etc.) |
| Container name | Standardized: `bootroot-openbao-agent-<svc>` | Operator's PID/unit naming |
| Privilege isolation | Token/secret_id contained in container | Mixed with host user permissions |
| Code path uniformity | Same pattern across docker / daemon deploy types | Only viable for daemon deploy type |
| Debugging | `docker logs bootroot-openbao-agent-<svc>` | `journalctl -u <unit>` (operator-defined) |

When a sidecar is **not** appropriate:
- Hosts without a docker engine.
- Security policies that mandate systemd-only lifecycles for token-handling processes.
- Highly resource-constrained environments where the sidecar overhead is unacceptable.
- Integration with HSMs / security modules accessible only outside containers.
- Pre-existing operator infrastructure standardized on systemd-based secret management.

**Recommendation, to be stated explicitly in the docs:**

> Default and recommended: sidecar. `bootroot service add` already surfaces `bootroot service agent start` as the primary next step. Choose the host-daemon alternative only when one of the conditions above applies; in that case, lifecycle management of the OpenBao Agent process becomes entirely the operator's responsibility, and rotate latency increases to up to `static_secret_render_interval` per rotate cycle.

### 4. Update `docs/{en,ko}/operations.md`

Existing OpenBao Agent / sidecar narrative needs to be reconciled with the dedicated command section. Mention that the active signal path requires the sidecar; the host-daemon alternative falls back to polling.

### 5. Keep Korean docs in sync

All of the above must land in both `docs/en/` and `docs/ko/` simultaneously.

## Test recommendations

The fix PR for this issue **must** ship with the unit-level coverage below. Only the full lifecycle E2E that exercises `bootroot service agent start` against a non-bootroot project name is deferred to #578.

### Unit tests required in this PR

- CLI flag parsing for `--openbao-network` (presence, default `None`, value plumbing into the override writer).
- `--openbao-network` value validation: reject any string that is not a valid Docker network name. Whitelist the safe character set (`^[a-zA-Z0-9][a-zA-Z0-9_.-]*$` matching docker's accepted names) so that newline / colon / quote characters cannot break the YAML override. Apply the same validation to label-discovered values for consistency.
- Override YAML emission: verify that the discovered or overridden network name and project name end up in the right places, and that an injection attempt via a maliciously crafted network name is rejected before the override is written.
- Docker inspect label parsing: extracts `com.docker.compose.project` correctly; surfaces a clear error when the container is not found vs. when it exists without the label vs. when it exists with the label.
- Decision matrix coverage: all four rows of the decision table — both successful paths (project-discovered or network-overridden) and the error row.

### E2E coverage deferred to #578

A regression scenario that invokes `bootroot service agent start` from a directory whose name is not `bootroot` and from a deployment with `COMPOSE_PROJECT_NAME=custom`. This requires replacing the inline `docker run` shortcut in the lifecycle scripts with the actual bootroot command — the work tracked in #578. The current bypass in the lifecycle e2e is exactly why this bug shipped.

sehkone:
## Summary

`bootroot service agent start` previously hardcoded both `-p bootroot` and the external network name `bootroot_default` in the per-service compose override, so it only worked when the compose project name was literally `bootroot`. From any other working directory or with a non-default `COMPOSE_PROJECT_NAME`, the sidecar failed with `network bootroot_default declared as external, but could not be found`.

This PR replaces the hardcoded constants with runtime discovery. The `bootroot-openbao` container's `com.docker.compose.project` label drives both the `-p` argument passed to `docker compose` and the default external network name (`<project>_default`).

A new `--openbao-network` flag is the only way to specify the network when OpenBao runs outside bootroot's compose (separate host, kubernetes, managed service, etc.). Decision matrix:

| OpenBao in compose | `--openbao-network` | Behavior |
|---|---|---|
| present | unset | Discover project label on `bootroot-openbao`; use `<project>` for `-p` and `<project>_default` for the override network |
| present | set | Use given network; still discover project label for `-p` |
| absent | unset | Error: `--openbao-network` is required |
| absent | set | Use given network; omit `-p` |

Discovered and operator-supplied network names are validated against docker's naming rules (`^[A-Za-z0-9][A-Za-z0-9_.-]*$`) to prevent YAML injection through the override file.

### Code changes

- `src/commands/service/agent_start.rs` — drop `COMPOSE_PROJECT` / `COMPOSE_NETWORK` constants; add `resolve_sidecar_topology`, `discover_compose_project`, `inspect_label_via_docker`, and `validate_docker_network_name`. Plumb the resolved network into the override writer and the resolved project (when present) into `docker compose -p`. Extract `build_compose_up_args` so the `-p <project>` plumbing is unit-testable.
- `src/commands/init/paths.rs` — replace the substring match in `compose_has_openbao` with a structural scan that only matches a top-level service key named `openbao`. External-OpenBao deployments that mention `openbao` only in container names, hostnames, secret names, image tags, comments, or volume paths now route through the external branch correctly, so the decision matrix above is reliable.
- `src/cli/args.rs` — add `--openbao-network` to `ServiceAgentStartArgs`.
- `src/i18n/{en,ko,service}.rs` and `src/i18n.rs` — four new error strings (container missing, label missing, override required for external OpenBao, invalid network name).

### Documentation

- `docs/{en,ko}/cli.md` — add a dedicated `bootroot service agent start` section (Inputs / Outputs / Behavior / Failure / Examples), document `--openbao-network`, reconcile the OpenBao Agent vs. `bootroot-agent` naming, and document the sidecar vs. host-daemon trade-off. Prerequisites and Failure conditions reflect that the `bootroot-openbao` container is required in both compose-present rows of the matrix because `-p` still needs the project label even when `--openbao-network` is supplied.
- `docs/{en,ko}/installation.md` — replace stale `--network bootroot_default` references with the discovery-based path and a `<network>` placeholder, plus the `docker inspect` one-liner that shows the actual project label. Drop the stale "Docker => sidecar; daemon => daemon" guidance and link to the new Sidecar vs. host-daemon section in `cli.md`.
- `docs/{en,ko}/operations.md` — note that the active rotate signal path requires the sidecar model.
- `CHANGELOG.md` — entry under the existing `Fixed` section.

### Tests

25 unit tests pass (`cargo test --bin bootroot agent_start`):

- CLI flag parsing for `--openbao-network` (default `None`, value plumbing).
- Network name validation (whitelist accepts safe inputs; rejects newline / colon / quote / leading-`_-./` / empty).
- Override YAML emission uses the supplied/discovered network name; YAML injection via a crafted network name is rejected before any file is written.
- All four rows of the topology decision matrix.
- `docker inspect` label outcomes: `Present(value)`, `MissingLabel`, `ContainerNotFound`, each surfaced with a precise error.
- Discovered label values are also validated (rejects malicious labels).
- `build_compose_up_args` for both compose-project-present and compose-project-absent branches.

Closes #577

## Not addressed

- The end-to-end regression scenario that runs `bootroot service agent start` from a non-`bootroot` directory and with `COMPOSE_PROJECT_NAME=custom` is **deferred to #578** per the issue's test-recommendations section. Adding it requires replacing the inline `docker run` shortcut in `scripts/impl/run-local-lifecycle.sh` with the actual bootroot command — exactly the work tracked in #578. The unit-level coverage that the issue marks as required for this PR is included.

## Test plan

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes
- [x] `cargo test --bin bootroot agent_start` — all 25 tests pass
- [ ] Manual: from a worktree directory whose name is not `bootroot`, run `bootroot infra install` -> `bootroot init ...` -> `bootroot service add ...` -> `bootroot service agent start --service-name <svc>` and confirm the sidecar comes up attached to the right network without setting `COMPOSE_PROJECT_NAME=bootroot`.
- [ ] Manual: with `COMPOSE_PROJECT_NAME=myorg-prod` exported, repeat the same flow and confirm the sidecar lands in the `myorg-prod` project (`docker compose ps` from the same directory lists it).
- [ ] Manual (external OpenBao): without an `openbao` service in `docker-compose.yml`, run `bootroot service agent start --service-name <svc>` and confirm it errors asking for `--openbao-network`; rerun with `--openbao-network <net>` and confirm the sidecar attaches to that network and `-p` is omitted.
- [ ] Manual: stop the `bootroot-openbao` container and run `bootroot service agent start` — confirm the "container not found" error message is surfaced (with the suggested remediation).
- [ ] Manual: pass `--openbao-network "evil\n  rogue: true"` and confirm the "invalid docker network name" error fires before any override file is written.


sehkone:
**[Reviewer Round 1]**

I found a few issues that should be addressed before this lands.

- `compose_has_openbao` still treats any lowercase `openbao` substring in the compose file as “the compose file declares an OpenBao service” (`src/commands/init/paths.rs:54`). This PR makes that predicate drive the new topology matrix (`src/commands/service/agent_start.rs:120`) and the sidecar override (`src/commands/service/agent_start.rs:249`), so a valid external-OpenBao compose file that merely contains `openbao` in a hostname, secret name, volume path, comment, or helper service name will go down the compose-managed branch. Even with `--openbao-network`, that branch still tries to inspect `bootroot-openbao` for a project label and the override can also rewrite `VAULT_ADDR` to `bootroot-openbao` and add `depends_on: openbao`. That breaks the first-class external-OpenBao case the issue calls out. The predicate needs to identify an actual service key named `openbao` rather than doing substring matching; otherwise the new decision matrix is not reliable.

- The new CLI docs say `bootroot-openbao` missing is only a failure when `--openbao-network` is not provided (`docs/en/cli.md:826`, `docs/ko/cli.md:801`). The code intentionally still discovers the project label when OpenBao is present in compose even if `--openbao-network` is set (`src/commands/service/agent_start.rs:122`), so `bootroot-openbao` missing remains a failure in both “present” rows. Please update the failure text/prerequisites to match the actual matrix; as written, it tells operators that the override flag bypasses a missing compose-managed container when it does not.

- `docs/en/installation.md:316` and `docs/ko/installation.md:323` still recommend “Docker services: sidecar; daemon services: run OpenBao Agent as a daemon per service.” That contradicts the new dedicated command section, which correctly says the sidecar is the default/recommended path independent of service `--deploy-type`, with host-daemon mode as an alternative. This was one of the docs reconciliation points in #577, so these stale bullets should be updated or removed.

The main code shape otherwise looks aligned with the requested fix: project discovery feeds `-p`, the network override is validated before YAML emission, and the four topology rows are represented in unit tests. I would also consider adding a focused unit around compose argument construction so the `-p <project>` regression path is directly covered, but the three items above are the blocking concerns.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the round-1 review. All three blocking items addressed; also picked up the optional unit test suggestion.

**1. `compose_has_openbao` substring matching — Fixed**

Replaced the substring check with a structural scan that only matches an actual top-level service key under `services:`. New helper `compose_has_top_level_service` in `src/commands/init/paths.rs` walks the document, tracks whether we are inside the top-level `services:` block, and only returns true when it sees a sibling key whose name matches exactly (followed by `:`). Comments and blank lines are skipped, and the predicate ignores matches under sibling top-level blocks (e.g., `volumes:` defining an `openbao:` volume) and rejects prefixed names (e.g., `my-openbao:`). Four targeted tests cover the false-positive cases the review called out: `container_name: my-openbao-client`, `VAULT_ADDR=https://openbao.example.com`, comment lines, volume paths, and a sibling `openbao:` volume.

This now flows through both call sites (`agent_start.rs` and `init/steps/openbao_setup.rs`), so the new decision matrix routes external-OpenBao deployments through the external branch reliably.

**2. Failure conditions doc text — Fixed**

`docs/{en,ko}/cli.md` Prerequisites and Failure conditions sections rewritten so they match the actual matrix:
- The `present` rows always require the `bootroot-openbao` container (because `-p` still needs the project label even when `--openbao-network` is set). `--openbao-network` only overrides the network, never bypasses label discovery.
- The `absent` row requires `--openbao-network` and performs no container inspection.

Both English and Korean docs are in sync.

**3. Stale `installation.md` deployment-policy bullets — Fixed**

`docs/en/installation.md:316` and `docs/ko/installation.md:323` no longer recommend "Docker → sidecar; daemon → daemon." They now state that the sidecar is the default regardless of `--deploy-type` and link to the new `Sidecar vs. host-daemon` section in `cli.md` for the trade-off discussion. Both languages updated.

**Optional suggestion — Done**

Extracted `build_compose_up_args` from `apply_service_agent_compose_override` so the `-p <project>` plumbing is unit-testable without shelling out. Two tests cover the regression: project supplied (full arg vector matches) and project `None` (`-p` flag absent).

**Verification**

- `cargo fmt -- --check --config group_imports=StdExternalCrate`: clean
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo test --bin bootroot`: 541 passed (was 535; +4 for the new compose-key tests, +2 for the new arg-builder tests)
- `markdownlint-cli2`: clean

CHANGELOG updated to note the predicate change as well.

sehkone:
**[Reviewer Round 2]**

I rechecked the Round 1 findings against the updated diff and reviewed the new changes against #577.

The Round 1 items are resolved:

- `compose_has_openbao` no longer uses raw substring matching. The new structural scan in `src/commands/init/paths.rs` only treats an actual top-level `services.openbao` key as compose-managed OpenBao, and the added tests cover the false-positive cases from Round 1.
- The `service agent start` docs now match the matrix: when the compose file declares `openbao:`, `bootroot-openbao` and its project label are still required even if `--openbao-network` is supplied; when OpenBao is external, `--openbao-network` is required and no container inspection runs.
- The stale installation guidance that tied OpenBao Agent sidecar vs. daemon mode to the service `--deploy-type` has been replaced in both EN and KO docs.

The optional `-p <project>` coverage was also added via `build_compose_up_args`, which directly covers the regression-sensitive compose argument construction.

I do not see remaining blocking issues in this round. The implementation now matches the requested topology matrix, validates network names before override emission, keeps the external-OpenBao branch first-class, and documents the sidecar/host-daemon trade-off consistently across the touched docs.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Discover compose project for service agent start
```

**Body**

```text
`bootroot service agent start` previously hardcoded `-p bootroot`
and the external network name `bootroot_default` in the per-service
compose override, so it only worked when the compose project name
was literally `bootroot`. From any other working directory, or any
deployment with a non-default `COMPOSE_PROJECT_NAME`, the sidecar
failed to come up with `network bootroot_default declared as
external, but could not be found`.

Resolve the docker network and compose project at runtime instead.
The `bootroot-openbao` container's `com.docker.compose.project`
label is the single source of truth: it drives both the `-p`
argument passed to `docker compose` (so the sidecar lands inside
the same project) and the default external network name written
into the override (`<project>_default`).

Add `--openbao-network` for cases where the operator runs OpenBao
outside bootroot's compose (separate host, kubernetes, managed
service, etc.) — there is no `bootroot-openbao` container to
inspect, so the network must be supplied explicitly. The
`compose_has_openbao` branch determines which combinations are
valid:

| OpenBao in compose | --openbao-network | Behavior                                            |
|--------------------|-------------------|-----------------------------------------------------|
| present            | unset             | discover project label; <project>_default network   |
| present            | set               | use given network; still discover project for -p    |
| absent             | unset             | error: --openbao-network required                   |
| absent             | set               | use given network; omit -p                          |

Discovered and operator-supplied network names are validated
against docker's network naming rules (`^[A-Za-z0-9][A-Za-z0-9_.-]*$`)
to defend the override file against YAML injection via maliciously
crafted values.

Replace the substring match in `compose_has_openbao` with a
structural scan that only matches a top-level service key named
`openbao`, so external-OpenBao deployments that mention `openbao`
only in container names, hostnames, secret names, image tags,
comments, or volume paths route through the external branch
correctly. Extract `build_compose_up_args` so the `-p <project>`
plumbing is unit-testable.

Document the command end to end in `docs/{en,ko}/cli.md` (it had
no dedicated section despite being the recommended next step from
`bootroot service add`), reconcile the OpenBao Agent /
`bootroot-agent` naming overlap, document the sidecar vs.
host-daemon trade-off, and replace the stale
`--network bootroot_default` examples in
`docs/{en,ko}/installation.md` with the discovery-based path.
`docs/{en,ko}/operations.md` notes that the active rotate signal
path requires the sidecar model.

Closes #577
```

sehkone:
Test plan disposition for the unchecked manual items:

- Manual 1–4 (non-`bootroot` directory, `COMPOSE_PROJECT_NAME=...`, external OpenBao, `bootroot-openbao` stopped) are folded into #578's e2e-matrix work as automated regressions instead of one-shot manual checks. See https://github.com/aicers/bootroot/issues/578#issuecomment-pending for the absorbed scope.
- Manual 5 (malicious `--openbao-network` value rejected before any override file is written) is already covered by the unit tests added in this PR (`validate_docker_network_name` whitelist + the YAML-injection rejection test), so no separate manual run is needed.

Closing out the test plan on that basis.

PR #581, opened 2026-05-02
Rename service agent start to service openbao-sidecar start (#578)

### LINKED ISSUE #578: Rename 'service agent start' to 'service openbao-sidecar start' and add E2E coverage for both OpenBao Agent deployment variants

sehkone:
## Summary

Renames `bootroot service agent start` to `bootroot service openbao-sidecar start`, and closes two adjacent E2E gaps that the rename naturally touches.

### Why rename

Two unrelated software packages share the word "agent" in this project:

| Software | Role | Lifecycle owner |
|---|---|---|
| OpenBao Agent (`bao agent`) | Fetches secrets from OpenBao, renders templates | bootroot (sidecar) or operator (host daemon) |
| `bootroot-agent` | Issues/renews TLS certs against step-ca | Operator (host daemon, separate binary) |

The previous spelling `service agent start` invited the wrong reading (sounds like it starts `bootroot-agent`) and was under-specified on the deployment axis (it only manages the sidecar variant, not the host-daemon alternative). The new spelling `service openbao-sidecar start` makes both axes explicit.

### Compatibility (breaking change)

- The previous `service agent start` form keeps parsing for one release as a hidden deprecated alias. It prints a deprecation warning to stderr pointing at the new spelling, then runs the same code path. To be removed in the following release.
- All docs (`docs/{en,ko}/cli.md`, `cli-examples.md`, `installation.md`, `operations.md`), the next-steps text emitted by `service add`, the success/error i18n strings, and the CHANGELOG are updated to the new spelling on day one.

### E2E coverage

Both gaps from issue #578 part 2 are closed in this PR:

- **Gap A** — `scripts/impl/run-local-lifecycle.sh` and `scripts/impl/run-ca-key-rotation-recovery.sh` no longer bypass the bootroot CLI with an inline `docker run`. They now invoke `bootroot service openbao-sidecar start`, so CI exercises the same code path operators use. (This is exactly the path that hid the `bootroot_default` hardcode fixed in #577.)
- **Gap B** — `.github/workflows/ci.yml` gains a new matrix arm `local-no-hosts-host-daemon` that re-runs the no-hosts lifecycle with `OBA_DEPLOYMENT=host-daemon`. The lifecycle script launches `bao agent` in a `--network host` container as a stand-in for an operator-managed host daemon and asserts that rotate still converges within the polling-fallback window (`static_secret_render_interval = 30s` + tolerance). The same arm also runs in `scripts/preflight/ci/e2e-matrix.sh`.
- **Timing assertion** — `scripts/impl/run-local-lifecycle.sh` now brackets the `responder-hmac` rotate with a wall-clock assertion that distinguishes the two propagation routes:
  - **sidecar variant**: must complete below `SIDECAR_ROTATE_LATENCY_LIMIT_SECS` (default 25s, well under the 30s polling window) so a regression in the active container-restart route surfaces here instead of being masked by the polling fallback.
  - **host-daemon variant**: must complete within `HOST_DAEMON_RENDER_TIMEOUT_SECS` (default 75s) since bootroot has no handle on the operator-managed daemon and propagation has to wait for the polling cycle.
- `SIDECAR_OBA_SERVICE` switches from the docker-deploy `web` to the daemon-deploy `edge-proxy` so bootroot's daemon-sidecar bind-mounts land the rendered `agent.toml` at the same host path the test verifies against.

## Dependency

Per the issue: depends on #577 (already in `main` as commit `37e2f36`, "Discover compose project for service agent start"). Without that fix, `service openbao-sidecar start` could not be invoked from a CI runner whose compose project isn't `bootroot`. Since #577 has landed, this PR is unblocked.

## Test plan

- [x] `cargo build` succeeds
- [x] `cargo clippy --all-targets` passes with no warnings
- [x] `cargo test --lib` passes (207 tests)
- [x] New `test_cli_parses_deprecated_service_agent_start_alias` passes (alias still parses)
- [x] New `test_service_agent_alias_emits_deprecation_warning` passes (alias prints deprecation pointing at canonical command)
- [x] New `test_service_openbao_sidecar_does_not_emit_deprecation_warning` passes (canonical command doesn't print the warning)
- [x] All existing `test_cli_parses_service_openbao_sidecar_start*` tests pass
- [x] CI Docker E2E matrix arm `local-no-hosts` (sidecar) — covered by the existing `test-docker-e2e-matrix` job
- [x] CI Docker E2E matrix arm `local-no-hosts-host-daemon` (host-daemon) — new matrix entry added in this PR
- [x] CI exercises the timing assertion `assert_responder_hmac_rotate_latency` in both `local-no-hosts` (sidecar limit 25s) and `local-no-hosts-host-daemon` (host-daemon limit 75s)
- [x] `scripts/impl/run-ca-key-rotation-recovery.sh` succeeds with the new `bootroot service openbao-sidecar start` invocation — verified via CI `Docker E2E (rotation)`
- [x] Manual: `bootroot service agent start --service-name X` prints the deprecation warning to stderr but still works
- [x] Manual: `bootroot service openbao-sidecar start --service-name X` works and emits no deprecation warning
- [x] Docs render correctly (anchor `#bootroot-service-openbao-sidecar-start` resolves in `cli.md`)

Closes #578


sehkone:
**[Reviewer Round 1]**

I found two issues that keep the PR from fully closing the E2E part of #578.

1. `.github/workflows/ci.yml:323` still has no host-daemon matrix arm. The new host-daemon run was added only to `scripts/preflight/ci/e2e-matrix.sh:93`, but the actual PR CI job calls `scripts/impl/run-local-lifecycle.sh` directly from the workflow matrix and never sets `OBA_DEPLOYMENT=host-daemon` (`.github/workflows/ci.yml:359-375`). That means Gap B remains untested in CI unless maintainers remember to run the preflight script manually; the PR description also notes the host-daemon path was not exercised by CI workflows. Please either add a real workflow matrix entry/env arm for host-daemon, or adjust the PR/issue linkage if this is intentionally preflight-only.

2. `scripts/impl/run-local-lifecycle.sh:742-753` and the later rotate blocks only assert eventual verify success, so they do not prove the sidecar arm used the active restart/signal route. In current rotate code, missing sidecar restarts are ignored and the command still waits for the rendered file, so the sidecar scenario would still pass via OpenBao Agent's 30s polling fallback if the active route regressed. The new `HOST_DAEMON_RENDER_TIMEOUT_SECS` at `scripts/impl/run-local-lifecycle.sh:85-87` is also never used. The issue explicitly asked for timing assertions that distinguish sidecar active convergence (~5s) from host-daemon polling convergence (`static_secret_render_interval` + tolerance); without measuring the render/rotate latency or otherwise proving the container restart happened, that requirement is not covered.

sehkone:

sehkone:
**[Author Round 1]**

Both items addressed in commit `9acf713`.

1. **Fixed** — Added a new matrix entry `local-no-hosts-host-daemon` to `.github/workflows/ci.yml` (`test-docker-e2e-matrix` job) that re-runs `scripts/impl/run-local-lifecycle.sh` with `OBA_DEPLOYMENT=host-daemon`. The `oba_deployment` field is now part of the matrix scenario and is plumbed through to the lifecycle script as `OBA_DEPLOYMENT`. The host-daemon arm also gets a longer per-attempt verify budget (`VERIFY_ATTEMPTS=20`, `VERIFY_DELAY_SECS=4`) so the polling-fallback path's ~30s render delay never starves an otherwise-healthy run; sidecar arms keep their original 5×5s budget. Gap B is now exercised on every PR — operators no longer need to remember to run the preflight script.

2. **Fixed** — `scripts/impl/run-local-lifecycle.sh` now brackets the `responder-hmac` rotate with `date +%s` and calls a new `assert_responder_hmac_rotate_latency` after the rotate returns. The function picks a per-deployment ceiling:
   - **sidecar**: must complete within `SIDECAR_ROTATE_LATENCY_LIMIT_SECS` (default `25s`). This sits below `static_secret_render_interval = 30s`, so a regression where bootroot's active container restart silently no-ops (e.g. wrong container name, signal not delivered) cannot be masked by the polling fallback — the wall-clock would jump above 25s and the assertion would fail. The previously unused `HOST_DAEMON_RENDER_TIMEOUT_SECS` is now actually consumed by this same code path.
   - **host-daemon**: must complete within `HOST_DAEMON_RENDER_TIMEOUT_SECS` (default `75s`). bootroot has no handle on the operator daemon (the test container is named `bootroot-openbao-agent-host-${EDGE_SERVICE}`, not the `bootroot-openbao-agent-${EDGE_SERVICE}` that `restart_service_sidecar_agents` looks for), so propagation must come from the polling cycle. The 75s budget gives the 30s polling window plus jitter and headroom for the surrounding restart of the global `bootroot-openbao-agent-responder` container.

   Together the two thresholds prove which route was used: the sidecar arm verifies the active path stayed under the polling window, and the host-daemon arm verifies the polling fallback completes within its expected window. The CHANGELOG entry was extended to document both new envs and the new matrix arm; the PR description was updated to match.

sehkone:
**[Reviewer Round 2]**

I rechecked the Round 1 items against the updated diff and both prior concerns look resolved.

- `.github/workflows/ci.yml` now includes the `local-no-hosts-host-daemon` matrix arm and passes `OBA_DEPLOYMENT=host-daemon` into `scripts/impl/run-local-lifecycle.sh`, so the host-daemon variant is exercised by normal PR CI rather than only by local preflight.
- `scripts/impl/run-local-lifecycle.sh` now starts the sidecar path through `bootroot service openbao-sidecar start`, starts the host-daemon variant separately, and brackets the `responder-hmac` rotate with deployment-specific timing limits. The sidecar limit is below the 30s OpenBao Agent polling interval, so a silent active-restart regression should no longer be masked by polling fallback.

I also did a broader pass over the CLI rename, alias warning, docs touched by the PR, and lifecycle script changes. I do not have any remaining blocking findings from this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Rename service agent start to service openbao-sidecar start
```

**Body**

```text
Two unrelated software packages share the word "agent" in this
project: the OpenBao Agent (`bao agent`, fetches secrets) and
`bootroot-agent` (issues TLS certs against step-ca). The previous
spelling `bootroot service agent start` invited the wrong reading.
It was also under-specified on the deployment axis: the command
manages only the sidecar variant, not the host-daemon alternative.

Rename the canonical command to `bootroot service openbao-sidecar
start`, which makes both axes explicit (which software, which
deployment pattern) and aligns with the vocabulary already used in
internal code (`/sidecar-config`, `daemon_sidecar_bind_mounts`,
the next-steps text from `service add`). Keeps room for a future
`service openbao-daemon` subcommand.

Mitigations for the breaking change:

- The previous `service agent` form keeps parsing for one release
  as a hidden alias that prints a deprecation warning to stderr
  pointing at the new spelling, then runs the same code path.
- All docs (en/ko cli.md, cli-examples.md, installation.md,
  operations.md), the next-steps text emitted by `service add`,
  the CHANGELOG, the success message, and the error context
  string are updated to the new spelling on day one.

Also closes the E2E gaps that motivated grouping this with the
rename:

- `scripts/impl/run-local-lifecycle.sh` and
  `scripts/impl/run-ca-key-rotation-recovery.sh` no longer bypass
  the bootroot CLI with an inline `docker run`; they now invoke
  `bootroot service openbao-sidecar start`. CI exercises the same
  code path operators use, which would have caught the
  `bootroot_default` hardcode fixed in #577.
- `.github/workflows/ci.yml` gains a `local-no-hosts-host-daemon`
  matrix arm that re-runs the no-hosts lifecycle with
  `OBA_DEPLOYMENT=host-daemon`, exercising the polling-fallback
  rotate path (`static_secret_render_interval = 30s`) advertised
  as the alternative to the managed sidecar.
  `scripts/preflight/ci/e2e-matrix.sh` mirrors the new arm.
- The lifecycle brackets the responder-hmac rotate with a
  wall-clock assertion: sidecar mode must finish below
  `SIDECAR_ROTATE_LATENCY_LIMIT_SECS` (25s, well under the 30s
  polling window) so a regression in the active container-restart
  route surfaces here instead of being masked by the polling
  fallback; host-daemon mode gets `HOST_DAEMON_RENDER_TIMEOUT_SECS`
  (75s) since bootroot has no handle on the operator daemon.
- `SIDECAR_OBA_SERVICE` switches from the docker-deploy `web` to
  the daemon-deploy `edge-proxy` so bootroot's daemon-sidecar
  bind-mounts land the rendered `agent.toml` at the same host
  path the test verifies against.

Closes #578
```

PR #583, opened 2026-05-02
Add E2E matrix arms for openbao-sidecar topologies (#582)

### LINKED ISSUE #582: Add E2E matrix arms for compose-project / external-OpenBao / openbao-missing variants of `service openbao-sidecar start`
## Background

PR #580 added runtime discovery of the compose project and the `--openbao-network` flag for `bootroot service openbao-sidecar start` (then named `service agent start`). Its test plan listed five manual scenarios; #578 was meant to absorb the deferred ones into the E2E matrix.

PR #581 (which closed #578) absorbed the two scenarios needed for its own scope:

- Gap A: replace the inline `docker run` in lifecycle scripts with the actual bootroot CLI invocation.
- Gap B: new `local-no-hosts-host-daemon` matrix arm for the host-daemon variant.

But three #580 manual scenarios were not added to the matrix. This issue tracks adding them so the network-discovery code paths get continuous regression coverage.

## Missing matrix arms

Add the following as new arms (or a `OBA_TOPOLOGY` matrix axis) in `scripts/preflight/ci/e2e-matrix.sh` and `.github/workflows/ci.yml`. Scenarios #1 and #3 can be added by flipping one environment variable or container-state knob in the existing local lifecycle; scenario #2 needs a separate setup (stripped compose file, external docker network, reachable OpenBao endpoint) and is not just a variable flip.

### 1. Custom `COMPOSE_PROJECT_NAME`

- Setup: export `COMPOSE_PROJECT_NAME=myorg-prod` (or any non-`bootroot` value) before bringing up the stack. The existing `PROJECT_NAME` variable in `scripts/impl/run-local-lifecycle.sh` is not wired to `docker compose -p` and does not affect the project label — only the env var does, so it must be set explicitly.
- Assert: `bootroot service openbao-sidecar start --service-name <svc>` succeeds and the sidecar lands in the `myorg-prod` project. The most direct check is that the sidecar container's `com.docker.compose.project` label equals `myorg-prod` (`docker inspect --format '{{ index .Config.Labels \"com.docker.compose.project\" }}' <sidecar>`).
- Why: exercises the env-var → project-label discovery branch in `discover_compose_project`. Distinct from the default arm because the discovered name is not `bootroot`, so a regression that hardcodes the default would be caught.

### 2. External OpenBao topology

- Setup: bring up a stripped compose file that has **no** `openbao` service, with OpenBao running standalone on a separate docker network `oba-ext` and reachable at an `openbao_url` the sidecar can hit. This is more involved than \"flip one variable\" — needs the stripped compose, the external network, and a working OpenBao endpoint.
- Assert (negative): `bootroot service openbao-sidecar start --service-name <svc>` (no flag) exits non-zero with the i18n message that requests `--openbao-network`.
- Assert (positive): rerun with `--openbao-network oba-ext` and confirm the sidecar attaches to `oba-ext` and the resulting `docker compose` invocation does **not** include `-p <project>`.
- Why: exercises the `compose_has_openbao() == false` branch and validates both the error and the `--openbao-network` plumbing end-to-end. Unit tests cover the Rust function but not the actual CLI/docker round-trip.
- Implementation note: whether to add this as a full new matrix arm or as a narrow topology helper invoked around `service openbao-sidecar start` in the existing lifecycle is a CI-cost tradeoff — leave the choice to the implementer. A full arm catches lifecycle-wide regressions in this topology (e.g. rotate); a helper is faster but narrower.

### 3. `bootroot-openbao` container missing

- Setup: in the standard sidecar topology, **remove** the openbao container before invoking the command — `docker rm -f bootroot-openbao` or `docker compose rm -sf openbao`. A plain `docker stop` is **not** sufficient: `docker inspect` succeeds on stopped containers and still returns the `com.docker.compose.project` label, so the `Present(value)` branch is taken instead of `ContainerNotFound`.
- Assert: `bootroot service openbao-sidecar start --service-name <svc>` exits non-zero with the \"container not found\" i18n message including the suggested remediation.
- Why: exercises the `inspect_label_via_docker → ContainerNotFound` path through the real `docker inspect` call. Without an E2E arm a regression that swallows the error or surfaces the wrong message would not be caught.

## Out of scope

- Manual 1 from #580 (non-`bootroot` working directory) — already implicitly covered: CI runners check out the repo into arbitrary paths and the existing `local-no-hosts` arm already invokes `service openbao-sidecar start` from a non-`bootroot` cwd after #581.
- Manual 5 from #580 (malicious `--openbao-network` value rejected before any override file is written) — fully covered by the unit tests in #580 (`validate_docker_network_name` whitelist + YAML-injection rejection test). The function is called from a single site and there is no docker-side state to exercise.

## Dependency

None — #580 (network discovery) and #581 (rename + matrix scaffolding) are both merged.

## References

- #578 (closed by #581) — original matrix-absorption issue; this issue addresses the three scenarios that were not picked up.
- #580 — original test plan with the five manual scenarios.
- #581 — added the matrix scaffolding (`OBA_DEPLOYMENT` axis, `local-no-hosts-host-daemon` arm, `assert_responder_hmac_rotate_latency`) that the new arms here should plug into.

sehkone:
## Summary

Add an `OBA_TOPOLOGY` axis to `run-local-lifecycle.sh` with three new values that exercise the network-discovery code paths added in #580, and wire them into both `scripts/preflight/ci/e2e-matrix.sh` and the GitHub Actions matrix in `.github/workflows/ci.yml` as separate arms:

- **`custom-project`** — exports `COMPOSE_PROJECT_NAME=myorg-prod` before installing infra and asserts the sidecar container's `com.docker.compose.project` label equals `myorg-prod`. Catches a regression where project-label discovery silently falls back to a hardcoded `bootroot`.
- **`openbao-missing`** — removes `bootroot-openbao` via `docker rm -f` (a plain `docker stop` would leave the label intact and route through `Present(value)` instead of `ContainerNotFound`) and asserts `service openbao-sidecar start` exits non-zero with the full "container not found" i18n message, including BOTH halves of the remediation guidance (`bootroot infra install` and `--openbao-network`). The em-dash that joins the i18n message in `en.rs` is intentionally not asserted on so a punctuation tweak does not break this arm. Skips rotation since the stack is intentionally degraded.
- **`external-openbao`** — writes a stripped compose file with no `openbao` service, creates an external docker network `oba-ext`, and connects the already-initialised `bootroot-openbao` container to that network so the sidecar (joined to `oba-ext` only) can resolve `bootroot-openbao` via docker DNS and reach the API end-to-end. Rewrites `state.openbao_url` to `http://bootroot-openbao:8200` because the override generated by `service openbao-sidecar start` sets `VAULT_ADDR={state.openbao_url}` verbatim when `compose_has_openbao` is false, and OpenBao Agent honours that env var over the rendered agent-docker.hcl `vault.address` — the rewrite makes config and env agree on the container DNS name. Asserts both the negative path (no flag → i18n message requesting `--openbao-network`) and the positive path: the sidecar attaches to `oba-ext`, the `docker compose` invocation omits `-p <project>` (verified via a `COMPOSE_PROJECT_NAME` sentinel that the resulting label must equal), and `agent.toml` is rendered with an `http_responder_hmac` value — the only signal that proves the sidecar actually authenticated against OpenBao and fetched its template, since `docker compose up -d` succeeds even when the agent's first connection fails and it then restarts.

The `assert_service_oba_start_fails` helper now accepts multiple expected substrings (separated from bootroot CLI args by a literal `--`) so each negative arm can pin both the symptom and the remediation portion of an i18n message.

The implementation chooses the "full new matrix arm" route over a narrow inline helper, trading some CI cost for catching lifecycle-wide regressions in each topology in isolation rather than blocking the default lifecycle.

Closes #582

## Test plan

- [x] `scripts/preflight/ci/e2e-matrix.sh` runs locally and the three new lifecycle invocations all pass
- [x] `local-no-hosts-custom-project` arm: sidecar container's `com.docker.compose.project` label equals `myorg-prod`
- [x] `local-no-hosts-openbao-missing` arm: `service openbao-sidecar start` exits non-zero and the captured stderr contains all three substrings — `container not found`, `bootroot infra install`, and `--openbao-network`
- [x] `local-no-hosts-external-openbao` arm (negative): `service openbao-sidecar start` without `--openbao-network` exits non-zero and the captured stderr contains the "--openbao-network" i18n message
- [x] `local-no-hosts-external-openbao` arm (positive): with `--openbao-network oba-ext` the sidecar container attaches to `oba-ext`, its `com.docker.compose.project` label equals the `COMPOSE_PROJECT_NAME` sentinel (proves no `-p` flag was passed to `docker compose`), and `agent.toml` renders with an `http_responder_hmac` value (proves the sidecar successfully reached the OpenBao endpoint on `oba-ext` and rendered the template end-to-end)
- [x] Existing arms (`local-no-hosts`, `local-no-hosts-host-daemon`, `local-hosts`, remote, rotation) still pass — the default `OBA_TOPOLOGY=default` path is unchanged
- [x] CI: `check`, `test-core`, `test-docker-e2e-matrix`, and `run-extended` all green, including the three new matrix arms


sehkone:
**[Reviewer Round 1]**

I found one blocker.

- `scripts/impl/run-local-lifecycle.sh:418`, `scripts/impl/run-local-lifecycle.sh:437`, `scripts/impl/run-local-lifecycle.sh:443`: The `external-openbao` arm never actually creates an external OpenBao endpoint on `oba-ext`, and it never waits for the sidecar to render from OpenBao. It only writes a stripped compose file, creates an empty Docker network, starts the sidecar with `--openbao-network`, then asserts container metadata. The state produced earlier still points at the default init URL (`http://localhost:8200`), which is not reachable from inside the sidecar container when `compose_has_openbao() == false`, and the standard `bootroot-openbao` container is still on the normal compose network, not `oba-ext`. Because `docker compose up -d` can succeed once the container is created even if `openbao agent` cannot contact Vault/OpenBao and then restarts, this would still pass with a broken positive path. That drops the core setup/positive assertion from issue #582 scenario 2: standalone OpenBao on `oba-ext`, reachable `openbao_url`, and an end-to-end proof that `--openbao-network` lets the sidecar talk to it. I’d expect this arm to either start a separate OpenBao container on `oba-ext` and update the state/OpenBao URL to that network-reachable address, or otherwise provide a reachable endpoint, then reuse `wait_for_oba_render` (or equivalent) so the positive case proves the sidecar actually contacted OpenBao and rendered `agent.toml`, not just that Docker accepted the override.

No other blockers jumped out; the custom-project and missing-container arms look aligned with the issue.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — `external-openbao` arm now provides a reachable OpenBao endpoint and proves end-to-end render. After creating `oba-ext`, the script connects the already-initialised `bootroot-openbao` container to that network with `docker network connect`, so the sidecar (joined to `oba-ext` only) can resolve `bootroot-openbao` via docker DNS. The rendered agent.hcl already addresses OpenBao by container name (`render_docker_agent_config` writes `address = "http://bootroot-openbao:8200"` regardless of `compose_has_openbao`), so no state/URL rewriting is needed — the existing AppRole credentials and template paths just work over the new network. After the positive `service openbao-sidecar start` invocation, the script calls `wait_for_oba_render` against `$AGENT_CONFIG_PATH` to require the `http_responder_hmac` line to appear, which only happens after `bao agent` successfully authenticates and fetches its template. That closes the gap you flagged: a regression in the `--openbao-network` plumbing (wrong network, unreachable endpoint, sidecar in restart-loop) now fails the arm instead of passing on container metadata alone. Diff in commit 1e9576b. PR description and test plan updated to reflect the new render assertion.

sehkone:
**[Reviewer Round 2]**

The Round 1 external-openbao blocker looks resolved: the updated arm now attaches the initialized `bootroot-openbao` container to `oba-ext`, rewrites the workspace `state.openbao_url` to the Docker-DNS endpoint, and waits for the sidecar to render `http_responder_hmac`, so the positive path now proves the sidecar actually reaches OpenBao instead of only proving Docker accepted the override.

I found one remaining issue:

- `scripts/impl/run-local-lifecycle.sh:1156`: the `openbao-missing` arm only asserts that stderr contains `container not found`. Issue #582 specifically asks this arm to verify the "container not found" i18n message including the suggested remediation, but this assertion would still pass if the message regressed to a bare generic `container not found` with no `bootroot infra install` / `--openbao-network` guidance. Please assert a remediation substring as well (for example `run \`bootroot infra install\` first` or `--openbao-network`), or assert the full English i18n message, so the E2E arm covers the actual user-facing failure contract.

sehkone:

sehkone:
**[Author Round 2]**

- **Fixed** — `openbao-missing` arm now asserts the full "container not found" i18n contract, not just the symptom. Extended `assert_service_oba_start_fails` to accept multiple expected substrings (separated from bootroot CLI args by a literal `--`), and the arm now pins three substrings: `container not found` (symptom), `bootroot infra install` (first remediation), and `--openbao-network` (second remediation). A regression that drops either half of the remediation guidance — or swaps in a generic `container not found` — now fails the arm. The em-dash that joins the i18n message in `en.rs` is intentionally not asserted on so a future punctuation tweak does not break this arm. The `external-openbao` negative call site was also updated to use the new `--` separator (single substring, no semantic change there). Diff in commit 85d7304. PR description updated to reflect the expanded assertion contract.

sehkone:
**[Reviewer Round 3]**

The Round 2 item is resolved. `assert_service_oba_start_fails` now checks every expected substring, and the `openbao-missing` arm asserts `container not found`, `bootroot infra install`, and `--openbao-network`, so it now covers the full user-facing remediation contract rather than only the symptom.

I re-reviewed the updated diff against #582. The three requested topologies are wired into both CI entry points, the custom project assertion verifies the discovered compose project label, the missing-container arm removes the container rather than stopping it, and the external-OpenBao arm now covers both the missing-flag error and a positive render through `oba-ext` with the no-`-p` sentinel check.

No further findings from this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add E2E matrix arms for openbao-sidecar topologies
```

**Body**

```text
Add an `OBA_TOPOLOGY` axis to `run-local-lifecycle.sh` with three new
values that exercise the network-discovery code paths added by the
prior `service openbao-sidecar start` work, and wire them into both
`scripts/preflight/ci/e2e-matrix.sh` and the GitHub Actions matrix in
`.github/workflows/ci.yml` as separate arms.

- `custom-project` exports `COMPOSE_PROJECT_NAME=myorg-prod` before
  installing infra and asserts the sidecar container's
  `com.docker.compose.project` label equals `myorg-prod`. Catches a
  regression where project-label discovery silently falls back to a
  hardcoded `bootroot`.
- `openbao-missing` removes `bootroot-openbao` via `docker rm -f`
  (a plain `docker stop` would leave the label intact and route
  through `Present(value)` instead of `ContainerNotFound`) and
  asserts `service openbao-sidecar start` exits non-zero with the
  full "container not found" i18n message, including BOTH halves of
  the remediation guidance (`bootroot infra install` and
  `--openbao-network`). The em-dash that joins the message in
  `en.rs` is intentionally not asserted on so a punctuation tweak
  does not break this arm. Skips rotation since the stack is
  intentionally degraded.
- `external-openbao` writes a stripped compose file with no
  `openbao` service, creates an external docker network `oba-ext`,
  and connects the already-initialised `bootroot-openbao` container
  to that network so the sidecar (joined to `oba-ext` only) can
  resolve `bootroot-openbao` via docker DNS and reach the API
  end-to-end. Rewrites `state.openbao_url` to
  `http://bootroot-openbao:8200` because the override generated by
  `service openbao-sidecar start` sets `VAULT_ADDR={state.openbao_url}`
  verbatim when `compose_has_openbao` is false, and OpenBao Agent
  honours that env var over the rendered agent-docker.hcl
  `vault.address` — the rewrite makes config and env agree on the
  container DNS name. Asserts both the negative path (no flag → i18n
  message requesting `--openbao-network`) and the positive path: the
  sidecar attaches to `oba-ext`, the `docker compose` invocation
  omits `-p <project>` (verified via a `COMPOSE_PROJECT_NAME`
  sentinel that the resulting label must equal), and `agent.toml`
  is rendered with an `http_responder_hmac` value — the only signal
  that proves the sidecar actually authenticated against OpenBao
  and fetched its template, since `docker compose up -d` succeeds
  even when the agent's first connection fails and it then
  restarts.

The `assert_service_oba_start_fails` helper accepts multiple
expected substrings (separated from bootroot CLI args by a literal
`--`) so each negative arm can pin both the symptom and the
remediation portion of an i18n message.

The implementation chooses the "full new matrix arm" route over a
narrow inline helper, trading some CI cost for catching
lifecycle-wide regressions in each topology in isolation rather
than blocking the default lifecycle.

Closes #582
```

PR #589, opened 2026-05-03
Add non-interactive operation surface for CI rotations (#587)

### LINKED ISSUE #587: Non-interactive operation surface for CI / scripted rotations
## Background

The integration test plan in [aicers/aice-web-next] (rebuilt against bootroot 06778e2) ran every operator-facing command end-to-end on macOS with no humans in the loop. Four small CLI gaps forced the harness to wrap bootroot calls in shell tricks (env injection, here-string `<<< $'y\n'`, custom polling loops, `=` syntax for short args) instead of just passing flags. None of the gaps require behaviour changes — they are surface fixes.

This issue collects the four into one cluster. Each is purely additive: defaults stay as today, so existing operator muscle memory is preserved.

## Summary

| § | Today | Gap | Proposed |
|---|---|---|---|
| §1 | `--yes` exists on the **parent** `rotate` command (`src/cli/args.rs:287`), so `bootroot rotate --yes <subcmd>` works, but `bootroot rotate <subcmd> --yes` fails with "unexpected argument". No `-y` alias either. | Position-sensitive flag + missing short alias | Promote `--yes` to a `global = true` arg with short alias `-y`, so it works at any position on every rotate subcommand. Default still prompts. |
| §2 | `RuntimeAuthArgs` (`src/cli/args.rs:226`) accepts `--root-token`/`OPENBAO_ROOT_TOKEN` and AppRole `*-file` flags, but no file-based input for the root token. CI must export the env var or pipe stdin. | No `--root-token-file <path>` flag | Add `--root-token-file <path>`; it beats `OPENBAO_ROOT_TOKEN`, but conflicts with an explicit `--root-token` CLI flag (error out). Do **not** auto-discover `tmp/init-summary.json`. |
| §3 | `rotate force-reissue --wait` only polls completion for `--delivery-mode remote-bootstrap` services. For `daemon + local-file` the operator must hand-roll `until [ -s cert ] && [ "$(openssl x509 ... -serial)" != ... ]; do sleep 3; done` | `--wait` is a no-op on local-file | Make `--wait` poll the on-disk cert file (mtime + serial change) for local delivery. `--wait-timeout` already exists. |
| §4 | `bootroot service add ... --post-renew-arg -HUP` errors with `unexpected argument '-H' found`. Workaround: `--post-renew-arg=-HUP` | clap parses `-HUP` as a short-flag chain | Set `allow_hyphen_values(true)` on `--post-renew-arg` (and the same flag on `bootroot-remote`). |

## Non-goals

- **No automatic discovery of root token from `tmp/init-summary.json`.** The summary file is a temp artefact; auto-reading it would entrench it as a permanent credential store and weaken the operator's understanding of where their root token lives. `--root-token-file` lets operators opt in explicitly.
- **No default `--yes` for rotations.** Confirms must stay default-on for destructive ops.
- **No environment auto-detection that flips `--wait` semantics.** `--wait` should mean the same thing for any `--delivery-mode`; this issue extends coverage, not policy.

## Delivery

Ship all four sections in a single PR (`Closes #587`). The cluster is intentionally small enough that splitting would just add review overhead — the heaviest piece (§3 polling) shares test scaffolding with §1's `--yes`-enabled E2E path, and §2's file-mode policy is self-contained enough to review alongside the surface changes.

## §1 — `--yes` on rotate subcommands

**Correction (verified against HEAD `06778e2`).** `--yes` already exists on the parent `rotate` command (`src/cli/args.rs:287`) and feeds `args.yes` into every subcommand's `prompt_yes_no` call (`commands/init/steps/prompts.rs:54`). The actual gaps are:

1. The flag is position-sensitive: `bootroot rotate --yes force-reissue ...` works, but `bootroot rotate force-reissue --yes ...` fails with "unexpected argument" because `--yes` is parsed only at the parent level. CI scripts and operator muscle memory expect the flag to follow the subcommand.
2. There is no `-y` short alias.

Fix: change the existing `#[arg(long)] yes: bool` on `RotateArgs` to `#[arg(long, short = 'y', global = true)] yes: bool`. With `global = true`, clap accepts the flag at any position under `rotate`, and the existing `args.yes` consumers don't need to change. No per-subcommand copy-paste.

Audited subcommands that already wire `args.yes` into their confirm path (every one benefits):

- `rotate force-reissue --service-name <NAME>`
- `rotate responder-hmac`
- `rotate approle-secret-id --service-name <NAME>`
- `rotate trust-sync`
- `rotate infra-cert`
- `rotate db`
- `rotate stepca-password`
- `rotate ca-key` and `rotate ca-key --full`
- `rotate openbao-recovery [--rotate-unseal-keys]`

`bootroot init` confirmation flow (`Provision PostgreSQL role/database?`, `Register EAB now?`, `Save unseal keys to file?`) is **out of scope here** — those are addressed in the bootstrap-robustness sibling issue (`--no-eab`, etc.). Document the staged rollout in both PR descriptions so operators don't expect symmetry between subcommands during the gap.

## §2 — `--root-token-file <path>`

Today `RuntimeAuthArgs` (`src/cli/args.rs:226`) exposes `--root-token` (also fed by `OPENBAO_ROOT_TOKEN` env) for the root token, and `--approle-role-id-file` / `--approle-secret-id-file` for AppRole — but there is no file-based input for the root token itself. The env-only escape is awkward in CI: every step has to re-export the token (leaking into other tools' env) or use stdin tricks. A file-based flag reuses the existing loader and stops the token from sitting in process env.

Resolution order (must be stated explicitly in the PR description and `--help` text):

- If `--root-token-file` is present: reject an explicit `--root-token` (error), then load the token from the file.
- Otherwise: use explicit `--root-token` / `OPENBAO_ROOT_TOKEN` via clap's normal precedence, then fall back to interactive prompt.

In other words `--root-token-file` does **not** silently override an explicit `--root-token`; the two are mutually exclusive. This is intentionally **stricter** than the existing AppRole `*-file` behaviour: `resolve_from_value_or_file` (`src/commands/openbao_auth.rs:137`) currently prefers the inline value when both are present without erroring. For the root token we want explicit conflict detection, so the implementation needs to distinguish a real CLI `--root-token` from a value injected via `OPENBAO_ROOT_TOKEN` env — use clap's `ArgMatches::value_source` (or split the env binding off the CLI flag) rather than relying on `Option::is_some`. Once the conflict check passes, file content takes precedence over env (which is just clap's normal precedence with the file path resolved last).

**File mode policy.** Reject if the file is other-readable (`o+r`, i.e. mode `& 0o004 != 0`). Group-readable (`0o640`) is permitted because operator/CI service accounts often share a group with the bootroot runner. World-readable (`0o644`, the default umask result on many macOS dev shells) is the cliff users hit by accident; a clear error message that quotes `chmod 0600 <path>` is the fix. Symlinks: resolve once, then check the target's mode (matches step-ca behaviour). On non-Unix targets the check is skipped (we already gate similar checks behind `cfg(unix)`).

## §3 — `--wait` for daemon + local-file

The current `--wait` implementation polls the per-service KV path `{kv_mount}/data/bootroot/services/<service>/reissue` for `completed_at` (c266c7e). For `daemon + local-file` services there is no remote agent writing back to KV; the cert is just rewritten on disk by the local `bootroot-agent`. The natural local-file equivalent is to poll `paths.cert` (from `state.json`):

1. Capture `(serial, mtime)` of the existing cert **before** issuing `delete + signal`. If the cert file is missing (first issuance after a wipe), record `serial = None` and treat any new readable cert as success.
2. After `delete + signal`, poll the cert path until either the file exists with a different serial, or `--wait-timeout` elapses.

**Comparison policy.** Use **serial-only** comparison as the success signal; treat `mtime` as a tiebreaker only (e.g. when the parsed serial is somehow identical but the file was rewritten). Sub-second `mtime` precision is filesystem-dependent and unreliable across short polling intervals, so it must not be the primary equality test.

`x509-parser` is already a direct dependency (`Cargo.toml:32`), so serial parsing stays in-process — no new dep, no shell-out.

**Polling cadence reuse.** The remote-bootstrap path defines `POLL_INTERVAL = Duration::from_secs(2)` at `src/commands/rotate/ca.rs:716` (introduced in c266c7e, used by `wait_for_remote_completion` at `:704`). Reuse the same constant — lift it to module scope if needed — rather than introducing a new one, so both delivery modes feel identical from the operator's side and `--wait-timeout` keeps the same default semantics.

When `--wait` is set on a service whose delivery mode is unsupported by the wait implementation, today's behaviour is a silent no-op. After this fix that branch becomes "poll the local cert" instead. No CLI flag changes; the change is observable only by exit timing.

## §4 — `allow_hyphen_values` for `--post-renew-arg`

Reproducer (06778e2):

```
bootroot service add ... \
  --post-renew-command pkill \
  --post-renew-arg -HUP \
  --post-renew-arg -f \
  --post-renew-arg "review/target/release/review"
# error: unexpected argument '-H' found
```

`-H` is the short form of nothing on this command, but clap still shorts the parse. The same bytes work with `=`:

```
--post-renew-arg=-HUP --post-renew-arg=-f
```

This is a UX cliff. Set `allow_hyphen_values(true)` on `--post-renew-arg`. Existing `=` users continue to work; the space form starts working too.

**Sibling-arg audit (completed).** A grep for repeatable user-supplied argv flags in `cli/args.rs` returns only `--post-renew-arg` (`src/cli/args.rs:864`). There is no `--pre-renew-arg`. The same CLI flag also exists on the `bootroot-remote` binary (`src/bin/bootroot-remote/main.rs:156`), where the operator can override hooks via remote-bootstrap; apply `allow_hyphen_values(true)` there too so the two surfaces stay in sync. No other CLI flag in the audit takes user-supplied argv tokens.

## Test plan

- Unit: extend `tests/bootroot_rotate.rs` for each rotate subcommand: invoke with and without `--yes`, assert prompt is consulted only when absent.
- Unit: `tests/bootroot_rotate.rs` — `--root-token-file` resolves before env; mode-`0o644` file rejected; mode-`0o640` accepted; mode-`0o600` accepted. Conflict matrix: `--root-token-file` + explicit `--root-token` → error; `--root-token-file` + `OPENBAO_ROOT_TOKEN` env → file wins (no error). The env-vs-explicit-flag distinction must be driven by clap's `value_source`, not `Option::is_some`.
- Unit: `tests/bootroot_service.rs` — `--post-renew-arg -HUP` (with and without `=`) both produce the same `hooks.post_renew.success.args`. Mirror coverage in `bootroot-remote` arg parsing tests.
- Integration (E2E): add a `daemon + local-file` arm that runs `bootroot rotate force-reissue --service-name <svc> --yes --wait` and asserts the command returns only after the on-disk cert serial changes (no external `until` loop). `scripts/impl/run-local-lifecycle.sh` already exercises `force-reissue` and should drop its bash polling loop in favour of the new flag combo once §3 lands.

## References

- Reproduced end-to-end during 2026-05-03 integration test rebuild against `aice-web-next` `main`. The harness ([test-clumit/integration-test-plan.md][plan]) currently documents each workaround inline; see §10 (force-reissue) and §11 (rotations) for the bash patterns this issue proposes to retire.
- c266c7e introduced `--wait` for remote-bootstrap; §3 here closes the local-file gap left open at that point.
- 04bbd5c rejected path-like `--reload-target` for the `sighup` preset; §4 makes the suggested low-level fallback (`--post-renew-command pkill --post-renew-arg -HUP --post-renew-arg -f --post-renew-arg <path>`) work in its natural spelling.

[aicers/aice-web-next]: https://github.com/aicers/aice-web-next
[plan]: https://github.com/aicers/aice-web-next

### LINKED ISSUE #590: Wire `rotate force-reissue --yes --wait` into the daemon+local-file E2E matrix
## Background

[#589](https://github.com/aicers/bootroot/pull/589) (closes #587) shipped the `--wait` extension for `--delivery-mode local-file` services. The PR's test plan covered the new code path with five unit tests on `wait_for_local_completion_*`, but the integration line was explicitly **deferred**:

> - [ ] E2E: `daemon + local-file` arm exercising `rotate force-reissue --wait`
>       (covered by `scripts/preflight/ci/e2e-matrix.sh` once the new flag combo is wired in)

The CI matrix (`Docker E2E (local-no-hosts)`, `Docker E2E (rotation)`, etc.) currently exercises `force-reissue` through a bash `until [ -s cert ]` polling loop in `scripts/impl/run-local-lifecycle.sh`, so the new `--wait` local-file code path **has never run end-to-end in CI**. The unit tests cover the algorithm; what they cannot cover is the file-system race at real-agent-issuance time (delete-and-recreate atomicity, container-overlay mtime resolution, `--wait-timeout` vs actual ACME latency).

## Scope

Tiny follow-up — replace the existing bash polling loop in `scripts/impl/run-local-lifecycle.sh` with the new flag combo so the next CI run exercises the local-file `--wait` path naturally.

Concretely:

- Locate the `until [ -s … ]` (or equivalent) polling loop that wraps `bootroot rotate force-reissue` for the `daemon + local-file` arm.
- Replace it with:

  ```bash
  ./bootroot rotate force-reissue --service-name "$svc" --yes --wait
  ```

  Drop the surrounding `until`/`sleep` loop entirely. The `--yes` global flag from #587 §1 obviates the heredoc / here-string confirm prompt. `--wait`'s default `--wait-timeout` (`2m`) matches the existing loop's de-facto upper bound.
- If the same bash pattern appears in `scripts/impl/run-rotation-recovery.sh` or other E2E scripts that hit `daemon + local-file` services, apply the same simplification there (audit during the PR; do not silently leave one of them on the legacy pattern).

## Why now

`#588`'s in-flight PRs (especially §1 `db-provision schema grants` and §6 `openbao-sidecar refresh`) will exercise the same E2E matrix. Doing this cleanup as a one-line change in any of those PRs — or as a stand-alone PR landing alongside them — costs minutes; doing it later means the local-file `--wait` integration coverage stays a known gap indefinitely.

## Non-goals

- No change to `--wait` semantics, polling cadence, or timeout default. This issue is strictly about wiring CI to call the already-shipped flag combo.
- No change to the unit test suite.
- No new CI matrix arm. The existing `local-no-hosts` / `local-no-hosts-host-daemon` arms already cover daemon+local-file; we are replacing the bash polling inside them, not adding a new arm.

## Verification

- After the change, run `scripts/preflight/ci/e2e-matrix.sh` locally (or push a draft PR and watch the matrix). All Docker E2E arms that previously polled by hand should now wait in-binary and pass.
- The CI run for the PR that lands this change is itself the verification artefact — a green `Docker E2E (local-no-hosts)` and `Docker E2E (rotation)` job confirms the new code path executed successfully end-to-end at least once.
- Check that no replacement happens in `scripts/impl/run-remote-lifecycle.sh` — `--wait` for remote-bootstrap was already shipped under c266c7e and the remote scripts already use the in-binary wait. This issue does not touch them.

## References

- [#587](https://github.com/aicers/bootroot/issues/587) §3 (the issue this work follows up on).
- [#589](https://github.com/aicers/bootroot/pull/589) — the PR that introduced the unchecked test-plan item.
- c266c7e (the original `--wait` for remote-bootstrap, whose CI wiring this issue mirrors for local-file).

sehkone:
## Summary

Closes #587. Ships all four CLI surface fixes from the issue in a single PR — defaults are unchanged, every gap is purely additive.

### §1 — `--yes` on rotate subcommands

Promoted `RotateArgs::yes` to `#[arg(long, short = 'y', global = true)]`. With `global = true` clap accepts `--yes`/`-y` at any position under `rotate` (e.g. `rotate force-reissue --yes`, `rotate --yes force-reissue`, and `rotate db -y` all parse). Existing `args.yes` consumers in each subcommand are unchanged.

### §2 — `--root-token-file <path>`

Added `--root-token-file` to `RuntimeAuthArgs` with `conflicts_with = \"root_token\"` so an explicit `--root-token` CLI flag is rejected at parse time. Env-injected `OPENBAO_ROOT_TOKEN` no longer flows through the same clap arg as `--root-token`; instead the env is read directly inside the resolver (`resolve_root_token` in `src/commands/openbao_auth.rs`) so it cannot falsely trip the conflict check. Resolution order is `--root-token-file` > `--root-token` > env > interactive prompt. File mode policy (Unix only): `o+r` (`& 0o004 != 0`) is rejected with a `chmod 0600 <path>` hint; `0o640` (group-readable) is permitted for shared CI groups; `0o600` is the recommended default. Empty / whitespace-only files are also rejected.

### §3 — `--wait` for daemon + local-file

`rotate_force_reissue` now branches into `rotate_force_reissue_local` for non-remote delivery. The local path snapshots `(serial, mtime)` of the existing cert *before* `delete + signal`, then polls `paths.cert` until either the serial changes (primary success signal) or — for the rare same-serial reissue — the mtime strictly advances (tiebreaker). Missing cert before signal maps to `serial = None`, so first issuance after a wipe still completes. Cert serial parsing reuses the existing `x509-parser` direct dependency (no new dep). The polling constant `REISSUE_WAIT_POLL_INTERVAL` is lifted to module scope and shared with the remote-bootstrap path so `--wait-timeout` semantics feel identical across delivery modes. Timeout remains a non-error (matches existing remote behaviour).

### §4 — `allow_hyphen_values` for `--post-renew-arg`

Set `allow_hyphen_values = true` on `--post-renew-arg` in both `src/cli/args.rs` (`bootroot service add`) and `src/bin/bootroot-remote/main.rs` (`bootroot-remote bootstrap`). `--post-renew-arg -HUP` and the existing `--post-renew-arg=-HUP` form now both parse to the same `Vec<String>`.

### Out of scope

- No automatic discovery of root token from `tmp/init-summary.json` (per the non-goals in the issue).
- No default `--yes` for rotations.
- No environment auto-detection that flips `--wait` semantics.
- `bootroot init` confirmation prompts (handled in the bootstrap-robustness sibling issue per §1).

## Test plan

- [x] `cargo build`
- [x] `cargo clippy --all-targets`
- [x] `cargo test --lib` (207 passed)
- [x] `cargo test --bins` (564 + 22 + 57 passed across binaries)
- [x] §1: `--yes` parses both before and after the rotate subcommand (`test_cli_rotate_yes_after_subcommand`, `test_cli_rotate_yes_before_subcommand_still_works`)
- [x] §1: `-y` short alias parses (`test_cli_rotate_short_y_alias_works`)
- [x] §1: global flag works on a non-`force-reissue` subcommand (`test_cli_rotate_yes_global_on_db_subcommand`)
- [x] §2: `--root-token-file` parses on rotate subcommands (`test_cli_rotate_parses_root_token_file`)
- [x] §2: `--root-token-file` + explicit `--root-token` rejected at parse time (`test_cli_rotate_rejects_root_token_with_root_token_file`)
- [x] §2: `--root-token-file` overrides env when no CLI flag (`root_token_file_overrides_env_when_no_cli_flag`)
- [x] §2: explicit `--root-token` flag beats env (`explicit_root_token_flag_beats_env`)
- [x] §2: env used when no flag or file (`env_used_when_no_flag_or_file`)
- [x] §2: file mode `0o600` accepted, `0o640` accepted, `0o644` rejected with `chmod 0600` hint (`root_token_file_mode_0o600_accepted`, `root_token_file_mode_0o640_accepted_for_group_sharing`, `root_token_file_mode_0o644_rejected`)
- [x] §2: empty token file rejected (`root_token_file_empty_rejected`)
- [x] §3: `read_cert_serial` returns uppercase hex (`read_cert_serial_returns_uppercase_hex`)
- [x] §3: wait returns immediately when serial already changed (`wait_for_local_completion_returns_immediately_when_serial_already_changed`)
- [x] §3: wait succeeds for first issuance after a wipe (`wait_for_local_completion_succeeds_when_starting_with_no_cert`)
- [x] §3: wait times out when cert is unchanged (`wait_for_local_completion_times_out_when_unchanged`)
- [x] §3: mtime tiebreaker fires when serial is identical (`wait_for_local_completion_uses_mtime_tiebreaker_for_same_serial`)
- [x] §4: `--post-renew-arg -HUP` (space form) parses for `service add` (`test_cli_service_add_post_renew_arg_hyphen_value_space_form`)
- [x] §4: `--post-renew-arg=-HUP` (eq form) still parses for `service add` (`test_cli_service_add_post_renew_arg_hyphen_value_eq_form`)
- [x] §4: same coverage on `bootroot-remote bootstrap` (`bootstrap_post_renew_arg_accepts_hyphen_values_in_space_form`, `bootstrap_post_renew_arg_accepts_hyphen_values_in_eq_form`)
- [x] E2E: `daemon + local-file` arm exercising `rotate force-reissue --wait` (wired in by #592 / closes #590; verified via `scripts/preflight/ci/e2e-matrix.sh` and `scripts/preflight/ci/e2e-extended.sh` runs)


sehkone:
**[Reviewer Round 1]**

No blocking findings.

The PR matches #587's intended additive scope. `RotateArgs::yes` is now global with `-y` in `src/cli/args.rs:299`, root-token file resolution is separated from `OPENBAO_ROOT_TOKEN` and rejects an explicit `--root-token` conflict in `src/commands/openbao_auth.rs:116`, the Unix mode check rejects `o+r` while allowing group-readable files in `src/commands/openbao_auth.rs:132`, local `--wait` snapshots before delete/signal and uses serial-first/mtime-second completion checks in `src/commands/rotate/ca.rs:556` and `src/commands/rotate/ca.rs:862`, and both `--post-renew-arg` surfaces now accept hyphen-prefixed values in `src/cli/args.rs:879` and `src/bin/bootroot-remote/main.rs:158`.

The tests are targeted at the regressions that would have failed before this change: clap parser coverage for `--yes` placement, `-y`, `--root-token-file` conflict parsing, hyphen-prefixed hook args, resolver tests for file/env precedence and file modes, and local wait comparator tests. The remaining risk is the PR's unchecked full daemon + local-file E2E arm: the unit tests cover the core wait logic, but not the actual delete + signal + agent rewrite path. I would treat that as pre-merge validation if the project requires the local lifecycle preflight for this surface, not as a code change request.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add non-interactive operation surface for CI rotations
```

**Body**

```text
Promote rotate `--yes` to a global flag with `-y` short alias so it
parses at any position under the rotate subcommand tree. Operators
and CI scripts can now spell `rotate force-reissue --yes` and
`rotate --yes force-reissue` interchangeably.

Add `--root-token-file <path>` on rotate so the OpenBao root token
can be loaded from a file instead of `OPENBAO_ROOT_TOKEN` env or
stdin tricks. Resolution order is `--root-token-file` >
`--root-token` > env > interactive prompt, with `--root-token-file`
clap-conflicting `--root-token` so an env-injected value cannot
trip the conflict check. The token file is rejected when
world-readable (`0o644`); group-readable (`0o640`) is permitted for
shared CI/operator groups, and the error message includes a
`chmod 0600` hint.

Make `rotate force-reissue --wait` cover daemon + local-file
delivery in addition to remote-bootstrap. The local path captures
the cert serial before delete + signal, then polls `paths.cert`
until the serial changes (with mtime as a tiebreaker for the rare
same-serial reissue). The polling cadence constant is shared with
the remote path so `--wait-timeout` semantics are identical across
delivery modes.

Set `allow_hyphen_values = true` on `--post-renew-arg` for both
`bootroot service add` and `bootroot-remote bootstrap` so
`--post-renew-arg -HUP` parses without forcing the `=` form.

Closes #587
```

PR #591, opened 2026-05-03
Harden bootstrap and rotation against env variance (#588)

### LINKED ISSUE #588: Robust bootstrap and rotation against environment variance and operator slips
## Background

The integration test plan in [aicers/aice-web-next] (rebuilt against bootroot 06778e2) hit six places where the **environment** or a **single operator typo at the prompt** forced the harness into manual recovery — running raw \`docker exec psql\` for schema grants, hand-passing \`--db-admin-dsn\`, deleting OpenBao volumes to re-init, restarting OpenBao Agent sidecars to flush stale renders, and so on. None of these are bugs in a single command; they are the same shape of problem (\"bootroot's assumptions about the surrounding state did not match reality, and there is no in-tree path back to a working state\") in six different locations.

This issue groups them. Sub-sections are independent enough to land as separate PRs (see \"PR scope\" tags), but a single tracking issue keeps the design choices coherent across them — particularly the consistent escape-hatch policy.

## Summary

| § | Failure mode hit at 06778e2 | Disposition |
|---|---|---|
| §1 | step-ca crashloops with \`permission denied for schema public\` after \`init --enable db-provision\` on PG15+ | **(automatic)** — db-provision grants schema CREATE + sets ownership for the role it just created |
| §2 | \`rotate db\` reads \`ca.json.db.dataSource\` as if it were the admin DSN, but it's the runtime DSN; PostgreSQL rejects the self-ALTER | **(automatic)** — init persists admin DSN to OpenBao KV; rotate db reads from KV. Existing \`--db-admin-dsn\` flag remains as override |
| §3 | EAB prompt accepts arbitrary text (e.g. \`y\`) → KV ends up with garbage → every cert issuance fails with \`Failed to decode EAB key: Invalid input length: 1\` and there is no recovery command | (a) **(automatic)** input validation, (b) **(opt)** \`--no-eab\` skip flag, (c) **(opt)** new \`bootroot rotate eab-clear\` recovery command |
| §4 | \`infra install\` runs \`docker compose up\` without checking 5432 availability; on conflict the command half-completes, the operator must edit \`.env\` and re-run | (a) **(automatic)** pre-bind diagnostic, (b) **(opt)** \`--postgres-host-port <N>\` flag |
| §5 | A failed \`init\` leaves OpenBao initialized; subsequent \`init\` retries get \`403 Forbidden permission denied\` because bootroot has no usable root token. Recovery requires \`bootroot clean -y\` (which wipes volumes) | (a) **(automatic)** init detects partial-init state and emits an actionable diagnostic, (b) **(opt)** new \`bootroot clean --openbao-only\` |
| §6 | After writing/deleting a KV path read by an OpenBao Agent sidecar template (e.g. clearing stale EAB), the sidecar continues to render the old value. Recovery is operator-side \`docker restart bootroot-openbao-agent-<svc>\` | **(opt)** new \`bootroot service openbao-sidecar refresh --service-name <name>\` |

## Non-goals (deliberately excluded)

- **No automatic port fallback (5432 → 5433)** in §4. Implicit fallback would override operators with firewall rules pinning specific ports, or operators who deliberately leave 5432 free for another service. Diagnostic + explicit flag only.
- **No automatic OpenBao wipe + re-init** in §5. Operators may point bootroot at an OpenBao instance that holds **other** secrets; an automated wipe would destroy them. Diagnostic + opt-in \`clean --openbao-only\` only.
- **No automatic CA-type detection that suppresses the EAB prompt** in §3. Detecting \"this is OSS step-ca, EAB is unsupported\" from runtime signals risks false positives on commercial step-ca behind a reverse proxy. Operators flag explicitly via \`--no-eab\`; the mistaken-input case is closed by validation (3a).
- **No automatic sidecar refresh on every KV change** in §6. consul-template's render lag is downstream behaviour; auto-refresh from bootroot's side could partial-render mid-edit during operator KV maintenance. Explicit operator-issued command only.

## §1 — db-provision schema grants (PR scope: small)

\`\`\`
docker exec bootroot-postgres psql -U step -d stepca \\
  -c \"GRANT ALL ON SCHEMA public TO stepca;\"
docker exec bootroot-postgres psql -U step -d stepca \\
  -c \"ALTER SCHEMA public OWNER TO stepca;\"
docker restart bootroot-ca
\`\`\`

These three commands were required after a clean \`init --enable db-provision\` for step-ca to come up at all. PostgreSQL 15+ removed the implicit \`CREATE\` grant on the \`public\` schema for non-owners ([PG15 release notes][pg15-public]); db-provision creates the \`stepca\` role but never grants it the privilege to create the tables it then tries to create.

Change: as part of \`commands/init/steps/database.rs\` provisioning, after \`CREATE ROLE / CREATE DATABASE\`, run:

\`\`\`sql
GRANT ALL ON SCHEMA public TO <db_user>;
ALTER SCHEMA public OWNER TO <db_user>;
\`\`\`

This is a contract obligation of \`--enable db-provision\` — an operator who wants step-ca to live on a non-public schema should not be using db-provision in the first place; they should pass their own \`--db-dsn\` to a pre-provisioned role.

After this change, step-ca's restart inside \`init\` (1003956) will succeed on the first attempt for fresh PG15+ installs.

## §2 — rotate db admin DSN (PR scope: small-to-medium)

\`\`\`
./bootroot rotate db
# bootroot rotate failed
# details: ERROR: permission denied to alter role
# DETAIL: Only roles with the CREATEROLE attribute and the
#         ADMIN option on role \"stepca\" may alter this role.
\`\`\`

Today (beb2e50) \`rotate db\` reads \`ca.json.db.dataSource\` as the admin DSN. But that field holds the **runtime** DSN (the \`stepca\` role bootroot just provisioned), not the admin DSN (\`step\` superuser from \`.env\`). The result is bootroot logging in as \`stepca\` and trying to ALTER \`stepca\` — PostgreSQL refuses because a role cannot self-alter without explicit \`ADMIN OPTION\` on itself.

Workaround used in the integration test:

\`\`\`
PASS=$(grep POSTGRES_PASSWORD .env | cut -d= -f2)
./bootroot rotate db \\
  --db-admin-dsn \"postgresql://step:${PASS}@127.0.0.1:5433/stepca?sslmode=disable\"
\`\`\`

Change:

1. At \`init --enable db-provision\` time, persist the admin DSN bootroot used (the \`step\` superuser DSN) to a new OpenBao KV path, e.g. \`{kv_mount}/data/bootroot/stepca/db_admin\`. This is no more sensitive than the runtime DSN already in \`bootroot/stepca/db\` — they are both protected by the same OpenBao policies.
2. \`rotate db\` reads the admin DSN from KV when \`--db-admin-dsn\` is not supplied. Existing flag remains as override (operators with externally-managed admin credentials).
3. \`init\` without \`db-provision\` (operator-supplied DSNs) does **not** create this KV path; \`rotate db\` continues to require \`--db-admin-dsn\` in that mode (or a clear error when neither flag nor KV is available).

Update beb2e50's behaviour: when \`ca.json\` is unreadable AND no admin DSN is in KV AND \`--db-admin-dsn\` is not given, fail fast with a message naming all three sources (instead of falling through to the interactive DSN prompt with a mis-typed default).

## §3 — EAB prompt: validation, opt-out, recovery (PR scope: medium)

Three sub-changes that travel together because they each address a different stage of the same EAB-input failure mode:

**§3a (automatic)** — Validation on \`init\`'s EAB prompt. Empty input is allowed (no EAB). Non-empty input must satisfy: \`kid\` is non-empty, \`hmac\` is base64url-decodable to >= 16 bytes (typical ACME EAB HMAC). Reject otherwise; re-prompt with the failure reason. The current parser silently accepts \`y\` of length 1 and bootroot-agent only finds out at first ACME issuance, by which time the value is in KV and breaks every service.

**§3b (opt)** — \`--no-eab\` flag on \`bootroot init\`. When set, the EAB prompt is skipped and no EAB material is written to KV. Documented as the recommended setting for OSS step-ca (which does not implement EAB; CHANGELOG entry for 0ba4474 already states this) and for CI flows.

**§3c (opt)** — Re-introduce \`bootroot rotate eab-clear\` (small companion to the now-removed \`rotate eab\`). Behaviour: PUT empty-string \`{kid: \"\", hmac: \"\"}\` to all known EAB KV paths (\`bootroot/agent/eab\`, \`bootroot/services/<svc>/eab\` for every registered service), then signal each affected sidecar via §6's refresh command. After this command the templates render the \`{{ if .Data.data.kid }}...{{ end }}\` branch as empty, agent.toml no longer carries an \`[eab]\` block, and the next bootroot-agent cycle issues certs again.

Without §3c, an operator who hits §3a's validator only on a stale install (i.e. before §3a lands, then upgrades) has no in-tree path to clean up the KV.

## §4 — \`infra install\` port handling (PR scope: small)

Today's experience on a host already running PostgreSQL on 5432:

\`\`\`
./bootroot infra install
# Container bootroot-openbao Started
# Error response from daemon: ports are not available:
#   exposing port TCP 127.0.0.1:5432 -> 127.0.0.1:0:
#   listen tcp4 127.0.0.1:5432: bind: address already in use
# bootroot infra install failed
echo \"POSTGRES_HOST_PORT=5433\" >> .env
docker compose down
./bootroot infra install   # second attempt succeeds
\`\`\`

Two changes:

**§4a (automatic)** — Before invoking \`docker compose up\`, attempt a TCP bind on the to-be-published ports (5432 and any other ports declared by the active compose file). On collision, abort with a message naming the busy port, the responsible PID (best-effort via \`lsof\`/equivalent), and the recommended remediation (\`echo \"POSTGRES_HOST_PORT=N\" >> .env\` if the operator wants bootroot to move; the \`--postgres-host-port\` flag if invoking fresh). Avoid the half-up state where \`bootroot-openbao\` is running while postgres failed to bind.

**§4b (opt)** — \`--postgres-host-port <N>\` flag on \`bootroot infra install\`. Equivalent to setting \`POSTGRES_HOST_PORT\` in \`.env\`, but flag-driven so that scripted bootstraps don't need an out-of-band file edit between commands. Existing \`POSTGRES_HOST_PORT\` env / \`.env\` entry continue to work; the flag overrides them.

Explicitly out of scope (see Non-goals): automatically picking a different port on collision.

## §5 — Resumable / diagnosable init after partial failure (PR scope: medium)

\`\`\`
./bootroot init ...   # fails midway (e.g. DB step)
# rollback runs
./bootroot init ...   # retried
# bootroot init failed
# details: OpenBao KV mount check failed
# details: OpenBao API error (403 Forbidden):
#   {\"errors\":[\"permission denied\"]}
\`\`\`

The first run initialised OpenBao, generated a root token, attempted later steps, failed, and rolled back. Rollback un-staged the secrets dir but **left OpenBao initialised in its volume**. The retried run does not have a saved root token (rollback removed the in-flight artefacts), so it cannot authenticate and bails.

The integration test recovered via \`bootroot clean -y\` (which removes the OpenBao volume) and a clean re-run. That is the right answer for a sandbox; it is the wrong answer for any environment where OpenBao stores state bootroot does not own.

**§5a (automatic)** — On \`init\` startup, before initialising OpenBao, query \`/v1/sys/init\`. If \`initialized: true\` and bootroot has no usable root token (no \`tmp/init-summary.json\` from this directory, no env, no flag), emit a diagnostic naming the three recovery paths:

1. Re-run with \`--root-token-file <path>\` pointing at a saved token from the prior init.
2. \`bootroot clean --openbao-only\` if this OpenBao instance is dedicated to bootroot and may be wiped.
3. Manual OpenBao operator action if the instance is shared.

**§5b (opt)** — \`bootroot clean --openbao-only\` removes \`bootroot-openbao\` and its volume but leaves \`bootroot-postgres\`, \`bootroot-http01\`, \`bootroot-ca\` intact. Useful when only the OpenBao state is fouled.

These two together make the recovery story explicit and bounded; today the only escape is \`clean -y\`, which is too broad for production-shaped setups.

## §6 — Operator-issued sidecar refresh (PR scope: small)

After clearing a KV path that an OpenBao Agent sidecar templates from (e.g. \`bootroot/services/<svc>/eab\`), consul-template's \`{{ with secret \"...\" }}\` block continues to render the previously-cached value until the sidecar process restarts. The integration test recovered via:

\`\`\`
docker restart bootroot-openbao-agent-aice-web-next
\`\`\`

Change: add \`bootroot service openbao-sidecar refresh --service-name <name>\` (sibling of \`openbao-sidecar start\`). Implementation can be as simple as \`docker restart <recorded container_name>\` for the daemon/docker delivery modes; for remote-bootstrap services, emit guidance only (sidecar lives on the remote host).

This is the natural pair to §3c (\`rotate eab-clear\` calls it for each affected service) but is generally useful any time an operator needs to flush a stale render after manual KV maintenance.

## PR scoping note

Each section is independent at the code level:

- §1, §4a, §6 are isolated additions.
- §2 and §5 each touch \`commands/init/steps/orchestrator.rs\` but in non-overlapping branches.
- §3 (a/b/c) cluster around \`init\` prompts + a small new rotate subcommand.

A reviewer can land them as 6 PRs or fewer at their discretion. The non-goals list above must be respected across all PRs to keep the consistent escape-hatch policy.

## References

- Reproduced end-to-end during 2026-05-03 integration test rebuild against \`aice-web-next\` \`main\`. Each section's \"today\" snippet is a verbatim transcript from that run.
- §1: PG15 [public schema CHANGELOG entry][pg15-public].
- §2: beb2e50 (\"Auto-read admin DSN from ca.json in rotate db\") — this issue corrects the path from which it auto-reads.
- §3: 0ba4474 (\"Remove EAB auto-issuance and mandatory enforcement\") — recovery (3c) re-introduces a small, scoped form of the removed \`rotate eab\`.
- §5: 1003956 (\"Stop step-ca before init to fix permission race\") sits adjacent to this code path and may inform the diagnostic wording.

[aicers/aice-web-next]: https://github.com/aicers/aice-web-next
[pg15-public]: https://www.postgresql.org/docs/release/15.0/

sehkone:
## Summary

Groups six independent recovery papercuts surfaced during the 2026-05-03 integration-test rebuild against `bootroot 06778e2`. None are bugs in a single command on their own; they share the shape of "bootroot's assumptions about surrounding state did not match reality, with no in-tree path back to a working state". The non-goals from the issue (no implicit port fallback, no automatic OpenBao wipe, no auto-detection of CA type) are preserved.

- **§1 db-provision schema grants** — after `CREATE ROLE / CREATE DATABASE`, open a second connection to the target DB and `GRANT CREATE, USAGE ON SCHEMA public TO <db_user>`. Closes the PG15+ step-ca crashloop with `permission denied for schema public`. Ownership stays with `postgres`. The grant runs on every `db-provision` call (idempotent), so an operator who hit the original failure on a stale install can recover by re-running `init --enable db-provision` on the fixed binary.
- **§2 rotate-db admin DSN** — `init` persists the admin DSN to `bootroot/stepca/db_admin` (separate KV path, not readable by runtime/sidecar policies). `rotate db` reads from KV (or the explicit `--db-admin-dsn` flag, which always wins) and **does not** fall back to `ca.json.db.dataSource` — that field stores the runtime (`stepca`) DSN and using it as an admin DSN is the original §2 self-ALTER bug. When neither source is available the command fails fast with a message naming both recovery paths. In the bundled same-role topology (`POSTGRES_USER` and `--db-user` both `step`), `provision_db_sync` ALTERs the admin's own password during rotation, so `rotate db` also rewrites `bootroot/stepca/db_admin` with the post-ALTER credential — without that rewrite the *next* `rotate db` would read a stale DSN and fail authentication, leaving the new automatic §2 path working only once. The same rebuild also runs at the end of `init`'s post-bootstrap `.env` password rotation so the persisted admin DSN reflects the post-rotation password before any `rotate db` ever runs; that rotation now also resolves the host-side Postgres port from the compose dir's `.env` / process env (same precedence Docker Compose uses for `${POSTGRES_HOST_PORT:-5433}`) instead of reusing the compose-internal port from `ca.json`, which would otherwise silently skip the rotation on the new 5433 default. The auto-derived admin DSN built when `--db-admin-dsn` is not supplied (from `POSTGRES_USER`/`POSTGRES_PASSWORD` in the compose `.env`) follows the same `${POSTGRES_HOST_PORT:-5433}` precedence for its port and defaults its host to `127.0.0.1` (host-reachable, since `provision_db_sync` runs from the host) so the documented `init --enable db-provision` path after `infra install` works on the new 5433 default — and on a non-default `--postgres-host-port` — without a manual `--db-admin-dsn`.
- **§3 EAB hardening** — (a) prompt validates `kid` (non-empty after trim) and `hmac` (base64url-decodable to ≥16 bytes), re-prompting the missing field on blank input; (b) `bootroot init --no-eab` skips the prompt; (c) `bootroot rotate eab-clear` writes empty `{kid: "", hmac: ""}` unconditionally to the agent path and every state-derived service path (KV v2 PUTs are idempotent, so creating a missing path is safe — and necessary, because consul-template treats a missing secret as a transient error and would never render the empty branch). Each affected sidecar is refreshed via §6's command.
- **§4 infra install port handling** — (a) TCP-bind preflight on every host port the active compose stack publishes (postgres, openbao, step-ca, bootroot-http01) before `docker compose up`, with a best-effort `lsof` PID/command hint and per-service remediation. The preflight always checks the **localhost** ports because `infra install` runs `docker compose up` against the **base** compose file only — `--openbao-bind` / `--http01-admin-bind` write override files but those are not layered into the install-time `up` (they are applied first by `infra up` / `init`), so the install-time bind stays on `127.0.0.1` regardless of any recorded override intent. (b) `--postgres-host-port <N>` flag overrides `.env` *and* the docker compose subprocess environment so the flag wins over a shell-inherited `POSTGRES_HOST_PORT` (Docker Compose otherwise prefers shell env over `.env`, recreating the half-up collision the flag exists to avoid). (c) docker-compose.yml default published port moves from 5432 to 5433 (internal `postgres:5432` container address unchanged).
- **§5 partial-init detection** — before initialising OpenBao, query `sys/init`; if initialised but no usable root token exists, emit a diagnostic naming the three recovery paths instead of bubbling up the opaque `403`. `bootroot clean --openbao-only` removes only the `bootroot-openbao` container and volume.
- **§6 sidecar refresh** — `bootroot service openbao-sidecar refresh --service-name <name>` docker-restarts the recorded sidecar container so consul-template re-reads its KV sources. `RemoteBootstrap` services emit operator guidance only. §3c reuses this per affected service.

Closes #588

## Test plan

- [x] `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test` all pass.
- [x] **§1**: `provision_db_sync` runs `GRANT CREATE, USAGE ON SCHEMA public TO <role>` against the target DB on every call (no longer gated on first-create), so re-running `init --enable db-provision` on a stale install is now a recovery path; verified by `tests/db_provision_integration.rs` against a live `postgres:18` container.
- [x] **§2**: `init` persists the admin DSN to `bootroot/stepca/db_admin`; `rotate db` reads from KV when `--db-admin-dsn` is not supplied and **errors out** when neither is available (no `ca.json` fallback). Covered by `resolve_db_admin_dsn_prefers_kv_admin`, `resolve_db_admin_dsn_errors_when_neither_flag_nor_kv_present`, `resolve_db_admin_dsn_cli_flag_overrides_kv`, and `resolve_db_admin_dsn_uses_cli_arg`. Round-3 follow-up: `rotate db` rewrites the persisted KV admin DSN with the post-ALTER credential when admin user matches the runtime user being rotated, so a second consecutive `rotate db` succeeds. Covered by `kv_admin_dsn_to_persist_rebuilds_for_same_role`, `kv_admin_dsn_to_persist_returns_none_for_distinct_role`, and `kv_admin_dsn_to_persist_returns_none_when_flag_supplied`. Round-5 follow-up: `init`'s post-bootstrap `.env` password rotation (a) builds the bootstrap admin DSN with the host-side Postgres port resolved from compose dir's `.env`/process env (same precedence Docker Compose uses for `${POSTGRES_HOST_PORT:-5433}`) so the new 5433 default does not silently skip the rotation, and (b) mirrors `rotate db`'s same-role rebuild by rewriting `bootroot/stepca/db_admin` after the ALTER. Covered by `rebuilt_admin_dsn_for_kv_rebuilds_for_same_role`, `rebuilt_admin_dsn_for_kv_returns_none_for_distinct_role`, and `rebuilt_admin_dsn_for_kv_returns_none_when_path_absent`. Round-6 follow-up: `build_admin_dsn_from_env` defaults the host to `127.0.0.1` and the port to `resolve_postgres_host_port(compose_dir)` so the auto-derived admin DSN used by `init --enable db-provision` after `infra install` is host-reachable on the new 5433 default and on a non-default `--postgres-host-port`. Covered by `build_admin_dsn_from_env_uses_host_reachable_defaults`, `build_admin_dsn_from_env_resolves_postgres_host_port_from_dotenv`, and `build_admin_dsn_from_env_postgres_port_overrides_host_port`.
- [x] **§3a**: `validate_eab` rejects `kid="y"`-style HMACs and short base64url decodes; covered by `validate_eab_rejects_single_char_hmac`, `validate_eab_rejects_empty_kid`, `validate_eab_accepts_32_byte_base64url`. The prompt loop re-prompts on blank input rather than coercing to "no EAB".
- [x] **§3b**: `bootroot init --no-eab` skips the EAB prompt and writes no EAB material; CLI surface verified (`--no-eab` present, `conflicts_with_all = ["eab_kid", "eab_hmac"]` enforced).
- [x] **§3c**: `bootroot rotate eab-clear` enumerates services from state.json and writes empty `{kid:"", hmac:""}` **unconditionally** to agent + per-service KV paths, then refreshes each sidecar via §6's `refresh_service_sidecar`. CLI surface verified.
- [x] **§4a**: `preflight_host_port` aborts with an actionable message naming the busy port, the per-service remediation (env var + CLI flag for postgres; "free port N or stop the listener" for the others), and a best-effort `lsof` PID/command hint. `preflight_compose_published_ports` always preflights the localhost ports for every core service in `args.services` because `infra install` only runs the base compose file (no override layering), matching the actual install-time bind. Covered by `preflight_host_port_aborts_when_already_bound`, `preflight_host_port_succeeds_when_free`, and `preflight_compose_published_ports_checks_openbao_localhost_during_install`.
- [x] **§4b**: `bootroot infra install --postgres-host-port <N>` overrides `.env` and the docker compose subprocess environment (via the new `run_docker_with_env` helper); CLI surface verified (`--postgres-host-port` flag present).
- [x] **§4c**: `docker-compose.yml` published default is `${POSTGRES_HOST_PORT:-5433}:5432` and `DEFAULT_POSTGRES_HOST_PORT=5433`; live `docker compose config` confirms `published: "5433"` with no env var, and `published: "5432"` when explicitly set.
- [x] **§5a**: `diagnose_partial_init` queries `/v1/sys/init` and bails with the three-path diagnostic naming the URL, `--root-token`, and `--openbao-only` when initialised without a usable token; covered by `diagnose_partial_init_bails_when_initialized_without_token`, `..._passes_when_token_supplied`, `..._passes_when_not_initialized`.
- [x] **§5b**: `bootroot clean --openbao-only` removes only the openbao container and its named volumes (`openbao-data`, `openbao-audit`), discovering the compose project from the container's `com.docker.compose.project` label and falling back to the compose-dir basename. Covered by `resolve_compose_project_*` unit tests.
- [x] **§6**: `bootroot service openbao-sidecar refresh --service-name <svc>` runs `docker restart bootroot-openbao-agent-<svc>` for `LocalFile`; emits operator guidance for `RemoteBootstrap` (covered by `refresh_remote_bootstrap_returns_ok_without_docker_call`). CLI surface verified.
- [x] `scripts/preflight/run-all.sh` paths exercised in CI via the green Docker E2E matrix (`local-hosts`, `local-no-hosts`, `local-no-hosts-custom-project`, `local-no-hosts-external-openbao`, `local-no-hosts-host-daemon`, `local-no-hosts-openbao-missing`, `remote-hosts`, `remote-no-hosts`, `rotation`) plus `Unit & CLI Smoke` — all passing on this branch.


sehkone:
**[Reviewer Round 1]**

Findings:

- `src/commands/rotate/db.rs:143`: `rotate db` still falls back to `ca.json` and treats `db.dataSource` as an admin DSN when `--db-admin-dsn` is absent and `bootroot/stepca/db_admin` is missing or unreadable. That preserves the exact §2 failure mode for operator-supplied DSNs, pre-existing installs without the new KV key, and AppRole runs that 403 on `db_admin`: the value in `ca.json` is the runtime `stepca` DSN, so `provision_db_sync` can still log in as the role it is trying to alter. The issue explicitly says init without `db-provision` should continue to require `--db-admin-dsn` (or fail clearly when neither flag nor KV is available), not use `ca.json` as a legacy fallback. The tests around `resolve_db_admin_dsn_reads_from_ca_json_when_flag_absent` currently encode the old bug; this should become an error path or an explicit prompt path that does not default to the runtime DSN.

- `src/commands/infra.rs:319`: `--postgres-host-port` is documented and described as overriding `.env` and the process environment, but the implementation only writes/updates `.env` and then runs `docker compose` with the inherited environment. Docker Compose gives shell environment variables precedence over `.env`, so `POSTGRES_HOST_PORT=5432 bootroot infra install --postgres-host-port 5433` will preflight 5433 at `src/commands/infra.rs:337` but `docker compose up` at `src/commands/infra.rs:367` can still publish 5432. That recreates the half-up/collision path the flag is meant to avoid. The selected flag value needs to be passed into the compose process environment, or the inherited `POSTGRES_HOST_PORT` needs to be overridden for the docker invocations.

- `src/commands/rotate/eab_clear.rs:40` and `src/commands/rotate/eab_clear.rs:58`: `rotate eab-clear` only writes empty `{kid:"", hmac:""}` when the EAB KV path already exists. §3c says to PUT the empty value to the agent path and every registered service path enumerated from state. Skipping missing paths matters for stale installs and partial states: a service whose `bootroot/services/<svc>/eab` path is absent will still leave consul-template on a missing secret after the refresh, so the command is not a complete recovery path. This should unconditionally write the empty agent path and each state-derived service path, then refresh the sidecars.

- `src/commands/infra.rs:340`: the port preflight only checks PostgreSQL. The issue’s §4a asks for a pre-bind diagnostic on 5432/5433 and any other ports declared by the active compose file, and this compose file still publishes OpenBao (`docker-compose.yml:6`), step-ca (`docker-compose.yml:49`), and HTTP-01 (`docker-compose.yml:72`). Conflicts on those ports will still be discovered by `docker compose up` after earlier services may already have started, which is the half-completed state §4a is trying to avoid. The diagnostic also does not attempt the best-effort PID lookup requested in the issue.

- `src/db.rs:346`: the schema grant only runs when the role or database was just created. If an operator hit the old PG15+ failure before this PR, the role and DB can already exist while the `public` schema grant is still missing; re-running `init --enable db-provision` on the fixed binary will skip the grant and step-ca can crashloop again. `GRANT CREATE, USAGE ON SCHEMA public TO <role>` is idempotent, so it should run whenever db-provision is asked to manage that target role/database, not only on first creation.

I did not run the suite; this review is based on the diff and the issue contract. The PR has good coverage in several new units, but a few tests currently assert the behavior above instead of the requested recovery semantics.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the review. All five items accepted; new commit 84b2cc1.

- **`src/commands/rotate/db.rs:143` (ca.json fallback) — Fixed.** Removed the `ca.json.db.dataSource` legacy fallback entirely. Resolution is now: `--db-admin-dsn` → `bootroot/stepca/db_admin` in KV → hard error naming both recovery paths. The old `resolve_db_admin_dsn_reads_from_ca_json_when_flag_absent` test was rewritten as `resolve_db_admin_dsn_errors_when_neither_flag_nor_kv_present`, and the now-unused `read_ca_json_dsn_if_present` / interactive-prompt fallthrough were dropped. CHANGELOG and `docs/{en,ko}/cli.md` updated to reflect the new resolution order and explain why `ca.json` is intentionally not consulted.

- **`src/commands/infra.rs:319` (`--postgres-host-port` not winning over shell env) — Fixed.** Added `run_docker_with_env` and used it for both `docker compose pull` and `docker compose up --build` in `run_infra_install`. When `--postgres-host-port <N>` is set, the flag value is injected into the compose subprocess environment so it overrides any inherited `POSTGRES_HOST_PORT`, defeating Docker Compose's "shell env wins over `.env`" precedence. The preflight already used the flag value, so the flag now wins consistently across preflight + compose up. CHANGELOG entry expanded to spell out the precedence detail.

- **`src/commands/rotate/eab_clear.rs:40,58` (skip-on-missing) — Fixed.** Removed both `kv_exists` guards. The empty `{kid:"", hmac:""}` value is now written unconditionally to `bootroot/agent/eab` and to every `bootroot/services/<svc>/eab` derived from `state.services`. KV v2 PUTs are idempotent and create the path if missing, which is exactly the recovery path needed for stale installs and partial-state services where consul-template is otherwise stuck on a missing-secret error.

- **`src/commands/infra.rs:340` (preflight only checks PostgreSQL) — Fixed.** Replaced the single `preflight_host_port` call with `preflight_compose_published_ports`, which iterates `args.services` and pre-binds the host port for each core service that publishes one (postgres @ `POSTGRES_HOST_PORT`, openbao @ 8200, step-ca @ 9000, bootroot-http01 @ 8080). Services with an active non-loopback bind override are skipped because the override moves the bind address (and the `validate_*` gates already rejected unsafe configurations there). Added `best_effort_listening_pid` which shells out to `lsof -nP -iTCP:PORT -sTCP:LISTEN -Fpcn` and surfaces `pid N (cmd)` in the diagnostic when available, silently degrading on systems without `lsof`. New `preflight_compose_published_ports_skips_overridden_services` test covers the override-skip path.

- **`src/db.rs:346` (grant gated on first-create) — Fixed.** Removed the `if report.role_created || report.db_created` gate around `grant_public_schema`. The `GRANT CREATE, USAGE ON SCHEMA public TO <role>` now runs on every `provision_db_sync` call. The grant is idempotent, the second admin connection is the price of the recovery path, and operators stuck with a pre-fix role/DB combination can now recover by re-running `init --enable db-provision` on the patched binary. The PR description test-plan entry for §1 was reworded to reflect the new "every call" semantics.

CI is green locally (`cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, `cargo test`).

sehkone:
**[Reviewer Round 2]**

Round 1 follow-up: the five prior findings are resolved in the updated diff. I verified the `ca.json` admin fallback is gone, `--postgres-host-port` is passed into the compose subprocess environment, `rotate eab-clear` now writes missing KV paths, the install preflight covers the core published ports, and the public-schema grant runs unconditionally.

New findings:

- `src/commands/init/steps/database.rs:40` / `src/commands/init/steps/database.rs:56` and `src/commands/init/steps/orchestrator.rs:234`: the admin DSN persisted to `bootroot/stepca/db_admin` can be stale as soon as `init --enable db-provision` finishes when the provisioning admin role is also the runtime role. `resolve_db_dsn_for_init` clones `inputs.admin_dsn` into `admin_dsn_for_kv` before calling `provision_db_sync`; `provision_db_sync` then runs `ALTER ROLE <db_user> WITH PASSWORD <db_password>` (`src/db.rs:314`), and the latest commit already acknowledges the `admin_user == db_user` case for the second grant connection. But the DSN returned for KV still contains the pre-ALTER admin password and is persisted unchanged. This is the topology used by the E2E scripts (`--db-user step` with the bundled `.env` admin user `step`), so a later `rotate db` without `--db-admin-dsn` will read the new KV path, try the old password, and fail authentication instead of closing §2. The persisted admin DSN needs to be rebuilt with `db_password` when `admin.user == inputs.db_user` (or otherwise derived after provisioning from the effective credentials), and covered by a unit test for the same-role case.

- `src/commands/rotate/eab_clear.rs:64`: `rotate eab-clear` reports success even when a local sidecar refresh fails. The stale-render failure in §6 is fixed only after `refresh_service_sidecar` restarts the local sidecar; currently that error is downgraded to a warning and the command still prints “EAB clear completed”, leaving the operator with old rendered EAB material and a zero exit status. Remote-bootstrap guidance can remain non-fatal because `refresh_service_sidecar` returns `Ok(())` for that branch, but LocalFile restart failures should make `eab-clear` fail (ideally after attempting the remaining services and summarizing which refreshes failed).

I did not rerun the suite; this round is based on the updated diff and the issue contract.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the second pass. Both findings accepted; new commit d720753.

- **`src/commands/init/steps/database.rs:40,56` / `src/commands/init/steps/orchestrator.rs:234` (stale admin DSN persisted to KV when `admin.user == db_user`) — Fixed.** Added `effective_admin_dsn_for_kv` and used it to derive `admin_dsn_for_kv` before `provision_db_sync` runs. When the admin DSN's user equals `db_user`, the helper rebuilds the DSN with `db_password` (the post-ALTER credential); otherwise it returns the input unchanged. The persisted `bootroot/stepca/db_admin` value now stays usable across the bundled E2E topology (`POSTGRES_USER=step` + `--db-user step`) so a later `rotate db` without `--db-admin-dsn` reads the new password rather than authenticating with the pre-ALTER one. Covered by `test_effective_admin_dsn_for_kv_rebuilds_when_same_role` (asserts post-ALTER password is what reaches KV) and `test_effective_admin_dsn_for_kv_unchanged_when_distinct_role` (asserts no rewrite when admin and runtime are separate roles). The Round 1 CHANGELOG entry already mentioned the same-role topology for the second-grant connection inside `provision_db_sync`; it now also covers the persisted-to-KV path.

- **`src/commands/rotate/eab_clear.rs:64` (refresh failures silently downgraded) — Fixed.** The loop now collects per-service refresh errors instead of warning-and-continuing-as-success: every service's KV write and refresh is still attempted (so no service is left with stale KV because an earlier sidecar restart failed), but if any LocalFile refresh fails the command exits non-zero with the count and the names of the services whose sidecars are still rendering the old EAB plus the `docker restart bootroot-openbao-agent-<svc>` recovery hint. RemoteBootstrap services return `Ok(())` without shelling out (per `refresh_remote_bootstrap_returns_ok_without_docker_call`), so they never appear in the failure list. Pre-existing eab-clear tests would require mocking `OpenBaoClient` and the docker call to exercise the new aggregation; I left that as a follow-up rather than a one-off scaffolding here, but the failure-collection logic is straightforward and covered by reading code review. CHANGELOG and `docs/{en,ko}/cli.md`'s `rotate eab-clear` sections now spell out the "fatal on local-file refresh failure" behaviour.

CI is green locally (`cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, `cargo test`).

sehkone:
**[Reviewer Round 3]**

Round 2 follow-up: both prior items are resolved in the updated diff. `effective_admin_dsn_for_kv` now persists the post-ALTER password during `init --enable db-provision` for the same-role topology, and `rotate eab-clear` now treats local sidecar refresh failures as fatal after attempting the remaining services.

New finding:

- `src/commands/rotate/db.rs:44` / `src/commands/rotate/db.rs:64` / `src/commands/rotate/db.rs:91`: the same-role admin DSN still becomes stale after the first successful `rotate db`. In the bundled topology, `bootroot/stepca/db_admin` is `step:<current>` and `ca.json.db.dataSource` is also user `step`. `rotate db` reads the KV admin DSN, calls `provision_db_sync`, and `provision_db_sync` runs `ALTER ROLE step WITH PASSWORD <new>`; the runtime DSN at `bootroot/stepca/db` is then updated, but `bootroot/stepca/db_admin` is left with the old password. The next `rotate db` without `--db-admin-dsn` will read that stale KV admin DSN and fail PostgreSQL authentication, so the new automatic §2 path only works once. This needs the same effective-credential treatment as the Round 2 init fix: when the admin DSN user matches the runtime DB user being rotated, update `bootroot/stepca/db_admin` to the admin DSN rebuilt with the new password after the DB provisioning succeeds (at least for the KV-backed path, and preferably whenever a persisted db_admin path exists). A regression test should cover two consecutive same-role rotations or the helper-level equivalent.

I did not rerun the suite; this round is based on the updated diff and the issue contract.

Review verdict: not approved until the repeat-rotation credential path is fixed.

sehkone:

sehkone:
**[Author Round 3]**

Thanks for the catch — the round 2 fix only closed half the loop. Pushed `a191c26` to address the finding.

- **Stale persisted KV admin DSN after first same-role rotate db** — **Fixed.** `src/commands/rotate/db.rs:91` now mirrors the round 2 init treatment on the rotation side: after `provision_db_sync` succeeds, when the persisted KV admin DSN's user matches the runtime user being rotated (the bundled `step` same-role topology), `bootroot/stepca/db_admin` is rewritten with the post-ALTER credential, preserving the original (compose-internal) host/port form. The rewrite triggers only for the KV-backed path; `--db-admin-dsn` overrides do not touch KV (operators with externally-managed admin credentials should not have their persisted DSN rewritten under them). The `effective_admin_dsn_for_kv` helper graduates from `commands/init/steps/database.rs` to `bootroot::db` so both call sites share the same implementation. Three new unit tests cover the same-role rebuild (`kv_admin_dsn_to_persist_rebuilds_for_same_role`), the distinct-role no-op (`kv_admin_dsn_to_persist_returns_none_for_distinct_role`), and the `--db-admin-dsn` override no-op (`kv_admin_dsn_to_persist_returns_none_when_flag_supplied`). CHANGELOG and docs/{en,ko}/cli.md updated.

`cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test` all pass locally; rotation E2E coverage remains via the existing CI matrix (`rotation` arm).

sehkone:
**[Reviewer Round 4]**

Round 3 follow-up: the prior same-role repeat-rotation finding is resolved in the updated diff. `rotate db` now rewrites `bootroot/stepca/db_admin` after a KV-backed same-role rotation, and the helper-level tests cover the same-role, distinct-role, and explicit-flag cases.

New finding:

- `src/commands/infra.rs:341` / `src/commands/infra.rs:389`: the §4 preflight skips OpenBao and HTTP-01 when a non-loopback bind intent is present, but `infra install` does not apply those override files in the `docker compose up` it actually runs. The code writes the override and explicitly says it is “NOT applied yet” (`src/commands/infra.rs:272`), then invokes only `docker compose -f <base> up --build -d ...` at `src/commands/infra.rs:389`. The base compose file still publishes `127.0.0.1:8200:8200` and `127.0.0.1:8080:8080`, so `bootroot infra install --openbao-bind 192.168.1.10:8200` or `--http01-admin-bind 192.168.1.10:8080` can still hit a localhost port collision only after Compose has started earlier services. That reintroduces the half-up state §4a is meant to prevent. The skip test at `src/commands/infra.rs:1368` is encoding the wrong install-time behavior; either preflight the base localhost ports for the install command, or include the override files in the install-time compose invocation and preflight the actual override bind addresses.

Review verdict: not approved until the install-time preflight matches the compose invocation it is guarding.

sehkone:

sehkone:
**[Author Round 4]**

Thanks for the catch — the override-skip branch was reasoning about `infra up` / `init` semantics while sitting in the `infra install` code path. New commit `2b92c7d`.

- **`src/commands/infra.rs:341` / `src/commands/infra.rs:389` (preflight skips localhost ports the install actually binds) — Fixed.** Removed the `openbao_bind_override` / `http01_admin_bind_override` parameters from `preflight_compose_published_ports` and made the openbao / http01 cases unconditional. Rationale: `run_infra_install` invokes `docker compose -f <base> up --build -d` only — the openbao / http01 override files written above the preflight are explicitly NOT applied at install time (the comment at `src/commands/infra.rs:272`–`276` calls this out, and the install-time `up_args` at the new `src/commands/infra.rs:392` carries no `-f <override>`). So the install-time bind is always on `127.0.0.1:8200` / `127.0.0.1:8080` regardless of the recorded bind intent, and skipping those services left the half-up state §4a is meant to prevent. The override is first layered in by `run_infra_up` (`src/commands/infra.rs:103`–`114`) and by `init`'s post-TLS swap, which was the surface my earlier reasoning was applying to the wrong command. Inline doc comment on `preflight_compose_published_ports` now spells out the install-only call site and the "no override layering at install time" invariant so the next reader doesn't repeat the mistake.

- **Test rewritten — `preflight_compose_published_ports_skips_overridden_services` → `preflight_compose_published_ports_checks_openbao_localhost_during_install`.** The old test was encoding the install-time bug as expected behaviour. The new test asserts the localhost branch is checked unconditionally: when 8200 is free the preflight passes, when 8200 is busy the preflight aborts and names the port. The `--openbao-bind`-was-set parameter is gone from the call site, matching the new function signature.

- **Docs / CHANGELOG.** `CHANGELOG.md` §4a entry, `docs/en/cli.md`, and `docs/ko/cli.md` updated to (a) extend the preflight description from "PostgreSQL only" to all four core ports (postgres/openbao/step-ca/bootroot-http01) and (b) explicitly state that the install-time preflight always checks localhost because the bind override files are not layered in until `infra up` / `init`. The PR description's §4 bullet and §4a test-plan entry were rewritten on the same basis (the prior wording carried over the now-incorrect "services with an active non-loopback bind override are skipped" claim).

CI / local checks: `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test` all pass locally; the rotation and remote-hosts E2E arms remain green via the existing matrix.

sehkone:
**[Reviewer Round 5]**

Round 4 follow-up: the prior install-time preflight finding is resolved. `preflight_compose_published_ports` now always checks the base-compose localhost ports for OpenBao and HTTP-01, and the rewritten test no longer encodes the override-skip behavior.

New findings:

- `src/commands/init/steps/orchestrator.rs:696` still builds the post-init DB password-rotation admin DSN with `localhost:<parsed ca.json port>`. After this PR's §4c change, `ca.json` correctly contains the compose-internal runtime DSN (`postgres:5432`), while the default host-side published port is now 5433 (`docker-compose.yml:34`). So on a default `infra install` with no explicit `POSTGRES_HOST_PORT`, this helper tries `localhost:5432`, fails authentication/connection, and silently skips the `.env` password rotation via the warning path at `src/commands/init/steps/orchestrator.rs:722`. This is the same host-vs-compose translation problem the PR fixed for `rotate db`; this init helper should resolve the host port from the compose directory (or route the DSN through the existing host-runtime translator) instead of reusing the compose-internal port from `ca.json`. The current E2E scripts pin `POSTGRES_HOST_PORT=5432`, so they will not catch the default-5433 regression.

- `src/commands/init/steps/orchestrator.rs:743`: when `maybe_rotate_env_db_password` does succeed, it updates only `bootroot/stepca/db`. It does not update the newly introduced `bootroot/stepca/db_admin` path that was written earlier at `src/commands/init/steps/orchestrator.rs:234`. In the same-role topology (`POSTGRES_USER == db_user`), this has the same stale-admin-DSN shape as the Round 3 `rotate db` bug: `provision_db_sync` changes the admin/runtime role password, the runtime DSN is rewritten with the new password, but the persisted admin DSN still carries the pre-rotation password. The fix should mirror `rotate db`'s KV-backed same-role rewrite so the automatic §2 path works after an init run that performs this final password rotation.

Review verdict: not approved until the init-time password-rotation path uses the new host-port default correctly and keeps `db_admin` in sync when it changes the same role's password.

sehkone:

sehkone:
**[Author Round 5]**

Both findings accepted; new commit `459cadb`.

- **`src/commands/init/steps/orchestrator.rs:696` (admin DSN built with `localhost:<compose-internal port>`) — Fixed.** Replaced the hard-coded `parsed.port` (which is 5432 because `ca.json`'s DSN was rewritten through `for_compose_runtime`) with `bootroot::db::resolve_postgres_host_port(compose_dir)`. That helper is the same one `for_host_runtime` uses, with the precedence Docker Compose itself applies to `${POSTGRES_HOST_PORT:-5433}` (process env → `compose_dir/.env` → 5433 default). On the new default-5433 install, the bootstrap admin connection now goes to `localhost:5433` and the `.env` password rotation actually runs instead of silently skipping via the warning path. Inline comment spells out why `parsed.port` is wrong here so the next reader does not regress.

- **`src/commands/init/steps/orchestrator.rs:743` (post-init password rotation does not sync `bootroot/stepca/db_admin`) — Fixed.** After writing the new runtime DSN to `PATH_STEPCA_DB`, the rotation now mirrors the Round 3 `rotate db` same-role rewrite: read the persisted `bootroot/stepca/db_admin`, and when its user matches the runtime user just rotated, write it back with the post-ALTER password (host/port preserved). Implemented as `rebuilt_admin_dsn_for_kv`, which is best-effort (KV read errors degrade to "no rewrite needed") because the primary work has already succeeded; the worst case if the helper falls through is one extra `--db-admin-dsn` on the next `rotate db`. The KV write itself is also best-effort (`let _ = ...`) for the same reason.

- **Tests.** Added three new tests for `rebuilt_admin_dsn_for_kv` using wiremock-mocked OpenBao responses:
  - `rebuilt_admin_dsn_for_kv_rebuilds_for_same_role` — persisted user `step`, runtime user `step`, asserts the rebuilt DSN swaps the password and preserves the compose-internal `postgres:5432` host/port.
  - `rebuilt_admin_dsn_for_kv_returns_none_for_distinct_role` — persisted user `step`, runtime user `stepca`, asserts no rewrite (the admin DSN is not stale).
  - `rebuilt_admin_dsn_for_kv_returns_none_when_path_absent` — 404 on `kv_exists`, asserts no rewrite.

- **Docs / CHANGELOG.** `CHANGELOG.md` `bootroot init persists the admin DSN ...` entry extended to cover the init-time post-rotation rebuild and the host-port resolution. `docs/en/cli.md` and `docs/ko/cli.md` updated with the same two clarifications. PR description's §2 bullet and §2 test-plan entry updated.

CI / local checks: `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test` all pass locally.

sehkone:
**[Reviewer Round 6]**

Round 5 follow-up: both prior findings are resolved. `maybe_rotate_env_db_password` now resolves the host-side Postgres port from the compose dir instead of `ca.json`, and it attempts to keep `bootroot/stepca/db_admin` in sync after the same-role post-init password rotation.

New finding:

- `src/commands/init/steps/database.rs:145` still builds the provisioning admin DSN from `.env` using only `POSTGRES_HOST` / `POSTGRES_PORT`, defaulting to `postgres:5432`. That is the compose-internal address, but `provision_db_sync` runs from the host. After this PR's §4c default move, `bootroot infra install` writes `.env` with `POSTGRES_USER`, `POSTGRES_PASSWORD`, and `POSTGRES_DB` (and only writes `POSTGRES_HOST_PORT` when `--postgres-host-port` is passed; see `src/commands/infra.rs:313`), so the documented `bootroot init --enable db-provision` path after `infra install` has no `POSTGRES_HOST` / `POSTGRES_PORT`. The result is an admin DSN of `postgresql://step:<pass>@postgres:5432/postgres?...`, which ignores the new `${POSTGRES_HOST_PORT:-5433}` published default and also ignores an explicit `.env` `POSTGRES_HOST_PORT=...`. The Round 5 fix closed this host-vs-compose bug for the later `.env` password rotation, but the earlier db-provision call can still fail before reaching the new PG15 schema grant. The admin DSN auto-discovery for db-provision should use the host-side published port (`POSTGRES_HOST_PORT` / `resolve_postgres_host_port(compose_dir)`, default 5433) and a host-reachable address when deriving credentials from the compose `.env`; tests should cover the no-manual-DSN path after `infra install`, including a non-default `--postgres-host-port` value.

Review verdict: not approved until the init db-provision admin DSN path is aligned with the new 5433 / `POSTGRES_HOST_PORT` behavior.

sehkone:

sehkone:
**[Author Round 6]**

Finding accepted; new commit `396bb04`.

- **`src/commands/init/steps/database.rs:145` (auto-derived admin DSN defaults to `postgres:5432`) — Fixed.** `build_admin_dsn_from_env` now defaults the host to `DB_HOST_RUNTIME_HOST` (`127.0.0.1`) and resolves the port via `bootroot::db::resolve_postgres_host_port(compose_dir)` — the same `${POSTGRES_HOST_PORT:-5433}` precedence Docker Compose uses (process env → `compose_dir/.env` → 5433). `provision_db_sync` runs from the host (it shells out via the `postgres` crate, not from inside the compose network), so the prior `postgres:5432` defaults could never authenticate after the §4c default move. `POSTGRES_HOST` / `POSTGRES_PORT` remain explicit overrides for operator-supplied topologies (and `POSTGRES_PORT`, when set, still wins over the resolved `POSTGRES_HOST_PORT` — that env var is the historical operator override and stays authoritative). Threaded `compose_dir` through `resolve_db_dsn_for_init` and `resolve_db_provision_inputs` so the helper can read the compose `.env`; the orchestrator already had the path computed.

- **Tests.** Three new tests in `src/commands/init/steps/database.rs`:
  - `build_admin_dsn_from_env_uses_host_reachable_defaults` — `infra install` writes only `POSTGRES_USER`/`POSTGRES_PASSWORD` (the documented post-§4c topology); asserts the auto-derived DSN is `127.0.0.1:5433`.
  - `build_admin_dsn_from_env_resolves_postgres_host_port_from_dotenv` — non-default `--postgres-host-port` produces `POSTGRES_HOST_PORT=6543` in `compose_dir/.env`; asserts the auto-derived DSN port is `6543`.
  - `build_admin_dsn_from_env_postgres_port_overrides_host_port` — explicit `POSTGRES_PORT=7777` (operator-supplied) wins over a `POSTGRES_HOST_PORT=6543` in `.env`; asserts port is `7777`.

- **Docs / CHANGELOG.** `CHANGELOG.md` §2 entry extended to cover the auto-derived admin DSN's host-reachable defaults and the host vs. port precedence. `docs/en/cli.md` and `docs/ko/cli.md` updated with the same clarification. PR description's §2 bullet and §2 test-plan entry updated.

CI / local checks: `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test` all pass locally.

sehkone:
**[Reviewer Round 7]**

Round 6 follow-up: resolved. I verified that `resolve_db_dsn_for_init` now passes the compose directory into the db-provision input path, and `build_admin_dsn_from_env` now derives a host-reachable admin DSN (`127.0.0.1` plus `resolve_postgres_host_port(compose_dir)`) instead of the compose-internal `postgres:5432`. The new tests cover the default 5433 path, `.env` `POSTGRES_HOST_PORT`, and explicit `POSTGRES_PORT` override.

I also re-checked the broader §1-§6 surfaces touched by this PR and did not find any remaining blocking issues from the prior rounds.

Non-blocking cleanup: `src/db.rs` still has a Rustdoc line describing Docker Compose precedence for `${POSTGRES_HOST_PORT:-5432}` even though the compose default is now 5433. The implementation is correct (`DEFAULT_POSTGRES_HOST_PORT = 5433`); this is just a stale comment worth fixing when touching the file next.

Review verdict: no objections from me.

sehkone:
[Review Verdict Round 7: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Harden bootstrap and rotation against env variance
```

**Body**

```text
Group six independent recovery papercuts hit during the 2026-05-03
integration-test rebuild against bootroot 06778e2. None are bugs in
a single command on their own; they share the shape of "bootroot's
assumptions about surrounding state did not match reality, with no
in-tree path back to a working state". The consistent escape-hatch
policy across sections (no implicit port fallback, no automatic
OpenBao wipe, no auto-detection of CA type) is preserved.

§1 db-provision schema grants. After CREATE ROLE / CREATE DATABASE,
open a second connection to the target DB and grant CREATE, USAGE
on schema public to the new role. PG15+ removed the implicit grant
on public, so step-ca crashlooped with permission denied for schema
public on first CREATE TABLE. Ownership stays with postgres,
matching aice-web-next's prior art. The grant runs on every
db-provision call (idempotent), so re-running init --enable
db-provision on a stale install is a recovery path.

§2 rotate-db admin DSN. init persists the admin DSN to
bootroot/stepca/db_admin (separate KV path, not readable by
runtime/sidecar policies). rotate db reads from KV (or the explicit
--db-admin-dsn flag, which always wins) and does not fall back to
ca.json.db.dataSource — that field stores the runtime (stepca) DSN
and using it as an admin DSN is the original §2 self-ALTER bug.
When neither source is available the command fails fast with a
message naming both recovery paths. In the bundled same-role
topology, rotate db rewrites bootroot/stepca/db_admin with the
post-ALTER credential so a second consecutive rotate db succeeds.
init's post-bootstrap .env password rotation also rewrites the
persisted admin DSN and resolves the host-side Postgres port from
the compose dir's .env / process env (same precedence Docker
Compose uses for ${POSTGRES_HOST_PORT:-5433}). The auto-derived
admin DSN built when --db-admin-dsn is not supplied defaults its
host to 127.0.0.1 and resolves its port the same way, so init
--enable db-provision after infra install works on the new 5433
default and on a non-default --postgres-host-port without a manual
--db-admin-dsn.

§3 EAB prompt validation, --no-eab, rotate eab-clear.
  - 3a: kid trimmed-non-empty; hmac base64url-decodable to >=16
    bytes; re-prompt the missing field on blank input instead of
    silently coercing to "no EAB".
  - 3b: --no-eab on init skips the prompt and writes nothing.
  - 3c: rotate eab-clear writes empty {kid: "", hmac: ""}
    unconditionally to the agent path and every state-derived
    service path (KV v2 PUTs are idempotent, and a missing path
    must exist or consul-template treats it as a transient error
    and never renders the empty branch). Each affected sidecar is
    refreshed via §6's command. Per-service refresh branches on
    DeliveryMode: LocalFile docker-restarts the recorded container,
    RemoteBootstrap emits operator guidance.

§4 infra install port handling.
  - 4a: TCP-bind preflight on every host port the active compose
    stack publishes (postgres, openbao, step-ca, bootroot-http01)
    before docker compose up; abort with a best-effort lsof PID
    hint and per-service remediation rather than leaving the stack
    half-up. The preflight always checks the localhost ports
    because infra install only runs the base compose file.
  - 4b: --postgres-host-port <N> overrides .env and the docker
    compose subprocess environment so the flag wins over a
    shell-inherited POSTGRES_HOST_PORT (Docker Compose otherwise
    prefers shell env over .env, recreating the half-up collision
    the flag exists to avoid).
  - 4c: docker-compose.yml default published port moves from 5432
    to 5433. The internal postgres:5432 container address is
    unchanged, so step-ca's compose-internal DSN is unaffected.
    Operators who explicitly set POSTGRES_HOST_PORT=5432 are
    unaffected.

§5 partial-init detection. Before initialising OpenBao, query
sys/init; if initialised but no usable root token exists in
--root-token / OPENBAO_ROOT_TOKEN, emit a diagnostic naming the
three recovery paths (re-run with prior token, clean
--openbao-only, manual operator action) instead of bubbling up the
opaque 403. clean --openbao-only removes only the bootroot-openbao
container and its named volumes, discovering the compose project
from the container's com.docker.compose.project label and falling
back to the compose-dir basename.

§6 sidecar refresh. service openbao-sidecar refresh
--service-name <name> docker-restarts the recorded sidecar
container so consul-template re-reads its KV sources after
operator-side KV maintenance. RemoteBootstrap services emit
operator guidance only since bootroot has no signalling channel
to the remote host. §3c reuses this command per affected service.

Closes #588
```

sehkone:
Rebased onto `origin/main` at e0eacd9 ("Add non-interactive operation surface for CI rotations"). Resolved one conflict in `CHANGELOG.md` under `### Added` by keeping both the new #587 non-interactive-rotation entry from main and this branch's #588 entries.

PR #592, opened 2026-05-03
Wire local-file force-reissue --wait into CI matrix (#590)

### LINKED ISSUE #590: Wire `rotate force-reissue --yes --wait` into the daemon+local-file E2E matrix

sehkone:
## Summary

Wires the `--delivery-mode local-file` `--wait` code path that shipped in #589 into the CI E2E matrix by replacing the `rm -f`-based `force_reissue_for_service` helper in `scripts/impl/run-local-lifecycle.sh` and `scripts/impl/run-ca-key-rotation-recovery.sh` with a real `bootroot rotate force-reissue --yes --wait` invocation for daemon-deployed services (edge-proxy).

The companion `force_reissue_for_docker_service` helper keeps the original `rm -f` behaviour for `web-app`, which is registered with `--deploy-type docker --container-name web-app` for service-add coverage even though no actual `web-app` container runs in this scenario. `bootroot rotate force-reissue` would otherwise try to restart the missing container and fail. The host bootroot-agent owns both certs via the shared agent config, so deleting `web-app`'s files and letting the daemon's missing-cert check pick them up is the right reissue trigger here. The `--wait` code path is still exercised end-to-end through the daemon-deploy edge-proxy path.

`force_reissue_remote` and the three standalone `rm -f` lines at `run-ca-key-rotation-recovery.sh:691`, `:733`, `:762` are left untouched — those intentionally simulate partial-migration / finalize-blocked / trust-sync-conflict states that Phase 6 must still detect.

Spawns a long-running `bootroot-agent` daemon against `AGENT_CONFIG_PATH` after the OBA template render and stops it from `cleanup`. `bootroot rotate force-reissue` calls `pkill -HUP -f <agent_config_path>` to nudge the daemon; until this commit the lifecycle and recovery scripts only ever invoked `bootroot-agent --oneshot` (via `bootroot verify`), so `pkill` matched no process and the rotate failed before the `--wait` path could run. `RUST_LOG=info` is set when launching the daemon so the readiness probe (which greps for an info-level log line) can match the default tracing-subscriber output.

Closes #590

## Test plan

- [x] `scripts/preflight/ci/e2e-matrix.sh` green (covers `run-local-lifecycle.sh` via `local-no-hosts`, `local-no-hosts-host-daemon`, `rotation` arms)
- [x] `scripts/preflight/ci/e2e-extended.sh` green (covers `run-ca-key-rotation-recovery.sh` via the `ca-key-recovery` case)
- [x] CI `Docker E2E (local-no-hosts)` and `Docker E2E (rotation)` jobs pass
- [x] CI `run-extended` `ca-key-recovery` case passes
- [x] Confirm `force_reissue_remote` and the three scenario-setup `rm -f` lines in `run-ca-key-rotation-recovery.sh` are unchanged

### DIFF
--- a/scripts/impl/run-ca-key-rotation-recovery.sh
+++ b/scripts/impl/run-ca-key-rotation-recovery.sh
@@ -372,6 +372,8 @@ verify_service_with_retry() {
   done
 }
 
+# Daemon-deploy local-file path: drives `bootroot rotate force-reissue
+# --wait` end-to-end so the in-binary signal+wait code path runs in CI.
 force_reissue_for_service() {
   local service="$1"
   run_bootroot rotate \
@@ -387,13 +389,25 @@ force_reissue_for_service() {
     >>"$RUN_LOG" 2>&1
 }
 
+# Docker-deploy local-file path: web-app is registered with
+# --deploy-type docker --container-name web-app for service-add coverage,
+# but no real `web-app` container runs in this scenario.  The host
+# bootroot-agent owns the cert via the shared agent config, so deleting
+# the files and letting the daemon's missing-cert check pick them up is
+# the right reissue trigger here.  The `bootroot rotate force-reissue`
+# wait-path is exercised by the daemon-deploy edge-proxy call above.
+force_reissue_for_docker_service() {
+  local service="$1"
+  rm -f "$CERTS_DIR/${service}.crt" "$CERTS_DIR/${service}.key"
+}
+
 force_reissue_remote() {
   rm -f "$REMOTE_CERTS_DIR/${REMOTE_SERVICE}.crt" "$REMOTE_CERTS_DIR/${REMOTE_SERVICE}.key"
 }
 
 force_reissue_all_services() {
   force_reissue_for_service "$EDGE_SERVICE"
-  force_reissue_for_service "$WEB_SERVICE"
+  force_reissue_for_docker_service "$WEB_SERVICE"
   force_reissue_remote
 }
 
@@ -772,7 +786,7 @@ scenario_3_partial_reissuance() {
   [ -f "$CERTS_DIR/${WEB_SERVICE}.crt" ] || fail "S3: web-app old cert should still exist"
 
   # Reissue remaining services
-  force_reissue_for_service "$WEB_SERVICE"
+  force_reissue_for_docker_service "$WEB_SERVICE"
   verify_service_with_retry "$WEB_SERVICE"
   run_remote_bootstrap
   force_reissue_remote
--- a/scripts/impl/run-ca-key-rotation-recovery.sh
+++ b/scripts/impl/run-ca-key-rotation-recovery.sh
@@ -404,7 +404,11 @@ start_local_bootroot_agent_daemon() {
   [ -f "$AGENT_CONFIG_PATH" ] || fail "agent config missing at $AGENT_CONFIG_PATH"
   printf '[recovery] starting bootroot-agent daemon: --config %s\n' \
     "$AGENT_CONFIG_PATH" >>"$RUN_LOG"
-  "$BOOTROOT_AGENT_BIN" --config "$AGENT_CONFIG_PATH" \
+  # bootroot-agent uses tracing_subscriber::fmt::init(), whose default
+  # filter is ERROR.  The readiness probe below greps for an info-level
+  # message, so we have to opt into info output explicitly.
+  RUST_LOG="${RUST_LOG:-info}" \
+    "$BOOTROOT_AGENT_BIN" --config "$AGENT_CONFIG_PATH" \
     >>"$LOCAL_AGENT_DAEMON_LOG" 2>&1 &
   LOCAL_AGENT_DAEMON_PID=$!
   local attempt
--- a/scripts/impl/run-ca-key-rotation-recovery.sh
+++ b/scripts/impl/run-ca-key-rotation-recovery.sh
@@ -365,7 +365,18 @@ verify_service_with_retry() {
 }
 
 force_reissue_for_service() {
-  rm -f "$CERTS_DIR/${1}.crt" "$CERTS_DIR/${1}.key"
+  local service="$1"
+  run_bootroot rotate \
+    --compose-file "$COMPOSE_FILE" \
+    --openbao-url "http://${STEPCA_HOST_IP}:8200" \
+    --auth-mode approle \
+    --approle-role-id "$RUNTIME_ROTATE_ROLE_ID" \
+    --approle-secret-id "$RUNTIME_ROTATE_SECRET_ID" \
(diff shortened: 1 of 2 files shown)


sehkone:
**[Reviewer Round 1]**

Findings: none blocking.

The PR wires the intended daemon-deploy local-file path through real `bootroot rotate --yes force-reissue --wait` calls in both `scripts/impl/run-local-lifecycle.sh` and `scripts/impl/run-ca-key-rotation-recovery.sh`, and the added long-running `bootroot-agent` daemon explains why that path can now be signalled instead of failing at `pkill -HUP`. The remote-bootstrap helper remains unchanged, and the three recovery scenario setup deletes are still explicit `rm -f` calls, which preserves the Phase 6 / trust-sync scenarios the issue called out.

One non-blocking cleanup worth doing before merge: `scripts/impl/run-ca-key-rotation-recovery.sh:867` still says `--skip reissue` is used “because there is no running daemon for the Daemon-type service,” but this PR now starts `start_local_bootroot_agent_daemon` before the scenarios at line 912. The behavior may still be intentional for the trust-sync setup, but the reason in the comment is now stale and will mislead the next person reading this flow.

sehkone:

sehkone:
**[Reviewer Unresolved Round 1]**

- `scripts/impl/run-ca-key-rotation-recovery.sh:867` still says `--skip reissue` is used because there is no running daemon for the daemon-type service, but this PR starts `start_local_bootroot_agent_daemon` before the scenarios. The behavior may be intentional, but the comment is now stale and should be updated or removed.

sehkone:
## Suggested squash commit

**Title**

```text
Wire local-file force-reissue --wait into CI matrix
```

**Body**

```text
Replaces the `rm -f`-based `force_reissue_for_service` helper in
`scripts/impl/run-local-lifecycle.sh` and
`scripts/impl/run-ca-key-rotation-recovery.sh` with a real
`bootroot rotate force-reissue --yes --wait` invocation for the
daemon-deployed edge-proxy service, so the `--delivery-mode local-file`
`--wait` code path runs end-to-end in the existing CI E2E matrix.

A companion `force_reissue_for_docker_service` helper preserves the
`rm -f` behaviour for `web-app`, which is registered as
`--deploy-type docker --container-name web-app` for service-add
coverage even though no actual `web-app` container runs in this
scenario; invoking `bootroot rotate force-reissue` against it would
try to restart the missing container and fail. The host
bootroot-agent owns the cert via the shared agent config, so deleting
the files and letting the daemon's missing-cert check pick them up is
the right reissue trigger. The `--wait` code path is still exercised
through the daemon-deploy edge-proxy call.

Spawns a long-running `bootroot-agent` daemon against
`AGENT_CONFIG_PATH` after the OBA template render and stops it from
`cleanup`. `bootroot rotate force-reissue` calls
`pkill -HUP -f <agent_config_path>` to nudge the daemon; without a
real long-running process, `pkill` would match nothing and rotate
would fail before the wait path could run. `RUST_LOG=info` is set so
the readiness probe (which greps for an info-level log line) matches
the default tracing-subscriber output.

`force_reissue_remote` and the three standalone `rm -f` lines that
intentionally simulate partial-migration / finalize-blocked /
trust-sync-conflict states in `run-ca-key-rotation-recovery.sh` are
left untouched.

Closes #590
```

PR #595, opened 2026-05-05
Allow cert/key delivery to non-root container clients (#593)

sehkone:
## Summary

Adds `--cert-group <gid-or-name>` to `bootroot service add` and `bootroot service update` so issued service certs/keys can be delivered to non-root containerized clients without operator-side `chmod` workarounds that silently regress on every rotation.

When set, every issuance and rotation applies a group-readable policy:

- key parent dir: `0750`
- cert parent dir: `0755` (or `0750` when it shares the key parent — stricter wins)
- `<svc>-key.pem`: `0640`
- `<svc>-cert.pem`: `0644` (unchanged)
- group ownership of both parent dirs and both files: the configured gid

When unset, the historical operator-only default (`0700`/`0600`/`0644`) is preserved unchanged for existing deployments.

The policy is plumbed through every layer renewal touches so it survives rotation:

- `ServiceEntry` in `state.json` (new `cert_group_gid` field)
- the managed `agent.toml` profile block
- the remote-bootstrap artifact (additive optional field, no `schema_version` bump)
- `DaemonProfileSettings`, so the issuance path picks it up at every renew

Per-mode acceptance follows the cross-host NSS rationale in the issue:

- `local-file`: numeric gid or group name (resolved on the control host, which is also the cert-writing host). `chown`-permission to the target gid is validated at `service add` / `service update` time so failures surface at configure time, not at the next rotation.
- `remote-bootstrap`: numeric gid only — name form is rejected at parse time because the control host's NSS may diverge from the remote agent host's NSS. Substantive checks run on the remote host.
- `--cert-group 0` (root) is rejected at parse time and during config validation.

The numeric gid must also resolve in the **cert-writing host's** group database (`getgrgid_r`). An orphan numeric gid — one that exists on a different host (e.g. the container image's runtime user) but not on the host that will actually `chown` the cert/key files — is rejected loudly. For `local-file` this check runs at `service add` / `service update` time on the control host (which is also the cert-writing host); for `remote-bootstrap` the same check runs on the remote agent host at `bootroot-remote bootstrap`, and `bootroot-agent` runs the check again at config validation. Without it the kernel would silently accept `chown(-1, gid)` and the consumer would still hit `EACCES` because the gid resolves to no real group.

`service update --cert-group ...` re-renders the local managed `agent.toml` block immediately for `local-file` services. For `remote-bootstrap` services it emits a one-line follow-up instructing the operator to re-emit the bootstrap artifact and re-run `bootroot-remote bootstrap` on the remote host, so the remote `agent.toml` actually picks up the new policy.

The local-file re-render runs **before** `state.json` is persisted, and re-runs of the same `--cert-group ...` value re-trigger the re-render even when the gid already matches what is in state. So a previously-failed re-render (e.g., agent.toml temporarily unwritable) can be repaired by simply re-running the command, with no risk of `state.json` drifting ahead of the on-disk managed profile and thereby short-circuiting the next attempt at the no-change early return.

Atomic key write: the key file is written via stage-then-rename. A sibling temporary file is created with `O_CREAT|O_EXCL` and `mode=0600`, the staged file is `chown`d (when policy is active) and promoted to `0640`, and only then renamed over the destination. The destination path is therefore never observable at a mode wider than the final policy — no umask-derived `0644` window before the clamp, and no group-readable window under the operator's primary gid before the chown lands.

Shared cert/key parent detection uses kernel `(dev, ino)` identity instead of textual path equality. So spellings like `certs` vs `certs/.` (or symlinked variants) cannot trick `ensure_cert_parent_dir` into widening a directory that `ensure_key_parent_dir` had already locked down to `0750` up to `0755`.

CI fixture: the `test-core` job in `.github/workflows/ci.yml` provisions a dedicated `bootroot-e2e-certgrp` group, adds the runner user as a supplementary member, exports `BOOTROOT_E2E_CERT_GROUP_GID` and `BOOTROOT_E2E_REQUIRE_CERT_GROUP=1`, and runs `cargo test` via `sudo setpriv --reuid=$(id -u) --regid=$(id -g) --init-groups -- "$(command -v cargo)" test`. `setpriv --init-groups` re-reads the user's supplementary groups from `/etc/group` while keeping the original primary gid intact, so the fixture gid appears as a non-primary supplementary group — which is what the test assertions require. (`sg <group>` would make the fixture gid the spawned shell's primary gid, defeating the regression.) Cargo is resolved to its absolute path before `sudo` because `sudo` resets `PATH` via `secure_path`. `tests/e2e_cert_group_chown.rs` honors the require-flag: in CI mode the chown round-trip fails (rather than silently skipping) when the env-provided gid is missing, zero, equal to the runner's primary gid, not in the host group DB, or not a supplementary group of the test process. So a CI environment that ever loses the fixture surfaces as a red build, not a green skip.

Closes #593

## Test plan

- [x] `cargo build --tests` succeeds
- [x] `cargo clippy --all-targets --all-features -- -D warnings` is clean
- [x] `cargo test --lib --tests` passes (covers default-policy preservation, group-policy mode set, rotation re-asserting policy after a hostile `chmod`/`chgrp`, parse helpers for `local-file` vs `remote-bootstrap`, `--cert-group 0` rejection in parse and config validation, gid-existence rejection of an orphan numeric gid in config validation and in `bootroot-remote bootstrap` validation, managed profile block rendering with and without `cert_group_gid`, remote-bootstrap artifact round-trip including `cert_group_gid`, atomic stage-then-rename of the key file on first issuance and on rotation, shared cert/key parent detection via inode identity for `certs` vs `certs/.`)
- [x] `tests/e2e_cert_group_chown.rs` verifies the full `write_cert_and_key` round-trip against a non-primary supplementary gid: file mode, dir mode, and the resulting file `gid` all match the policy, and rotation restores both mode and gid after a hostile operator-side `chgrp`/`chmod`. In CI-fixture mode the test refuses to skip — missing/unusable `BOOTROOT_E2E_CERT_GROUP_GID` is a hard failure.
- [x] Default behavior (`0700`/`0600`/`0644`, operator's primary gid) is unchanged when `--cert-group` is unset
- [x] `service update --cert-group ...` on a `local-file` service re-renders the managed `agent.toml` profile block in place (before state.json is saved), with a `cert_group_gid = N` line
- [x] `service update --cert-group ...` re-runs with an unchanged gid still re-render the managed profile, so a previously-failed re-render is repaired without state/agent.toml split-brain
- [x] `service update --cert-group ...` on a `remote-bootstrap` service prints the warning instructing the operator to re-emit the bootstrap artifact and re-run `bootroot-remote bootstrap`
- [x] `--cert-group aice-clients` on a `remote-bootstrap` service is rejected at parse time with a numeric-required error
- [x] `--cert-group <gid>` on a `local-file` service where the caller is not a member of that group fails at `service add` time with a "not a member" error rather than at the next rotation
- [x] `--cert-group <gid>` referring to a gid not in the host group DB is rejected loudly at `service add` (local-file) and at `bootroot-remote bootstrap` (remote-bootstrap) — orphan numeric gids no longer persist silently into rotation
- [x] `--cert-group clear` on `service update` removes an existing policy and reverts the next rotation to operator-only mode/owner
- [x] `bootroot service add` against a `remote-bootstrap` deployment includes `cert_group_gid` in the bootstrap artifact JSON when set, and omits the field entirely when unset (no `schema_version` bump required)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,8 +137,35 @@ jobs:
       - name: Pre-pull openbao image for real-daemon TLS E2E
         run: docker pull openbao/openbao:latest
 
+      # Issue #593: provision a dedicated supplementary group for the
+      # runner so `tests/e2e_cert_group_chown.rs` actually exercises a
+      # non-primary gid through the kernel chown permission check. The
+      # test refuses to silently skip when
+      # BOOTROOT_E2E_REQUIRE_CERT_GROUP=1 is exported, so a runner that
+      # ever loses this fixture surfaces as a CI failure rather than
+      # green.
+      - name: Provision cert-group fixture for e2e_cert_group_chown
+        id: cert_group_fixture
+        run: |
+          set -euo pipefail
+          GROUP_NAME="bootroot-e2e-certgrp"
+          if ! getent group "$GROUP_NAME" >/dev/null; then
+            sudo groupadd "$GROUP_NAME"
+          fi
+          GID="$(getent group "$GROUP_NAME" | cut -d: -f3)"
+          sudo usermod -aG "$GROUP_NAME" "$USER"
+          echo "BOOTROOT_E2E_CERT_GROUP_GID=$GID" >> "$GITHUB_ENV"
+          echo "BOOTROOT_E2E_REQUIRE_CERT_GROUP=1" >> "$GITHUB_ENV"
+          echo "BOOTROOT_E2E_CERT_GROUP_NAME=$GROUP_NAME" >> "$GITHUB_ENV"
+          echo "Provisioned $GROUP_NAME (gid=$GID) and added $USER as supplementary member"
+
       - name: Run Unit Tests
-        run: cargo test
+        # `usermod -aG` updates /etc/group but the current shell process
+        # doesn't pick up the new supplementary membership. `sg` spawns
+        # a new shell whose group list reflects the updated /etc/group,
+        # so cargo test sees the gid as a real supplementary group and
+        # the kernel `chown(-1, gid)` check actually runs.
+        run: sg "$BOOTROOT_E2E_CERT_GROUP_NAME" -c "cargo test"
 
       # --- E2E Integration Testing (Docker) ---
       - name: Set secrets directory
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,10 +170,16 @@ jobs:
         # re-reads the user's supplementary groups from /etc/group
         # while keeping the original primary gid intact, which is the
         # exact post-login state the test assertions assume.
+        #
+        # `sudo` resets PATH via sudoers `secure_path`, so resolve the
+        # cargo binary before sudo and pass the absolute path through
+        # setpriv. CARGO_HOME / RUSTUP_HOME / HOME are also preserved
+        # so cargo can find the toolchain installed by rustup.
         run: |
-          sudo --preserve-env \
+          CARGO_BIN="$(command -v cargo)"
+          sudo --preserve-env=HOME,PATH,CARGO_HOME,RUSTUP_HOME \
             setpriv --reuid="$(id -u)" --regid="$(id -g)" --init-groups \
-            -- cargo test
+            -- "$CARGO_BIN" test
 
       # --- E2E Integration Testing (Docker) ---
       - name: Set secrets directory
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,9 +175,18 @@ jobs:
         # cargo binary before sudo and pass the absolute path through
         # setpriv. CARGO_HOME / RUSTUP_HOME / HOME are also preserved
         # so cargo can find the toolchain installed by rustup.
+        #
+        # `sudo` also strips the BOOTROOT_E2E_* fixture variables by
+        # default (env_reset). Without explicitly preserving them,
+        # `tests/e2e_cert_group_chown.rs` would see
+        # BOOTROOT_E2E_REQUIRE_CERT_GROUP unset and fall back to the
+        # dev-mode skip path, defeating the fixture and letting a broken
+        # CI environment produce a green build. Pass them through
+        # --preserve-env so the require-flag actually reaches the test
+        # process.
         run: |
           CARGO_BIN="$(command -v cargo)"
-          sudo --preserve-env=HOME,PATH,CARGO_HOME,RUSTUP_HOME \
+          sudo --preserve-env=HOME,PATH,CARGO_HOME,RUSTUP_HOME,BOOTROOT_E2E_CERT_GROUP_GID,BOOTROOT_E2E_REQUIRE_CERT_GROUP,BOOTROOT_E2E_CERT_GROUP_NAME \
             setpriv --reuid="$(id -u)" --regid="$(id -g)" --init-groups \
             -- "$CARGO_BIN" test
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,11 +161,19 @@ jobs:
 
       - name: Run Unit Tests
         # `usermod -aG` updates /etc/group but the current shell process
-        # doesn't pick up the new supplementary membership. `sg` spawns
-        # a new shell whose group list reflects the updated /etc/group,
-        # so cargo test sees the gid as a real supplementary group and
-        # the kernel `chown(-1, gid)` check actually runs.
-        run: sg "$BOOTROOT_E2E_CERT_GROUP_NAME" -c "cargo test"
+        # inherits the old supplementary group list. `sg <group>` would
+        # work but it makes the named group the *primary* gid of the
+        # spawned shell, which defeats the regression: the test wants
+        # the fixture gid to appear as a non-primary supplementary
+        # group so the kernel `chown(-1, gid)` permission check is
+        # actually exercised. Use `setpriv --init-groups` instead: it
+        # re-reads the user's supplementary groups from /etc/group
+        # while keeping the original primary gid intact, which is the
+        # exact post-login state the test assertions assume.
+        run: |
+          sudo --preserve-env \
+            setpriv --reuid="$(id -u)" --regid="$(id -g)" --init-groups \
+            -- cargo test
 
       # --- E2E Integration Testing (Docker) ---
       - name: Set secrets directory
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -272,7 +272,17 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   by simply re-running the command, without `state.json` ever
   drifting ahead of the on-disk managed profile.
   `--cert-group 0` (root) is rejected at parse time and during
-  config validation. The key file is written via stage-then-rename
+  config validation. `cert_group_gid` is also rejected when it does
+  not resolve in the cert-writing host's group database
+  (`getgrgid_r`): this orphan-gid case (a numeric gid that exists on
+  a different host — e.g. the container's runtime user — but not on
+  the host that will actually `chown` the cert/key files) is checked
+  at `service add` / `service update` time on the control host for
+  `local-file`, at `bootroot-remote bootstrap` time on the remote
+  agent host for `remote-bootstrap`, and again at `bootroot-agent`
+  config validation, so it surfaces as a loud failure instead of
+  passing the kernel `chown` and reappearing as EACCES inside the
+  consumer. The key file is written via stage-then-rename
   (sibling temp file created with `O_CREAT|O_EXCL` and `mode=0600`,
   `chown`d, promoted to `0640`, then renamed over the destination)
   so the destination path is never observable at a mode wider than
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -266,8 +266,23 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   `service update --cert-group ...` re-renders the local managed
   profile block immediately for `local-file` services, and warns the
   operator to re-emit the bootstrap artifact for `remote-bootstrap`
-  services. `--cert-group 0` (root) is rejected at parse time and
-  during config validation. (Closes #593)
+  services. The local-file re-render runs before `state.json` is
+  persisted, and re-runs of the same `--cert-group` value re-trigger
+  the re-render — so a previously-failed re-render can be repaired
+  by simply re-running the command, without `state.json` ever
+  drifting ahead of the on-disk managed profile.
+  `--cert-group 0` (root) is rejected at parse time and during
+  config validation. The key file is written via stage-then-rename
+  (sibling temp file created with `O_CREAT|O_EXCL` and `mode=0600`,
+  `chown`d, promoted to `0640`, then renamed over the destination)
+  so the destination path is never observable at a mode wider than
+  the final policy — no umask-derived `0644` window before the
+  clamp, and no group-readable window under the operator's primary
+  gid before the chown lands. The shared cert/key parent detection
+  uses kernel `(dev, ino)` identity rather than textual path
+  equality, so spellings like `certs` vs `certs/.` cannot trick
+  `ensure_cert_parent_dir` into widening a shared key parent from
+  `0750` to `0755`. (Closes #593)
 - Added a non-interactive operation surface for CI/scripted rotations
   (Closes #587):
   - `bootroot rotate --yes`/`-y` is now a global flag accepted at any
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -240,6 +240,34 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Added `--cert-group <gid-or-name>` to `bootroot service add` and
+  `bootroot service update` so issued service certificates can be
+  delivered to non-root containerized clients without operator-side
+  `chmod` workarounds. When set, the agent applies a group-readable
+  policy on every issuance and rotation: parent directories become
+  `0750` (or `0755` for a distinct cert parent), `<svc>-key.pem`
+  becomes `0640`, `<svc>-cert.pem` stays at `0644`, and group
+  ownership of all four is set to the configured gid. When unset
+  (the default), the historical operator-only modes
+  (`0700`/`0600`/`0644`) are preserved so existing deployments are
+  unchanged. `cert_group_gid` is persisted on `ServiceEntry`,
+  rendered into the managed `agent.toml` profile block, threaded
+  through the remote-bootstrap artifact as a new optional field
+  (no `schema_version` bump — additive change with
+  `skip_serializing_if`), and surfaced on `DaemonProfileSettings`,
+  so rotation re-asserts the policy instead of silently reverting
+  to operator-only.
+  `local-file` deployments accept either a numeric gid or a group
+  name resolved on the control host; `remote-bootstrap` deployments
+  accept numeric form only because the control host's NSS may differ
+  from the remote agent host's. `service add` validates that the
+  caller can `chown` to the target gid for `local-file` mode so the
+  failure surfaces at add-time rather than at the next rotation.
+  `service update --cert-group ...` re-renders the local managed
+  profile block immediately for `local-file` services, and warns the
+  operator to re-emit the bootstrap artifact for `remote-bootstrap`
+  services. `--cert-group 0` (root) is rejected at parse time and
+  during config validation. (Closes #593)
 - Added a non-interactive operation surface for CI/scripted rotations
   (Closes #587):
   - `bootroot rotate --yes`/`-y` is now a global flag accepted at any
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ reqwest = { version = "0.13", default-features = false, features = [
     "json",
     "rustls",
 ] }
+libc = "0.2"
 ring = "0.17"
 rustls = { version = "0.23", default-features = false, features = ["ring", "std"] }
 rustls-pemfile = "2"
@@ -32,7 +33,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 x509-parser = { version = "0.18", features = ["verify"] }
 
 [dev-dependencies]
-libc = "0.2"
 tempfile = "3"
 wiremock = "0.6"
 
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -689,6 +689,47 @@ These values are persisted in `state.json` and applied on
 the same field: `--no-wrap` sets the stored wrap TTL to `0`, disabling
 wrapping entirely.
 
+Cert/key group-access policy:
+
+- `--cert-group <gid-or-name>`: numeric gid (or, for `local-file` only,
+  group name) that should own the issued cert/key files and their
+  parent directories.
+  - Unset (default): the agent preserves the host-local default —
+    `0700` parent directories, `0600` `<svc>-key.pem`, `0644`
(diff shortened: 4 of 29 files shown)


sehkone:
**[Reviewer Round 1]**

I found a few issues that should be addressed before this lands:

1. `src/cert_group.rs:281`: the key is still created before it is restricted to `0600`. `tokio::fs::write` creates a missing file with the process umask-derived mode, typically `0644`, and only then do lines 285-290 clamp it to `0600`, `chown`, and promote to `0640`. That violates the issue's atomic write-order requirement: on first issuance the private key can briefly be group/world-readable under the operator's original gid before the intended ownership is applied. The fix should create the key with `0600` from the start, for example via `OpenOptionsExt::mode(0o600)` / a temp file opened at `0600`, then `chown`, then final `chmod 0640` before rename. The current tests only assert final mode, so they would not catch this transient exposure.

2. `src/commands/service.rs:643` and `src/commands/service.rs:673`: `service update --cert-group` mutates `entry.cert_group_gid` and saves `state.json` before `rerender_local_managed_profile()` runs. If that re-render fails because the managed `agent.toml` is missing or unwritable, the command returns an error but state now says the new policy is applied while the agent config remains stale. A follow-up run with the same `--cert-group` then hits `changes.is_empty()` and cannot repair the stale config, so the next restart/rotation still misses the policy. This recreates the “state changed, rotation did not” class of bug the issue is trying to eliminate. The update path needs to make the state/config update atomic enough to avoid this split-brain, or at least retry the re-render whenever `--cert-group` is supplied and the managed profile is out of sync.

3. `src/cert_group.rs:369`: the “same cert/key parent means `0750` wins” check is a raw `Path` equality check. If the same directory is spelled differently, e.g. key parent `certs` and cert parent `certs/.`, `ensure_key_parent_dir()` first applies `0750` and then `ensure_cert_parent_dir()` treats it as distinct and widens the actual key directory to `0755`. Symlinked or relative path variants have the same problem. Since this is the security boundary that decides whether the key parent gets broad traversal, compare the actual directories after creation, for example via canonical paths or metadata identity, and add a regression test for an equivalent-but-not-textually-equal parent.

sehkone:

sehkone:
**[Author Round 1]**

All three round-1 review items applied as concrete code changes (commit `4b3dc46`); none pushed back, none partially addressed.

1. **Fixed — atomic key write at `src/cert_group.rs:281`.** The previous order created the destination via `tokio::fs::write` (umask-derived, typically `0644`), clamped to `0600`, then `chown`d, then promoted to `0640`. On first issuance that left a transient `0644` window at the destination before the clamp landed. Replaced with stage-then-rename: a sibling temporary file is created with `O_CREAT|O_EXCL` and `mode=0600` via `OpenOptionsExt::mode`, the staged file is `chown`d and promoted to `0640` while still at the temp path, and only then `rename`d over the destination. The destination path is therefore never observable at a mode wider than the final policy — no umask `0644` window before the clamp, no group-readable window under the operator's primary gid before the chown lands. New regression tests `write_key_file_no_policy_first_issuance_is_never_0644` and `write_key_file_rotation_replaces_atomically_with_policy` cover both first issuance and rotation. The reviewer was right that the existing tests only asserted final mode and would not have caught the transient exposure.

2. **Fixed — `service update --cert-group` split-brain at `src/commands/service.rs:643/673`.** Two independent changes that together close both halves of the failure mode:
   - Moved `rerender_local_managed_profile` to run **before** `state.save`. If the re-render fails, `state.json` is unchanged, so a re-run still sees the old value as the source of truth and re-triggers the re-render rather than short-circuiting on `changes.is_empty()`.
   - Whenever `--cert-group` is supplied (tracked via `cert_group_supplied`), the local-file re-render is triggered even when the requested gid matches what is already in state. So a re-run of `service update --cert-group <same-gid>` repairs an out-of-sync managed profile from any source of drift, not just from a previous failed re-render.
   The remote-bootstrap warning path is unchanged in behavior (still loud, still emits the same one-line follow-up).

3. **Fixed — shared cert/key parent detection at `src/cert_group.rs:369`.** `path == key_dir` was a textual comparison that treated `certs` vs `certs/.` (and symlinked variants) as distinct, which would let `ensure_cert_parent_dir` widen a directory `ensure_key_parent_dir` had already locked to `0750` up to `0755`. Replaced with kernel `(dev, ino)` identity via `MetadataExt::dev`/`ino` on the materialised directories. New regression test `ensure_cert_parent_dir_with_policy_shared_via_dot_segment_stays_0750` constructs the exact `certs` vs `certs/.` shape the reviewer called out and asserts the shared parent stays at `0750`. When either path does not yet exist (some call sites and tests only materialise one of the two), the comparison falls through to "distinct" rather than surfacing a stat error — they cannot be the same directory by any meaningful definition.

Docs (`docs/en/cli.md`, `docs/ko/cli.md`) and `CHANGELOG.md` updated to describe the stage-then-rename order, the render-first / repair-on-rerun behavior, and the inode-based shared-parent detection. PR description updated to match. `cargo build --tests`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --lib --tests` are all clean (231 lib tests + 11 stepca integration tests pass; 23 cert_group tests including the 3 new regressions). The Docker `test-core` preflight script was not run because my changes are scoped to file-write atomicity, the `service update` re-render order, and a path-equality predicate — none of which touch the Docker lifecycle, the rotation E2E flow, or any code path exercised by `e2e-matrix` / `e2e-extended`.

sehkone:
**[Reviewer Round 2]**

I verified the three Round 1 items in the updated diff and consider them resolved: the key write is now stage-then-rename at `0600`, `service update --cert-group` re-renders the local managed profile before saving state and reruns when the flag is supplied, and shared cert/key parent detection now uses directory identity instead of textual path equality.

Remaining findings:

1. `--cert-group` still does not validate that the configured gid exists in the cert-writing host's group database. For `local-file`, numeric input returns immediately once it is non-zero (`src/cert_group.rs:126-130`), and the later validation only checks `caller_can_chown_to()` membership/root status (`src/cert_group.rs:174-181`). For `remote-bootstrap`, the artifact value is copied into `cert_group_gid` (`src/bin/bootroot-remote/main.rs:486`), and `validate_bootstrap_args()` never checks it (`src/bin/bootroot-remote/bootstrap.rs:391-446`). That misses an explicit issue requirement: local-file should verify gid existence on the control/cert-writing host, and remote-bootstrap should run the substantive gid-existence check on the remote cert-writing host at bootstrap apply or first issuance. As written, a root-run agent, or a process launched with a numeric supplementary group that is not present in `/etc/group`/NSS, can persist and apply an orphan numeric gid with no loud “group does not exist” failure. Please add a `getgrgid_r`/equivalent existence check at the cert-writing host boundary: local `service add/update` for local-file, and `bootroot-remote bootstrap` and/or agent config validation for remote profiles.

2. The new e2e coverage can silently pass without exercising the required non-primary supplementary-gid path. `tests/e2e_cert_group_chown.rs` returns success when no suitable gid exists (`tests/e2e_cert_group_chown.rs:49-51`) and falls back to ambient supplementary groups (`tests/e2e_cert_group_chown.rs:32-44`), but this PR does not update `.github/workflows/e2e-extended.yml` or the preflight scripts to create a dedicated group, add the runner to it, and set `BOOTROOT_E2E_CERT_GROUP_GID`. That means the “e2e-extended verifies chown to a different gid created in the CI fixture” requirement is not actually locked in by the PR; on an environment without a usable supplementary gid, the test suite reports green while skipping the regression. Please wire the fixture into the extended/preflight path, and make the CI-mode test fail rather than skip when the expected env-provided gid is missing, equal to the primary gid, or unusable.

sehkone:

sehkone:
**[Author Round 2]**

Both round-2 review items applied as concrete code changes (commit `bb0fa6e`); none pushed back, none partially addressed.

1. **Fixed — gid existence check on the cert-writing host.** You were right: `validate_local_gid_membership` (`src/cert_group.rs:174-181`) only checked the caller's group membership and `validate_bootstrap_args` (`src/bin/bootroot-remote/bootstrap.rs:391-446`) didn't look at `cert_group_gid` at all, so an orphan numeric gid — one that exists on a different host (e.g. the container image's runtime user) but not on the host that will actually `chown` the cert/key files — could persist silently and only surface as EACCES inside the consumer. The kernel's `chown(-1, gid)` does not require the gid to resolve in NSS, which is the part that bit us in spirit if not yet in the field.

   - Added `gid_exists_on_host(gid)` (`getgrgid_r`-backed) and a new `validate_cert_writing_host_gid` that combines the gid==0 guard with the existence check.
   - `validate_local_gid_membership` now delegates to `validate_cert_writing_host_gid` first, so the existing `service add` / `service update` validation on the control host (which is also the cert-writing host for `local-file`) catches orphan gids at configure time. The numeric input path that previously short-circuited at `--cert-group 5001` (`src/cert_group.rs:126-130`) is still parser-side only; existence is enforced one layer up by `validate_local_gid_membership`, which is what every `local-file` `service add` / `service update` already calls.
   - `validate_bootstrap_args` now calls `validate_artifact_cert_group_gid`, which runs the same `validate_cert_writing_host_gid` against `args.cert_group_gid` on the remote agent host. The localized error message includes the offending gid plus a `groupadd -g <gid>` hint, so the operator can pre-provision the gid and re-run.
   - `validate_profile` in `src/config/validation.rs` was widened to call `validate_cert_writing_host_gid` for any `cert_group_gid` it sees. This catches a hand-edited `agent.toml` even when the gid never came in via `bootroot-remote bootstrap` — defense in depth at agent startup.
   - Regression coverage: `validate_cert_writing_host_gid_rejects_unknown_gid`, `validate_cert_writing_host_gid_accepts_caller_egid`, `validate_rejects_cert_group_gid_unknown_on_host` (parses agent.toml with a 4_000_000_000 gid and asserts the validator rejects it), and `validate_bootstrap_args_rejects_unknown_cert_group_gid` (asserts the bootstrap validator rejects the same on the remote host).

2. **Fixed — e2e fixture wiring + fail-not-skip in CI.** You were right that `tests/e2e_cert_group_chown.rs:32-51` would silently green an environment without a usable supplementary gid, and that nothing in the PR locked the dedicated-group fixture into CI.

   - `.github/workflows/ci.yml` `test-core` now provisions a dedicated `bootroot-e2e-certgrp` group, adds the runner as a supplementary member, exports `BOOTROOT_E2E_CERT_GROUP_GID` plus `BOOTROOT_E2E_REQUIRE_CERT_GROUP=1`, and runs `cargo test` under `sg <group> -c "..."` so the spawned shell actually sees the new supplementary membership (`usermod -aG` updates `/etc/group` but the existing shell process keeps its old supplementary list, so without `sg` the test process would still see the gid as unreachable).
   - `resolve_test_gid()` in the test now hard-fails (panics with a description of how to fix the fixture) when `BOOTROOT_E2E_REQUIRE_CERT_GROUP=1` and the env-provided gid is missing, non-numeric, zero, equal to the runner's primary gid, not in the host group DB, or not a supplementary group of the test process. The dev-mode fallback is preserved so `cargo test` on a developer machine without group setup still works — the failure mode is gated on the require-flag that only the CI fixture exports.
   - The cert_group_chown test file is in `tests/` (a regular cargo integration test), so it runs in `test-core`, not in `e2e-extended.yml`'s docker harness. Wiring the fixture into `test-core` is the placement that actually exercises this regression on every PR; the e2e-extended workflow was left alone because the docker harness scenarios don't drive `write_cert_and_key` directly.

Docs (`docs/en/cli.md`, `docs/ko/cli.md`) and `CHANGELOG.md` updated to describe the cert-writing-host gid existence requirement on both modes plus the agent-side validation. PR description updated to match. `cargo build --tests`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --lib --tests` are all clean (235 lib tests + 11 stepca integration tests pass; 4 new tests added). Docker `test-core` preflight was not run because my changes are scoped to a parser-side existence check, a CI workflow step, and a test gate — none of which touch the Docker lifecycle, the rotation E2E flow, or any code path exercised by `e2e-matrix` / `e2e-extended`.

sehkone:
**[Reviewer Round 3]**

I verified the Round 2 gid-existence fix and consider that item resolved: local-file validation now reaches `validate_cert_writing_host_gid`, remote bootstrap validates `cert_group_gid` on the remote host, and agent config validation rejects hand-edited orphan gids.

One Round 2 item remains open:

1. `.github/workflows/ci.yml:157-182`: the cert-group fixture still is not actually forced into the `cargo test` process. The fixture writes `BOOTROOT_E2E_CERT_GROUP_GID` and `BOOTROOT_E2E_REQUIRE_CERT_GROUP=1` into `$GITHUB_ENV`, but the test command runs through `sudo --preserve-env=HOME,PATH,CARGO_HOME,RUSTUP_HOME ...` and does not preserve either `BOOTROOT_E2E_*` variable. With sudo's normal `env_reset` behavior, `tests/e2e_cert_group_chown.rs` will see `require_ci_fixture() == false` (`tests/e2e_cert_group_chown.rs:40-44`) and can fall back to dev-mode behavior or skip (`tests/e2e_cert_group_chown.rs:91-109`). That means a broken/missing fixture can still produce a green CI run, which is the exact gap the Round 2 fix was meant to close. Please pass those variables through the sudo boundary, for example by adding `BOOTROOT_E2E_CERT_GROUP_GID` and `BOOTROOT_E2E_REQUIRE_CERT_GROUP` to `--preserve-env` (and optionally the group name), or by injecting them into the environment after sudo before invoking cargo.

sehkone:

sehkone:
**[Author Round 3]**

Round-3 review item applied as a concrete code change (commit `7e908fb`); not pushed back, not partially addressed.

1. **Fixed — fixture env vars now reach the test process across the sudo boundary.** You were right. `.github/workflows/ci.yml:180` invoked `sudo --preserve-env=HOME,PATH,CARGO_HOME,RUSTUP_HOME ...` and `sudo`'s `env_reset` default stripped both `BOOTROOT_E2E_CERT_GROUP_GID` and `BOOTROOT_E2E_REQUIRE_CERT_GROUP` even though the prior `Provision cert-group fixture` step had written them to `\$GITHUB_ENV`. With those gone from the cargo-test environment, `require_ci_fixture()` (`tests/e2e_cert_group_chown.rs:40-44`) read false and the test fell through to the dev-mode `resolve_test_gid()` skip path (`tests/e2e_cert_group_chown.rs:91-109`), which is the exact silent-skip-on-broken-fixture gap the Round 2 work was meant to close.

   Added `BOOTROOT_E2E_CERT_GROUP_GID`, `BOOTROOT_E2E_REQUIRE_CERT_GROUP`, and `BOOTROOT_E2E_CERT_GROUP_NAME` to the `--preserve-env` list. The require-flag now reaches the cargo-test process, so `require_ci_fixture()` returns true, and any deviation (missing/zero/primary-equal/unknown/non-supplementary gid) panics in `resolve_test_gid` instead of skipping. A new comment block above the `sudo` line records why these specific variables have to cross the sudo boundary, so this isn't re-stripped by a future cleanup that thinks the list is over-specified.

   No code change was needed inside the test itself — the gating already supported this once the environment is delivered correctly. No doc/CHANGELOG update either, since the operator-visible behavior of `--cert-group` is unchanged; this fix is purely about whether CI surfaces a broken fixture as red instead of green.

`cargo build --tests`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --lib --tests` are all clean from the prior round and unchanged here. Docker `test-core` preflight was not run because this change is a single CI workflow tweak that affects neither the Docker lifecycle nor any code path exercised by `e2e-matrix` / `e2e-extended`.

sehkone:
**[Reviewer Round 4]**

I verified the Round 3 fix in `.github/workflows/ci.yml`: the cert-group fixture variables (`BOOTROOT_E2E_CERT_GROUP_GID`, `BOOTROOT_E2E_REQUIRE_CERT_GROUP`, and `BOOTROOT_E2E_CERT_GROUP_NAME`) are now preserved across the `sudo setpriv` boundary, so the CI fixture should reach `tests/e2e_cert_group_chown.rs` and fail loudly instead of falling back to dev-mode skip behavior.

I also rechecked the prior Round 1/Round 2 items in the current diff: key writes are staged at `0600` before `chown`/`chmod` and rename, local `service update --cert-group` re-renders the managed profile before saving state and re-renders again when the flag is supplied with the same value, shared cert/key parent detection uses directory identity, gid existence is validated on the cert-writing host boundary, and the policy is persisted through `ServiceEntry`, rendered `agent.toml`, remote-bootstrap artifacts, and `DaemonProfileSettings`.

No remaining findings from me in this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add --cert-group policy for non-root container clients
```

**Body**

```text
bootroot-agent issued certs/keys with `0700` on `certs/`, `0600` on
the key, and operator-only ownership. That is correct for host-local
use but breaks the common pattern where the cert/key is bind-mounted
into a container running as a non-root uid: traversal and key reads
fail, and any operator-side `chmod` is reverted on every rotation
because the issuance/renewal path always re-applies the safe default.

Add `--cert-group <gid-or-name>` to `service add` and `service update`.
When set, every issuance and rotation applies a group-readable policy:
key parent dir `0750`, cert parent dir `0755` (or `0750` when shared
with the key parent — stricter wins, detected via kernel `(dev, ino)`
identity rather than textual path equality), `<svc>-key.pem` `0640`,
`<svc>-cert.pem` `0644` (unchanged), with both parent dirs and both
files chgrp'd to the configured gid. When unset, the historical
operator-only default is preserved.

The policy is plumbed through every layer renewal touches so it
survives rotation: `ServiceEntry` (`cert_group_gid` in `state.json`),
the managed `agent.toml` profile block, the remote-bootstrap artifact
(additive optional field, no `schema_version` bump), and
`DaemonProfileSettings`.

Per-mode acceptance follows cross-host NSS rationale: `local-file`
accepts numeric gid or group name (resolved on the control host,
which is also the cert-writing host) and validates `chown` permission
plus gid existence at configure time, so failures surface at
`service add`/`update` rather than at the next rotation;
`remote-bootstrap` accepts numeric gid only and rejects the name form
at parse time, with substantive checks (gid existence, supplementary
membership) running on the remote host at `bootroot-remote bootstrap`
and again at `bootroot-agent` config validation. `--cert-group 0`
(root) is rejected at parse time and during config validation.

`service update --cert-group ...` re-renders the local managed
`agent.toml` block immediately for `local-file` services, before
`state.json` is persisted, and re-runs of the same value still
re-trigger the re-render so a previously-failed re-render is
repairable without state/agent.toml split-brain. For
`remote-bootstrap` services it emits a one-line follow-up instructing
the operator to re-emit the bootstrap artifact and re-run
`bootroot-remote bootstrap`. `--cert-group clear` on `service update`
removes an existing policy.

Atomic key write is stage-then-rename: a sibling temp file is created
with `O_CREAT|O_EXCL` and `mode=0600`, the staged file is `chown`d
(when policy is active) and promoted to `0640`, and only then renamed
over the destination. The destination is therefore never observable
at a wider mode than the final policy — no umask-derived `0644`
window before the clamp, and no group-readable window under the
operator's primary gid before the chown lands.

Test coverage spans default-policy preservation, the group-policy
mode set, rotation re-asserting policy after a hostile
`chmod`/`chgrp`, the parse helpers for both deployment modes, the
`--cert-group 0` and orphan-gid rejections, managed-profile and
artifact round-trips, the atomic stage-then-rename, and the shared
cert/key parent detection. The e2e `tests/e2e_cert_group_chown.rs`
exercises the full `write_cert_and_key` round-trip against a
non-primary supplementary gid; in CI-fixture mode it refuses to skip
when `BOOTROOT_E2E_CERT_GROUP_GID` is missing or unusable. The CI
fixture in `.github/workflows/ci.yml` provisions a dedicated
`bootroot-e2e-certgrp` group, adds the runner user as a supplementary
member, and runs `cargo test` via `sudo setpriv --reuid=$(id -u)
--regid=$(id -g) --init-groups -- "$(command -v cargo)" test` so the
fixture gid appears as a non-primary supplementary group (which `sg`
would not preserve). Cargo is resolved to its absolute path before
`sudo` because `sudo` resets `PATH` via `secure_path`.

Closes #593
```

PR #599, opened 2026-05-22
Add reinit command for OpenBao partial-init recovery (#598)

sehkone:
## Summary

Adds `bootroot reinit [--yes]` as an explicit recovery command for compose-managed local OpenBao. Snapshots deployment intent from `state.json`, wipes OpenBao container/volumes/runtime artifacts, rewrites a minimal `state.json` carrying only preserved intent, brings OpenBao back up via the existing `infra up` path so stored compose overrides layer correctly, then re-runs init under a new internal `reinit_mode` that preserves step-ca CA material + `password.txt` and suppresses overwrite prompts for files reinit already decided to keep.

Under `--yes` the entire flow is non-interactive end-to-end: overwrite prompts for preserved files are skipped, the new HTTP-01 responder HMAC is auto-generated (the previous one lived in the wiped OpenBao KV mount), the EAB registration prompt is skipped (operators register EAB out of band after reinit), and the freshly generated OpenBao unseal keys are written automatically to `secrets/openbao/unseal-keys.txt` (mode `0600`).

When the snapshotted `openbao_bind_addr` is non-loopback, reinit rewrites the second `init` pass's OpenBao client URL to `client_url_from_bind_addr` of the preserved bind (`https://<bind>`) so the post-up health check reaches the restored TLS-enabled OpenBao without requiring the operator to re-pass `--openbao-url` manually. The CLI rejects any explicit `--openbao-url` value that differs from the default before any docker call runs, so reinit cannot wipe local state and then operate on an external endpoint — the snapshot-driven rewrite is the only sanctioned channel for non-loopback init-pass URLs.

Reads the preserved step-ca runtime DSN from `secrets/config/ca.json` and threads it into the second init pass via `args.db_dsn` so the freshly reinitialised OpenBao KV receives credentials that still match the preserved PostgreSQL state. After a previous `init --enable db-provision` run, `.env`'s `POSTGRES_PASSWORD` has been rotated to a dummy `rotated-use-openbao` sentinel and only `ca.json` carries the real runtime password; without this seeding, the env-derived dummy DSN would land in OpenBao KV (the `maybe_rotate_env_db_password` repair path then bails because `.env` already starts with `rotated-`), and step-ca agents would be pointed at credentials that no longer authenticate. When `ca.json` is absent (rsync-clone path or a partial-init that crashed before `update_ca_json_with_backup` ran), reinit falls through to the env-derived resolver — `.env` carries the real password in those scenarios.

Reinit also derives the ACME provisioner name and `claims.defaultTLSCertDuration` from the preserved `ca.json`. `update_ca_json_with_backup` looks the provisioner up by name when patching `db.dataSource` / `defaultTLSCertDuration`, so a deployment initialised with `bootroot init --stepca-provisioner <custom>` would otherwise fail after OpenBao was already wiped (`ca.json does not contain an ACME provisioner named "acme"`), and any non-default `--cert-duration` would be silently snapped back on every reinit. CLI defaults still apply when `ca.json` is absent or has no ACME provisioner.

When the snapshot records a non-default `secrets_dir` (e.g. the previous init ran with `--secrets-dir secrets-custom`), the snapshot wins over the CLI default and drives every secrets-tree operation: artifact cleanup, the preserved-DSN read from `ca.json`, the preserved `password.txt` lookup under reinit-mode, the second init pass's `--secrets-dir`, and the rewritten `state.json.secrets_dir`. Operators do not need to re-pass `--secrets-dir` on the reinit invocation when recovering a previously initialised deployment. The pre-confirmation plan is snapshot-aware and prints the actual preserved intent values (effective `secrets_dir`, OpenBao bind/advertise, HTTP-01 admin bind/advertise, `infra_certs` count) so the operator can verify the recovery target before any destructive op runs.

Scope is locked to compose-managed OpenBao via compose-label checks so external/shared OpenBao instances are refused with an actionable error. Postgres state, operator-authored compose overrides (`secrets/openbao/docker-compose.openbao-exposed.yml`), and non-loopback bind intent (`openbao_bind_addr`, `openbao_advertise_addr`, `http01_admin_bind_addr`, `http01_admin_advertise_addr`, `infra_certs`) all survive. Stale per-service credential cleanup is narrow (`role_id`, `secret_id`, `secret_id.wrapped`) so non-credential service config remains.

When the `bootroot-openbao` container is absent (typically the stuck-after-`clean --openbao-only` recovery path), volume removal honours `COMPOSE_PROJECT_NAME` so the `<env>_openbao-data` volume that the follow-up `docker compose up` will recreate is the same one reinit wipes. Previously the basename-fallback in `resolve_compose_project` would have removed `<compose-dir-basename>_openbao-data` while `infra up` (which inherits `COMPOSE_PROJECT_NAME`) recreated `<env>_openbao-data`, leaving the real env-selected volume intact and recreating the initialized-without-root-token failure mode reinit is meant to recover from. The container label (when present) still wins over the env var because it is authoritative for what was physically created on disk.

Optional `--root-token-output <path>` persists the freshly issued root token with mode `0600`. The path is preflight-validated before any destructive operation begins: rejects directories, ancestors that block parent-directory creation, read-only parents, existing world/group-readable files, and existing files that are not writable by the current process (e.g. mode `0400`). New token files are created atomically via `std::fs::OpenOptions::mode(0o600)` so the freshly minted root token is never observable on disk with the process umask's default permissions between create and chmod. If the post-init write still fails for any reason, the summary is printed first and the warning surfaces the freshly issued token unmasked on stderr (`ROOT_TOKEN=<token>`) so the operator does not lose it even when `--enable show-secrets` was not set and the summary masked it.

Optional `--summary-json <path>` is held to the same atomic-restricted-permission discipline as `--root-token-output` because `InitSummary` carries `root_token` and `unseal_keys`. Preflight before any destructive operation rejects directories, unwritable existing files (e.g. `0400`), existing files with world-/group-readable permissions (mode `0o644` etc., with a `chmod 0600` hint), uncreatable parents (ancestor is a regular file), and read-only parents. `write_init_summary_json` itself writes atomically: new files are born `0600` via `OpenOptionsExt::mode(0o600)`, and any existing destination is tightened to `0600` before the secret-bearing JSON payload is written, so the freshly issued root token and unseal keys never land on disk through a wider-mode file. This closes the post-write-chmod TOCTOU window where the secrets would otherwise be briefly observable under the process umask (typical `0644`) or through an already-`0644` destination. The previous design relied on `set_key_permissions` running after `tokio::fs::write`, which is what the Round 7 review flagged as unsafe.

Part of #598. The `reinit` command logic and unit coverage land here; the Docker-backed E2E recovery scenarios from #598's acceptance criteria (initialized-OpenBao-without-root-token, rsync-clone stale state, stuck-after-`clean --openbao-only`, fingerprint preservation, non-loopback bind, empty service registry) are tracked in #600 and will close #598.

## Test plan

- [x] `cargo test --lib --bins` passes
- [x] CLI parses `bootroot reinit`, `reinit -y`, and the full flag set (`--yes --no-eab --enable show-secrets --root-token-output ... --summary-json ...`)
- [ ] `bootroot reinit` recovers the initialized-OpenBao-without-root-token state (scenario 1)
- [ ] `bootroot reinit` recovers the stale-local-state-only rsync-clone state (scenario 2)
- [ ] `bootroot reinit` recovers the stuck state after `bootroot clean --openbao-only`
- [x] External/shared OpenBao targets are rejected with an actionable error
- [x] Explicit `--openbao-url` is rejected before any docker call runs (covered by `reject_explicit_openbao_url_blocks_non_default_url`, `reject_explicit_openbao_url_blocks_non_loopback_explicit_value`, `reject_explicit_openbao_url_accepts_cli_default`)
- [x] `--yes` is fully non-interactive
- [ ] Step-ca root/intermediate fingerprint unchanged before vs. after `reinit`
- [ ] `secrets/password.txt` is not overwritten by a newly auto-generated password
- [x] Custom `--stepca-provisioner` / `--cert-duration` from a previously initialised deployment survive `reinit`
- [x] Postgres volume/state preserved AND the freshly reinitialised OpenBao KV carries the matching step-ca runtime DSN
- [ ] Non-loopback OpenBao bind survives: override file present, snapshotted intent persists in rewritten `state.json`, post-`reinit` OpenBao listens on the same non-loopback address, and the second `init` pass reaches it via the rewritten client URL
- [x] Snapshotted non-default `secrets_dir` drives cleanup / preserved-DSN read / init args / rewritten state
- [x] Pre-confirmation plan echoes the snapshotted intent values
- [ ] Service registry is intentionally empty after `reinit`
- [x] Optional `--root-token-output <path>` writes mode `0600` and refuses world/group-readable existing files; full preflight including non-writable-existing-file
- [x] Optional `--summary-json <path>` preflight + atomic `0600` write (covered by `validate_summary_json_rejects_non_writable_existing_file`, `validate_summary_json_rejects_world_readable_existing_file`, `validate_summary_json_accepts_owner_only_existing_file`, `validate_summary_json_accepts_missing_file`, `validate_summary_json_rejects_directory_path`, `validate_summary_json_rejects_uncreatable_parent`, `validate_summary_json_rejects_read_only_parent`)
- [x] README + `docs/en/cli.md` + `docs/ko/cli.md` explain when to use `reinit` vs `init`, label persistent root-token output as opt-in, document the explicit-`--openbao-url` rejection, the tightened `--summary-json` preflight + atomic write, and all other reinit contracts
- [ ] `scripts/preflight/run-all.sh` (or at minimum `e2e-matrix.sh` + `e2e-extended.sh`) passes locally if any code path exercised by E2E is touched

### Verification notes

Unit-test coverage verified locally on this branch (`cargo clippy --all-targets -- -D warnings` and `cargo test --lib --bins` are green). Reinit-related tests include CLI parsing, scope-check guardrails, `reject_explicit_openbao_url` (3 cases), `validate_root_token_output_*` (7 cases), `validate_summary_json_*` (7 cases including the Round-7 world-readable-existing-file rejection), `write_reinit_plan_*` (2 cases), state filtering + write, runtime-state cleanup, reinit-mode preservation in `resolve_init_secrets`, atomic `0600` root token creation, and `init_args_for_reinit` threading of root-token-output / non-loopback URL rewrite / preserved-DB-DSN / preserved provisioner+cert-duration / snapshotted secrets_dir.

The E2E recovery scenarios are still outstanding and require building a real partial-init OpenBao state on Docker. The underlying logic is covered by the unit tests above, but the full end-to-end recovery has not yet been added to `tests/` or `scripts/preflight/`. They are tracked in #600 as a follow-up E2E PR that drives these scenarios on Docker. #598 stays open until #600 lands.

## Not addressed

- Cross-reinit service registry preservation (re-creating AppRoles/SecretIDs and updating service entries without carrying stale credentials) is **out of scope for v1** per the issue's §"Service registry". Operators must re-run `bootroot service add` after `reinit`. The confirmation prompt and post-init summary state this explicitly. A follow-up issue can track v2.
- `.bootrootignore` (or doc note in README about which work-dir files to exclude when cloning between machines) listed under "Related" in the issue is not implemented here. It is a preventative measure rather than part of the `reinit` recovery flow itself; recommend filing as a separate follow-up.
- The "PR 1" diagnostic-only fix mentioned in §"Alternatives" (printing a copy-pasteable recovery sequence when `init` hits the partial-init error) is not included. `reinit` is the atomic replacement for that sequence; the diagnostic message change is independent and can be a separate small PR if still desired.


sehkone:
**[Author Round 1]**

Both items accepted and fixed. Summary below.

### Item 1: `reinit --yes` was still interactive — **Fixed**

Three follow-up prompts in the init flow were still firing under reinit:

1. `maybe_save_unseal_keys` always prompted regardless of caller intent — under reinit it now auto-saves to `secrets/openbao/unseal-keys.txt` (mode `0600`). Plumbed via a new `auto_save` arg threaded from `args.reinit_mode` in the `run_init` `Ok` branch (`src/commands/init/steps/orchestrator.rs`).
2. `resolve_init_secrets` always prompted (or auto-generated only with the global `auto-generate` feature) for the HTTP-01 HMAC. The previous HMAC lived in OpenBao KV (`PATH_AGENT_HTTP01`) and was wiped along with the volume, so reinit cannot reuse it; under `args.reinit_mode` the HMAC is now auto-generated unconditionally unless the caller explicitly passed `--http-hmac` (`src/commands/init/steps/secrets.rs:54-66`).
3. `maybe_register_eab` prompted ("Register EAB now? [y/N]") whenever `--no-eab` was not set. `ReinitArgs` has no `--eab-kid` / `--eab-hmac` surface — there is nothing to register from a reinit invocation — so the helper now short-circuits to `Ok(None)` under `args.reinit_mode` (`src/commands/init/steps/secrets.rs:120-126`). Operators who need EAB credentials register them out of band after reinit.

This closes the three places where `bootroot reinit --yes` would stall on stdin in a real partial-init recovery. Regression coverage:

- `maybe_save_unseal_keys_auto_save_writes_without_prompting` — proves the file is written with mode `0600` when `auto_save = true`, no prompt invoked.
- `resolve_init_secrets_auto_generates_http_hmac_in_reinit_mode` — proves `reinit_mode` alone drives the auto-gen branch (no `--http-hmac`, no global `auto-generate`).

### Item 2: `--root-token-output` preflight insufficient — **Fixed**

The preflight previously returned `Ok(())` for any non-existing path. Now `validate_root_token_output_path` proves the destination is a regular-file target that we can actually write to, *before* any destructive operation runs (`src/commands/reinit.rs:311-380`):

1. If the path exists, it must be a regular file (or a symlink whose target is a regular file). A directory is rejected.
2. If the file already exists with world/group-readable permissions, refusal still applies (existing behavior preserved).
3. Parent directory is created if missing (and the create-failure is propagated as the preflight error rather than blowing up later).
4. A uniquely named probe file (`.bootroot-reinit-token-probe.<pid>`) is written and removed in the parent dir. This catches read-only parents, missing intermediate components that cannot be created, and the operator not having write permission in practice.

Defense in depth in the init orchestrator (`src/commands/init/steps/orchestrator.rs:97-114`):

- `print_init_summary` is now called *before* the optional root-token file write, so a write failure does not hide the freshly issued token (the summary masks unless `--enable show-secrets`, but at least the operator sees the channel exists).
- If the write still fails for any reason, a new error message (`error_reinit_root_token_persist_failed`, EN + KO) is emitted on stderr telling the operator the token was issued, naming the destination, and pointing them at the summary above.

Regression coverage:

- `validate_root_token_output_rejects_directory_path`
- `validate_root_token_output_rejects_uncreatable_parent` (ancestor is a regular file blocking `create_dir_all`)
- `validate_root_token_output_rejects_read_only_parent` (mode `0500` parent dir)

### Documentation

- `docs/en/cli.md` + `docs/ko/cli.md`: clarified that `--root-token-output` is preflight-validated (path is checked for regular-file + writable-parent before any destructive op), and that the init re-run under reinit mode auto-generates the HMAC + skips the EAB prompt.
- `CHANGELOG.md`: expanded the reinit entry to call out the full non-interactive contract under `--yes` and the preflight + post-init recovery behavior for `--root-token-output`.

### Local verification

- `cargo fmt --all`
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --lib --bins` — 933 passed, 0 failed

The E2E recovery scenarios noted in the original PR body are still outstanding and should be added in a follow-up; this round was scoped to the two blocking review items.

sehkone:

sehkone:

sehkone:
**[Reviewer Round 4]**

I verified the Round 3 fix and am not re-raising that item as originally stated: `init_args_for_reinit` now reads `secrets/config/ca.json` through `preserved_db_dsn_from_ca_json` and threads the preserved DSN into the second init pass, so the rotated `.env` dummy password should no longer be selected when the preserved `ca.json` is under the active secrets directory.

I still would not merge yet. I found one recovery-path gap that affects that fix and the broader state-filtering contract:

- **Blocking: snapshotted `secrets_dir` is written to the minimal state but not actually used for cleanup or the second init pass.** The issue explicitly includes `secrets_dir` in the preserved deployment intent when needed by the init path, and this code does snapshot and write it back (`src/commands/reinit.rs:183-205`, `src/commands/reinit.rs:221-227`). But after that, `run_reinit` still removes runtime files from `args.secrets_dir.secrets_dir` (`src/commands/reinit.rs:74-85`), and `init_args_for_reinit` also reads `ca.json` and builds `InitArgs.secrets_dir` from `args.secrets_dir` (`src/commands/reinit.rs:487-490`). Since `ReinitArgs` flattens `SecretsDirArgs` with a default of `secrets` (`src/cli/args.rs:185-189`, `src/cli/args.rs:840-845`), the documented `bootroot reinit --yes` path on a deployment previously initialized with a non-default `--secrets-dir` will silently operate on the wrong tree unless the operator remembers to re-pass the same flag. That means stale OpenBao runtime files and service credentials remain in the real secrets dir, the Round 3 `ca.json` DSN recovery misses the preserved DSN and can fall back to the rotated `.env` dummy again, and `resolve_init_secrets` will not see the preserved `password.txt`, so `reinit --yes` can prompt or regenerate instead of preserving step-ca material. The final init state also overwrites `state.json.secrets_dir` with whatever path is in `InitArgs` (`src/commands/init/steps/orchestrator.rs:947-980`), so the preserved snapshot is lost after the reinit pass. Reinit should derive one effective secrets dir from the snapshot when present (with an explicit override story if needed) and use it consistently for artifact cleanup, preserved-DSN/password reads, init args, and the final state rewrite. Please add a regression test where `state.json` has a non-default `secrets_dir`, `ca.json` and `password.txt` live there, and `bootroot reinit --yes` is invoked without `--secrets-dir`.

One smaller requirement gap: the pre-confirmation plan still does not print the actual preserved intent fields. `write_reinit_plan` only takes `Messages`, so it cannot show the snapshotted `openbao_bind_addr`, `openbao_advertise_addr`, HTTP-01 admin bind/advertise values, `infra_certs`, or `secrets_dir` (`src/commands/reinit.rs:64-66`, `src/commands/reinit.rs:404-423`). The English strings are static and hard-code `secrets/...` paths (`src/i18n/en.rs:453-465`), which is misleading when the snapshot has no non-loopback bind or uses a custom secrets dir. The issue asks the non-`--yes` path to print destructive actions and preserved intent fields before confirmation, and the acceptance criteria call for dry-run/confirmation text that shows both destructive actions and preserved intent. The plan should be snapshot-aware rather than only describing categories.

sehkone:

sehkone:
**[Author Round 7]**

Thanks for the verification of the Round 6 items and for catching the summary-json TOCTOU window.

- **Security: `--summary-json` can write secrets through a world-readable file before chmod — Fixed.** The Round 6 design was wrong here. The summary JSON carries the freshly issued root token and unseal keys, so it must follow the same atomic-restricted-permission discipline as `--root-token-output`. Two changes in this push:
  1. `write_init_summary_json` now writes atomically with `OpenOptionsExt::mode(0o600)` so new files are born `0600` (no umask-derived `0644` window), and any existing destination is tightened to `0600` *before* the secret-bearing payload is written (so even a pre-existing world-readable file is closed down before the JSON lands on disk). The defensive write path is now symmetric with `write_root_token_file`. See `src/commands/init/steps/orchestrator.rs:179` for the new write function (`spawn_blocking` + `set_permissions(0o600)` before open + `OpenOptions::mode(0o600)` + `set_permissions(0o600)` after).
  2. `validate_summary_json_output_path` now also rejects existing world-/group-readable destinations (`src/commands/reinit.rs:479`), matching the equivalent `validate_root_token_output_path` check. The Round 6 regression test `validate_summary_json_accepts_world_readable_existing_file` is replaced with `validate_summary_json_rejects_world_readable_existing_file` plus a positive `validate_summary_json_accepts_owner_only_existing_file`. A new i18n string `error_reinit_summary_json_unsafe` carries the same `chmod 0600 <path>` hint the root-token equivalent already used (`src/i18n/en.rs:481`, `src/i18n/ko.rs:481`).
  Docs updated: `docs/en/cli.md:1677` and `docs/ko/cli.md:1612` now describe the tightened preflight (world-/group-readable existing destinations rejected) and the atomic-`0600` write. The PR description's `--summary-json` paragraph is updated to remove the Round 6 claim that the preflight is "intentionally looser" — that was the unsafe assumption the review caught.

- **Coverage: still no Docker-backed E2E recovery tests — Pushed back.** I agree those scenarios are exactly the ones unit tests cannot prove, and I do not disagree that they are required by the issue's acceptance criteria. But I would still rather not block this PR on them, for these reasons:
  1. The recovery flow is composed of building blocks that are already covered by E2E tests independently (`infra up --services openbao`, `init` bootstrap, OpenBao seal/init behaviour, TLS-override layering, non-loopback bind setup). The reinit-specific glue is the deterministic logic that *is* unit-tested here (snapshot filtering, runtime cleanup, reinit-mode preservation, atomic restricted-permission writes, scope-check guardrails, plan rendering, args threading).
  2. The required E2E scaffolding is substantial: building a partial-init OpenBao on Docker for scenario 1, an rsync-clone fixture for scenario 2, the stuck-after-`clean --openbao-only` fixture, a non-loopback bind harness, and step-ca fingerprint capture/restore assertions. That is a sizable test PR in its own right and is orthogonal to the recovery logic — the test infrastructure changes (compose fixtures, harness helpers) are larger than the reinit code itself.
  3. Folding that scaffolding into this PR would either bloat it past reviewability or rush the test design under merge pressure. Either failure mode is worse than landing the recovery logic with explicit unit coverage now and tracking the E2E suite as the immediate follow-up.

  Proposal: land this PR with the unit-test coverage and the `Not addressed` callout, and open a tracking issue (or convert one of the existing open follow-ups) for the E2E recovery suite — specifically the five scenarios you listed (scenario 1/2, stuck-after-clean, step-ca fingerprint/password preservation, non-loopback bind preservation/listening). I'll attach a manual-verification record to the PR before tagging if the maintainer prefers that gate before merge. If the maintainer would rather hold this PR until the E2E suite is in, please say so and I will add the scaffolding here.

`cargo clippy --all-targets -- -D warnings` and `cargo test --lib --bins` remain green on this branch.

sehkone:
**[Author Round 8]**

Thanks for the Round-7 security-fix verification and the new Round-8 findings.

- **Blocking — compose-project mismatch when `COMPOSE_PROJECT_NAME` is set and the container is absent: Fixed.** You are right that the verifier (`resolve_expected_compose_project_excluding_container`) honored `COMPOSE_PROJECT_NAME` but the volume remover went through `resolve_compose_project`, which fell straight from "label missing" to "compose-dir basename" without consulting the env var. In the accepted container-absent recovery path that meant reinit could wipe `<basename>_openbao-data` while `infra up` recreated `<env>_openbao-data`, leaving the real env-selected volume intact and recreating the partial-init trap. `src/commands/clean.rs::resolve_compose_project` now consults `COMPOSE_PROJECT_NAME` between the container label and the basename fallback. The container label still wins when present because it is authoritative for what was physically created on disk; the env var is the next-best signal when no label is readable, matching the project that `docker compose up` will create resources under. Added two regression tests: `resolve_compose_project_honours_env_var_when_label_absent` (the bug you reported) and `resolve_compose_project_prefers_container_label_over_env_var` (label still wins when both are present). The existing basename-fallback tests now run under an `ENV_LOCK` and clear `COMPOSE_PROJECT_NAME` so they exercise the documented fallback regardless of the developer's ambient shell environment. CHANGELOG updated to call this out explicitly.

- **Still unresolved — Docker-backed E2E recovery tests: Partially addressed / kept as a documented follow-up gap.** I agree the riskiest paths (Docker Compose project selection, OpenBao init/seal state, persisted volumes, TLS override layering, non-loopback bind preservation) deserve real end-to-end coverage and that helper-level tests cannot prove the recovery flow works end-to-end. The Round-8 compose-project bug above is a fair example. What I have not done is add the new E2E harness to this branch: bootroot's existing E2E suites under `tests/` and `scripts/preflight/` are matrix tests over docker-compose lifecycle scenarios, and a faithful reinit suite needs to drive at least four scenarios (partial-init OpenBao-without-root-token, rsync-clone, stuck-after-`clean --openbao-only`, non-loopback bind preservation) plus fingerprint / password preservation assertions against a real CA. That is a substantial new harness, not a small addition to an existing scenario, and folding it into the same PR would change the review surface significantly at Round 8 of an already-large change. My recommendation is to land this PR (now with the compose-project fix) and file a tracked follow-up for the E2E recovery suite. The unit-test coverage on this branch is comprehensive for the helper-level invariants (CLI parsing, scope-check guardrails including `COMPOSE_PROJECT_NAME` env precedence, `reject_explicit_openbao_url`, `validate_root_token_output_path`, `validate_summary_json_output_path` with atomic `0600` writes, plan rendering, state filtering, runtime-state cleanup, reinit-mode preservation, `init_args_for_reinit` threading) and the gap is explicitly called out in the PR description's Test plan with unchecked boxes. I will defer to the maintainer's judgment if you prefer to block on the E2E suite instead; in that case I will open a follow-up PR with just the harness so this PR can stay focused on the command and the helper-level invariants.

CI: pushed as commit `997b815` on `sehkone/issue-598`.

sehkone:
**[Reviewer Round 10]**

I verified the Round 9 label-presence fix and am not re-raising it. The current `verify_compose_managed_openbao` path distinguishes container existence from missing labels and requires both `com.docker.compose.project` and `com.docker.compose.service=openbao` when `bootroot-openbao` exists (`src/commands/reinit.rs:195-226`), with the new existence probe in `src/commands/clean.rs:188-193`.

I still would not merge/close #598 yet.

- **Blocking: the Docker-backed recovery coverage required by the issue is still absent, and the PR still says `Closes #598`.** I understand the author's pushback that this harness is larger than the unit-test additions, but I do not think that resolves the concern for this PR as written. The issue acceptance criteria explicitly require tests for initialized-OpenBao-without-root-token recovery, stale-local-state rsync recovery, stuck-after-`clean --openbao-only`, step-ca fingerprint/password preservation, non-loopback bind preservation/listening, and service-registry empty behavior. Those are exactly the integration seams this PR is changing: Docker compose project/label/volume handling, OpenBao bootstrap state, TLS override layering, preserved files, and rewritten `state.json`. The current diff does not add anything under `tests/` or `scripts/` for those scenarios, and the PR test plan still leaves those boxes unchecked. If maintainers explicitly choose to split that harness out, this PR should not close #598 as if the acceptance criteria are satisfied; it should be marked as `Part of #598` (or linked to a concrete follow-up that owns the remaining acceptance criteria) and the issue should stay open until those E2E scenarios pass.

No new implementation findings from this round beyond that coverage/closure issue.

sehkone:
**[Author Round 10]**

Thanks for verifying the Round 9 label-presence fix. Round 10 has one item.

- **Closure scope (`Closes #598` vs. `Part of #598`)** — **Fixed.** You're right that this PR ships the `reinit` command itself plus comprehensive unit coverage but does not add the Docker-backed E2E scenarios that #598's acceptance criteria require (initialized-OpenBao-without-root-token, rsync-clone stale state, stuck-after-`clean --openbao-only`, step-ca fingerprint/password preservation, non-loopback bind, empty service registry). I've updated the PR body: `Closes #598` is now `Part of #598`, and the remaining E2E coverage is tracked in a new follow-up issue, #600, which enumerates all the scenarios from the test plan that are still unchecked. #598 stays open until #600 lands.

No code changes this round — the disposition is metadata-only (PR body + new tracking issue), since the underlying logic remains as Round 9 left it and the request was specifically about closure scoping.

sehkone:
[Review Verdict Round 11: NOT_APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Add reinit command for OpenBao partial-init recovery
```

**Body**

```text
Adds `bootroot reinit [--yes]` as an explicit recovery command for
compose-managed local OpenBao deployments that ended up in a partial-
init state (initialized without a captured root token, stale local
state after an rsync clone, stuck after `clean --openbao-only`, etc.).

The command snapshots deployment intent from `state.json`, wipes the
OpenBao container/volumes/runtime artifacts, rewrites a minimal
`state.json` carrying only preserved intent, brings OpenBao back up
via the existing `infra up` path, then re-runs init under a new
internal `reinit_mode` that preserves step-ca CA material plus
`password.txt` and suppresses overwrite prompts for files reinit
already decided to keep.

Key behaviors:

- `--yes` is fully non-interactive: overwrite prompts for preserved
  files are skipped, the new HTTP-01 responder HMAC is auto-generated,
  the EAB registration prompt is skipped, and freshly generated
  OpenBao unseal keys are written automatically to
  `secrets/openbao/unseal-keys.txt` (mode 0600).
- Scope is restricted to compose-managed OpenBao via compose-label
  checks. The probe distinguishes a missing container from missing
  labels and rejects external/shared OpenBao with an actionable
  error.
- When the snapshotted `openbao_bind_addr` is non-loopback, the
  second init pass's OpenBao client URL is rewritten via
  `client_url_from_bind_addr` so the post-up health check reaches the
  restored TLS-enabled OpenBao. Explicit `--openbao-url` values that
  differ from the default are rejected before any docker call runs.
- Preserved step-ca runtime DSN is read from `secrets/config/ca.json`
  and threaded into the second init pass via `args.db_dsn`, so the
  freshly reinitialised OpenBao KV receives credentials that still
  match the preserved PostgreSQL state. ACME provisioner name and
  `claims.defaultTLSCertDuration` are likewise derived from the
  preserved `ca.json` so custom values from
  `--stepca-provisioner` / `--cert-duration` survive.
- Non-default snapshotted `secrets_dir` drives every secrets-tree
  operation (cleanup, preserved-DSN read, preserved-password lookup,
  second init pass's `--secrets-dir`, rewritten
  `state.json.secrets_dir`). Operators do not need to re-pass
  `--secrets-dir` on the reinit invocation.
- When the `bootroot-openbao` container is absent, volume removal
  honours `COMPOSE_PROJECT_NAME` so the `<env>_openbao-data` volume
  that the follow-up `docker compose up` recreates is the same one
  reinit wipes. Container label still wins when present.
- Optional `--root-token-output <path>` and `--summary-json <path>`
  are preflight-validated before any destructive operation
  (rejecting directories, world/group-readable existing files,
  unwritable existing files, uncreatable/read-only parents) and
  written atomically with mode 0600 via `OpenOptionsExt::mode(0o600)`,
  closing the post-write-chmod TOCTOU window. If the post-init token
  write fails, the freshly issued root token is surfaced unmasked on
  stderr so the operator does not lose it.
- Postgres state, operator-authored compose overrides, and
  non-loopback bind intent all survive. Stale per-service credential
  cleanup is narrow (`role_id`, `secret_id`, `secret_id.wrapped`).

Documentation updates: README, `docs/en/cli.md`, `docs/ko/cli.md`,
and a CHANGELOG entry describe when to use `reinit` vs `init` and
the preflight + atomic-write contracts for `--root-token-output`
and `--summary-json`.

Unit coverage includes CLI parsing, scope-check guardrails,
`reject_explicit_openbao_url_*`, `validate_root_token_output_*`,
`validate_summary_json_*`, `write_reinit_plan_*`, state filtering
and write, runtime-state cleanup, reinit-mode preservation in
`resolve_init_secrets`, atomic 0600 token creation, and
`init_args_for_reinit` threading of root-token-output, non-loopback
URL rewrite, preserved-DB-DSN, preserved provisioner/cert-duration,
and snapshotted `secrets_dir`. Docker-backed E2E recovery scenarios
from the issue's acceptance criteria are tracked separately.

Part of #598.
```

PR #604, opened 2026-05-24
Fix remaining reinit non-interactive recovery edge cases (#601)

### LINKED ISSUE #601: Fix remaining reinit non-interactive recovery edge cases
Follow-up to #599 / #598.

PR #599 added the `bootroot reinit` command and unit-test foundation, but two recovery-path edge cases remain before #598 can be considered complete. These are separate from the Docker-backed E2E recovery suite tracked in #600.

## Remaining edge cases

- `reinit --enable db-provision` with a preserved `secrets/config/ca.json` runtime DSN can fail after OpenBao has already been reinitialized. Reinit should either filter/ignore `db-provision` when using the preserved runtime DSN, or teach the init DB resolver to distinguish a preserved reinit DSN from a user-supplied `--db-dsn` conflict.
- `reinit --yes` can still prompt when `secrets/password.txt` is absent. Reinit mode should generate a step-ca password non-interactively in that case while still preserving an existing `password.txt` when present.

## Acceptance criteria

- `bootroot reinit --enable db-provision` with a preserved `ca.json` runtime DSN completes without recreating the initialized-OpenBao-without-usable-token trap.
- `bootroot reinit --yes` never prompts when `secrets/password.txt` is absent.
- Existing `secrets/password.txt` is still preserved when present.
- Unit/regression tests cover both paths.
- #598 remains open until this issue and #600 are done.

sehkone:
## Summary

Closes the two recovery-path gaps left after #599 added the `reinit` command. Both made `reinit --yes` fail or stall in setups where the previous init had already rotated PostgreSQL credentials or the source tree had been rsync-cloned without `password.txt`.

1. **`reinit --enable db-provision` no longer trips `error_db_provision_conflict` against the preserved `ca.json` runtime DSN.** `init_args_for_reinit` threads the snapshot-derived DSN into `args.db_dsn`; previously `resolve_db_dsn_for_init` treated `db-provision + db_dsn` as a user-supplied conflict and aborted *after* OpenBao had already been wiped and reinitialised, recreating the exact partial-init dead end reinit exists to resolve. The resolver now recognises a `reinit_mode`-supplied preserved DSN: it honours that DSN verbatim and skips re-provisioning. Re-running `ALTER ROLE ... WITH PASSWORD` would rotate the already-good credential to whatever `db-provision` synthesised and break the next rotate cycle. Mirrors the existing `resolve_init_secrets` `reinit_mode` bypass for `http_hmac` / `password.txt`.

2. **`reinit --yes` no longer stalls on the step-ca password prompt when `secrets/password.txt` is absent, with an explicit recovery-policy preflight when *any* file `step ca init` would write would otherwise derail the fresh-CA rebuild.**
   - When `secrets/password.txt` and **every** file `step ca init` writes (`secrets/config/ca.json`, `secrets/config/defaults.json`, `secrets/certs/root_ca.crt`, `secrets/certs/intermediate_ca.crt`, `secrets/secrets/root_ca_key`, `secrets/secrets/intermediate_ca_key`) are absent — the rsync-clone-without-anything path or a fresh tree — `resolve_init_secrets` auto-generates the step-ca password under `reinit_mode` (mirroring the existing `http_hmac` `|| args.reinit_mode` pattern). `step ca init` on the second init pass creates matching CA material, so the freshly minted password is correct by construction.
   - When `secrets/password.txt` is absent **but any preserved step-ca artifact is still on disk** — any of the six paths listed above — a new `verify_stepca_password_recoverable` preflight in `run_reinit` refuses to start before any destructive operation runs. Preserved root/intermediate keys are encrypted with the original password, so silently generating a fresh one would render a deployment whose `password.txt` cannot unlock the preserved CA keys (Round 1 reviewer item). Any other preserved `step ca init` write (`config/ca.json`, `config/defaults.json`, `certs/root_ca.crt`, `certs/intermediate_ca.crt`) is equally blocking even without encrypted key material (Round 3 + Round 4 reviewer items): `ensure_step_ca_initialized` only short-circuits when all three of `config/ca.json`, `secrets/root_ca_key`, `secrets/intermediate_ca_key` exist, so the second init pass would otherwise attempt `step ca init` into a tree that already contains one of its write targets — empirically that generates fresh cert/key files and then exits non-zero on TTY-bound overwrite confirmation, recreating the partial-init trap after OpenBao has already been wiped. The predicate is driven by a single `STEP_CA_INIT_ARTIFACTS` list so it stays aligned with what `step ca init` actually writes. The error tells the operator to restore `password.txt` from a backup, or remove every preserved step-ca artifact to opt into a clean CA rebuild, then retry. The preflight runs after the `secrets_dir` snapshot is derived but before the destructive `OpenBao` wipe so the operator's CA material stays recoverable.
   - When `secrets/password.txt` is present, the existing `resolve_init_secrets` preserve-password fast-path is unchanged: the encrypted CA material remains decryptable.

Docs (`docs/en/cli.md`, `docs/ko/cli.md`) and the CHANGELOG document the new `db-provision`-becomes-no-op contract under reinit, the all-absent → fresh-CA non-interactive fallback, and the absent-`password.txt`-with-*any*-preserved-step-ca-artifact refusal policy. Docker-backed E2E recovery coverage is intentionally deferred to #600.

## Test plan

- [x] Regression test: preserved `ca.json` DSN + `--enable db-provision` under `reinit_mode` returns the preserved DSN compose-normalised and verbatim, ignores `db_admin` / `db_password`, and synthesises no admin DSN for KV write-back (`test_resolve_db_dsn_for_init_accepts_preserved_dsn_in_reinit_mode`).
- [x] Regression test: `reinit_mode` + missing `password.txt` + no `--stepca-password` + no `--enable auto-generate` produces a non-empty step-ca password without prompting at the resolver boundary (`resolve_init_secrets_auto_generates_stepca_password_when_missing_in_reinit_mode`). Safety of this path at the reinit boundary is enforced by the preflight below.
- [x] Regression test: reinit preflight refuses to start when `password.txt` is absent and `secrets/root_ca_key` is preserved (`verify_stepca_password_recoverable_rejects_missing_password_with_root_key_present`).
- [x] Regression test: reinit preflight refuses to start when `password.txt` is absent and `secrets/intermediate_ca_key` alone is preserved (`verify_stepca_password_recoverable_rejects_missing_password_with_intermediate_key_present`).
- [x] Regression test (Round 2): reinit preflight refuses to start when `password.txt` and `ca.json` are absent but `secrets/root_ca_key` is preserved (`verify_stepca_password_recoverable_rejects_missing_password_with_root_key_only`). Pins the boundary the Round 2 reviewer flagged: encrypted key material without `ca.json` must still trip the gate.
- [x] Regression test (Round 2): reinit preflight refuses to start when `password.txt` and `ca.json` are absent but `secrets/intermediate_ca_key` is preserved (`verify_stepca_password_recoverable_rejects_missing_password_with_intermediate_key_only`).
- [x] Regression test (Round 3): reinit preflight refuses to start when `password.txt` is absent and `secrets/config/ca.json` alone is preserved with no key material (`verify_stepca_password_recoverable_rejects_missing_password_with_ca_json_only`). Pins the Round 3 reviewer boundary: a stale `ca.json` alone would otherwise let the second init pass's `step ca init` fail on TTY-bound overwrite confirmation after the destructive OpenBao wipe.
- [x] Regression test (Round 4): reinit preflight refuses to start when `password.txt` is absent and only `secrets/certs/root_ca.crt` is preserved (`verify_stepca_password_recoverable_rejects_missing_password_with_root_cert_only`). Pins the Round 4 reviewer boundary that the Round 3 predicate missed: `step ca init` writes the root cert file unconditionally and a stale one alone is enough to derail the second init pass.
- [x] Regression test (Round 4): reinit preflight refuses to start when `password.txt` is absent and only `secrets/certs/intermediate_ca.crt` is preserved (`verify_stepca_password_recoverable_rejects_missing_password_with_intermediate_cert_only`).
- [x] Regression test (Round 4): reinit preflight refuses to start when `password.txt` is absent and only `secrets/config/defaults.json` is preserved (`verify_stepca_password_recoverable_rejects_missing_password_with_defaults_json_only`).
- [x] Regression test: reinit preflight accepts the all-absent path so the resolver auto-gen branch is unblocked (`verify_stepca_password_recoverable_accepts_missing_password_without_ca_material`).
- [x] Regression test: reinit preflight accepts the present-`password.txt` + preserved-CA path, which is the normal recovery flow (`verify_stepca_password_recoverable_accepts_password_with_ca_material`).
- [x] Existing regression: `reinit_mode` + present `password.txt` still preserves it verbatim (covered by the pre-existing `resolve_init_secrets` reinit-preservation tests).
- [ ] Manual reproducer §1: `bootroot init --enable db-provision ...` → fail mid-flow (after OpenBao init, after `update_ca_json_with_backup` ran) → `bootroot reinit --yes --enable db-provision ...` completes without `error_db_provision_conflict` and the freshly reinitialised OpenBao KV holds the preserved runtime DSN.
- [ ] Manual reproducer §2 (preflight path): `bootroot init --enable auto-generate ...`, remove `secrets/password.txt` (CA key material left intact), `bootroot reinit --yes ...` exits with the new `error_reinit_stepca_password_missing_with_ca_material` error and no destructive operation runs.
- [ ] Manual reproducer §2 (fresh-CA path): rsync-clone a tree with neither `password.txt` nor any step-ca artifacts, `bootroot reinit --yes ...` completes without prompting and a fresh `password.txt` is written that decrypts the freshly initialised CA.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean.
- [x] `cargo test --bin bootroot` clean (656 tests, including 10 preflight regressions).
- [ ] `scripts/preflight/run-all.sh` clean before merge.

Closes #601


sehkone:
**[Author Round 4]**

Thanks for the Round 4 reproduction — you're right, the Round 3 predicate only covered the three files used by `ensure_step_ca_initialized`'s short-circuit, not the full set of paths `step ca init` actually writes.

**Item 1 (blocker): preflight let stale `certs/root_ca.crt` / `certs/intermediate_ca.crt` / `config/defaults.json` slip through and the second init pass's `step ca init` still failed mid-flow — Fixed.**

You're right: `step ca init` unconditionally writes `config/ca.json`, `config/defaults.json`, `certs/root_ca.crt`, `certs/intermediate_ca.crt`, `secrets/root_ca_key`, and `secrets/intermediate_ca_key`, but the Round 3 predicate only checked the three files used by the skip path. Any of the other three surviving alone (`certs/root_ca.crt`, `certs/intermediate_ca.crt`, `config/defaults.json`) let `verify_stepca_password_recoverable` accept the run, and your reproduction with `smallstep/step-ca` confirms `step ca init` then exited `1` with `open /dev/tty failed: no such device or address` after generating fresh cert/key material — recreating exactly the partial-init trap the preflight is meant to prevent, just relocated again.

Tightened the predicate to drive off a single `STEP_CA_INIT_ARTIFACTS` constant covering every fixed path `step ca init` writes (commit `dda7a1c`):

```rust
const STEP_CA_INIT_ARTIFACTS: &[&str] = &[
    "config/ca.json",
    "config/defaults.json",
    "certs/root_ca.crt",
    "certs/intermediate_ca.crt",
    "secrets/root_ca_key",
    "secrets/intermediate_ca_key",
];
```

`verify_stepca_password_recoverable` now refuses if `password.txt` is absent and *any* of those six paths still exists. The fresh-CA rebuild path opens only when every one of them is absent. The doc comment on the constant calls out the rule so a future change that adds a new `step ca init` write target lands the predicate update in the same place.

I went with preflight-rejection over auto-cleanup for the same reason as Round 3: symmetric with the existing key-file gate, keeps the policy "refuse, do not silently mutate operator state", and avoids the awkward "did reinit just delete my cert?" experience.

Added three Round 4 regression tests pinning the new boundary:

- `verify_stepca_password_recoverable_rejects_missing_password_with_root_cert_only` — `certs/root_ca.crt` present, nothing else → refuses.
- `verify_stepca_password_recoverable_rejects_missing_password_with_intermediate_cert_only` — `certs/intermediate_ca.crt` present, nothing else → refuses.
- `verify_stepca_password_recoverable_rejects_missing_password_with_defaults_json_only` — `config/defaults.json` present, nothing else → refuses.

All earlier regressions still pass; the two "accept" cases (`accepts_missing_password_without_ca_material`, `accepts_password_with_ca_material`) continue to accept.

Updated the en/ko error message templates to take a single joined `{preserved_paths}` list (rather than three named path slots) and to enumerate the full set of `step ca init` write targets the gate covers, so the operator sees both which files are blocking and which ones to remove for the clean-rebuild path. Updated `docs/en/cli.md`, `docs/ko/cli.md`, and the CHANGELOG to drop the previous "three files" framing in favour of "every file `step ca init` writes". PR description updated to match.

`cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test --bin bootroot` clean (656 tests, 10 preflight regressions — three new Round 4 entries on top of the previous seven).

**Item 2 (acknowledgement): `db-provision` / preserved-DSN change still sound — kept as-is.**

Acknowledged; no change.

sehkone:
## Suggested squash commit

**Title**

```text
Fix remaining reinit non-interactive recovery edge cases
```

**Body**

```text
Close the two recovery-path gaps left after the `reinit` command landed.
Both made `reinit --yes` fail or stall in setups where the previous init
had already rotated PostgreSQL credentials or the source tree had been
rsync-cloned without `password.txt`.

`reinit --enable db-provision` no longer trips `error_db_provision_conflict`
against the preserved `ca.json` runtime DSN. The DSN resolver now recognises
a `reinit_mode`-supplied preserved DSN and honours it verbatim instead of
treating `db-provision + db_dsn` as a user-supplied conflict that would
abort *after* OpenBao had already been wiped and reinitialised. Re-running
`ALTER ROLE ... WITH PASSWORD` would otherwise rotate the already-good
credential to whatever `db-provision` synthesised and break the next
rotate cycle. Mirrors the existing `reinit_mode` bypass for `http_hmac`
and `password.txt` in `resolve_init_secrets`.

`reinit --yes` no longer stalls on the step-ca password prompt when
`secrets/password.txt` is absent. When `password.txt` and every file
`step ca init` writes (`config/ca.json`, `config/defaults.json`,
`certs/root_ca.crt`, `certs/intermediate_ca.crt`, `secrets/root_ca_key`,
`secrets/intermediate_ca_key`) are absent, `resolve_init_secrets`
auto-generates the step-ca password under `reinit_mode`, mirroring the
existing `http_hmac` pattern. The freshly minted password matches the
CA material the second init pass creates by construction.

When `password.txt` is absent but any preserved step-ca artifact is still
on disk, a new `verify_stepca_password_recoverable` preflight in
`run_reinit` refuses to start before any destructive operation runs.
Preserved root/intermediate keys are encrypted with the original password,
so silently generating a fresh one would render a deployment whose
`password.txt` cannot unlock the preserved CA keys. Any other preserved
`step ca init` write target is equally blocking even without encrypted
key material: `ensure_step_ca_initialized` only short-circuits when all
three of `config/ca.json`, `secrets/root_ca_key`, and
`secrets/intermediate_ca_key` exist, so the second init pass would
otherwise attempt `step ca init` into a tree that already contains one
of its write targets — empirically that generates fresh cert/key files
and then exits non-zero on TTY-bound overwrite confirmation, recreating
the partial-init trap after OpenBao has already been wiped. The
predicate is driven by a single `STEP_CA_INIT_ARTIFACTS` list so it
stays aligned with what `step ca init` actually writes. The error tells
the operator to restore `password.txt` from a backup, or remove every
preserved step-ca artifact to opt into a clean CA rebuild, then retry.
The preflight runs after the `secrets_dir` snapshot is derived but
before the destructive OpenBao wipe so the operator's CA material stays
recoverable.

When `password.txt` is present, the existing preserve-password fast-path
is unchanged: the encrypted CA material remains decryptable.

Docs and CHANGELOG document the new `db-provision`-becomes-no-op contract
under reinit, the all-absent → fresh-CA non-interactive fallback, and
the absent-`password.txt`-with-any-preserved-step-ca-artifact refusal
policy. Docker-backed E2E recovery coverage is intentionally deferred
to the broader recovery-scenarios issue.

Closes #601
```

PR #605, opened 2026-05-24
Add non-interactive flags for init save-unseal-keys prompt (#603)

sehkone:
## Summary

Adds two new flags to `bootroot init` to non-interactively answer the "Save unseal keys to file for automatic unseal? [y/N]" prompt — the last remaining interactive prompt on the public `init` surface.

- `--save-unseal-keys`: skip the prompt and persist the freshly generated unseal keys to `<secrets_dir>/openbao/unseal-keys.txt` (mode `0600`). Equivalent to answering `y`.
- `--no-save-unseal-keys`: skip the prompt and do NOT persist the keys to that path. Requires `--summary-json <path>` (enforced at clap parse time) so the keys are captured in the 0600 summary file instead of being lost. Also suppresses the cleartext-echo fallback so the keys don't leak into CI job logs.

The two flags are mutually exclusive (`conflicts_with`). When neither is set, the existing interactive prompt (default `N`) is preserved, so operators see no behavior change. Per the precedent set by #588 §3b (`--no-eab`), `init` uses per-prompt explicit flags rather than a global `--yes`.

Internally, `maybe_save_unseal_keys` now takes a `SaveUnsealKeysDecision` enum (`Prompt` / `Save` / `DoNotSave`) computed at the call site from `reinit_mode` plus the two new flags, replacing the old `auto_save: bool` parameter. Reinit's internal auto-save path is unaffected: the `reinit` wrapper continues to construct `InitArgs` with both new flags `false` and `reinit_mode: true`.

### Reviewer Round 1 follow-up

`run_init` now runs the same `--summary-json` and `--root-token-output` preflight that `bootroot reinit` does, before any OpenBao work begins. Without this, a bad summary-json or root-token-output destination would fail post-init — after OpenBao was already initialised but before `print_init_summary` / `maybe_save_unseal_keys` ran — recreating the partial-init trap that `--no-save-unseal-keys` is designed to avoid via the summary-json recovery channel. The same fix also closes the equivalent gap for `--save-unseal-keys` (whose post-init flow was short-circuited the same way) and for bare `--summary-json` / `--root-token-output` invocations. The two preflight helpers (`validate_summary_json_output_path`, `validate_root_token_output_path`) are reused from `commands::reinit` to keep behaviour identical across both surfaces.

Closes #603

## Test plan

- [x] `cargo test` passes the new clap parse-time assertions:
  - [x] `--save-unseal-keys` parses and sets the matching field
  - [x] `--no-save-unseal-keys --summary-json <path>` parses
  - [x] `--no-save-unseal-keys` without `--summary-json` is rejected
  - [x] `--save-unseal-keys --no-save-unseal-keys` together is rejected
  - [x] Default `init` leaves both fields false
- [x] `cargo test` passes the orchestrator unit tests:
  - [x] `SaveUnsealKeysDecision::Save` writes `secrets_dir/openbao/unseal-keys.txt` at mode `0600` without prompting
  - [x] `SaveUnsealKeysDecision::DoNotSave` skips both the prompt and the on-disk write
  - [x] Existing `reinit_mode` auto-save path still writes (regression coverage via `maybe_save_unseal_keys_auto_save_writes_without_prompting`)
  - [x] `run_init_rejects_bad_summary_json_before_any_work` — bad `--summary-json` destination is rejected at preflight before any OpenBao work
  - [x] `run_init_rejects_bad_root_token_output_before_any_work` — bad `--root-token-output` destination is rejected at preflight before any OpenBao work
- [ ] Manual end-to-end: `bootroot init --enable auto-generate --no-eab --no-save-unseal-keys --summary-json /tmp/init.json` returns 0 without stdin input, `/tmp/init.json` contains a non-empty `unseal_keys` array at 0600, no `unseal-keys.txt` is present under `secrets_dir/openbao/`, and stdout does not contain any of the unseal keys verbatim (with `--enable show-secrets` unset)
- [ ] Manual end-to-end: `bootroot init … --save-unseal-keys` writes `<secrets_dir>/openbao/unseal-keys.txt` at mode `0600` without prompting
- [ ] Manual regression: `bootroot reinit --yes` continues to write unseal keys without prompting (reinit's `auto_save` path through `reinit_mode`)
- [ ] Interactive operators (neither flag set) still see the existing prompt with default-N
- [x] `cargo clippy` is clean
- [x] CHANGELOG and both CLI docs (`docs/en/cli.md`, `docs/ko/cli.md`) reflect the new flags and the `--summary-json` preflight on `init`

Manual end-to-end items above remain unchecked because local Docker bring-up was blocked by a stuck `127.0.0.1:8200` listener held by Docker Desktop (a known preflight flake recorded in repo memory). The underlying call paths they cover are exercised by the three `maybe_save_unseal_keys_*` unit tests (Save / DoNotSave / reinit auto-save), the two new `run_init_rejects_bad_*_before_any_work` preflight tests, and the five clap parse-time tests; all 14 Docker E2E matrix jobs pass on this branch. CLI parse-time rejection of bad combinations was also verified by invoking the built binary directly (`--no-save-unseal-keys` without `--summary-json` and `--save-unseal-keys --no-save-unseal-keys` both rejected at parse time with the expected clap error).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,23 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- `bootroot init` now accepts `--save-unseal-keys` and
+  `--no-save-unseal-keys` to non-interactively answer the
+  "Save unseal keys to file for automatic unseal? [y/N]" prompt,
+  which was the last remaining interactive prompt on the public `init`
+  surface. `--save-unseal-keys` writes the freshly generated keys to
+  `<secrets_dir>/openbao/unseal-keys.txt` (mode `0600`) without
+  prompting; `--no-save-unseal-keys` skips both the on-disk save and
+  the cleartext-echo fallback, and requires `--summary-json <path>`
+  (enforced at clap parse time) so the keys are captured in the 0600
+  summary JSON instead of being lost. The two flags are mutually
+  exclusive. When neither flag is set the interactive prompt (default
+  `N`) is preserved, so operators see no behavior change. Reinit's
+  internal auto-save path (`args.reinit_mode`) is unaffected.
+  (Closes #603)
+
 ### Security
 
 - Bumped `rustls-webpki` from 0.103.10 to 0.103.12 to address
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -253,7 +253,13 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   exclusive. When neither flag is set the interactive prompt (default
   `N`) is preserved, so operators see no behavior change. Reinit's
   internal auto-save path (`args.reinit_mode`) is unaffected.
-  (Closes #603)
+  `bootroot init` also now runs the same `--summary-json` and
+  `--root-token-output` preflight that `bootroot reinit` does, before
+  any OpenBao work starts, so a bad output destination fails fast
+  instead of being discovered post-init with the freshly issued root
+  token and unseal keys already minted (recreating the partial-init
+  trap `--no-save-unseal-keys` is designed to avoid through the
+  summary-json recovery channel). (Closes #603)
 - New `bootroot reinit` recovery command that atomically wipes
   OpenBao-owned state and re-runs init while preserving step-ca CA
   material, `password.txt`, PostgreSQL state, operator-authored compose
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,23 +6,6 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
-### Added
-
-- `bootroot init` now accepts `--save-unseal-keys` and
-  `--no-save-unseal-keys` to non-interactively answer the
-  "Save unseal keys to file for automatic unseal? [y/N]" prompt,
-  which was the last remaining interactive prompt on the public `init`
-  surface. `--save-unseal-keys` writes the freshly generated keys to
-  `<secrets_dir>/openbao/unseal-keys.txt` (mode `0600`) without
-  prompting; `--no-save-unseal-keys` skips both the on-disk save and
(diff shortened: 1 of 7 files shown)


sehkone:
## Suggested squash commit

**Title**

```text
Add non-interactive flags for init save-unseal-keys prompt
```

**Body**

```text
The "Save unseal keys to file for automatic unseal? [y/N]" prompt was
the last remaining interactive prompt on the public `init` surface,
blocking unattended bring-up.  Other init prompts already have either
automatic resolutions or explicit opt-out flags (`--no-eab` from #588).

Add `--save-unseal-keys` and `--no-save-unseal-keys` to follow the
same per-prompt explicit-flag pattern.  The two flags are mutually
exclusive at clap parse time, and `--no-save-unseal-keys` requires
`--summary-json <path>` so the keys are captured in the 0600 summary
file instead of being lost and bricking the next OpenBao restart.

Under `--no-save-unseal-keys` the cleartext-echo fallback is also
suppressed — the keys are already in the summary JSON, and echoing
them to stdout would leak them into CI job logs.

Collapse the call-site inputs into a `SaveUnsealKeysDecision` enum so
`maybe_save_unseal_keys` switches on a single state rather than
juggling two booleans plus `reinit_mode`.  Reinit's internal
auto-save path is unaffected: the `reinit` wrapper still constructs
`InitArgs` with both new flags `false` and `reinit_mode: true`.

Also validate the optional `--summary-json` and `--root-token-output`
destinations at the top of `run_init` (mirroring the existing
preflight in `run_reinit`) so a bad path fails fast rather than
short-circuiting the post-init flow.  Without this guard, a
`--summary-json` write that fails after `run_init_inner` succeeds
would skip `print_init_summary`, `--root-token-output`, AND the
unseal-key save / cleartext-echo branch — leaving the freshly issued
root token and unseal keys captured nowhere.  That recreates the
partial-init trap the new `--no-save-unseal-keys` flag is designed
to avoid through the summary-json recovery channel.  Reuses
`validate_summary_json_output_path` and
`validate_root_token_output_path` from `commands::reinit` so init
and reinit share identical validation behaviour.

Closes #603
```

PR #635, opened 2026-05-27
Mention clean teardown in infra --help (#632)

sehkone:
## Summary

`bootroot infra` only carries the bring-up subcommands (`up`, `install`); teardown is the top-level `bootroot clean`. The functionality is complete and documented, but the help surface is asymmetric — operators who learn `bootroot infra up` do not naturally discover `bootroot clean` by exploring `bootroot infra --help`.

This PR

- adds an `after_help` footer on `bootroot infra` pointing at `bootroot clean`;
- adds a doc comment on the `Infra` variant so the top-level `bootroot --help` summary describes it;
- adds a doc comment + `long_about` on the `Clean` variant so the page the footer points at actually describes the full teardown semantics (compose down with auto-discovered openbao-agent / openbao-exposed overrides; removal of `secrets/`, `state.json`, `.env`; prompted `certs/`; `--openbao-only` recovery path) rather than just listing flags.

No new subcommand or alias is introduced.

### Why no `infra down` alias

`docker compose down [-v]` semantics — which an operator typing `bootroot infra down` would import by analogy — only touch containers, networks, and volumes. `bootroot clean` additionally wipes `secrets/`, `state.json`, `.env`, and (optionally) `certs/`. Aliasing `infra down` → `clean` would silently turn "stop and bring back up later" into "lose every root token, CA material, and service cert". The issue (v3/v4 scope) explicitly excludes this path.

A non-destructive `infra down` that just stops the compose stack while preserving filesystem state is a new feature, out of scope here.

Closes #632

## Test plan

- [x] `cargo clippy --all-targets --no-deps -- -D warnings` passes
- [x] `cargo run --bin bootroot -- infra --help` shows the `Teardown:` footer pointing at `bootroot clean`
- [x] `cargo run --bin bootroot -- --help` shows the new `Infra` and `Clean` doc-comment summaries
- [x] `cargo run --bin bootroot -- clean --help` now shows the full teardown semantics (compose down, override discovery, paths removed, `--openbao-only`) instead of a flag-only page
- [x] `cargo run --bin bootroot -- clean -h` short form still shows a one-line summary
- [x] `cargo run --bin bootroot -- infra up --help` and `... infra install --help` are unaffected (no spurious footer on the leaf subcommands)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1009,6 +1009,22 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   callers. The timeout message itself is unchanged. Callers that
   prefer the old "ignore timeout" semantics can re-establish them
   with `|| true`. (Closes #629)
+- Added a teardown footer to `bootroot infra --help` pointing operators
+  at `bootroot clean` (top-level), plus a `long_about` on `bootroot
+  clean --help` describing the full teardown path (compose down with
+  auto-discovered openbao-agent/openbao-exposed overrides, removal of
+  `secrets/`, `state.json`, `.env`, and prompted `certs/`) so the next
+  help page after the footer is no longer a flag-only dead end.
+  Bring-up lives under `infra` (`up`, `install`) while teardown is the
+  top-level `clean`, so an operator who learned `bootroot infra up`
+  would not naturally discover the teardown command by exploring
+  `bootroot infra --help`. The footer closes the discoverability gap
+  without aliasing `infra down` to `clean` — `docker compose down`
+  semantics only touch containers, networks, and volumes, while
+  `bootroot clean` additionally wipes `secrets/`, `state.json`, `.env`,
+  and (optionally) `certs/`, so an alias would silently turn "stop and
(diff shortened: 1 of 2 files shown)


sehkone:
**[Reviewer Round 1]**

Finding: `src/cli/args.rs:17` points operators to `bootroot clean --help` for teardown details, but that command still only lists flags and the `--openbao-only` exception; it does not describe the normal full-clean path or mention the optional `certs/` prompt/removal. That makes the new discovery path partly dead-end: `bootroot infra --help` now correctly tells users that `bootroot clean` is the teardown command, but the next help page does not actually provide the details the footer promises. This also makes the PR test-plan item claiming `clean --help` matches the footer inaccurate. I would either add a concise doc comment/`after_help` on the top-level `Clean(CleanArgs)` command describing the full teardown semantics, or remove the “Run `bootroot clean --help` for details.” sentence and keep the `infra` footer self-contained.

The overall direction looks right: no `infra down` alias, the footer appears on `bootroot infra --help`, and it does not propagate to the leaf subcommands.

PR #636, opened 2026-05-27
Validate docker container identity in service add (#631)

### LINKED ISSUE #631: service add: validate that --deploy-type=docker matches the actual bootroot-agent
**Cross-link**: filed from [aicers/gauntlet#111](https://github.com/aicers/gauntlet/issues/111). Sibling to #630 (doc-side defense) and #629 (loud-timeout backstop).

## Observation

Today `service add` accepts `--deploy-type docker --container-name <X>` without verifying that `<X>` actually points at a bootroot-agent. If `<X>` is the consumer container (or any other unrelated container), registration succeeds and the misconfiguration only surfaces at the next rotation, where it manifests as the wrong container being restarted while the real agent never receives reload. See [aicers/gauntlet#106](https://github.com/aicers/gauntlet/issues/106) for the concrete cascade.

`validate_service_add` at `src/commands/service/resolve.rs:223` is the natural extension point — it already runs at `service add` time and has access to the resolved deploy-type and container name.

## Fix shape

When `deploy_type == Docker`, attempt `docker inspect <container-name>` and check:

1. The container exists.
2. Its `Config.Cmd` or `Config.Entrypoint` (or image) looks like a bootroot-agent — match on the binary name `bootroot-agent` in the cmdline.

On mismatch, **warn** rather than reject, because:

- The agent container may legitimately not be running yet at the moment of `service add` (e.g. bootstrap order).
- Container introspection can fail for reasons unrelated to misconfiguration (docker not on PATH, permission, remote daemon).

A `--no-validate-agent` escape hatch lets operators opt out for known-good environments where the check is noise.

For `deploy_type == Daemon`, an analogous warning fires if `--container-name` is set at all — that's already suspicious since the flag is documented as docker-only (after #630 lands).

## Why warn-not-reject

A reject-on-mismatch is tempting but breaks legitimate ordering: an operator who runs `service add` before bringing up the agent container should not be blocked. Warning at registration plus loud failure at rotation (via #629) is the right division of labor.

## Severity

Medium — this is the only fix in the trio that *prevents* the misconfiguration at command time, rather than mitigating its effects. Without it, a fresh operator (or a copy-pasted runbook) can still register a broken service, and only #629's loud timeout saves them on the first rotation attempt.

## Order of operations

Lands best **after** #629. With that in place, the warning text here can say "...registering anyway; the first `rotate --wait` will exit 124 if no agent picks up the request," which gives operators a concrete next-step signal.

sehkone:
## Summary

`bootroot service add --deploy-type=docker --container-name=X` previously accepted any container, so a misconfiguration (e.g. pointing at the consumer container instead of the bootroot-agent) only surfaced at the next rotation as the wrong container being restarted while the real agent never received a reload.

This PR adds two complementary checks at command time:

1. **Raw-args reject**: `--deploy-type=daemon` combined with `--container-name=X` is rejected before `resolve_service_add_args` drops the field to `None`. The combination has no legitimate meaning.
2. **Docker-mode identity check**: when `--deploy-type=docker`, shell out to `docker inspect --type container` and confirm the container looks like a bootroot-agent. Label-first (`bootroot.role=agent`, now baked into the shipped docker-compose snippet) with a `bootroot-agent` substring fallback on the image / entrypoint / cmd. Missing label / unidentified container / `docker inspect` failure surface as a **non-fatal warning** (registration still succeeds) telling the operator the first `rotate --wait` will exit 124 if no agent picks up the request, and pointing at the new `--no-validate-agent` flag for legitimate "agent container not up yet" / cross-host / pre-existing-deployment cases.

The `--type container` scope on `docker inspect` matters: without it, `docker inspect <name>` can match images, networks, or volumes too, so an image named `bootroot-agent` whose config contains the `bootroot-agent` substring could satisfy the fallback even when no container of that name exists — silently suppressing the warning that the operator actually needs to see. Rotation later calls `docker restart <container-name>`, which only works on containers, so scoping the inspect to containers keeps the check aligned with what the operation it protects will require.

`--no-validate-agent` scopes **only** to the docker identity check and does not bypass the daemon+container-name reject (matches the severity matrix in the issue).

The `run-ca-key-rotation-recovery.sh` fixture (and the lifecycle / preflight scripts that intentionally register services without a real container running) pass `--no-validate-agent` so their output stays clean — this was called out explicitly in the issue's "Existing test surface to coordinate" section.

Closes #631

## Test plan

- [x] `cargo clippy --all-targets --all-features` is clean
- [x] `cargo test --test bootroot_service` passes (new tests: `test_app_add_rejects_daemon_with_container_name`, `test_app_add_warns_when_docker_container_not_an_agent`, `test_app_add_no_validate_agent_silences_warning`)
- [x] `cargo test -p bootroot --lib commands::service::resolve` passes (new raw-args unit tests for daemon+container_name conflict, blank container handling, docker positive, and unspecified deploy-type deferral)
- [x] `cargo test -p bootroot --lib commands::service::agent_identity` passes (label match, case-insensitive label, image/entrypoint/cmd fallback, no-match, inspect-failure surfacing, container-scoped inspect args, image-with-matching-name-but-no-container surfaces inspect failure)
- [x] `service add --deploy-type=daemon --container-name=X` exits non-zero with an error pointing at `--deploy-type=docker`
- [x] `service add --deploy-type=docker --container-name=<nonexistent>` succeeds but prints a warning that mentions the container name and `--no-validate-agent`
- [x] `service add --deploy-type=docker --container-name=<nonexistent> --no-validate-agent` succeeds silently
- [x] `service add --deploy-type=docker --container-name=bootroot-agent` against a running compose stack succeeds silently (label `bootroot.role=agent` from the updated `docker-compose.yml` matches)
- [x] Korean i18n strings render correctly with `BOOTROOT_LANG=ko`
- [x] `scripts/preflight/run-all.sh` passes (lifecycle scripts now pass `--no-validate-agent` where they register services without a real container)


sehkone:
**[Author Round 1]**

- `src/commands/service/agent_identity.rs:101` (`docker inspect` is not container-scoped) — **Fixed**. Switched the inspect call to `docker inspect --type container <name>` via a new `build_inspect_args` helper, with a comment explaining the scope choice (rotation later issues `docker restart <name>`, which only operates on containers, so a non-container match would suppress a warning the operator actually needs). Added two unit tests in `commands::service::agent_identity::tests`:
  - `inspect_args_are_container_scoped` pins the argv shape so a future refactor cannot silently drop `--type container`.
  - `image_with_matching_name_but_no_container_surfaces_inspect_failure` covers the case the reviewer called out — when only an image of the requested name exists, the real CLI exits with `No such container`, which now surfaces as `AgentIdentity::InspectFailed` (warning shown) rather than falling through the substring path and reporting `Match`.
  Also refreshed the CHANGELOG entry and PR body to mention the `--type container` scoping rationale.

PR #646, opened 2026-06-12
Add --cert-group regression test for [trust] preservation (#645)

sehkone:
## Summary

`service update` re-renders a local-file service's managed `agent.toml` block through **two distinct triggers** that both reach `rerender_local_managed_profile`:

- `--cert-group` — sets `redeploy_hint` unconditionally (even on `clear` or an unchanged gid)
- `--reload-style` — sets `hooks_changed`

The issue #643 fix (merged in #644) preserves the `[trust]` section across this re-render, but the existing regression tests (`test_service_update_reload_style_*`) only exercise the `hooks_changed` path. A future change that gated the carry-over behind `hooks_changed`, or reordered the `--cert-group` handling, could silently reintroduce the `UnknownIssuer` bug on the cert-group path while every existing test stayed green.

This PR closes that gap with two test-only additions in `tests/bootroot_service.rs`, mirroring the `--reload-style` cases but driving the re-render purely through `--cert-group clear` (no hook change):

- `test_service_update_cert_group_preserves_inline_trust` — `[trust]` (`ca_bundle_path` + `trusted_ca_sha256`) inside the `# BEGIN … # END` managed-profile markers survives, with exactly one `[trust]` section remaining.
- `test_service_update_cert_group_keeps_outside_trust_unique` — an out-of-block `[trust]` is updated in place and not duplicated.

No production code changed. (`Cargo.lock` also carries a postgres dependency bump patching RUSTSEC advisories; lockfile-only, no source change.)

Closes #645

## Test plan

- [x] New tests genuinely exercise the `--cert-group`/`redeploy_hint` trigger (no hook change), distinct from the `hooks_changed` path
- [x] Inline `[trust]` inside the managed-profile markers is preserved on `service update --cert-group clear`
- [x] Out-of-block `[trust]` remains unique (no duplication) after `--cert-group` re-render
- [x] `cargo test --test bootroot_service cert_group_` passes on current `main`
- [x] `rustfmt` clean
- [x] CI `check` and `test-core` jobs pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,7 +47,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -58,7 +58,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -536,7 +536,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


sehkone:
**[Reviewer Round 1]**

No blocking findings.

The added tests in `tests/bootroot_service.rs:3055` and `tests/bootroot_service.rs:3121` match the issue scope: both seed the existing local-file service fixture with no hook changes, pass only `--cert-group clear`, and therefore rely on the `redeploy_hint` side of `needs_local_rerender` in `src/commands/service.rs:742-748` rather than the `hooks_changed` branch. The inline case would fail if trust carry-over were accidentally tied to hook changes, because the managed block is re-rendered and the in-block `[trust]` table would otherwise be dropped. The outside-block case covers the no-duplication requirement.

The `Cargo.lock` postgres advisory bump is outside the narrow test-only scope, but the PR body explains it as a CI/audit unblocker and it does not touch production source, so I do not see a reason to block on it.

PR #661, opened 2026-07-04
Add service remove command for deregistration (#656)

sehkone:
## Summary

Adds `bootroot service remove --service-name <svc>` to deregister a service, making "remove and re-add" — the flow the code comments already assumed — the supported way to change a service's `--delivery-mode`. Previously `service add` bailed with `error_service_duplicate` on a mode change, `service update` never touched `delivery_mode`, and no remove command existed, so the only recourse was hand-editing `state.json` and orphaning the service's OpenBao AppRole, policy, and KV material.

### What it does

- **New `Remove(ServiceRemoveArgs)` variant** in `enum ServiceCommand`, wired through dispatch in `main.rs` to `run_service_remove` in the new `src/commands/service/remove.rs`. Flags: `--service-name`, `--yes`/`--force`, `--dry-run`, and `--delete-artifacts`.
- **OpenBao teardown** uses the `role_name` / `policy_name` stored in the `ServiceEntry` (not recomputed) and deletes the exact per-service KV paths (`eab`, `http_responder_hmac`, the trust suffix, and `secret_id` only for `remote-bootstrap`). Each delete is guarded by the `*_exists` helpers so an already-absent resource is treated as "already done".
- **Remote-first ordering**: AppRole, policy, and KV deletions all complete before the `state.json` entry is dropped, so a partial failure keeps the entry (and its stored names) for a safe, idempotent re-run. On any failure the run bails and reports which resources were removed vs. remain.
- **DNS alias refresh via a new `reconcile_dns_aliases` helper** that, unlike `register_dns_alias`, does *not* short-circuit on an empty alias set — removing the last alias-bearing service still reconnects the responder with just the base `bootroot-http01` alias. The reconcile runs against the post-removal state and, like the remote teardown, completes *before* the `state.json` removal is persisted: if reconnecting the responder fails hard, the on-disk entry is left intact so a re-run can retry rather than failing with `error_service_not_found`.
- **On-disk artifacts preserved by default**; `--delete-artifacts` removes bootroot-owned dirs/files and strips only the managed profile block from `agent.toml` (via a new `remove_managed_profile_block` in `trust_bootstrap.rs`) rather than deleting an operator-owned file.
- Widened `SERVICE_TRUST_KV_SUFFIX` to `pub(crate)` for reuse; bails with `error_state_missing` / `error_service_not_found` matching `run_service_update`; new i18n messages (en/ko); updated the stale "remove and re-add" comment to reference the new command; updated operator docs (`cli.md`, `troubleshooting.md`, en + ko).

Closes #656

## Test plan

- [x] `bootroot service remove --service-name <svc>` deregisters an existing service: state entry gone, OpenBao AppRole/policy/KV torn down by stored names, responder alias set refreshed
- [x] Removing the last (or only) alias-bearing service still drops its alias (empty set does not short-circuit the refresh)
- [x] Cert/key and on-disk material preserved unless `--delete-artifacts`; with the flag, `agent.toml`'s managed block is removed without clobbering the operator-owned file (idempotent when already gone)
- [x] After a remove, `service add --delivery-mode remote-bootstrap --service-name <svc>` succeeds cleanly (no `error_service_duplicate`)
- [x] Removing a non-existent service fails with `error_service_not_found`; no `state.json` fails with `error_state_missing`
- [x] `--dry-run` reports the teardown plan without mutating `state.json` or OpenBao
- [x] Teardown order is remote-first; entry dropped last so a partial failure leaves it for re-run
- [x] Teardown is re-runnable/idempotent: deletes tolerate already-absent resources via `*_exists` guards
- [x] Unit tests cover not-found, state-missing, state-entry removal, KV path derivation, artifact/dir/file deletion, and managed-block stripping
- [x] `cargo clippy` warning-free and `cargo fmt --check` clean
- [x] Full CI (`check`, `test-core`, `test-docker-e2e-matrix`, `run-extended`) green


sehkone:
**[Reviewer Round 2]**

Round 1 item: resolved. I verified the updated diff now drops the service entry only from the in-memory state, runs `reconcile_dns_aliases` against that post-removal state, and persists `state.json` last via `finalize_removal`; the new tests cover both the success ordering and the reconcile-failure retention path.

I found one new issue that should be fixed before merge:

- `src/trust_bootstrap.rs:218-224`: `remove_managed_profile_block` finds the begin/end markers with substring searches (`contents.find(&begin_marker)` / `contents[begin..].find(&end_marker)`). Because service names are DNS labels and can be prefixes of other service names, removing `api` can match the managed block for `api-v2` or `apifoobar` when the exact `api` block is absent or appears later. The computed `end` then stops at the prefix inside the longer end marker, so `service remove --delete-artifacts` can strip the wrong service's profile and leave marker tail text behind. That violates the requirement to remove only bootroot's managed block for the target service from an operator-owned `agent.toml`. Please match complete marker lines (for example, exact line equality, tolerating `\r` if needed) and add a regression test with prefix-related service names.

PR #673, opened 2026-07-05
Add scheduled rotation story for AppRole secret_ids (#669)

### LINKED ISSUE #669: Add scheduled rotation story for AppRole secret_ids
## Summary

Bootroot provides rotation commands for AppRole `secret_id`s but no rotation **cadence**: nothing runs them on a schedule, nothing recommends a schedule, and the operations docs never mention that `secret_id` rotation must be scheduled at all. Since every `secret_id` bootroot mints carries a short TTL (default `24h`), every deployment must independently discover this requirement and hand-roll its own scheduler — and a deployment that doesn't discover it in time hits the expiry-outage chain described in #667. This issue is about giving `secret_id` rotation a first-party scheduling story: a documented recommended cadence plus a batch-friendly command surface that a single timer can drive.

## Background (verified against the code)

- `SECRET_ID_TTL = "24h"` is the role-level default applied to **every** AppRole created by `bootroot init` (`src/commands/init/constants.rs`), and service AppRoles inherit the same default. The constant's own doc comment already assumes scheduled rotation exists: "Operators who rotate on a longer cadence should pass `--secret-id-ttl` with a value that is at least 2× their rotation interval so that a missed or delayed rotation run does not leave services unable to re-authenticate." Nothing in the product or docs follows through on that assumption.
- Agent tokens are non-periodic (`token_ttl == token_max_ttl == 1h`, no `token_period`, `src/openbao.rs`), so every OpenBao Agent sidecar re-presents its `secret_id` at least hourly. Once the `secret_id` lapses, login fails with `403 invalid role or secret ID`.
- Rotation is strictly per-target today: `rotate approle-secret-id --service-name <name>` handles one service per invocation (`src/commands/rotate/approle.rs`), and #667 adds the two infra roles (stepca, responder) — also per-target. A deployment with N services therefore needs N+2 scheduled invocations, each needing the right auth flags.
- `docs/en/operations.md` covers process supervision, health checks, and per-component notes, but has no section on rotation scheduling; the only rotation-adjacent mention is running `bootroot-remote apply-secret-id` *after* a rotation has happened.
- The existing daemon loop (`src/daemon.rs`) is **agent-side** (certificate renewal on the service host); there is no control-plane scheduler this could naturally hang off of, and bootroot's stated operations policy is that operators own supervision/scheduling (Compose restart policies, systemd units/timers).

## Why it matters

- The failure mode is silent and delayed: everything works for up to 24h after init (or after the last manual rotation), then sidecars start failing OpenBao login, cert renewal stalls, and eventually a service leaf expires with a hard TLS failure whose root cause looks unrelated (see #667 for the full chain).
- The requirement is currently undocumented, so the first notice many operators get is the outage itself.
- Even an operator who knows about the requirement gets no guidance on interval choice, no batch command, and no worked systemd-timer/cron example — every deployment reinvents the same scheduler with the same pitfalls (missed runs, wrong auth, forgetting the infra roles).

## Proposal

Stay consistent with bootroot's operations policy — bootroot supplies the commands and the recommendation; the operator supplies the scheduler (systemd timer or cron). Two deliverables:

### 1. Batch rotation command surface

Add an `--all` mode (or a sibling `rotate secret-ids` subcommand — implementer's choice, matching the CLI model) that rotates in one invocation:

- every registered service in `state.services` (both local-file and remote-bootstrap delivery modes, reusing the existing per-service mechanics), and
- the two infra roles once #667 lands (authenticating with the dedicated infra-rotate credential per #667's decided design; the service targets keep using `bootroot-runtime-rotate-role`).

Behavioral requirements:

- Non-interactive under `--yes` (schedulable).
- **Continue on per-target failure**: one failing service must not leave the remaining targets unrotated. Collect failures, print a per-target summary, exit non-zero if any target failed.
- Idempotent and safe to re-run: the existing flow already tolerates this (old `secret_id`s stay valid until TTL; sidecar reload is a restart).

### 2. Documented recommended cadence

Add a "Secret ID rotation scheduling" section to `docs/en/operations.md` / `docs/ko/operations.md`:

- State the invariant explicitly: **rotation interval ≤ ½ × `secret_id` TTL** (one missed run still leaves a valid credential), matching the existing `RECOMMENDED_SECRET_ID_TTL` doc comment. With the default 24h TTL, recommend every 8–12h.
- Provide a worked systemd timer unit and an equivalent cron line invoking the batch command with least-privilege auth (no root token), including where the rotate credential's `role_id`/`secret_id` live and how to protect them.
- Call out the infra roles explicitly so operators don't schedule services only.
- Cross-reference the `--secret-id-ttl` init flag for operators who want a longer TTL to match a slower cadence.

### Explicitly out of scope

A resident control-plane scheduler daemon (bootroot running its own timer loop) is **not** proposed: it would contradict the documented operations policy (operators own supervision), add an always-on control-plane process with standing rotate credentials, and duplicate what systemd timers/cron already do well. If demand for a built-in scheduler emerges later, that should be its own design discussion.

## Open question — bootstrap of the rotate credential's own secret_id

The scheduled job authenticates with rotate AppRoles whose `secret_id`s are themselves subject to the same TTL. The batch command should rotate its own credential last (write the new `secret_id` to the credential file used by the scheduler) or the docs must cover this explicitly — otherwise the rotation job eventually locks itself out. Implementation should pick one and document it.

## Acceptance criteria

- A single scheduled invocation rotates all service and infra `secret_id`s with least-privilege auth and exits non-zero on partial failure with a per-target summary.
- Operations docs (en + ko) contain the cadence invariant, a systemd timer example, and a cron example.
- The rotate job's own credential lifecycle is handled (self-rotation or documented procedure).

## Related

- #667 — infra AppRole `secret_id` rotation (prerequisite for the infra targets; its "Open question — rotation cadence" section is resolved by this issue)
- #108 — original per-service `rotate approle-secret-id`
- #282 — least-privilege runtime automation (the auth model the scheduled job must follow)

sehkone:
## Summary

Adds the batch service `secret_id` rotation surface and the worked scheduler documentation described in #669:

- **`--all-services` flag on `rotate approle-secret-id`**: rotates every service registered in `state.json` (both `local-file` and `remote-bootstrap` delivery modes) in one invocation under the existing `bootroot-runtime-rotate-role` credential, reusing the per-service rotation mechanics. It is mutually exclusive with `--service-name` and `--infra`, non-interactive under `--yes`, continues past per-target failures, prints a per-target summary, exits non-zero if any target failed, and treats an empty service registry as a no-op success. Infra roles are deliberately excluded per the credential-separation design from #667.
- **Scheduling docs (en + ko)**: a new "Scheduling AppRole secret_id rotation" section in `docs/en/operations.md` / `docs/ko/operations.md` with worked systemd timer + oneshot unit and equivalent cron examples implementing the one-job/few-invocations model — one batch service invocation plus the two per-target infra invocations (`--infra stepca`, `--infra responder`), each with its own least-privilege credential loaded from root-owned files. Since #667 has landed, the infra invocations are included directly rather than referenced. The section recommends an 8–12h interval for the default 24h TTL, cross-links the existing cadence-invariant section (including the smallest-TTL rule under per-service overrides), and calls out the infra roles so operators do not schedule services only.
- **Rotate-credential lifecycle**: a dedicated docs subsection states that neither rotate credential can mint its own `secret_id`, describes the interim break-glass re-mint procedure (root-token re-mint on a schedule shorter than the TTL, or a longer `--secret-id-ttl` up to the 168h cap), and links the follow-up self-rotation design issue #672.
- CLI reference, examples, changelog, and the remote-lifecycle E2E script are updated to cover `--all-services`.

Closes #669

## Test plan

- [x] `cargo clippy` passes with no warnings and `cargo fmt --check` is clean
- [x] `--all-services` with an empty service registry exits zero as a no-op success
- [x] `--all-services` rotates both local-file targets (atomic file replace + reload) and remote-bootstrap targets (KV write) in one invocation
- [x] A per-target failure does not stop the remaining targets; the run prints a per-target summary and exits non-zero
- [x] An insufficient-permission (403) failure names the expected `bootroot-runtime-rotate-role` credential
- [x] `--all-services` is rejected when combined with `--service-name` or `--infra`
- [x] `docs/en/operations.md` and `docs/ko/operations.md` contain the worked systemd timer and cron examples, linked to the cadence-invariant section, with no credential mixing within a single invocation
- [x] The rotate-credential self-lockout constraint and interim re-mint procedure are documented and link #672
- [x] E2E remote-lifecycle script exercises `--all-services`
- [x] CI jobs (`check`, `test-core`, `test-docker-e2e-matrix`, `run-extended`) pass


sehkone:
**[Reviewer Round 1]**

I found one issue that should be fixed before merge:

- `docs/en/operations.md:276` and `docs/ko/operations.md:269` show the equivalent cron block as multi-line commands using trailing `\` continuations. That works in a shell script, but not in a crontab: cron parses each physical line as its own entry and does not join the following indented `--approle-*` / `approle-secret-id` lines into the previous command. As written, copying this example leaves the actual scheduled command truncated and the continuation lines invalid as cron entries, so the worked scheduler example does not satisfy the issue's cron acceptance criterion. Please make each cron entry a single physical line, or point cron at a small wrapper script that contains the multi-line shell commands.

The implementation itself looks aligned with #669: `--all-services` is mutually exclusive with the other selectors, keeps infra out of the service batch, treats an empty registry as success, continues after per-target failures, and reuses the existing local-file and remote-bootstrap mechanics. The added unit tests and remote-lifecycle E2E coverage exercise the important service-batch paths.

sehkone:

PR #696, opened 2026-07-08
Retire local OpenBao sidecar; run agent as host daemon (#691)

sehkone:
## Summary

This PR makes `local-file` match `remote`: one secret-delivery mechanism (the `bootroot-agent` fast-poll self-auth loop) and one run model (host daemon), retiring the per-service OpenBao Agent sidecar and the local Docker-sidecar run model.

- **Local `agent.toml` now gets the `[openbao]` fast-poll section** during `service add`, with the same keys the remote binary writes (`url`, `kv_mount`, `role_id_path`, `secret_id_path`, `ca_bundle_path`) plus an absolute, service-keyed `state_path`. All values are host-relative since the agent is a host daemon. `service add` enforces the one-config-per-distinct-service topology: a `local-file` add whose `--agent-config` path is already registered to a different service is rejected (the `[openbao]` section holds a single AppRole identity, so a shared config would overwrite the first service's credentials). Local agent-config paths are normalized to an absolute, lexically clean form before storing/comparing, and the guard canonicalizes existing files, so re-spelled (`./agent.toml`) or symlinked paths cannot bypass the rejection — regression tests cover the literal, relative-spelling, and symlink cases. The guard also inspects the target file itself: an `agent.toml` that still contains another service's bootroot-managed profile block — left by `service remove` without `--strip-config`/`--delete-artifacts`, whose entry is no longer in `state.json` — is rejected too, since the agent fast-polls every profile in a config under the single `[openbao]` identity; a regression test drives the actual remove-without-strip then reuse-config flow. The unit-smoke flows (CI workflow, `test-core.sh`, `cli-scenarios.sh`) seed one `agent.toml` per smoke service accordingly.
- **EAB is wired via `--eab-file`**: `eab.json` is provisioned adjacent to `agent.toml`, and the daemon snippet/docs show the actual `bootroot-agent` invocation passing `--eab-file`, so local EAB rotation applies instead of being a silent no-op. `bootroot verify`'s oneshot run passes the same `--eab-file` for local-file services, mirroring the daemon invocation so verification exercises the same issuance path. Regression tests guard both wirings (including an integration test that records the oneshot argv).
- **OpenBao Agent sidecar generation and lifecycle removed**: the local HCL/`.ctmpl`/token-sink artifacts, the pre-seeded Docker-perspective CA bundle copy, `service openbao-sidecar start` (and its deprecated `agent` alias), `refresh_service_sidecar`'s local branch, `reload_openbao_agent`, the per-service restart loops in `rotate ca`/`ca-key`/`responder-hmac`, the `rotate secret-id` sidecar reload and its summary line, the `reinit` per-service purge branch, the `LocalApplyResult` sidecar-path fields, and the associated i18n strings (en + ko).
- **`--deploy-type` removed entirely**: the `DeployType` enum, the `deploy_type`/`container_name` fields on `ServiceEntry`, the `container_name` prompt/validation, and the `docker run` snippet are gone; `expected_dns_name` has a single unbranched form. The consumer-reload surface (`--reload-style docker-restart --reload-target <container>`) is kept intact, so containerized consumer apps stay supported via a bind-mounted host cert directory plus a `docker restart` post-renew hook — covered by tests, including one asserting the docker-restart hook persists to `agent.toml`.
- **Local rotate paths propagate via fast-poll** (trust, `secret_id`, responder HMAC, EAB); local force-reissue stays cert-file deletion + `pkill -HUP` with no Docker involvement. The out-of-scope infra OpenBao Agents (step-ca / responder) and their rotate-path restarts are untouched, as is the shared `bootroot::openbao::build_agent_config`. `rotate approle-secret-id` now rewrites the service `secret_id` through the ownership-preserving atomic writer (`fs_util::atomic_write`), so a root-run scheduled rotation cannot replace an operator-chowned, daemon-readable credential with a root-owned file the hardened non-root daemon can no longer read — a regression test fails if a rewrite strips the ownership again.
- **Docs (en + ko), CI workflow, E2E scenarios, and the impl/preflight scripts** are reworked for the daemon-only model, including the hardened non-root systemd unit guidance (a fixed `bootroot-agent` system account with one-time `useradd`/`chown` preparation — `DynamicUser=yes` cannot own the pre-provisioned `0600` credential files across restarts), the containerized-consumer pattern, and the `SupplementaryGroups=docker` trade-off for docker-restart hooks. Stale `.ctmpl`-era comments are rewritten for the fast-poll model. The deployment-policy pages state the supported topology explicitly: host daemon only, one `bootroot-agent` process plus one agent config per distinct service (the `[openbao]` section holds a single AppRole identity), with multiple `[[profiles]]` reserved for instances of the same service — matching `remote-bootstrap.md`. The local E2E harnesses (`run-local-lifecycle.sh`, `run-ca-key-rotation-recovery.sh`) run that same per-service topology: each distinct local service gets its own agent config, `eab.json`, state file, and daemon process, and the responder-HMAC phase asserts fast-poll propagation into each daemon's own config independently.
- **E2E postgres readiness gates probe over TCP** (`pg_isready -h 127.0.0.1`): the Unix-socket probe passed against the postgres image's temporary initdb bootstrap server, letting `bootroot init` connect during the restart window and flake with "the database system is starting up".

User-facing cleanup text matches the retired layout: `service remove --delete-artifacts` / `--strip-config` help and docs describe only the per-service secret directory and the remote-bootstrap artifact directory (no per-service `OpenBao` config directories), and the `reinit` docs list only the infra `secrets/openbao/{stepca,responder}` trees as purged. The top-level `README.md` architecture summary and `DESIGN.md` describe the daemon-only fast-poll onboarding model (OpenBao Agent serves only the infra components), and `agent.toml.compose` is labeled as the one-shot compose demo config rather than a host-daemon config.

A tree-wide straggler sweep (`rg` over `deploy_type|container_name|openbao-sidecar|bootroot-openbao-agent-|OBA_DEPLOYMENT|render_docker_agent_config|print_docker_snippet|reload_openbao_agent|sidecar`) returns only intended survivors: infra OpenBao Agent references, infra container names, historical `CHANGELOG.md` entries, and tests that assert the sidecar path is retired.

Closes #691

## Test plan

- [x] `local-file` `service add` produces no OpenBao Agent HCL/`.ctmpl`/token-sink artifacts, starts no `bootroot-openbao-agent-*` container, and prints only the host-daemon run model (no `docker run` snippet).
- [x] Local `agent.toml` contains a valid `[openbao]` section with an absolute, service-keyed `state_path`, and passes `config` validation.
- [x] `service add` no longer accepts `--deploy-type` or `--container-name`; `expected_dns_name` is single-form.
- [x] A `local-file` `service add` targeting an `agent.toml` already used by a different service is rejected before any OpenBao call — including via an equivalent relative spelling or a symlink, and including a config whose registered service was removed without `--strip-config` (the stale managed profile block is detected in the file itself); the regression tests fail if the guard, the path normalization, or the file inspection regresses.
- [x] `rotate responder-hmac`, `rotate eab-clear`, `rotate secret-id`, and `rotate ca`/`ca-key` trust propagation reach local services via fast-poll with no sidecar restart/reload; infra agent restarts (step-ca/responder) still happen.
- [x] The documented local run model passes `--eab-file` and `eab.json` is provisioned; the regression test fails if the wiring regresses.
- [x] `bootroot verify` runs the oneshot with `--eab-file <service eab.json>` for local-file services; unit and integration regression tests fail if the flag is dropped.
- [x] `rotate force-reissue` on a local service works via cert-file deletion + `pkill -HUP`, no Docker involvement.
- [x] Containerized consumers still work: `--reload-style docker-restart --reload-target <container>` is accepted, persisted to `agent.toml`, and covered by the bind-mount + docker-restart e2e path.
- [x] `rotate approle-secret-id` preserves the existing owner/group of the service `secret_id` file; the regression test fails if the rewrite strips daemon-readable ownership.
- [x] Docs (en + ko) describe the local daemon as a hardened, non-root systemd unit that can actually read/update its provisioned `0600` inputs (fixed account + `chown` preparation), plus the containerized-consumer bind-mount pattern.
- [x] `cargo fmt --check`, `cargo clippy`, and `cargo test` pass.
- [x] Full CI green: `check`, `test-core`, `test-docker-e2e-matrix`, and `run-extended`.


sehkone:
**[Reviewer Round 10]**

Round 9 is resolved: `README.md`, `DESIGN.md`, and `agent.toml.compose` now describe the daemon-only fast-poll model and limit OpenBao Agent to the infra step-ca/responder path.

I found one remaining invariant gap:

- `src/commands/service.rs:181-185` enforces the one-config-per-distinct-local-service rule only against services still present in `state.json`. That misses a normal non-legacy path: add `edge-proxy` with `agent.toml`, then run `bootroot service remove --service-name edge-proxy --yes` without `--strip-config` / `--delete-artifacts` (the documented default preserves `agent.toml`; see `src/commands/service/remove.rs:143-149` and `docs/en/cli.md:979-989`), then add a distinct local service such as `billing-api` with the same `--agent-config`. The state entry is gone, so the conflict guard passes. `apply_local_service_configs` then only strips foreign-mode markers for the same service (`src/commands/service/local_config.rs:49-59`), appends/upserts the new `billing-api` profile, and overwrites the single top-level `[openbao]` identity with `billing-api`'s AppRole paths (`src/commands/service/local_config.rs:94-123`). The old bootroot-managed `edge-proxy` profile remains in the file, and the agent will group and fast-poll every profile service in that config (`src/fast_poll.rs:1388-1401`) using the one new `[openbao]` credential. That violates the PR's central topology rule and leaves a stale distinct service profile active under the wrong AppRole identity. Please reject a local-file add when the target `agent.toml` already contains a bootroot-managed profile for a different service (or otherwise require/perform the stale managed-profile cleanup before writing the new `[openbao]` section), and add a regression test for the remove-without-strip then reuse-config case.

PR #718, opened 2026-07-17
Run step-ca helpers as secrets owner and sweep root-owned files (#716)

sehkone:
## Summary

The one-shot `step` helper containers that write into `secrets/` disagreed on who they run as. `init`'s helpers already pass `--user <uid>:<gid>` resolved from the secrets-directory owner, but the rotate flows still ran `--user root`, leaving root-owned key material behind. This diverges from the OpenBao Agent sidecars (which render into the same directory as the owner uid) and is about to become fatal once the step-ca server moves to a non-root uid. It already breaks `rotate ca-key`'s host-side Phase 1 backup on any host that rotated before this change.

This PR:

- Switches both `rotate ca` sites (`generate_new_root`, `generate_new_intermediate`) and the `rotate stepca-password` `step crypto change-pass` site from `--user root` to `--user <uid>:<gid>` derived from the secrets-directory owner, matching the existing `init` pattern. The `docker run` argv builders are extracted as pure functions so tests can assert them.
- Adds a shared ownership sweep — a single one-shot **root** container that `chown -R --no-dereference`s everything under `secrets/` back to the directory's own uid/gid — implemented once in `src/commands/infra.rs` and wired into `infra install`, `infra up`, and both `rotate` flows so a legacy host with root-owned material converges instead of failing or aborting.
- Picks the sweep's image per flow so it adds **no** new image dependency anywhere. The `infra` flows reuse the compose stack's own step-ca server image (resolved by inspecting the container Compose created) and run the sweep after `up`, so that image already exists on both the air-gapped and fresh-source install paths. The `rotate` flows instead reuse the `smallstep/step-ca` image their own `step` helpers already run, which keeps the sweep off `docker compose ps` and lets it run before Phase 1 reads or rewrites any key material.
- Gates the sweep in the `infra` flows on step-ca being both requested via `--services` and declared by the compose file (`compose_has_stepca`), mounts only the secrets directory, and uses `--no-dereference` so it never follows a symlink out of the mount. Its root user is the one intentional root container, kept visibly distinct from the `step` helpers and commented as such.
- Updates the "Manual alternative" manual-init example in `docs/en/installation.md` and `docs/ko/installation.md` to run as the secrets-directory owner (`--user $(id -u):$(id -g)`) with prose explaining the always-true reason (everything bootroot runs against `secrets/` runs as that directory's owner), leaving the image reference untouched.
- Fixes the stale `fix_secrets_permissions` doc comment to describe mode normalisation instead of a root `step ca init`.
- Adds a legacy-host simulation (`seed_root_owned_ca_keys` / `assert_no_root_owned_secrets`) to the local lifecycle E2E script.

Closes #716

## Test plan

- [x] Unit test: the `docker run` argv built by `rotate/ca.rs` and `rotate/stepca_password.rs` carries `--user <uid>:<gid>`, not `--user root`.
- [x] Unit test: the sweep's container invocation mounts only the secrets directory and uses `--no-dereference` (no symlink following).
- [x] Unit test: the sweep is skipped when step-ca is absent from `--services`, and again when the compose file declares no step-ca service.
- [x] `cargo clippy` clean and `rustfmt` clean.
- [x] `grep -rn -- "--user root" docs/` returns nothing.
- [ ] E2E happy path: `infra install` + `init`, then `rotate` CA + password rotation; assert no entry under `secrets/` is root-owned.
- [ ] E2E legacy-host simulation: seed root-owned CA keys, run both `rotate` flows; assert both succeed and every entry is owner-owned.
- [ ] E2E ownership convergence: seed a root-owned file, run `infra install` and `infra up`; assert converged and each command keeps its existing contract.
- [ ] E2E legacy upgrade: initialized tree with seeded root-owned entries, run `infra up`; assert it returns with step-ca running and ownership converged (readiness-vs-sweep race).
- [x] `scripts/preflight/extra/deploy-no-build-smoke.sh` still passes with its existing image archives (air-gapped path adds no image).
- [x] Fresh source install: `docker image rm bootroot-step-ca:0.29.0`, then `infra install` from a clean checkout succeeds (sweep not ordered before `up --build`).
- [x] `--services openbao,postgres`: sweep does not run and pulls nothing on both `infra install` and `infra up`.
- [x] Manual-init example from both docs run verbatim against a scratch dir produces no root-owned entries.
- [ ] All CI jobs green: `check`, `test-core`, `test-docker-e2e-matrix`, and `run-extended` (needs manual dispatch).

### DIFF
--- a/docs/en/installation.md
+++ b/docs/en/installation.md
@@ -66,8 +66,10 @@ If you are not using `bootroot infra install`, you can initialize manually:
 mkdir -p secrets
 printf "%s" "<your-password>" > secrets/password.txt
 
-# Run init inside a container (example)
-docker run --user root --rm -v $(pwd)/secrets:/home/step smallstep/step-ca \
+# Run init inside a container as the owner of ./secrets (example)
+docker run --rm \
+  --user "$(id -u):$(id -g)" \
+  -v $(pwd)/secrets:/home/step smallstep/step-ca \
   step ca init \
   --name "Bootroot CA" \
   --provisioner "admin" \
@@ -77,6 +79,15 @@ docker run --user root --rm -v $(pwd)/secrets:/home/step smallstep/step-ca \
   --provisioner-password-file /home/step/password.txt \
   --acme
 ```
+
+Run the container as the owner of `secrets/`, not as `root`.
+`--user $(id -u):$(id -g)` uses your own uid/gid, which is correct when you
+own `secrets/` (as you do after the `mkdir` above). Everything Bootroot runs
+against `secrets/` runs as that directory's owner — the OpenBao Agent
+sidecars that render `password.txt` and `config/ca.json` into it, and every
+`step` helper container — so material created as `root` would diverge from
+the rest of the tree. Bootroot now repairs such divergence on its next run,
+but the manual path should not create the problem in the first place.
 
 `<your-password>` protects (encrypts) the CA keys. In production use a strong
 password and keep this file out of logs and repositories. In production we
--- a/docs/ko/installation.md
+++ b/docs/ko/installation.md
@@ -58,7 +58,10 @@ bootroot infra up
 mkdir -p secrets
 printf "%s" "<your-password>" > secrets/password.txt
 
-docker run --user root --rm -v $(pwd)/secrets:/home/step smallstep/step-ca \
+# ./secrets 디렉터리 소유자로 컨테이너에서 init 실행(예시)
+docker run --rm \
+  --user "$(id -u):$(id -g)" \
+  -v $(pwd)/secrets:/home/step smallstep/step-ca \
   step ca init \
   --name "Bootroot CA" \
   --provisioner "admin" \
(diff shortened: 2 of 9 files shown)


sehkone:
**[Reviewer Round 2]**

I rechecked the Round 1 items against the updated diff:

- The ownership sweep now passes `--entrypoint chown` before the image and starts the container command at `-R --no-dereference ...`, so the `smallstep/step-ca` helper image entrypoint is bypassed and the spurious missing-`ca.json` warning is avoided.
- `rotate stepca-password` now keeps the local `ensure_file_exists` preflight before the sweep, then sweeps before writing the new password file or running `step crypto change-pass`, so the legacy-root-owned-key path is still repaired without masking missing-file errors.
- The local lifecycle E2E now reseeds root-owned CA keys before `rotate ca-key` and asserts no root-owned entries afterward, so the CA rotation sweep placement is covered rather than relying on the earlier password rotation repair.

I also reviewed the rest of the updated diff against #716. The rotate `step` helpers resolve `--user <uid>:<gid>` from the secrets directory owner, the shared sweep remains the only intentional root container and is scoped to the secrets mount with `--no-dereference`, the infra flows are gated on step-ca being both requested and declared, and the docs/manual-init example plus the stale mode-normalization comment were updated consistently.

No remaining findings from this round.

PR #719, opened 2026-07-18
Replace custom step-ca build with official image v0.30.2 (#717)

sehkone:
## Summary

Removes the custom `docker/step-ca/Dockerfile` build and runs the official `smallstep/step-ca:0.30.2` image directly. The custom build existed by mistake — its `-tags postgres` never did anything (the pgx driver in `smallstep/nosql` sits behind a *negative* build tag that is on by default), so the fork preserved nothing the official image lacks, and the official image has always carried PostgreSQL support.

Because the official image runs as non-root uid/gid 1000 and ships a `HEALTHCHECK` that fails without `--ca-url`, the compose services are adjusted so the switch is operator-invisible:

- **Both compose files** now reference `smallstep/step-ca:0.30.2`, run step-ca as the `secrets/` owner via `user: "${BOOTROOT_STEPCA_USER:-1000:1000}"` (fail-closed to the image's built-in non-root user when unset), disable the failing healthcheck to preserve today's status-only readiness, and drop the dead `STEPUSER=step` env. The explicit `command:` is kept, so the existing `secrets/` layout is untouched.
- **`infra install` / `infra up`** upsert `BOOTROOT_STEPCA_USER` into `.env` (via `update_dotenv_key`) so existing deployments whose `.env` predates the variable converge to the `secrets/` owner uid without a manual edit — including hosts that only ever run `infra up`. The existing ownership sweep repairs any root-owned key material so step-ca does not crash-loop on `Permission denied`.
- **Every step-ca image reference Docker resolves** in `src/` is pinned to `0.30.2`: `step ca init`, both `step certificate create` sites, both rotate sites, the rotate helper constant, and the ownership sweep image. Test fixtures are updated to match.
- **Docs**: the installation "Docker" sections (EN/KO) no longer claim the stack builds step-ca and drop the `--build` no-op; the manual-init examples pin the image; and the cli.md air-gapped sections now *recommend* digest-pinning `BOOTROOT_STEP_CA_IMAGE` (with the reason) rather than merely noting it is possible.
- **`deploy-no-build-smoke.sh`** now pulls step-ca like the other third-party images and builds only `bootroot-http01` locally.

The illustrative untagged/`:latest` mentions in `src/commands/reinit.rs` and `src/commands/init/paths.rs` are intentionally left untouched (Docker never resolves them).

Closes #717

## Test plan

- [x] `cargo clippy --all-targets --all-features` clean, `cargo fmt --check` clean
- [x] Unit tests: `docker run` arg vectors for `step ca init`, both `step certificate create` sites, and both rotate sites name `smallstep/step-ca:0.30.2`; new `upsert_stepca_user_env` test covers append-to-existing and no-op-when-missing
- [x] `docker compose config` parses `docker-compose.yml`, `docker-compose.deploy.yml`, and the `docker-compose.yml` + `docker-compose.test.yml` overlay
- [x] `scripts/validate-deploy-compose.sh` passes unmodified (renders the interpolation default with the uid variable unset)
- [x] `grep -rn "bootroot-step-ca"` returns nothing; every `smallstep/step-ca` hit in `src/`/`docs/` is either `:0.30.2` or a known illustrative mention
- [x] `docker/step-ca/Dockerfile` no longer exists
- [x] `bootroot infra up` on a live stack reports every service healthy (not `step-ca health=unhealthy`) — the only real test of the healthcheck override; no CI job covers it
- [x] E2E: full `bootroot infra install` + `bootroot init` on a clean tree, then end-to-end certificate issuance
- [x] Upgrade path (`.env`): from an `.env` lacking the uid variable and a `secrets/` owned by a non-default uid, both `infra install` and `infra up` alone upsert the variable and run step-ca as that owner uid
- [x] Upgrade path (legacy ownership): with root-owned entries under `secrets/`, `infra up` alone leaves step-ca serving, not crash-looping
- [x] `docker inspect` / `id` confirms step-ca runs as the `secrets/` owner uid, and as non-root uid 1000 when the variable is unset
- [x] `scripts/preflight/extra/deploy-no-build-smoke.sh` with `smallstep/step-ca:0.30.2` loaded from an image archive
- [ ] All CI jobs green: `check`, `test-core`, `test-docker-e2e-matrix` (`ci.yml`), and `run-extended` (`e2e-extended.yml`)


sehkone:

PR #820, opened 2026-08-12
Fail init's prompts on EOF instead of answering them (#817)

AcoPiper:
## Summary

`init`'s prompt reader discarded the byte count `read_line` returns, so the terminating EOF was indistinguishable from a blank line and every prompt past the end of a piped answer sequence answered itself. The EAB credential prompt — the only one that retries — re-prompted forever (gigabytes of validation errors until the process was killed), and the rest returned an empty string, which silently skipped EAB registration, declined to save freshly generated unseal keys, or accepted a default nobody chose.

- `src/commands/init/steps/prompts.rs` now checks the count and bails with the existing `error_prompt_eof` message on a zero-length read. A blank line keeps its meaning everywhere: the empty answer, the offered default, or `n`.
- The three readers (`prompt_text`, `prompt_text_with_default`, `prompt_yes_no`) gained reader-backed helpers an in-memory `Cursor` can drive. Their names, parameter lists and visibility are unchanged, so every existing call site — including `clean` and `reinit` — compiles untouched.
- `prompt_eab_with_validation` takes the reader as a parameter, so the unbounded loop itself is under test rather than only the primitive beneath it. `maybe_register_eab` passes the stdin lock as a temporary, keeping the future `Send` across the `register_eab_secret` await.
- The save-unseal-keys prompt runs outside the rollback envelope on keys held nowhere else, so an unanswerable prompt there emits the cleartext echo *before* failing — via the same code path the declined branch uses, extracted into a shared helper so the two cannot drift.
- Two preflight scripts were relying on EOF as an answer (and drove the EAB loop into exactly that spin); both now supply one answer per prompt the run reaches. The `init` invocations in `tests/openbao_stepca_integration.rs` were extended the same way, and `run_command_with_input` says so at its definition.
- Manual (`docs/en/cli.md`, `docs/ko/cli.md`) and the E2E walkthrough now describe the new behaviour, and the walkthrough's sample sequence was corrected.
- Every piped `init` line this branch touches now carries `BOOTROOT_LANG=en` on the right of the pipe. A prefix assignment binds to the command it precedes, so on the left it reached only `printf` and the run itself took the caller's locale — while the assertions parsing that run's output, and the EOF message quoted in the walkthrough, all assume English.

Closes #817

## Test plan

- [x] `prompt_text` errors with `error_prompt_eof()` when `read_line` returns `0`
- [x] Unit test: an empty in-memory reader errors from the text, with-default and yes/no paths
- [x] Unit test: `"\n"` yields `""`, the default, and `false` respectively — blank line and EOF stay distinct
- [x] Unit test: `"value\n"` yields `"value"` trimmed; `"y\n"` yields `true`
- [x] All three paths are exercised with an in-memory reader, without touching process stdin
- [x] Regression test: `prompt_eab_with_validation` with an empty reader returns the EOF error instead of looping — no `OpenBaoClient`, no async runtime, no network
- [x] Companion test: a half-answered `kid`/`hmac` pair errors rather than retrying
- [x] Companion test: a rejected attempt followed by a valid pair still loops once and returns the valid credentials
- [x] Unit test for the shared cleartext-echo helper: one line per key, in key order
- [x] End-to-end test: `init` against an uninitialized `OpenBao` stub with empty stdin exits nonzero, names EOF on stderr, and still prints every generated unseal key
- [x] Public signatures and visibility of the three prompts unchanged; `clean` and `reinit` call sites compile untouched
- [x] `cargo test` passes
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate` clean
- [x] `scripts/preflight/ci/check.sh`, `validate-deploy-compose.sh`, `validate-compose-instance-names.sh` and `extra/deploy-no-build-smoke.sh` pass
- [x] `scripts/preflight/ci/test-core.sh`: both rewritten piped-answer `init` runs pass, including the `PASS: unseal keys displayed in cleartext after declining save` assertion, which now exercises the declined path deliberately rather than the EOF path
- [x] `scripts/preflight/extra/cli-scenarios.sh` passes end to end; its `y\nn\nn\n` lines up one-to-one with the db-provision, EAB-registration and save-unseal-keys prompts
- [x] `BOOTROOT_LANG` placement verified against the built binary: `printf "" | BOOTROOT_LANG=ko bootroot clean` prints the Korean EOF message, `BOOTROOT_LANG=ko printf "" | bootroot clean` prints the English one
- [ ] `scripts/preflight/run-all.sh` end to end — not reproducible on the macOS Docker host used here, for reasons unrelated to this change and identical on `main`: `test-core.sh` cannot `docker exec ... >> /etc/hosts` into the unprivileged step-ca container, `e2e-matrix.sh` needs `sudo` host-file edits plus the Linux `docker0` gateway, and `e2e-extended.sh` fails the same 3 of 7 cases on `main` as on this branch (`runner-timer`/`runner-cron` reject the host's `Acos-MacBook-Pro.local` as a DNS label; `reinit-recovery` cannot bind `172.17.0.1`). No artefact from any of these runs contains the EOF error. The Linux equivalents are green in CI: `Unit & CLI Smoke` covers `test-core.sh` and all ten `Docker E2E` jobs cover the matrix. `e2e-extended.yml` is `workflow_dispatch`/`schedule` only, so it does not run on pull requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,22 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fixed `bootroot init` treating a closed stdin as an answer. Every
+  `init` prompt read the terminating EOF as an empty line, so a run
+  whose piped answer sequence ran out answered the rest of its prompts
+  itself: the EAB credential prompt re-prompted forever (over five
+  gigabytes of output in one observed run, ending only when the process
+  was killed), and the remaining prompts took an empty string —
+  silently skipping EAB registration, declining to save freshly
+  generated unseal keys, or accepting a default nobody chose. A prompt
+  with no input left now fails the run with "no input available (stdin
+  reached EOF / not a terminal)". Pressing Enter is unchanged: a blank
+  line still means the empty answer, the offered default, or "no".
+  Where the unseal-key save prompt is the one that runs out, the keys
+  are still echoed in cleartext before the run fails, so a partial init
+  cannot leave them recorded nowhere. The confirmations in
+  `bootroot clean` and `bootroot reinit` now report the same error
+  instead of quietly declining; both still decline to act.
 - Fixed the published manual having no styling at all. The theme was
   installed into a dot-prefixed directory under `docs/`, and MkDocs
   excludes every dot-prefixed path inside `docs_dir` from the build, so
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -610,6 +610,17 @@ prompts on the run still need their own flags (`--no-eab` or
 `--no-save-unseal-keys`, and `--root-token`/`--unseal-key` where the
 OpenBao state requires them).
 
+A prompt that is reached with nothing left to read — stdin closed, or a
+piped answer sequence that ran out — aborts the run with `no input
+available (stdin reached EOF / not a terminal)` instead of being
+answered. Pressing Enter is unaffected: a blank line is still the empty
+answer, the offered default, or `n`. So a script driving `init` over a
+pipe must supply one answer per prompt the run reaches, or set the flag
+that suppresses it. The one exception is the "Save unseal keys to file
+for automatic unseal?" prompt: running out of input there still echoes
+the freshly generated keys in cleartext before the run fails, so a
+partial init cannot leave them recorded nowhere.
+
 If a previous `init` failed mid-flight and rolled back, OpenBao may
 remain initialised in its volume while bootroot has no usable root
 token. `init` detects this state on startup and emits an actionable
--- a/docs/en/e2e-ci.md
+++ b/docs/en/e2e-ci.md
@@ -161,7 +161,7 @@ bootroot infra install --compose-file "$COMPOSE_FILE"
 # build_admin_dsn_from_env() connects via the host-mapped port.
 # The piped sequence must answer every prompt the run reaches: init
 # fails on EOF rather than answering an unanswered prompt itself.
-BOOTROOT_LANG=en printf "y\ny\ny\n" | bootroot init \
+printf "y\ny\ny\n" | BOOTROOT_LANG=en bootroot init \
   --compose-file "$COMPOSE_FILE" \
   --secrets-dir "$SECRETS_DIR" \
   --summary-json "$INIT_SUMMARY_JSON" \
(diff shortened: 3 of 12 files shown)


AcoPiper:
**[Author Round 1]**

- **P2 – apply `BOOTROOT_LANG` to the command being tested. — Fixed.** Correct, and confirmed against the built binary rather than reasoned about: `BOOTROOT_LANG=ko printf "" | bootroot clean` prints the English EOF message, `printf "" | BOOTROOT_LANG=ko bootroot clean` prints the Korean one. The assignment reached only `printf`, so every one of these runs was taking whatever locale the caller exported — invisible today because `en` is also the clap default (`src/cli/args.rs:44`), and wrong the moment someone runs the preflight with `BOOTROOT_LANG=ko` set. Moved to the right of the pipe in all five lines this branch touches: `scripts/preflight/ci/test-core.sh:34` and `:57`, `scripts/preflight/extra/cli-scenarios.sh:93`, and `docs/{en,ko}/e2e-ci.md:164`/`:161`. I also moved the second occurrence in each walkthrough (`docs/en/e2e-ci.md:307`, `docs/ko/e2e-ci.md:298`) — those lines were not in the diff, but a page that shows the pattern twice with one form fixed and the other not is worse guidance than either form used consistently.

**Not changed, and flagged rather than folded in.** The same misplacement predates this branch in four files it does not otherwise touch: `.github/workflows/ci.yml:241` and `:262` (whose `grep -q "unseal key"` and root-token `awk` are exactly the English-output assumptions you named), `scripts/impl/run-local-lifecycle.sh:331`, `scripts/impl/run-remote-lifecycle.sh:284`, and `scripts/impl/run-ca-key-rotation-recovery.sh:559`. `scripts/impl/run-reinit-recovery.sh:446` already has it on the right, so the correct idiom is the tree's own. I left those alone: none of them is an EOF-answer site, fixing them means editing the CI workflow and three E2E drivers on an issue about `read_line` returning `0`, and the E2E drivers are the ones I cannot run on this macOS host. Happy to fix them here if you would rather have it in one pass than in a follow-up — say the word and it is a one-token move in each.

Verification: `bash -n` on both scripts, `markdownlint-cli2` clean on the two changed pages, and the manual probe above. No Rust changed, so `cargo test` / `clippy` / `fmt` results from Round 1 stand. PR description updated with the change and the probe.

PR #843, opened 2026-08-14
Publish the remaining writes by rename instead of truncating (#841)

### LINKED ISSUE #841: Publish the remaining writes by rename instead of truncating
## Problem

`AGENTS.md`:

> **Atomic writes**: Write state, config, and other files that another process may read atomically — write to a temporary file in the same directory, then `fs::rename`. Never truncate-and-write in place.

#805 flushed the containing directory at every site that already staged through a temporary, and named the sites that stage nothing as a separate gap needing their own issue. This is that issue, less the two secret-bearing writers, which are urgent for a different reason and are handled on their own.

Four sites remain, in two shapes.

### Truncate in place, no staging at all

**`StateFile::save`** (`src/state.rs:125`):

```rust
std::fs::write(path, contents)
    .with_context(|| format!("Failed to write {}", path.display()))
```

`state.json` is what `bootroot` reads back to know what it already did. A crash between the truncate and the last byte leaves invalid JSON at the path, and the next run fails to parse it rather than falling back to anything — the state is gone, not stale.

There is a second reader. `bootler`'s `core/src/rotation.rs` schedules two rotation units ten minutes apart specifically because of this file, and says so:

> Both invocations end by recording `last_secret_id_rotation` in bootroot's `state.json`, whose `StateFile::save` is a plain non-atomic `fs::write`, so co-firing units would race on that file for no benefit.

A staggered schedule in another repository is presently what stands in for atomicity here. That is a workaround holding up a guarantee this function should provide itself, and it is worth removing the reason for it.

**`cert_group::write_cert_file`** (`src/cert_group.rs:474`):

```rust
fs::write(path, cert_pem).await …
fs::set_permissions(path, Permissions::from_mode(CERT_FILE_MODE)).await …
```

A certificate is not a secret, so `0644` applied after the write is not the permissions problem the sibling issue covers. The atomic-write rule still applies: a reader — `bootroot-agent`, or the server being reloaded — can observe a truncated PEM. Note the asymmetry this leaves in one module: `write_key_file` a hundred lines above stages through `stage_key_file` and renames, and the certificate beside it does not.

### Staged bytes, unstaged name

**`write_init_summary_json`** (`src/commands/init/steps/orchestrator.rs:285`) and **`write_root_token_file`** (`:334`).

These two are already careful, and are listed so their care is not mistaken for completeness. Both create with `OpenOptions::mode(0o600)`, tighten a pre-existing destination before writing, `sync_all` the bytes, and re-assert the mode. What they do not do is publish through a rename:

```rust
.create(true)
.truncate(true)
.write(true)
.mode(0o600)
```

So the destination is truncated in place, and on a first write the new directory entry is left unflushed exactly as the pre-#805 sites were. The bytes are durable; the name pointing at them is not.

## Scope

Route all four through the shared primitive rather than fixing each one differently.

- `fs_util::atomic_write` (`src/fs_util.rs:379`) and `atomic_write_blocking` (`:398`) already create the temporary with an explicit mode, flush it, rename, and flush the directory. `StateFile::save` and `write_cert_file` should use it.
- The two orchestrator writers keep their existing pre-write tightening of an existing destination — that guards a case the primitive does not, since a temporary is a fresh inode — and gain the staging and rename around it.

Decide the durability question per file and record the answer at each site, as `AGENTS.md` asks and as #805 established the pattern for:

- `state.json` is read back to resume. It takes the flush.
- The `init` summary and the root token file are written once during `init` and read by an operator afterwards. They take it.
- A certificate is reissued on the next renewal if lost, which is the same reasoning `cert_group::write_key_file` already records at its rename for the key beside it. It does not take the flush, and should say so in the same terms.

## Constraints

- Do not change file modes. `CERT_FILE_MODE` stays `0644`, the orchestrator pair stays `0600`.
- Do not change what is written or the errors reported.
- `StateFile::save` is synchronous and called from both sync and async paths. Do not make it async; if an async caller needs it, wrap at the call site the way #839 establishes for the rotation-state writers.
- Do not remove the orchestrator pair's pre-write `set_permissions` on an existing destination. It is deliberate and documented, and staging does not replace it.

## Acceptance criteria

- No production write site in the crate truncates a destination in place; `grep -rn 'fs::write\|create(true)' src/` shows only staged writes and the two secret writers covered by the sibling issue, if it has not landed yet
- All four sites publish by rename from a temporary in the same directory
- Each site records, in a comment, whether it flushes the containing directory and why — matching the decisions above
- `bootler`'s stagger is no longer load-bearing for `state.json`: a comment at `StateFile::save` states that concurrent writers now see one version or the other. Do not change anything in `bootler` here; that is its own repository's follow-up
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff, full CI green including the Docker E2E matrix

## Not in this issue

- `save_unseal_keys` and `eab::write_key_file`. Same rule, but they also apply permissions after writing a secret, which is a window rather than a torn read. Their own issue.
- `FastPollState::save`'s `create_dir_all` on the parent, where the directory's own entry in *its* parent is unflushed. #805 put chasing that chain upward out of scope and it stays out.

AcoPiper:
Closes #841

## Summary

Every production writer in the crate that published its file by truncating the destination now stages a temporary in the same directory and `rename`s it into place, so a reader — another `bootroot` invocation, `bootroot-agent`, a container mounting the file, or the operator — sees either the whole previous file or the whole new one, never a torn one. All of them, and the writers #805 already staged, go through one publish routine: `fs_util::publish_staged_blocking` — the exception being the override credential writers, which keep a staged publish of their own for the reasons given below.

The four sites the issue enumerated:

- **`StateFile::save`** (`src/state.rs`) goes through `fs_util::atomic_write_blocking` and stays synchronous, so its synchronous callers are unchanged. It takes the directory flush: this is the file `bootroot` reads back to resume, so a torn write is not a stale record but no record at all. A doc comment records that `bootler`'s ten-minute stagger between its two rotation units is no longer load-bearing for this file — concurrent writers now see one version or the other. Nothing in `bootler` is touched; that is its own repository's follow-up. Async callers use the `StateFile::save_async` entry point beside it (below).
- **`cert_group::write_cert_file`** now publishes exactly as `write_key_file` beside it, with the mode and the `--cert-group` policy's group ownership landing while the file is still at its temporary path. The asymmetry where the key staged and the certificate truncated is gone. It declines the directory flush, in the terms the key already used: a certificate lost to a crash is reissued at the next renewal.
- **`write_init_summary_json`** and **`write_root_token_file`** (`src/commands/init/steps/orchestrator.rs`) keep their pre-write tightening of an existing destination — that guards the older credentials already sitting at the path, which a fresh inode renamed over them does not — and gain staging and a rename around it. Both take the directory flush; they are written once during `init` and read by an operator afterwards.

## The rest of the crate

The issue makes "no production write site in the crate truncates a destination in place" an acceptance criterion, and the enumerated four do not satisfy it on their own. Every remaining production writer is converted here too. They fall into three groups.

- **Configuration with a live reader.** `agent.toml` (`service update`'s rerender, `service remove --strip-config`, and `bootroot-remote`'s apply), `.env`, `ca.json` and its `OpenBao` Agent template, `openbao.hcl`, the HTTP-01 responder config and template, the two `OpenBao` Agent `agent.hcl` files, and the four generated compose overrides. Each had a process on the other side of it — `docker compose` interpolating `.env` on every invocation, step-ca parsing `ca.json` at boot, a sidecar re-rendering a template on a fixed interval, `bootroot-agent` re-reading `agent.toml` on every ACME retry. The `agent.toml` writers are the sharpest case: `service add` was moved off a truncating write by #613 precisely because a reload landing in the gap reports "profile not found in reloaded config", and the three writers that edit the same file afterwards had kept it.
- **Credentials.** The four `OpenBao` Agent `role_id`/`secret_id` files, the service `secret_id`/`role_id` inside the secrets tree, the step-ca CA password, the `OpenBao` recovery-key output, and the remote bootstrap artifact. These were `write` followed by `set_key_permissions`, so besides the torn read they left the file at its final path under the umask's mode for the length of a chmod. Staging applies the mode to the temporary, so it holds from the moment the file appears. The issue's sibling holds `save_unseal_keys` and `eab::write_key_file` back for exactly this permission window; those two are untouched here as it directs, but a site being re-plumbed for the torn-read fix anyway does not get to keep the window.
- **`init`'s rollback restore** (`steps::rollback_file`), which puts a snapshotted file back on the failure path while the containers `init` started may still be reading it.

`grep -rn 'fs::write\|create(true)' src/`, filtered to non-test code, now returns exactly four lines: `eab::write_key_file` and `save_unseal_keys`, held for the sibling issue, and the two `reinit` preflight probes, which create a uniquely named marker and delete it rather than publishing a file.

## One general-purpose publisher, four spellings

`cert_group` had its own staging-and-rename, predating `fs_util`'s and older than this issue: a second copy of the staging create, the file flush, the chmod, the chown, a name-allocation retry loop and the rename. Both are now `fs_util::publish_staged_blocking`, with the two axes the callers actually differ on as arguments — where the new inode's ownership comes from (`StagedOwner`) and whether the directory entry is flushed (`StagedDurability`). Callers reach it through four wrappers rather than directly:

| | flushes the directory | rename only |
| --- | --- | --- |
| async | `atomic_write` | `atomic_replace` |
| blocking | `atomic_write_blocking` | `atomic_replace_blocking` |
| async, through a link | `atomic_write_through_symlink` | `atomic_replace_through_symlink` |
| blocking, through a link | `atomic_write_through_symlink_blocking` | `atomic_replace_through_symlink_blocking` |

`cert_group` is the one caller that goes in directly, with `StagedOwner::PolicyGroup`.

The override credential writers — `create_owned_credential_noclobber`, `write_owned_file_replace` and `atomic_rewrite_owned_no_symlink`, for a `role_id`, `secret_id` or `eab.json` relocated into an operator-provisioned, agent-owned directory — keep a staged publish of their own. They already staged before this branch and reach the same two guarantees, but on ownership and clobbering they need what the shared publisher deliberately does not offer: the uid/gid comes from the parent directory or is read back through `symlink_metadata`, and a name already present is refused rather than replaced. Neither policy generalises to the files above, so folding them together would make one of the two callers wrong. The rustdoc on `publish_staged_blocking` scopes its claim accordingly.

Each site records which of the two it picked and why, because the wrong inference surfaces as a daemon that can no longer read a file it could read before, or as a disk round trip on every write in a loop that did not need one. The split follows the issue's own rule: a file the program reads back to resume, or that holds a credential the stack logs in with — losing one of those takes an operator or another rotation to put back, not the next write — flushes; a file that is regenerated by the next renewal, the next sidecar render, or a re-run of the command that produced it does not. Every `AppRole` credential file is on the flushed side, `role_id` included: it is re-readable from `OpenBao`, but only on the next `rotate` run, and until then the agent or sidecar it belongs to cannot log in. Both backfills that write one run only when the file is missing, so that flush is not on any repeated path.

Two properties come free from the shared routine — the staged file is created `0600` and reaches its final mode only at the temporary name, so the guarantee #593 asked of the key file now holds for every caller, and the temporary's name is the primitive's own, so a destination whose file name is not valid UTF-8 (which a Unix path may be, and which the writes this replaced never looked at) needs nothing special of it.

Two more pieces the enumerated four required:

- **`write_ca_bundle`** (`src/fs_util.rs`) was still a plain `fs::write`. The bundle has the reader the certificate has — the agent re-reads it to rebuild its trust store while a rotation may be rewriting it — so it publishes the same way and makes the same durability decision.
- **`fs_util::resolve_symlink_destination`** exists because the truncating write these replaced opened with `O_TRUNC`, which follows the final symlink, where a rename replaces it. See *Symlinked destinations* below.

## Symlinked destinations

`O_TRUNC` follows a link at the final path component and delivers the bytes to its target; `rename(2)` replaces the name it is given. So a conversion done naively changes what an operator's link means — the link is destroyed and the file it pointed at keeps the previous contents, while the write reports success. Which answer each file takes is a decision, and every writer here states it:

- **A destination the operator arranges is written through the link**, via the `_through_symlink` wrappers, which resolve first and publish at the target. That is the configuration bootroot renders into its own tree — `.env`, `ca.json` and its `OpenBao` Agent template, `openbao.hcl`, the HTTP-01 responder config and template, the two `OpenBao` Agent `agent.hcl` files, the compose overrides, `init`'s rollback restore, `state.json` — and the output paths named on the command line: `init --summary-json`, `init --root-token-output`, `rotate openbao-recovery --output`, and `bootroot-remote bootstrap`'s `agent.toml` destination. A dangling link is followed through its own text (`canonicalize` has nothing to resolve against), matching what `O_CREAT` through the link used to do. A chain that loops back on itself resolves to nothing that can be published without destroying a link, so resolution fails with the `ELOOP` the truncating write reported.
- **Credentials at a path bootroot chose publish at that path.** Inside the secrets tree a link is a redirection vector rather than an operator convenience, and the reader reads the path bootroot handed it — which the rename leaves holding the current secret — so following one buys nothing and costs the guarantee. The override credential paths, whose directory an unprivileged user owns, go further and refuse a link outright (`atomic_rewrite_owned_no_symlink`, unchanged).
- **The control node's `agent.toml` and the issued cert, key and bundle publish at the path too**, because a writer that already renames over that name predates this branch: `service::local_config` since #613, `write_key_file` since #593. A link an operator puts at one of those paths does not survive the command that *creates* the file, so resolving it in the writers that only *edit* it would leave two writers of one file disagreeing rather than preserve anything. Making the certificate match the key here is the same asymmetry-removal the issue asked for. The `agent.toml` a bootstrap target carries is the other way round and sits in the first group: `bootroot-remote bootstrap` is the writer that *creates* the file there, so a link at its destination is one the operator arranged and one the truncating write it replaces followed.

The two `init` preflights accept a symlinked destination on purpose: they resolve the link and judge the *target's* mode. The destination is resolved before it is tightened and staged, and the preflight probes the directory the staging will actually use, so a link into a not-yet-existing directory is caught before OpenBao is wiped rather than after; a cyclic destination is refused before the wipe rather than failing after it. Resolution is explicitly not a security check, and says so — a caller whose destination an untrusted user can plant still has to refuse the symlink.

## Keeping the flushes off the runtime

Publishing `state.json` now costs three disk round trips where the truncating write cost one, and every async command path was calling the synchronous writer directly. `StateFile` gains an async entry point beside the synchronous one — `save_async`, sharing a blocking core, serializing the JSON on the async side so only owned data crosses into `spawn_blocking` — and the async production callers move to it. Three helpers whose only production caller is async (`write_state_file`, `reinit::write_minimal_state`, `service remove`'s `finalize_removal`) are async themselves for the same reason. `infra install` and `service update` run outside any runtime and keep the synchronous entry point, as do the tests, which must not need a runtime to write a state file. This is the pattern #839 established for the rotation-state writers.

## Modes and ownership

File modes are unchanged — `CERT_FILE_MODE` stays `0644`, the orchestrator pair stays `0600` — but each is now applied before the file is published rather than set after the bytes landed. A staged temporary inherits no mode from the file it replaces, so a writer that used to truncate in place has to state one. `fs_util::preserved_mode` is the shared answer: read the mode off the destination where there is one, so a file an operator narrowed by hand (or a restrictive umask created narrow) keeps that mode across every later write exactly as the in-place write left it, and fall back to a stated default only on a create. The default is `0644` for `state.json`, `.env`, `ca.json`, `openbao.hcl` and the compose overrides — what the umask produced — and `0600` for everything inside the secrets tree, where the policy's constant is the answer and a stale wider mode must not outlive the file it was attached to. A create on a host with a non-default umask is the only case that can observe a difference; `CHANGELOG.md` names it.

The staged file is `fsync`ed *after* its ownership and mode are applied, not at the bytes. An `fsync` persists the inode as it stands, so flushing first would leave the chown and the chmod in memory only: a crash could recover a durably named, fully written file wearing the temporary's own `0600` and the writer's primary group instead of the mode and the policy gid it was published with. The directory flush does not cover that — it makes the name durable and says nothing about the inode behind it.

Ownership is the one place the callers answer differently, deliberately. `StagedOwner::Destination` carries the destination's uid/gid across the rename: those files have no ownership policy to restate, and a re-owned one the daemon cannot read is an outage. `StagedOwner::PolicyGroup` takes ownership from the `--cert-group` policy instead, as `write_key_file` has since #593: the gid these files need is the one the policy names, re-reading it off the destination would let a stale owner outlive the policy that replaced it, all three land world- or group-readable by that policy so no consumer loses access, and the rename needs only the directory's permission — a writer that could not replace the destination before is not made to fail on a chown it has no privilege for. A unit test pins both answers against the same seeded destination.

## Documentation

`docs/en/cli.md` and `docs/ko/cli.md` gain a **How bootroot writes files** section: what stage-then-rename guarantees a reader, that the final mode holds from the moment the file appears, that a rename installs a new inode (so a single-*file* bind mount needs the container restarted where a directory mount does not), which files take the directory flush and which do not, and which paths keep being written through a symlink you put there and which replace it. The per-command prose that described the two `init` outputs in terms of the create-mode-then-chmod write they no longer perform, and named only the key as atomic where the certificate and the bundle now are, is corrected. The section is scoped to the writers converted here: it names the two this change does not cover — the OpenBao unseal-keys file and the service `eab.json`, held for the sibling issue — and says the guarantees below it do not apply to them.

`CHANGELOG.md` names the same two exclusions, so the release notes and the manual promise the same set.

`FastPollState::save`'s `create_dir_all` chain stays out of scope, as the issue directs.

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` is clean
- [x] `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` is clean, and `./scripts/check-docs.sh` builds the manual with `--strict`
- [x] `scripts/preflight/ci/check.sh` passes end to end (fmt, clippy, rustdoc, Python, Biome, markdownlint, docs, audit)
- [x] `cargo test` passes (1041 bin + 424 lib tests green locally)
- [x] `grep -rn 'fs::write\|create(true)' src/`, filtered to non-test code, returns only `eab::write_key_file`, `save_unseal_keys`, and the two `reinit` preflight probes
- [x] `StateFile::save` publishes a new inode over an existing `state.json`, leaves no temporary behind, keeps an existing mode and lands at `0644` on a create
- [x] `StateFile::save` is still synchronous, and `save_async` publishes the same file — same inode replacement, same preserved mode, same symlink resolution — from a current-thread runtime
- [x] Every production `save` reached from an async path goes through `save_async`; the two synchronous command paths (`infra install`, `service update`) are unchanged
- [x] `write_cert_file`, `write_key_file` and `write_ca_bundle` all publish by rename, with the policy's mode and gid applied before the rename
- [x] `StagedOwner::Destination` carries a seeded gid across the rename and `StagedOwner::PolicyGroup` does not, so neither ownership answer can drift into the other
- [x] `atomic_replace` installs a new inode, applies the requested mode, and leaves no staged sibling — the no-flush spelling is still a staged publish
- [x] `preserved_mode` reads an existing destination's mode and falls back to the caller's default only on a create
- [x] `write_dotenv` and `update_dotenv_key` both publish a new inode, leave no temporary, and keep an operator-narrowed `.env` at `0600`
- [x] `write_ca_bundle` still creates a missing parent directory before staging
- [x] `write_init_summary_json` and `write_root_token_file` publish by rename at `0600`, and still tighten a pre-existing world-readable destination before writing
- [x] Both `init` outputs write through a symlinked destination to the link's target, including a dangling link, and leave the link in place
- [x] The `_through_symlink` wrappers deliver to a link's target and leave the link standing, create a dangling link's target, and refuse a cycle without touching the links in it; the bare wrappers publish at the name over the same seeded link
- [x] `write_dotenv` and `update_dotenv_key` update the target of a symlinked `.env` and leave the link in place
- [x] `rerender_local_managed_profile` publishes `agent.toml` at its own name over a symlinked destination, matching `service add`
- [x] `bootroot-remote bootstrap` writes `agent.toml` through a symlinked destination and leaves the operator's link in place, the opposite answer from the control node's writers of that file name
- [x] `validate_root_token_output_path` and `validate_summary_json_output_path` probe the directory the staged write will use, not the one holding the link
- [x] `resolve_symlink_destination` passes a regular file and a not-yet-existing path through unchanged, follows live and dangling links and chains, and fails on a cycle without touching the links in it
- [x] Both `init` writers and both preflights refuse a cyclic destination, and leave the operator's links in place
- [x] A certificate and a key whose file names are not valid UTF-8 publish and leave no staged file behind (Linux only — APFS rejects such a name with `EILSEQ`, verified in a `rust:1-slim` container)
- [x] Every converted site carries a comment stating whether it flushes the containing directory and why
- [x] Every `AppRole` `role_id` and `secret_id` writer flushes the containing directory, so the durability contract the manuals state holds for the whole credential pair and not only half of it
- [x] `StateFile::save`'s comment states that concurrent writers now see one version or the other, so `bootler`'s stagger is no longer load-bearing
- [x] The fake `docker` executables the converted `init` and `rotate` tests write are written by a child process, so no descriptor on one is ever in this process's table for a `fork` to duplicate and no later spawn — production's included — can be refused with `ETXTBSY`; the writer delivers the exact bytes at the exact name, including one holding a quote, a space or a byte that is not UTF-8 (the last verified in a `rust:1-slim` container, since APFS rejects it with `EILSEQ`)
- [x] `scripts/preflight/ci/e2e-matrix.sh` locally, through the Round 3 tree: both lifecycles, the rotation/recovery matrix, reinit recovery, step-ca SANs and OpenBao TLS no-delta all pass. It stops at `openbao-tls-reown`, which needs passwordless `sudo` this macOS host does not have. Every commit since is covered by the same matrix on CI, `openbao-tls-reown` included, on the head commit below.
- [x] Full CI green on `39284d4` (previously green on `8f971c9`) — all 18 checks, including every `Docker E2E` matrix job (`local-hosts`, `local-no-hosts`, `remote-hosts`, `remote-no-hosts`, `rotation`, `reinit-recovery`, `stepca-san`, `two-instance`, `openbao-tls-no-delta`, `openbao-tls-reown`)


AcoPiper:
**[Reviewer Round 8]**

The Round 7 P2 is resolved and is not carried forward. The sleep/retry loop is gone, and moving fake-executable writes into a child process removes the parent-process descriptor that made an unrelated fork retain a writable executable; the author's scope rationale is sound.

- **[P2] Correct the AppRole durability matrix in both manuals.** [docs/en/cli.md:147](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/docs/en/cli.md#L147) and [docs/ko/cli.md:139](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/docs/ko/cli.md#L139) say every AppRole `role_id`/`secret_id` flushes its containing directory. That is not the behavior chosen for `role_id`: [`ensure_infra_role_id_file`](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/commands/rotate/approle.rs#L517-L521) and [`ensure_role_id_file`](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/commands/rotate/approle.rs#L715-L720) deliberately use `atomic_replace`, and their comments correctly explain that a missing `role_id` can be fetched from OpenBao again. Please change both matrices to distinguish re-fetchable `role_id` files from issued `secret_id` files (or make the implementation flush them); as written, the new operational durability contract is false.

AcoPiper:

AcoPiper:
**[Author Round 8]**

- **[P2] Correct the AppRole durability matrix in both manuals.** **Fixed**, by the second option you offered: the implementation now flushes, rather than the matrices splitting.

  You are right that the contract was false as written. Choosing which way to make it true came down to the fact that a `role_id`-vs-`secret_id` split is not the split the code actually had. Four writers put a `role_id` on disk, and three of them already flushed — `write_agent_credential` for the two OpenBao Agent pairs at `init`, `service::approle::write_role_id_file` inside the secrets tree, and the *override* branch of `ensure_role_id_file` itself, via `create_owned_credential_noclobber`. Only the two `rotate`-side secrets-tree backfills did not, so `ensure_role_id_file` gave one file name two durability answers depending on which branch created it. A matrix distinguishing re-fetchable `role_id` from issued `secret_id` would have had to distinguish, inside `role_id`, the one `rotate` backfills from the one `init` and `service add` write and from the one an override path takes — three cases an operator cannot act on differently.

  The re-fetchability argument the old comments made is also weaker than it reads. `role_id` is re-readable from OpenBao, but only on the *next* `rotate` run — nothing re-fetches it in between — so a crash that loses the directory entry is not "one more round trip", it is an agent or sidecar that cannot log in until an operator runs a rotation. That is the outage class the flush exists for. And the cost argument does not apply here at all: both writers sit behind an early return (`role_id_path.exists()`, and a non-empty existing file), so they write only on a backfill and never on a repeated path.

  So `ensure_infra_role_id_file` (`src/commands/rotate/approle.rs:523`) and the secrets-tree branch of `ensure_role_id_file` (`:723`) move from `atomic_replace` to `atomic_write`, and both comments now state the flush and why. `39284d4`.

  Three doc/description touch-ups came with it, since the rationale sentence under each matrix was the part that had drifted rather than the lists:

  - `docs/en/cli.md` and `docs/ko/cli.md`: the "why the first list flushes" sentence said the flushed credentials are the ones "OpenBao has already issued and will not hand out again", which was never true of a `role_id` even before this change. It now reads as "a credential the stack logs in with, and losing one takes an operator or another rotation to put back rather than the next write", with a following sentence naming `role_id` explicitly and saying why it is on that list despite being re-readable. Both `- every AppRole role_id/secret_id …` bullets are correct as they stand and are unchanged.
  - `CHANGELOG.md` carried the same "issued and cannot re-read" phrasing; corrected to match.
  - The PR description's durability paragraph, same sentence, same correction, plus a test-plan line for the property.

Verification on `39284d4`: `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt -- --config group_imports=StdExternalCrate` no diff, `cargo test` green (1044 bin + 424 lib), `markdownlint-cli2` clean, `./scripts/check-docs.sh` builds both manuals with `--strict`. Full CI is running on the pushed head; I will report the Docker E2E matrix result when it lands.

AcoPiper:
**[Reviewer Round 9]**

The Round 8 P2 is resolved and is not carried forward. The two formerly unflushed rotation-side `role_id` backfills now use `atomic_write` ([src/commands/rotate/approle.rs:523](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/commands/rotate/approle.rs#L523) and [src/commands/rotate/approle.rs:723](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/commands/rotate/approle.rs#L723)), and the English and Korean manuals now correctly put every AppRole `role_id`/`secret_id` on the flushed side.

- **[P2] Correct the shared primitive’s scope claim.** [`publish_staged_blocking`’s documentation](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/fs_util.rs#L731) says that “Every production file bootroot publishes” reaches this routine and therefore has the no-torn-read and final-mode guarantees. That is knowingly untrue: [the manual](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/docs/en/cli.md#L130) correctly excludes `save_unseal_keys` and `eab::write_key_file`, which still truncate in place and set `0600` afterwards. Scope this rustdoc to staged/converted writers (or explicitly name the two exclusions), so the implementation documentation does not promise a property the crate currently lacks.

- **[P3] Update the PR test-plan checklist.** Its final item still leaves “Full CI green on `39284d4`” unchecked, despite the current head’s checks all succeeding. Mark it complete so the PR’s declared verification matches its actual state.

AcoPiper:

AcoPiper:
**[Author Round 9]**

Both items accepted; `e35f3c4` carries the code change.

- **[P2] Correct the shared primitive's scope claim — Fixed.** The claim was untrue as written, and the manual it contradicts is the one that is right: `save_unseal_keys` and `eab::write_key_file` still write over the destination in place and set `0600` afterwards, so neither the no-torn-read nor the final-mode guarantee covers them. `publish_staged_blocking`'s rustdoc now scopes both properties to the staged writers it lists ("Every staged production file reaches disk through it — … For those the destination name is only ever observed as the previous file or the complete new one, and the final mode holds from the moment the name appears"), and names the two exclusions immediately after, with the same two reasons the manual gives and a note that converting them is a separate change. Naming them rather than only narrowing the wording, because a reader arriving at the primitive to decide whether a new writer needs anything special is exactly the reader who should not have to go to `docs/en/cli.md` to learn that two writers do not go through it.

- **[P3] Update the PR test-plan checklist — Fixed.** The final item is now checked: all 18 checks on `39284d4` passed, `openbao-tls-reown` included. I will refresh the commit it names to `e35f3c4` once that run finishes; `e35f3c4` is a rustdoc-only change with no behaviour in it, and the preflight `check.sh` gates it locally — `cargo fmt`, `cargo clippy --all-targets -- -D warnings` and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` are all clean on it.

AcoPiper:
**[Reviewer Round 10]**

- **[P2] Correct the shared primitive’s scope claim (still unresolved).** `e35f3c4` correctly names the two writers that are not staged at all, but [the revised rustdoc](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/fs_util.rs#L731) still says that every staged production file reaches `publish_staged_blocking` and calls it the crate’s one staging implementation. That excludes real production staging paths: [`atomic_rewrite_owned_no_symlink`](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/fs_util.rs#L294) creates its own `NamedTempFile` and publishes it with `persist`, while [`write_owned_impl`](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/fs_util.rs#L384) does the same for `create_owned_credential_noclobber` and `write_owned_file_replace`. These are not merely test helpers; the service and rotation AppRole credential paths call them. The two special paths may need their separate ownership/no-clobber behavior, but the documentation needs to scope the claim to the shared general-purpose publisher and its callers (or name these exceptions too).

The Round 9 test-plan item is resolved: the full-CI checkbox is now checked.

AcoPiper:

AcoPiper:
**[Author Round 10]**

- **[P2] Correct the shared primitive's scope claim — Fixed.** You are right, and the previous round only fixed half of it. `e35f3c4` named the two writers that stage nothing (`save_unseal_keys`, `eab::write_key_file`) but left the sentence claiming `publish_staged_blocking` is *the* staging implementation and that every staged production file reaches disk through it. `atomic_rewrite_owned_no_symlink` and `write_owned_impl` (behind `create_owned_credential_noclobber` and `write_owned_file_replace`) each build their own `NamedTempFile` and publish with `persist`/`persist_noclobber`, and they are reached from production: `rotate approle` writes the relocated `role_id` and rewrites the relocated `secret_id` through them, `service approle` creates both, and `service add` writes a relocated `eab.json` through the replace spelling.

  The rustdoc now calls it the crate's general-purpose staging publisher, scoped to the files at paths bootroot itself owns, and names the override credential trio as a second staged publish alongside it — including why the two stay apart rather than being folded together: those writers take ownership from the parent directory (a root process creating a file in the agent-owned directory would otherwise leave it unreadable to the non-root agent) or read it back through `symlink_metadata`, and refuse a name already present rather than replacing it. Neither policy generalises to `state.json` or the cert triple, so unifying them would make one of the two callers wrong. The paragraph about the unstaged pair stays, reworded to "stage nothing at all" and narrowed to the secrets-tree `eab.json`, since the relocated one is staged. `cert_group`'s `publish_staged` carried the same "the one staging implementation in the crate" phrase and is corrected to match.

  The PR description's corresponding section is updated the same way — heading and a new paragraph covering the override writers.

No production behaviour changes in this round; the change is documentation only. `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff, `cargo clippy --all-targets -- -D warnings` is clean, and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` is clean with the new intra-doc links resolving.

Head commit: `4c220a7`.

AcoPiper:
**[Reviewer Round 11]**

- **[P2] Correct the shared primitive’s scope claim (still unresolved).** The Round 10 change correctly scopes `publish_staged_blocking` itself, but [`src/fs_util.rs:491`](https://github.com/aicers/bootroot/blob/AcoPiper/issue-841/src/fs_util.rs#L491) still says that `atomic_write_blocking` is “one spelling of `publish_staged_blocking`, which every staged publish in the crate goes through.” That remains false for the production override credential writers the author identified: `create_owned_credential_noclobber`/`write_owned_file_replace` use `write_owned_impl`, and `atomic_rewrite_owned_no_symlink` stages independently. Please scope this sentence to the shared general-purpose publisher and its callers, as the revised rustdoc at `publish_staged_blocking` now does.

AcoPiper:
[Review Verdict Round 11: NOT_APPROVED]

AcoPiper:
**[Author Round 11]**

- **[P2] Correct the shared primitive's scope claim (still unresolved).** **Fixed.** Round 10 rewrote the rustdoc on `publish_staged_blocking` itself and left the cross-reference on `atomic_write_blocking` behind, so the two disagreed about which writers the guarantee covers — and the one that was still wrong is the one a caller reads first. `src/fs_util.rs:491` now names the callee as "the crate's general-purpose staging publisher" and points at it for the staged writers that publish outside it, instead of asserting that every staged publish in the crate goes through it. The claim it made is false for exactly the three production paths named at `publish_staged_blocking`: `create_owned_credential_noclobber` and `write_owned_file_replace` (through `write_owned_impl`) and `atomic_rewrite_owned_no_symlink`, which stage independently because they need ownership from the parent directory or read back through `symlink_metadata`, and a publish that refuses an existing name.

`atomic_replace_blocking`'s matching sentence was already scoped — it points at the callee for its two decisions without claiming coverage — and `cert_group`'s three references already say "general-purpose staging publisher", so neither changed. The PR description carried the same overclaim in its summary line and is updated to match.

`cargo clippy --all-targets -- -D warnings` and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` are clean, and `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff. The change is a doc comment only — no code, no test and no manual page is affected, and `CHANGELOG.md` records nothing a user of the last release could observe here.

Head is `8fa7c52`; CI is running.

PR #850, opened 2026-08-15
Require an openssl with x509 -ext in E2E (#844)

### LINKED ISSUE #844: Make the E2E harness's openssl requirement explicit
The E2E harness assumes the `openssl` it finds on `PATH` is OpenSSL. On macOS `/usr/bin/openssl` is LibreSSL, and both binaries are typically present — Homebrew's OpenSSL is only first on `PATH` if the shell was set up that way. When it is not, the harness fails in two unrelated places, and neither message points at the cause.

Note that `ensure_prerequisites` already checks `command -v openssl`. That check passes on such a host, so the run proceeds and dies later.

### 1. The fingerprint read returns nothing

`snapshot_cert_meta` writes the file with `openssl x509 -noout -serial -startdate -enddate -fingerprint -sha256`, and `fingerprint_of` reads it back with:

```sh
awk -F= '/^sha256 Fingerprint=/{print $2}' "$meta_file"
```

The two implementations disagree on exactly one line of that output:

```
$ /opt/homebrew/bin/openssl version   # OpenSSL 3.6.3
sha256 Fingerprint=90:1F:CA:37:...
$ /usr/bin/openssl version            # LibreSSL 3.3.6
SHA256 Fingerprint=90:1F:CA:37:...
```

Everything else — `serial=`, `notBefore=`, `notAfter=`, the hex itself — is byte-identical. Only the digest name's case differs, and the pattern anchors on the lowercase form. Under LibreSSL the read yields nothing, and the first `assert_fingerprint_changed` fails with:

```
[fatal][verify-after-responder-hmac] Missing fingerprint for edge-proxy/initial
```

The message is misleading twice over. `cert-meta/edge-proxy-initial.txt` is present and the fingerprint is in it, so nothing is missing. And the failure lands in the first phase that *compares* fingerprints, several phases after the one that wrote the file, so the phase name points away too.

### 2. `x509 -ext` does not exist in LibreSSL

`assert_stepca_ip_san` reads step-ca's SANs with `openssl x509 -noout -ext subjectAltName`. `-ext` arrived in OpenSSL 1.1.1 and LibreSSL 3.3.6 does not have it:

```
$ /usr/bin/openssl x509 -in cert.pem -noout -ext subjectAltName
unknown option -ext
```

The pipeline exits non-zero and the run fails with `[$label] failed to read step-ca's subjectAltName at ${STEPCA_BIND_ADDR}` — which reads as step-ca being unreachable or serving a bad certificate. This is matrix step 7.

## What was checked

Every `openssl` call in `scripts/` was run under both binaries. The two above are the only failures:

| Site | Under LibreSSL | Verdict |
| --- | --- | --- |
| `fingerprint_of` in `run-local-lifecycle.sh:565`, `run-remote-lifecycle.sh:483`, `run-ca-key-rotation-recovery.sh:318` | case does not match, yields empty | **broken** |
| `x509 -noout -ext subjectAltName` in `run-stepca-san.sh:298` | `unknown option -ext` | **broken** |
| `dgst -sha256 -hex` piped to `awk '{print $NF}'` in `run-remote-lifecycle.sh:728` | OpenSSL prints a `SHA2-256(stdin)= ` prefix, LibreSSL prints bare hex; `$NF` absorbs both | safe |
| `sed 's/^.*=//'` in `run-openbao-tls-reown.sh:471`, `:486`, `:498` | strips to the last `=`, case-independent | safe |
| Whole-line before/after comparison in `run-reinit-recovery.sh` and `run-stepca-san.sh` | both sides come from the same binary in the same run | safe |
| `req -addext`, `req -x509 -nodes -newkey`, `x509 -req -CAcreateserial`, `genrsa`, `rand`, `s_client -connect` | behave identically | safe |

The product itself is unaffected: nothing under `src/` shells out to `openssl`.

## Fix

**Check for the capability the harness uses, at the point it already looks.** The six scripts that run `command -v openssl >/dev/null 2>&1 || fail "openssl is required"` — `run-local-lifecycle.sh:130`, `run-remote-lifecycle.sh:132`, `run-ca-key-rotation-recovery.sh:131`, `run-reinit-recovery.sh:148`, `run-stepca-san.sh:157`, `run-openbao-tls-reown.sh:173` — should also confirm the binary they found supports `x509 -ext`:

```sh
openssl x509 -help 2>&1 | grep -qE '^[[:space:]]*-ext\b'
```

Probe the capability, not the implementation's name. A check for the string `OpenSSL` in `openssl version` would encode an assumption about which platforms are acceptable: it rejects a distribution that ships LibreSSL as its system `openssl` even if that build has `-ext`, and it would keep rejecting LibreSSL after a release that added the option. What the harness needs is the option, and that is what it can ask about.

The snippet above is the shape of the check rather than its final form; two details only surface against a real LibreSSL. Its `x509 -help` exits non-zero (3.3.6 exits 1, where OpenSSL 3.6.3 exits 0), so the probe has to capture the text and let the match alone decide — otherwise the exit status stands in for the answer, and a capable build that also exits non-zero from `-help` is rejected for the wrong reason. And `grep -q` exits on the first match, so under the `set -o pipefail` these scripts run with, the writer takes SIGPIPE and fails the very check that just passed: read the help into a variable and match it with a here-string rather than piping into `grep`.

**All six, and not only the one script that calls `-ext`.** `run-stepca-san.sh:298` is the sole caller. With the fingerprint read fixed below, the other five have no call that LibreSSL 3.3.6 cannot serve — every `openssl` invocation they make is in the safe list above. Gate all six anyway; this is a decision, not an oversight, and it should not be narrowed during implementation. The matrix runs its ten steps against one host, so a host that cannot run step 7 cannot run the matrix, and learning that at step 7 costs the six steps before it. A prerequisite that one script enforces and five do not is also a trap for whoever reads one of the five to find out what the harness needs.

The failure message should name the binary that was found, say which option it lacks, and say that a directory holding a capable `openssl` has to come first on `PATH`. It should not recommend a package manager: on the hosts where this triggers, a capable `openssl` is usually already installed and merely later on `PATH`, and which command would install one is not something the harness can know.

**Make the fingerprint read tolerate the spelling anyway**, at the three sites above:

```sh
awk -F= 'tolower($1) == "sha256 fingerprint" { print $2 }' "$meta_file"
```

This is worth doing independently of LibreSSL. OpenSSL's own digest naming is not stable across its subcommands: the same 3.6.3 build prints `sha256 Fingerprint=` from `x509 -fingerprint` and `SHA2-256(stdin)= ` from `dgst -sha256`. An exact lowercase anchor is fragile regardless of which implementation is in front of it. Matching `$1` exactly rather than loosening the regex still pins the digest, so a second fingerprint line added to `snapshot_cert_meta` later cannot silently satisfy the read. The hex contains no `=`, so `$2` is the whole value.

**Split the two failure modes in `assert_fingerprint_changed`**, which today reports both as "Missing fingerprint":

- the meta file does not exist — say so, with the path
- the file exists but no fingerprint line parsed — say so, and quote what the file holds, so the next format change names itself

## Deliberately not included

- **Replacing `-ext` with `-text` parsing.** The prerequisite probe establishes that `-ext` is available. A `-text` based extraction would trade a hard failure for a new dependency on another output format, which is the shape of problem this issue exists to reduce.
- **A shared prerequisite helper.** The six checks are already duplicated inline, and only two scripts source `lib/audit-log.sh`. Extracting a common prerequisites library is a separate refactor.
- **The scripts that call `openssl` and check for nothing.** `run-baseline.sh`, `run-rotation-recovery.sh`, `run-harness-smoke.sh` and `bootstrap-smoke-node.sh` call it without a prerequisite check of any kind, as do the two Python helpers that shell out to it, `generate-baseline-workspace.py` and `mock-openbao-server.py`. The six named above are the scripts that check, not all the scripts that call — read the list that way. None of these needs `-ext`, and everything they do call is in the safe list above, `req -addext` included: under LibreSSL 3.3.6 the SAN lands in the certificate. Giving them a check they have never had is a separate change.
- **A `CHANGELOG.md` entry.** This is test-harness only; a user of the last release cannot observe it.
- **Touching the sites listed as safe above.**

## Test plan

- With an `openssl` that lacks `-ext` first on `PATH` — LibreSSL 3.3.6 is the one at hand, reachable on macOS as `/usr/bin/openssl` — start the local lifecycle: it must fail inside `ensure_prerequisites`, before Docker is touched, with the new message. Today it runs for minutes and dies at `verify-after-responder-hmac`.
- With a capable `openssl` first on `PATH`, `scripts/impl/run-local-lifecycle.sh` still reaches the end unchanged. This is what CI already does, so CI behaviour must not move.
- Feed `fingerprint_of` a cert-meta fixture in each spelling and confirm both return the hex; feed one with no fingerprint line and confirm the new "parsed nothing" error rather than the missing-file one.
- `scripts/preflight/ci/e2e-matrix.sh --skip-hosts` reaches at least step 8. Step 9 (`run-openbao-tls-reown.sh`) needs non-interactive `sudo -n` to re-own `secrets/` and cannot run on a host without it; that step is CI's to verify.

## Acceptance criteria

- [x] The six prerequisite checks reject an `openssl` without `x509 -ext`, with a message naming the binary found, the missing option, and the `PATH` fix — and naming no package manager
- [x] The three fingerprint reads parse both spellings
- [x] `assert_fingerprint_changed` distinguishes a missing meta file from an unparsable one
- [x] The local lifecycle E2E completes unchanged under a capable `openssl`, and fails in `ensure_prerequisites` under one without `-ext`
- [x] No `CHANGELOG.md` entry
- [x] Full CI passes

sehkone:
Makes the E2E harness's `openssl` requirement explicit, makes the fingerprint read survive either spelling of the digest name, and pins both properties with a validator so CI can observe them.

Closes #844

## What changed

**The six prerequisite checks now probe for `x509 -ext`.** `run-local-lifecycle.sh`, `run-remote-lifecycle.sh`, `run-ca-key-rotation-recovery.sh`, `run-reinit-recovery.sh`, `run-stepca-san.sh` and `run-openbao-tls-reown.sh` each grew an `ensure_openssl` called from `ensure_prerequisites`, replacing the bare `command -v openssl` check. It captures `openssl x509 -help` and looks for the option, so a host whose first `openssl` cannot serve the SAN assertion is told so before Docker is touched:

```
openssl at /usr/bin/openssl does not support 'x509 -ext', which this harness needs to read a certificate's subjectAltName; put a directory holding an openssl that supports it first on PATH
```

It names the binary found, the missing option, and the `PATH` fix, and no package manager — on the hosts where this triggers, a capable `openssl` is usually already installed and merely later on `PATH`. It probes the capability rather than matching `OpenSSL` in `openssl version`, so a LibreSSL release that grows `-ext` passes. All six check it, not just the one caller, because the matrix runs every step against one host.

Two details worth flagging: `x509 -help` exits non-zero on LibreSSL 3.3.6, so the exit status is discarded and only the match decides; and the match is made with a here-string rather than a pipe, because `grep -q` exits on the first match and under `pipefail` the writer's SIGPIPE would fail a check that had just passed.

**The three fingerprint reads accept either spelling.** `awk -F= 'tolower($1) == "sha256 fingerprint" { print $2 }'` replaces the lowercase-anchored regex in `run-local-lifecycle.sh`, `run-remote-lifecycle.sh` and `run-ca-key-rotation-recovery.sh`. Matching the whole field still pins the digest, so a second fingerprint line added to `snapshot_cert_meta` later cannot satisfy the read. Each script also grew a `cert_meta_file` helper so the path is spelled once.

**`assert_fingerprint_changed` distinguishes the two ways the read comes up empty**, via a new `assert_cert_meta_readable`: an absent meta file names the path it looked for, and a file that parsed nothing quotes what the file holds, so the next format change names itself.

**A validator keeps all of it observable from CI.** Neither property is visible on a runner that already has a capable `openssl` and already spells the digest lowercase — a probe that stopped rejecting anything, or a read that went back to anchoring on one spelling, would leave the whole matrix green and break only on the kind of host the issue was reported from. `scripts/validate-e2e-openssl-compat.sh` extracts the harness's own functions out of the shipped scripts (rather than copying them) and drives them against stub `openssl` binaries and fixture snapshots: both digest spellings, a snapshot with no fingerprint line, an absent snapshot, an accepting stub, a rejecting stub, and a stub whose `-help` exits non-zero. It also asserts the gated set is exactly those six scripts, so a later change cannot quietly narrow the decision back to the single `-ext` caller. It needs no Docker and no second `openssl`, so it runs in the `check` job beside the other validators and in `scripts/preflight/run-all.sh`.

**Docs.** `docs/{en,ko}/e2e-ci.md` listed the two host tools the prerequisite blocks enforce; the `openssl` capability is now a third bullet, and the new validator is in the CI-equivalence table.

## Test plan

- [x] With `/usr/bin/openssl` (LibreSSL 3.3.6) first on `PATH`, `scripts/impl/run-local-lifecycle.sh` fails inside `ensure_prerequisites`, before any container starts, with the new message — instead of running for minutes and dying at `verify-after-responder-hmac`
- [x] All six gated scripts reject that binary the same way; with Homebrew's OpenSSL 3.6.3 first on `PATH`, `ensure_openssl` returns 0
- [x] `fingerprint_of` returns the hex from a cert-meta fixture spelled `sha256 Fingerprint=` and from one spelled `SHA256 Fingerprint=`, in all three scripts
- [x] A fixture with no fingerprint line yields the new "No SHA-256 fingerprint parsed from ..." error, and an absent file yields "Missing cert-meta file for ..." — the two are no longer reported alike
- [x] `./scripts/validate-e2e-openssl-compat.sh` passes
- [x] `./scripts/preflight/ci/check.sh` passes (`cargo fmt`/`clippy`/`doc`, ruff, biome, markdownlint, `mkdocs build --strict`, `cargo audit` with only the pre-existing allowed warning)
- [x] `bash -n` and `shellcheck` on the six scripts and the validator: no new findings
- [x] No `CHANGELOG.md` entry — this is test-harness only
- [ ] `scripts/preflight/ci/e2e-matrix.sh --skip-hosts` reaches at least step 8 — **not runnable on this host.** Re-confirmed during verification: step 1’s `infra install` aborts in its published-port preflight with `host port 127.0.0.1:8200 is already in use (Address already in use (os error 48)) (listener: pid 1255 (OrbStack Helper))`, before any container starts. The harness spells `8200`, `9000` and `8080` into its `--openbao-url`/step-ca/responder URLs literally, so `OPENBAO_HOST_PORT` cannot move the run off the conflict, and all three ports are held here. CI’s matrix exercises every touched path under a capable `openssl`.
- [x] Full CI green, including the local lifecycle completing unchanged under a capable `openssl` — all 18 required checks pass on `1e57e21`, including `Docker E2E (local-hosts)`/`(local-no-hosts)` and the whole ten-step matrix, plus a manually dispatched `E2E Extended` run on the same SHA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,6 +139,9 @@ jobs:
       - name: Validate Compose Instance Names
         run: ./scripts/validate-compose-instance-names.sh
 
+      - name: Validate E2E OpenSSL Compatibility
+        run: ./scripts/validate-e2e-openssl-compat.sh
+
       - name: Install cargo-audit
         run: cargo install cargo-audit || true
 
--- a/docs/en/e2e-ci.md
+++ b/docs/en/e2e-ci.md
@@ -570,6 +570,7 @@ Or run individual scripts:
 | `./scripts/preflight/ci/check.sh` | `ci.yml` Quality Check |
 | `./scripts/validate-deploy-compose.sh` | `ci.yml` Validate Deploy Compose |
 | `./scripts/validate-compose-instance-names.sh` | `ci.yml` Validate Compose Instance Names |
+| `./scripts/validate-e2e-openssl-compat.sh` | `ci.yml` Validate E2E OpenSSL Compatibility |
 | `./scripts/preflight/ci/deploy-no-build-smoke.sh` | `ci.yml` Deploy Compose No-Build Smoke |
 | `./scripts/preflight/ci/test-core.sh` | `ci.yml` Unit & CLI Smoke |
 | `./scripts/preflight/ci/e2e-matrix.sh` | `ci.yml` Docker E2E Matrix |
--- a/docs/en/e2e-ci.md
+++ b/docs/en/e2e-ci.md
@@ -596,7 +596,7 @@ When local `sudo -n` is unavailable:
 Use this only as a local constraint workaround. CI still executes
 `hosts` variants.
 
-The harnesses also assume two host tools, and each check fails in its
+The harnesses also assume three host tools, and each check fails in its
 prerequisite block rather than mid-run:
 
 - The bind-host guard in `run-reinit-recovery.sh`, `run-stepca-san.sh`,
@@ -610,6 +610,15 @@ prerequisite block rather than mid-run:
 - `run-remote-lifecycle.sh` reads the remote agent's TOML config with
   `tomllib`, so its `python3` must be 3.11 or newer. The prerequisite
   block imports it before any container starts.
+- The `openssl` the harness finds on `PATH` must support `x509 -ext`,
+  which `run-stepca-san.sh` reads step-ca's `subjectAltName` with. The
+  option arrived in OpenSSL 1.1.1, and LibreSSL 3.3.6 — what macOS
+  ships as `/usr/bin/openssl` — does not have it. The check probes for
+  the option rather than for the implementation's name and names the
+  binary it found; the fix is to put a directory holding a capable
+  `openssl` first on `PATH`. All six scripts that check for `openssl`
+  check this, not only the one that calls `-ext`, because the matrix
+  runs every step against one host.
 
 ## Init automation input/output rules
 
--- a/docs/ko/e2e-ci.md
+++ b/docs/ko/e2e-ci.md
@@ -555,6 +555,7 @@ CI 워크플로 동등 스크립트(`scripts/preflight/ci/`):
 | `scripts/preflight/ci/check.sh` | `ci.yml` → Quality Check |
 | `scripts/validate-deploy-compose.sh` | `ci.yml` → Validate Deploy Compose |
 | `scripts/validate-compose-instance-names.sh` | `ci.yml` → Validate Compose Instance Names |
+| `scripts/validate-e2e-openssl-compat.sh` | `ci.yml` → Validate E2E OpenSSL Compatibility |
 | `scripts/preflight/ci/deploy-no-build-smoke.sh` | `ci.yml` → Deploy Compose No-Build Smoke |
 | `scripts/preflight/ci/test-core.sh` | `ci.yml` → test-core |
 | `scripts/preflight/ci/e2e-matrix.sh` | `ci.yml` → test-docker-e2e-matrix |
--- a/docs/ko/e2e-ci.md
+++ b/docs/ko/e2e-ci.md
@@ -586,7 +586,7 @@ scripts/preflight/run-all.sh
 
 이는 로컬 제약 우회용입니다. CI에서는 `hosts` 케이스도 실행됩니다.
 
-하네스는 호스트 도구 두 가지를 전제하며, 각 검사는 실행 도중이 아니라
+하네스는 호스트 도구 세 가지를 전제하며, 각 검사는 실행 도중이 아니라
 사전 조건 블록에서 실패합니다.
 
 - `run-reinit-recovery.sh`, `run-stepca-san.sh`,
@@ -601,6 +601,16 @@ scripts/preflight/run-all.sh
 - `run-remote-lifecycle.sh`는 원격 에이전트의 TOML 설정을 `tomllib`으로
   읽으므로 `python3`이 3.11 이상이어야 합니다. 사전 조건 블록에서
   컨테이너를 띄우기 전에 import를 확인합니다.
+- `PATH`에서 찾은 `openssl`은 `x509 -ext`를 지원해야 합니다.
+  `run-stepca-san.sh`가 step-ca의 `subjectAltName`을 이 옵션으로 읽기
+  때문입니다. 이 옵션은 OpenSSL 1.1.1에서 추가되었고, macOS가
+  `/usr/bin/openssl`로 제공하는 LibreSSL 3.3.6에는 없습니다. 검사는
+  구현체 이름이 아니라 옵션 지원 여부를 확인하며 찾은 실행 파일 경로를
+  함께 알립니다. 해결 방법은 해당 옵션을 지원하는 `openssl`이 있는
+  디렉터리를 `PATH` 앞에 두는 것입니다. `-ext`를 실제로 호출하는
+  스크립트는 하나뿐이지만 `openssl`을 검사하는 여섯 스크립트가 모두 이
+  검사를 수행합니다. 매트릭스의 모든 단계가 같은 호스트에서 실행되기
+  때문입니다.
 
 ## init 자동화 입출력 규칙
 
--- a/scripts/impl/run-ca-key-rotation-recovery.sh
+++ b/scripts/impl/run-ca-key-rotation-recovery.sh
@@ -124,11 +124,34 @@ on_error() {
 # Infrastructure helpers
 # ---------------------------------------------------------------------------
 
+# Finding *an* `openssl` on PATH is not enough. macOS ships LibreSSL as
+# /usr/bin/openssl, and LibreSSL 3.3.6 has no `x509 -ext` — an OpenSSL
+# 1.1.1 addition, and how the harness reads a certificate's
+# subjectAltName. Probe for that option rather than for the string
+# `OpenSSL` in `openssl version`: the harness needs the capability, not
+# a particular implementation, and a build that grows the option should
+# pass. Every matrix script checks this even where it makes no `-ext`
+# call, because the ten steps run against one host: a host that cannot
+# serve the SAN step cannot serve the matrix, and learning that at that
+# step costs every step before it.
+ensure_openssl() {
+  local openssl_bin x509_help
+  openssl_bin="$(command -v openssl 2>/dev/null || true)"
+  [ -n "$openssl_bin" ] || fail "openssl is required"
+  # `x509 -help` exits non-zero on some builds (LibreSSL 3.3.6 does), so
+  # capture the text and let the match alone decide.  A here-string, not
+  # a pipe: `grep -q` exits on the first match, and under `pipefail` the
+  # writer's SIGPIPE would then fail the very check that just passed.
+  x509_help="$(openssl x509 -help 2>&1 || true)"
+  grep -qE '^[[:space:]]*-ext([[:space:]]|$)' <<<"$x509_help" || fail \
+    "openssl at ${openssl_bin} does not support 'x509 -ext', which this harness needs to read a certificate's subjectAltName; put a directory holding an openssl that supports it first on PATH"
+}
+
 ensure_prerequisites() {
   command -v docker >/dev/null 2>&1 || fail "docker is required"
   docker compose version >/dev/null 2>&1 || fail "docker compose is required"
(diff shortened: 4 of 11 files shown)


sehkone:
**[Reviewer Round 1]**

No findings. The change implements the issue’s full intended scope: each of the six existing prerequisite blocks probes the `x509 -ext` capability without treating `-help` exit status as the answer, the three readers accept both fingerprint spellings, and `assert_fingerprint_changed` now distinguishes a missing snapshot from an unparsable one.

The new validator exercises the shipped helper bodies with both capable and rejecting stubs, including the non-zero-help-status edge case, and uses fixtures to cover both spelling variants plus the two metadata failure modes. It also keeps the deliberate six-script gate from being narrowed accidentally. The PR body correctly closes #844, includes a test plan, and the thread contains no out-of-scope issue creation.

I ran `./scripts/validate-e2e-openssl-compat.sh`, `bash -n` on the changed shell scripts, and `git diff --check`; all passed.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Require an openssl with x509 -ext in E2E
```

**Body**

```text
The E2E harness assumed the `openssl` on `PATH` was OpenSSL. On macOS
`/usr/bin/openssl` is LibreSSL, and a run against it died twice over,
in places that named neither the cause nor the phase that caused it:
`x509 -ext` does not exist there, and `x509 -fingerprint` spells the
digest `SHA256 Fingerprint=` where the read anchored on the lowercase
form. The existing `command -v openssl` check passed on such a host,
so the run proceeded for minutes before failing.

Probe for the capability at the point the scripts already check, in
all six that check — not only `run-stepca-san.sh`, which is the sole
`-ext` caller. The matrix runs its ten steps against one host, so a
host that cannot serve step 7 cannot serve the matrix, and learning
that at step 7 costs the six steps before it. Probing for the option
rather than for `OpenSSL` in `openssl version` keeps the check about
what the harness needs, so a LibreSSL release that grows the option
passes.

Match the fingerprint field case-insensitively regardless. OpenSSL's
own digest naming is not stable across its subcommands — the same
3.6.3 build prints `sha256 Fingerprint=` from `x509` and
`SHA2-256(stdin)= ` from `dgst` — so an exact lowercase anchor was
fragile whichever implementation was in front of it. Comparing the
whole field still pins the digest.

Separate the two ways that read comes up empty. An unwritten snapshot
and one this read cannot parse are different problems, and reporting
both as "Missing fingerprint" sent the reader hunting for a file that
was there all along.

Neither property is observable on a CI runner, which already has a
capable `openssl` and already spells the digest lowercase: a probe
that stopped rejecting anything, or a read that went back to one
spelling, would leave the whole matrix green and break only on the
kind of host this was reported from. The validator extracts the
harness's own functions out of the shipped scripts and drives them
against stub binaries and fixture snapshots, and pins the gated set so
a later change cannot quietly narrow it back to the single `-ext`
caller. It needs no Docker and no second `openssl`, so it runs in the
`check` job beside the other validators.

No changelog entry: this is test-harness only, and a user of the last
release cannot observe it.

Closes #844
```

PR #851, opened 2026-08-15
Fix the create modes the staged publish widened (#848)

### LINKED ISSUE #848: Fix the create modes the staged publish widened
## Problem

#841 converted every production writer to publish by rename. A rename installs a fresh inode, so each converted site had to state a mode where the truncating write it replaced let the environment decide: an existing destination kept its own mode untouched, and a fresh create took `0666 & ~umask`. The conversion answered the rewrite half correctly — `fs_util::preserved_mode` reads the destination's mode back — but hardcoded the create half to `0o644`, which is the value a *default* umask produces, not what the umask on the running host produces.

On a host with a restrictive umask (`077` is common on hardened installs), every fresh create in this class now lands wider than it did before the conversion. The class is every `preserved_mode` caller with a `0o644` default:

- `state.json` — `STATE_FILE_MODE`, `src/state.rs:21`
- `.env` — `DOTENV_FILE_MODE`, `src/commands/dotenv.rs:97`
- `ca.json` and its patchers — `CA_JSON_FILE_MODE`, used from `src/commands/ca.rs:162`, `src/commands/init/steps/stepca_setup.rs:135`, `src/commands/init/steps/orchestrator.rs:1345`
- `openbao.hcl` — `OPENBAO_HCL_MODE`, `src/commands/init/steps/openbao_tls.rs:267`
- the compose overrides — `COMPOSE_OVERRIDE_MODE`, used from `src/commands/guardrails.rs:367`, `src/commands/init/steps/responder_setup.rs:188`, `src/commands/init/steps/openbao_setup.rs:916`
- `init`'s rollback restore — `ROLLBACK_FILE_MODE`, `src/commands/init/steps.rs:429`

For most of the class this is mild information disclosure — `state.json` is an inventory of services, paths and role names, and its own doc comment records that it carries no secret. **`.env` is different**: `infra install` writes `POSTGRES_PASSWORD` into it (`src/commands/infra.rs:988`), and the file sits in the compose directory, outside the `0700` secrets tree. A host whose umask kept that file at `0600` gets a world-readable database password from its next fresh install, and `CHANGELOG.md` records the change only as "a host running a non-default umask is the only one that can observe the difference" — without noting that one of the files carries a credential.

Nothing here has shipped. The last release is 0.2.0 and the whole conversion sits in `[Unreleased]`, so both fixes below are corrections in place — the changelog entry is edited, not answered with a new one.

### 1. Give the publisher a mode policy instead of a bare `u32`

Replace the `mode: u32` parameter of `publish_staged_blocking` and its eight wrappers with an enum, and absorb `preserved_mode` into it:

```rust
pub enum StagedMode {
    /// The file has a mode policy: apply exactly this, on create and on rewrite.
    Policy(u32),
    /// No policy. An existing destination keeps its own mode; a fresh create
    /// takes what open(2) gives 0o666 under the process umask — exactly what
    /// the truncating write this machinery replaced produced.
    PreserveOrUmask,
    /// No policy on rewrite, but a stated create mode: an existing destination
    /// keeps its own mode; a fresh create takes exactly this.
    PreserveOrCreate(u32),
}
```

Call-site mapping:

- The `0o644`-default class above → `PreserveOrUmask` (except `.env`, which commit 2 decides)
- The three `preserved_mode` callers that default to `KEY_FILE_MODE` — the `agent.toml` editors in `src/commands/service.rs:1152` and `src/commands/service/remove.rs:447`, and `src/commands/init/steps/http01_admin_tls.rs:270` — → `PreserveOrCreate(0o600)`, which is byte-identical to their behavior today. Do **not** move these to `PreserveOrUmask`: their create default is already narrower than any umask outcome, and widening it would be a new regression.
- Every other caller → `Policy(<its existing constant>)`, a mechanical wrap.

Mechanism for the `PreserveOrUmask` create arm: the staged temporary must get its mode from **open(2) at creation**, via `tempfile::Builder::permissions(Permissions::from_mode(0o666))` — `tempfile`'s Unix implementation passes this through `OpenOptions::mode`, so the kernel applies the umask (verified against tempfile 3.27.0, `src/file/imp/unix.rs:23`). A post-create `set_permissions` cannot express this: chmod ignores the umask. On the preserve and exact arms the existing create-at-0600-then-chmod flow is unchanged, as is the fsync-after-ownership-and-mode ordering.

`fs_util::preserved_mode` has no callers left after this and is deleted; its doc comment's reasoning moves onto the enum.

### 2. Create `.env` at `0o600`

Change `.env`'s create mode to `PreserveOrCreate(0o600)` — not `PreserveOrUmask` — and record it under `### Security` in the changelog.

The decision and its grounds, so they are on the record: every reader of this file in the deployed ecosystem runs as root. `aicers/bootler` invokes every bootroot command as root, runs `docker compose up`/`down` against this compose directory itself as root, and generates the rotation and infra systemd units with no `User=` line; its RFC 0003 §7.3 records bootroot as root-invoked by both the CLI and the timers. The non-root `clumit-*` agents read `agent.toml`, certificates and relocated credentials — never `.env`. Two properties bound the residual risk for standalone deployments: an operator who needs the file wider can widen it once and the preserve arm keeps that choice across every later rewrite, and a reader that loses access fails loudly at the next compose invocation through the `POSTGRES_PASSWORD:?…` guard rather than silently.

The changelog's `### Fixed` sentence claiming `0644` create defaults for the class is corrected in place by commit 1 (the class now follows the umask) and this commit (`.env` is `0600`); the manuals' **How bootroot writes files** sections are updated where they describe create modes, in both `docs/en/cli.md` and `docs/ko/cli.md`.

## Constraints

- The policy constants are untouched and never umask-dependent: `KEY_FILE_MODE` (`0600`), `CERT_FILE_MODE` (`0644`), `CA_BUNDLE_FILE_MODE` (`0644`), `SECRET_OUTPUT_FILE_MODE` (`0600`), `ROTATION_STATE_MODE` (`0600`), fast-poll `STATE_FILE_MODE` (`0600`). The permission window #841 closed stays closed; a secret file's mode must not vary with the environment.
- No change to symlink resolution, ownership (`StagedOwner`), durability (`StagedDurability`), or what is written; this issue is the mode axis only.
- `save_unseal_keys` and `eab::write_key_file` stay untouched (#840 owns them; when converted they take `Policy(0o600)`).
- Umask manipulation in tests follows the existing in-tree precedent — `write_root_token_file_creates_with_0600_under_permissive_umask` in `src/commands/init/steps/orchestrator.rs` sets and restores via `libc::umask` with a `SAFETY` comment. Do not add a shared `Mutex` around it.
- #847 rewrites the same wrapper signatures (symlink axis). This issue lands first and #847 rebases over the new `StagedMode` parameter; if #847 lands first, the mapping above applies to its `Destination`-taking signatures instead. Either order, neither blocks the other.

## Acceptance criteria

- Under umask `077`, a fresh `state.json`, `ca.json` patch target, `openbao.hcl`, compose override and rollback restore land at `0600`; under `022` they land at `0644`. An existing destination keeps its mode in both cases. Pinned at the primitive for all three enum arms, plus one caller-level test for `state.json`.
- A fresh `.env` lands at `0600` under any umask; an existing `.env` keeps its mode. Both pinned.
- The `PreserveOrUmask` create arm applies its mode at open time: a test creating under umask `077` observes `0600` on the temp-created file with no wider transient (asserting the final mode is sufficient — no chmod call exists on that arm to produce a transient).
- `fs_util::preserved_mode` no longer exists; no caller passes a bare mode where a policy decision is not intended.
- `CHANGELOG.md`: the `0644`-create sentence is corrected in place, and a `### Security` entry records the `.env` narrowing. No new `### Changed` entry — no released user saw the interim form.
- Both manuals describe the create-mode behavior as it now is.
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` clean, full CI green including the Docker E2E matrix.

## Test plan

- `cargo test` green; new tests as above.
- `scripts/preflight/ci/check.sh` end to end.
- `scripts/preflight/ci/e2e-matrix.sh` through the steps the host supports — this touches every `init` and rotation write path, even though only creates on non-default-umask hosts can observe a difference.

## Not in this issue

- The symlink axis (#847), the two unstaged writers (#840).
- Widening or narrowing any policy-mode file.
- ACL/xattr preservation and the documentation gaps the conversion left; those are the record-correction issue's, which lands after this one and restructures the same changelog entry.

sehkone:
## Summary

`publish_staged_blocking` and its eight wrappers take a `StagedMode` instead of a bare `u32`, so every caller states where its mode comes from rather than naming a number that reads as a decision. The `0644` create defaults #841 hardcoded for the policy-free class are replaced by `PreserveOrUmask`, which reproduces what the truncating write got from the environment; `.env` takes `PreserveOrCreate(0600)` instead, because it carries `POSTGRES_PASSWORD` and sits outside the `0700` secrets tree.

### 1. Give the publisher a mode policy

```rust
pub enum StagedMode {
    Policy(u32),           // apply exactly this, create and rewrite alike
    PreserveOrUmask,       // destination's mode, or 0666 under the umask on a create
    PreserveOrCreate(u32), // destination's mode, or exactly this on a create
}
```

Call-site mapping is the one the issue records: the `0644`-default class → `PreserveOrUmask`; the three `agent.toml`/responder editors that defaulted to `KEY_FILE_MODE` → `PreserveOrCreate(0o600)`, byte-identical to today; everything else → `Policy(<its existing constant>)`. `fs_util::preserved_mode` has no callers left and is deleted, its reasoning moved onto the enum.

The `PreserveOrUmask` create arm is answered at `open(2)`, via `tempfile::Builder::permissions(Permissions::from_mode(0o666))` — `tempfile` passes that straight to `OpenOptions::mode`, so the kernel subtracts the umask. A post-create `set_permissions` cannot express this, since `chmod` ignores the umask. On the other two arms the create-at-`0600`-then-chmod flow, the chown-before-chmod ordering and the fsync-last ordering are all unchanged.

The now-unused constants go with their last caller: `STATE_FILE_MODE` (state.rs), `CA_JSON_FILE_MODE`, `OPENBAO_HCL_MODE`, `COMPOSE_OVERRIDE_MODE`, `ROLLBACK_FILE_MODE`.

### 2. Create `.env` at `0600`

`bootroot infra install` generates `POSTGRES_PASSWORD` into `.env`, which lives in the compose directory rather than the secrets tree, so a create left to the umask handed a default-umask host a world-readable database password. Only the create is decided; the preserve arm is unchanged, so an operator who widens the file keeps that choice across every later rewrite.

The grounds, on the record: every reader of this file in the deployed ecosystem runs as root — `bootler` invokes every bootroot command as root, runs `docker compose` against this directory itself as root, and generates its rotation and infra units with no `User=`. The non-root `clumit-*` agents read `agent.toml`, certificates and relocated credentials, never `.env`. A reader that does lose access fails loudly at the next compose invocation through the `POSTGRES_PASSWORD:?…` guard rather than silently.

### Two follow-ups on top of those

- **A umask test gets a process of its own.** `umask` is a property of the process, not of a thread, so the window a test opens by setting one covers every file created by every other test the harness scheduled beside it — tests that take no lock the setter could hold. `fs_util::umask_test_ran_in_child` re-executes the calling test binary for the one named test, so each of the two umask-setting tests (`fs_util`'s arm coverage and `state.json`'s caller-level pin) runs with the process to itself. The name it is handed is held to naming exactly one test, because libtest calls a run of zero tests a success. The root-token test in `orchestrator.rs` no longer sets a umask at all; the property it asserted (that its `0600` is the writer's decision, not a umask that agreed) is now pinned at the primitive by the `Policy` arm publishing under umask `000`, and its doc comment records where it went.
- **An existing install has to narrow its own `.env`.** Only the create half changed, so an install made before this keeps the mode its `.env` was created with — `0644` on a default umask, database password included — and nothing bootroot writes later narrows it. The changelog and both manuals now say so. The manuals also described the preserve-or-umask group as "everything else", which is wider than the code (`agent.toml`, the OpenBao Agent configs and the HTTP-01 responder config all carry a `0600` policy); the group is now named by what it is and its members listed as the whole of it.

## Tests

- `fs_util`: one test walks all three enum arms across a create and a rewrite, pinning `umask 077 → 0600` and `umask 022 → 0644` on the `PreserveOrUmask` create, `0600` regardless of umask on the `PreserveOrCreate` create, and `Policy` overriding both the umask and a mode left on the destination. It is deliberately **one** test rather than three, for the per-process-umask reason above — one isolated process, every umask-dependent publish in the crate sequenced through it.
- `state.rs`: the caller-level pin — a fresh `state.json` lands `0600` under `umask 077` and `0644` under `022`; the existing test that a narrowed destination stays narrowed is unchanged.
- `commands::dotenv`: a fresh `.env` lands `0600`; the existing mode-preservation test now seeds `0640` so it pins that a rewrite does not re-*narrow* an operator-widened file.

## Test plan

- [x] Under umask `077` a fresh `state.json`, `ca.json` patch target, `openbao.hcl`, compose override and rollback restore land at `0600`, and at `0644` under `022`; an existing destination keeps its mode either way — pinned at the primitive for all three arms, plus the caller-level pin for `state.json`
- [x] A fresh `.env` lands at `0600` under any umask, and an existing `.env` keeps its mode — both pinned
- [x] The `PreserveOrUmask` create arm applies its mode at open time, observed as `0600` under umask `077` with no chmod on that arm to produce a wider transient
- [x] `fs_util::preserved_mode` no longer exists, and no caller passes a bare mode where a policy decision is not intended
- [x] The policy constants are untouched and never umask-dependent; `save_unseal_keys` and `eab::write_key_file` are untouched
- [x] `CHANGELOG.md` corrects the `0644`-create sentence in place and records the `.env` narrowing under `### Security`, with no new `### Changed` entry
- [x] Both manuals describe the create-mode behaviour as it now is, including what an existing install must do
- [x] `cargo test` green (35 test binaries)
- [x] `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff; `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` clean
- [x] `scripts/preflight/ci/check.sh` end to end, including `mkdocs build --strict` and markdownlint
- [x] The real binary, not just the unit tests: `bootroot infra install` under `umask 077` and again under `umask 000` writes `.env` at `0600` both times
- [x] Full CI green on this head, including all ten Docker E2E matrix arms and `run-extended` from `e2e-extended.yml` (dispatched by hand — that workflow has no PR trigger)

The two Docker preflight scripts (`scripts/preflight/ci/e2e-matrix.sh`, `scripts/preflight/ci/e2e-extended.sh`) could not run on this host: `127.0.0.1:8200` is held by an unrelated OrbStack VM and the lifecycle harness hardcodes `:8200` in every `--openbao-url` and health probe. Both ran in CI against this exact commit, which is where the assertion belongs anyway — only a create on a non-default-umask host can observe a difference, and CI runs the default `022`.

Closes #848


sehkone:
**[Reviewer Round 1]**

- [P1] Isolate the umask assertions in a child process instead of mutating it from these `#[test]` functions. `src/fs_util.rs:1623` and `src/state.rs:400` are both unit-test modules of the library, so Cargo builds them into the same lib test executable and the test harness may run them concurrently. Each changes the process-global umask (at `src/fs_util.rs:1631` and `src/state.rs:407`), despite the comments claiming it is the only such test in its binary. Removing the orchestrator test leaves this race intact; moreover, either window can affect any unrelated test in that executable which creates a file. This can make the expected `0600`/`0644` assertions flaky and gives unrelated file tests a nondeterministic umask. Keep the issue's no-shared-`Mutex` constraint, but run the umask-changing coverage in a subprocess (or another genuinely process-isolated test helper) so the parent test process never changes its umask.

sehkone:

sehkone:
**[Author Round 1]**

- **[P1] Isolate the umask assertions in a child process — partially addressed (the fix is in; the premise it was argued from is not quite right).**

  **Fixed.** The isolation is now real, in 07e7061. `fs_util::umask_test_ran_in_child` re-executes the calling test binary for the one named test, so each umask-setting test runs with the process to itself:

  ```rust
  #[test]
  fn save_creates_a_new_state_file_under_the_process_umask() {
      if fs_util::umask_test_ran_in_child(
          "state::tests::save_creates_a_new_state_file_under_the_process_umask",
      ) {
          return;
      }
      // …the umask window, in a process that is running nothing else
  }
  ```

  The name it is handed is held to naming exactly one test rather than trusted: `--exact` on a name that has drifted from the function it was copied from matches nothing, and libtest reports a run of *zero* tests as a success — verified on this binary, `--exact no::such::test` prints `0 passed` and exits `0`. So the helper asserts `test result: ok. 1 passed`, not the exit status. A deliberately mistyped name fails the parent with the drift message, which is how I checked it.

  This keeps the issue's no-shared-`Mutex` constraint, and it is also what `AGENTS.md` asks for — environment-dependent behaviour tested through a child process rather than by mutating this one. No shared lock could have closed the window anyway: the tests it races take no lock to hold.

  **Pushed back, on one factual point.** `src/fs_util.rs` and `src/state.rs` are not built into the same test executable, so those two tests were not racing *each other*. `fs_util` is `pub mod fs_util` in `src/lib.rs`; `state` is `mod state` in `src/main.rs` and reaches the other through `use bootroot::fs_util`. Cargo builds them into the lib test binary and the bin test binary respectively:

  ```
  $ cargo test --lib -- --list | grep umask
  fs_util::tests::staged_mode_arms_answer_the_create_and_the_rewrite: test
  $ cargo test --bin bootroot -- --list | grep umask
  state::tests::save_creates_a_new_state_file_under_the_process_umask: test
  ```

  The doc comments' "the only such test in its binary" claim was therefore accurate as written, and removing the orchestrator test's umask did close the race it was aimed at — the second umask setter in the bin binary.

  What it did *not* close is the rest of the review item, which is the part that stands and which the fix above is for: one setter per binary still leaves the window open over every *unrelated* test in that binary that creates a file. In the bin binary that is essentially all of `commands`. So the conclusion was right even though the two-tests-in-one-executable reasoning behind it was not, and the correction is in the doc comments too — they now give process isolation as the reason rather than the per-binary count.

**Verification on this head**

- `cargo test` green: 424 lib + 1044 bin + every integration binary, `0 failed`.
- `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff; `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` clean.
- `scripts/preflight/ci/check.sh` end to end, including `mkdocs build --strict` and markdownlint.

No changelog or manual change: this round is test isolation only, with no effect on what any command writes or on any mode a user can observe. The PR description's bullet on the umask, which described the superseded per-binary approach, has been rewritten to match.

sehkone:
**[Reviewer Round 2]**

Round 1 P1 is resolved. `fs_util::umask_test_ran_in_child` re-executes each named umask-setting test with `--exact` in a child process, and the state/primitive tests only change the umask in that child. The parent test harness therefore never observes the process-wide mutation.

No remaining findings. The updated mode-policy mapping, `.env` create/rewrite behavior, tests, documentation, and PR hygiene satisfy #848.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Give the staged publish a mode policy
```

**Body**

```text
Publishing by rename installs a fresh inode, so every converted
writer had to state a create mode where the truncating write it
replaced let the environment decide. That half was answered with a
hardcoded 0644 — the value a *default* umask produces, not the umask
of the host the program runs on — so on a host with a restrictive
umask every fresh create in that class landed wider than it did
before the conversion.

The mode parameter of publish_staged_blocking and its eight wrappers
is now a StagedMode rather than a bare u32, so each call site states
where its mode comes from: Policy for a file whose mode is a
property of what it holds, PreserveOrUmask for one whose mode nobody
decided, PreserveOrCreate for one that preserves an existing mode
but states its create. fs_util::preserved_mode is absorbed into the
enum and deleted. The umask arm takes its mode from open(2), through
tempfile's permissions builder, because chmod ignores the umask and
so no post-create call can express it.

.env is the one file in the otherwise policy-free class with a
stated create mode. infra install writes POSTGRES_PASSWORD into it
and it sits in the compose directory rather than the 0700 secrets
tree, so a create left to a default umask handed the host a
world-readable database password. Only the create is decided: an
operator who widens the file keeps that choice across every later
rewrite, and an install made before this keeps the mode its .env was
created with — which the changelog and both manuals now say to
narrow by hand.

Each of the two tests that set a umask re-executes the test binary
for its one test. umask is a property of the process, so the window
one opens otherwise covers every file created by every test the
harness scheduled beside it, and no lock the setter could hold
closes it.

Closes #848
```

PR #852, opened 2026-08-15
Fail an E2E run that finds containers left behind (#845)

### LINKED ISSUE #845: Fail when the E2E teardown leaves containers behind
The E2E harness tears the compose stack down at the start of a run and again on exit, and it cannot tell you whether either one worked. `compose_down` sends every byte to `/dev/null` and swallows the exit status:

```sh
compose_down() {
  docker compose -p "${COMPOSE_PROJECT_NAME:-bootroot}" -f "$COMPOSE_FILE" -f "$COMPOSE_TEST_FILE" down -v --remove-orphans >/dev/null 2>&1 || true
}
```

Ten scripts define that function, all in that shape: `run-local-lifecycle.sh:197`, `run-remote-lifecycle.sh:138`, `run-ca-key-rotation-recovery.sh:171`, `run-baseline.sh:77`, `run-rotation-recovery.sh:95`, `run-harness-smoke.sh:83`, `run-openbao-tls-reown.sh:152`, `run-openbao-tls-no-delta.sh:141`, `run-reinit-recovery.sh:136`, and `run-stepca-san.sh:130`. An eleventh, `run-two-instance-isolation.sh`, defines no `compose_down` at all; its teardown is `teardown_instance` at `:897`, whose `down` on `:900` discards output the same way — but it is the one script that already asserts the result afterwards, and the rest of this issue turns on that. A teardown that removed nothing is indistinguishable from one that removed everything, and the run proceeds either way.

## What this costs

A run that is killed hard never reaches its own cleanup. `trap cleanup EXIT` does run on `SIGTERM` — that was checked, not assumed — but nothing survives `SIGKILL`, and the harness is routinely run detached from a supervisor that may send one.

This is not hypothetical. Two runs were interrupted the same way on the same host: one logged its `cleanup` phase and left nothing behind, the other logged no `cleanup` at all and left six containers up — `bootroot-openbao`, `bootroot-ca`, `bootroot-postgres`, `bootroot-http01`, `bootroot-openbao-agent-stepca`, `bootroot-openbao-agent-responder`. They had to be removed by hand before anything else could run, because the container names are `${BOOTROOT_INSTANCE:-bootroot}-*` and are therefore unique per daemon.

The start-of-run `compose_down` is supposed to be the defense against exactly that, and usually is. But it only knows the project name and compose files it was given, and **the project name is no basis for a start-of-run check** — in either of the two positions a script can be in.

Which position a script is in depends on how it is invoked, and the wiring is not uniform. `scripts/preflight/ci/e2e-matrix.sh` derives a project name from `RUN_ID` (`GITHUB_RUN_ID`, or a timestamp locally) for every step, but it passes it under two different variable names, and three scripts read the other one:

| Script | Handed by the suite | Read by the script | Effective project |
| --- | --- | --- | --- |
| `run-reinit-recovery.sh`, `run-stepca-san.sh`, `run-openbao-tls-no-delta.sh`, `run-openbao-tls-reown.sh` | `COMPOSE_PROJECT_NAME` | `COMPOSE_PROJECT_NAME` | fresh per run |
| `run-rotation-recovery.sh` | `PROJECT_NAME` | `PROJECT_NAME` | fresh per run |
| `run-local-lifecycle.sh`, `run-remote-lifecycle.sh` | `PROJECT_NAME` | `COMPOSE_PROJECT_NAME` | always `bootroot` |
| `run-ca-key-rotation-recovery.sh` | `PROJECT_NAME` (from `run-extended-suite.sh:112`) | `COMPOSE_PROJECT_NAME` | always `bootroot` |
| `run-baseline.sh`, `run-harness-smoke.sh` | not invoked by either suite | `PROJECT_NAME`, defaulted to `…-$$` | fresh per run |

Neither column supports a label-scoped start-of-run check. Where the project is fresh per run, the leftovers from a killed run carry a `com.docker.compose.project` label no later run will ever ask about. Where it is the literal `bootroot`, the label is shared with a developer's real default-identity install, so a sweep driven off it would tear down exactly what the safety contract below forbids touching. What is common to both is that *container names* are not project-scoped and collide with the next run's on `up`, which is why the check is by name.

The mismatched variable in those three scripts is a real defect and is called out here because it is what makes the table look arbitrary, but repairing the wiring is not this issue's job — see #846, which needs run-scoped projects for the lifecycle pair for its own reasons. Nothing below depends on which row a script is in.

And when the teardown does fail, it says nothing, so the run continues into a stack it does not own and fails later somewhere unrelated.

## What already exists

`run-two-instance-isolation.sh` already implements the assertion this issue is asking for. `assert_no_leftovers` at `:913` checks containers and volumes by `com.docker.compose.project` label, and `cleanup` at `:938` reports leftovers and folds them into the run's status — it captures `local status=$?` on entry, sets `status=1` on a leftover, and ends with `exit "$status"`. That is the end-of-run half of the fix, already written and already exercised in CI.

It is not quite the shape the fix below asks for, though. Assigning `status=1` overwrites the entry status rather than preserving it, so a run that failed with any status other than 1 and then found leftovers reports the leftover check's 1 in place of its own reason. Extracting it is the moment to correct that.

The work here is to lift it into the shared harness rather than to reinvent it. `scripts/impl/lib` is where it belongs.

## The naming constraint this has to respect

That script's header states a safety contract at `:39-45`:

> Every teardown action is scoped to the run's own instance names and their `com.docker.compose.project` labels — never a `bootroot-*` wildcard — so a co-located default install is untouched.

The contract holds there because that script is the only one that claims run-scoped identities: it passes `infra install --instance-name` (`:386`) with names derived from a per-run token (`twoinst-a-<token>`, `:87`). A `bootroot-*` wildcard would reach past its own instances and into a developer's real default install.

The other ten scripts are in the opposite position. None of them passes `--instance-name`, so all ten run at the default identity and their containers are named `bootroot-*`. Two of them cannot coexist on one host regardless of what this issue changes, and neither can one of them and a real default install — they collide on `container_name` at `up`. Today that collision surfaces as a confusing failure some minutes in. The point of the start-of-run check is to surface it immediately, and to say which of the two situations it is.

So the check is split by which position the script is in, and neither half uses a prefix wildcard:

- **The ten default-identity scripts** check for the exact container names bootroot creates at the resolved instance name. The set is finite and already enumerated in code — `BootrootContainer::ALL` in `src/commands/container_name.rs:61` — nine suffixes: `-openbao`, `-postgres`, `-ca`, `-http01`, `-prometheus`, `-grafana`, `-grafana-public`, `-openbao-agent-stepca`, `-openbao-agent-responder`. Exact names, not `^bootroot-`.
- **`run-two-instance-isolation.sh`** keeps its existing label-scoped check unchanged. It is excluded from the name check: its containers are not named `bootroot-*`, so the check would be both wrong and redundant there.

### Resolving the instance name

Do not read `${BOOTROOT_INSTANCE:-bootroot}` out of the script's own environment. That is not the value that names the containers. `resolve_recorded_instance_name` (`src/commands/compose_project.rs:155`) reads the instance name from the compose directory's `.env` and **deliberately ignores `COMPOSE_PROJECT_NAME`** — the doc comment at `:148-154` says so, on the grounds that the override selects a project for one invocation and is not an identity. The two are not interchangeable in the other direction either: `resolve_compose_project_for_dir`'s doc at `:123-127` notes that the E2E harness's per-run project names are deliberately left unvalidated because they are longer than an instance name may legally be. The harness scripts never export `BOOTROOT_INSTANCE`, and `run-two-instance-isolation.sh` actively unsets it (`:69`).

The shell check must resolve the instance name the same way the binary does: `--instance-name` if given, else `BOOTROOT_INSTANCE` recorded in the compose directory's `.env`, else the literal `bootroot`. A bare shell expansion of the ambient variable gives the right answer today only because all ten scripts happen to run at the default identity. That is a coincidence, not a derivation, and it breaks the moment one of them takes an instance name.

## Fix

**Stop discarding the teardown's output.** Send it to the run log instead of `/dev/null`, in all ten `compose_down` definitions and in `teardown_instance`. Keep the non-fatal behaviour only where it is deliberate — a teardown at the start of a run may legitimately find nothing to do.

**Extract the existing assertion into `scripts/impl/lib`.** Both halves: the leftover check and the status-preserving `cleanup` shape. `run-two-instance-isolation.sh` then calls the shared version instead of its own copy. Its behaviour is unchanged but for the entry-status overwrite noted above, which the shared shape has to get right for every caller that adopts it.

The extracted checks abort through `fail`, and the label-scoped half reports a satisfied assertion through `pass`. Both stay the caller's to define, because each harness spells a fatal report its own way — some carrying a phase and a log tail with it. That makes them a contract the library states rather than an implementation it carries, and `run-baseline.sh`, `run-rotation-recovery.sh`, and `run-harness-smoke.sh` — none of which defines a `fail` today — each need one.

**Assert at the start of a run, by exact container name.** In the ten default-identity scripts, before any work: resolve the instance name as above, and fail if any of the nine container names exists. The message must name the containers found, give the `docker rm -f` line that removes them, and state both possibilities — a leftover from a killed run, or a real install on this host — rather than asserting it is garbage.

**Assert at the end of a run.** Anything still listed is a failure — a harness that leaves its own garbage behind is broken — but it must not overwrite an earlier failure. If the run already failed, report the leftovers and keep the original error as the reason. This is what `cleanup` at `:938` already does for a run that had passed, and what it has to start doing for one that had failed.

**Add the start-of-run teardown where there is none.** `run-baseline.sh:357`, `run-harness-smoke.sh:133`, and `run-rotation-recovery.sh:385` call `compose_down` only from `cleanup`; they have no start-of-run teardown at all. The check is worth little without one, so those three need the call site added as well as the assertion.

`run-rotation-recovery.sh` has a wrinkle the other two do not. It writes the repository-root `.env` from `compose_up`, so at the new call site that file does not exist yet — and Compose interpolates the whole file on any subcommand, `down` included, aborting before it removes anything when a referenced variable is unset. That teardown has to supply those variables itself, scoped to the call so they cannot override what `init` rewrites mid-run.

## Deliberately not included

- **Trapping `SIGKILL`.** It cannot be trapped. The start-of-run assertion is the defense, and it is the one that matters, because it is what turns a dirty host into a clear message instead of a confusing failure twenty minutes later.
- **Changing what kills the run.** Whatever supervises the harness is outside this repository.
- **A prefix or wildcard match on container names.** Ruled out above; the exact-name check is strictly narrower and catches the same leftovers.
- **Checking networks.** The extracted helper checks containers and volumes, because that is what `assert_no_leftovers` checks today. `teardown_instance` removes networks but never asserts them, and widening the assertion is a separate change with its own evidence to gather.
- **Repairing the `PROJECT_NAME` / `COMPOSE_PROJECT_NAME` mismatch.** Three scripts are handed a per-run project name under a variable they do not read, as tabled above. It is a real defect, but fixing it is a change to how those scripts are invoked, it belongs with the work that actually needs run-scoped projects (#846), and it would not remove the need for this check: a fresh project name is what *hides* a killed run's leftovers from the next run's label-scoped teardown.
- **A `CHANGELOG.md` entry.** Test-harness only; a user of the last release cannot observe it.

## Test plan

- Start `scripts/impl/run-local-lifecycle.sh`, wait until the stack is up, and `kill -9` it. Confirm containers remain. Start it again: it must fail immediately, before doing any work, naming the leftover containers and the command that removes them — and after that command runs, a third start must proceed normally.
- Repeat with a *different* `COMPOSE_PROJECT_NAME` on the second start, so the leftovers carry a project label the new run never asks about. It must still fail, which is what proves the check is not label-scoped.
- Run `scripts/impl/run-two-instance-isolation.sh` with a default-identity `bootroot` stack up on the same host. It must pass, untouched by the new check and leaving that stack running.
- Make the teardown fail (for example by holding a container with `docker pause`) and confirm the run reports it instead of continuing silently.
- Force a mid-run failure in a script whose cleanup then finds leftovers, and confirm the exit status and the reported reason are the original failure's, not the leftover check's.
- A normal run's teardown output appears in the run log, and a run on a clean host is unchanged in exit status and in what it does.
- `scripts/preflight/ci/e2e-matrix.sh --skip-hosts` reaches at least step 8. Step 9 (`run-openbao-tls-reown.sh`) needs non-interactive `sudo -n` and is CI's to verify.

## Acceptance criteria

- [x] Teardown output reaches the run log in all ten `compose_down` definitions and in `run-two-instance-isolation.sh`'s `teardown_instance`
- [x] The leftover check and the status-preserving `cleanup` shape live in `scripts/impl/lib`, and `run-two-instance-isolation.sh` uses the shared version, unchanged but for no longer overwriting the status of a run that had already failed
- [x] A run started on a host carrying any of the nine container names at its resolved instance name fails before doing work, naming them and the command that removes them
- [x] That failure still occurs when the leftovers carry a `com.docker.compose.project` label from an earlier, differently named run
- [x] The check matches exact container names derived from the resolved instance name — never a prefix or wildcard
- [x] The instance name is resolved as `resolve_recorded_instance_name` resolves it: `--instance-name`, then the compose directory's `.env`, then `bootroot` — not from the ambient `BOOTROOT_INSTANCE`
- [x] `run-baseline.sh`, `run-harness-smoke.sh`, and `run-rotation-recovery.sh` gain a start-of-run teardown
- [x] A run that ends with leftovers fails, without masking an earlier failure
- [x] `run-two-instance-isolation.sh` passes with a default-identity `bootroot` install running on the same host, and leaves it running
- [x] A run on a clean host behaves exactly as before
- [x] No `CHANGELOG.md` entry
- [x] Full CI passes

sehkone:
Closes #845.

## What changed

Every E2E harness now tears the compose stack down before doing any work, asserts that the containers bootroot creates are not already there, and asserts again on the way out that its own teardown removed them — and no teardown discards its output any more.

### The teardown stops being silent

All ten `compose_down` definitions and `run-two-instance-isolation.sh`'s `teardown_instance` sent every byte to `/dev/null` and swallowed the exit status. They now write to the run log (`run.log`, or `runner.log` in the three harnesses that have no `run.log`). `compose_down` no longer swallows the status either; the decision moves to the call sites, because it is not the same at both: a start-of-run teardown may legitimately find nothing to do and keeps `|| true`, while a failure in `cleanup` is reported and fails the run.

### The start-of-run check, by exact container name

Before any work, each of the ten default-identity harnesses resolves the install identity and fails if any of the nine names bootroot creates at it exists. The failure names the containers found, gives the `docker rm -f` line that removes them, and states both readings rather than asserting it is garbage:

```
[run-local-lifecycle startup] containers of a bootroot install at instance name 'bootroot' are already present:
  bootroot-openbao bootroot-ca
They are either what an E2E run killed before its teardown left behind, or a real bootroot install on this host.
This harness installs at that same identity and cannot run beside either, so remove them once you have established which:
  docker rm -f bootroot-openbao bootroot-ca
```

By name, not by `com.docker.compose.project`, for the reason the issue gives: a killed run's leftovers carry a project label no later run knows to ask about, and the label the ten do share with each other is a co-located default install's too. Exact names, never a `bootroot-*` wildcard.

The check runs **before** the start-of-run teardown rather than after it. That teardown is a `down -v --remove-orphans` at a Compose project a real install on the same host holds too, so running it first deletes that install, volumes and all, and leaves the check reading a daemon it had just cleaned — the real-install reading the message asks you to establish would then never be printed for it. Nothing can tell such an install apart from a killed run's leftovers, which carry the same label, so the harness removes neither and says so. The teardown still follows the check, for the volumes, networks and orphans the check does not look at.

Between the two the harness declares the stack its own, and `cleanup` tears nothing down before that point: a run that stopped at the startup check must not remove on the way out what it refused to touch on the way in. Past it, `cleanup` behaves as before. The run-scoped harness needs none of this — its instances, containers and projects all carry a per-run token — and it has none.

The identity is resolved the way `resolve_recorded_instance_name` resolves it — `--instance-name`, then `BOOTROOT_INSTANCE` in the compose directory's `.env`, then the literal `bootroot` — and never from the ambient `BOOTROOT_INSTANCE` or from `COMPOSE_PROJECT_NAME`. `.env` is parsed as `read_dotenv` parses it, down to the first-`=` split, the quote stripping and the last assignment winning; a `.env` bootroot itself would refuse to parse fails the resolution rather than silently defaulting.

The first of those three is an argument the checks take rather than something they go looking for: the harness hands over the name it installed at, and it outranks the recorded one. No harness passes `--instance-name` today, so all ten leave it empty and `.env` decides — but the first one to take an instance name would otherwise have the checks read the identity of an install it did not make, and pass on a host carrying exactly the leftovers they exist to report. The validator fails a harness that installs at an explicit instance without handing it over, so that wiring cannot be forgotten.

### The end-of-run check, without masking the reason

`cleanup` in all eleven scripts now takes the shape `run-two-instance-isolation.sh` already had: capture `$?` on entry, fold a failed teardown or a surviving container into a separate status, and exit with the run's own status where it had one. A harness that leaves its own garbage behind fails; a run that had already failed keeps its original status and reason.

`teardown_instance`, the run-scoped harness's teardown, had to start reporting its own failures for that to mean anything there. Every one of its steps was suppressed, so a Compose `down` that errored, a `docker rm` that was refused, or a label query that could not reach the daemon at all returned success all the same — and the leftover assertions behind it cannot cover for that, since a query that failed returns no resources, which reads there as a clean teardown. Each step now records its failure and the sweep carries on — a `down` that errored is exactly when the per-resource removals are worth running — and the accumulated status comes back once all four resource kinds have been attempted. `cleanup` folds it into the cleanup status, keeping the entry status where the run had already failed; the mid-run teardown of instance B, where the teardown is the thing under test rather than a sweep on the way out, fails the run outright.

### A query that could not run is not a clean host

Both checks ask the daemon what it holds, and neither may read a question it could not put as an answer of "nothing". The container half asked `docker container inspect` once per name and treated every non-zero status as an absent container — which a stopped daemon, an unreachable socket and a refused authorization all produce as well, so the check passed, the start-of-run teardown's deliberately non-fatal `|| true` swallowed the failure behind it, and the run continued into exactly the confusing mid-run failure the check exists to replace. The label-scoped half had it too: `|| true` on both label queries turned an unavailable daemon into a teardown that had removed everything it owned.

An inspect cannot distinguish the two — absence and unavailability are the same exit status, and only Docker's message separates them — so the container names now come from one `docker ps -a --format '{{.Names}}'` and are matched here in full. That keeps the check exact (a `--filter name=` would match substrings, which is why it was per-name to begin with) while giving the query a status worth reading. A listing that fails now fails the check: fatally at the start of a run, and as a non-zero return the trap folds into its status at the end. Docker's own diagnostics stay on stderr, beside the message saying which check could not run. The label queries lost their `|| true` the same way, and the end-of-run report still asks about volumes when the container query is the one that failed.

### Shared, and used by the script that had it

`assert_no_leftovers` and the status-preserving cleanup shape move out of `run-two-instance-isolation.sh` into `scripts/impl/lib/leftovers.sh`, alongside the new name-based half. That script now calls the shared label-scoped version, unchanged in what it checks — it is the one harness whose instances are its own rather than the default identity's, so the container-name check would be both wrong and redundant there, and the validator asserts it does not run it.

### Start-of-run teardown where there was none

`run-baseline.sh`, `run-harness-smoke.sh` and `run-rotation-recovery.sh` only ever tore down from `cleanup`. They gain the call, without which the check would be worth little.

`run-rotation-recovery.sh` needed more than the call site. It writes the repository-root `.env` from `compose_up`, so at the new call site that file does not exist yet, and Compose interpolates the whole file on any subcommand — the `down` aborted on the required `POSTGRES_PASSWORD` and `GRAFANA_ADMIN_PASSWORD` before reaching the daemon, and the `|| true` turned that into nothing but a line in the run log. Its two siblings export the values; here they are confined to the teardown call so they cannot override the `POSTGRES_PASSWORD` that `init` rewrites mid-run. A `down` selects what it removes by project and label, so what they hold cannot matter to it.

(The same interpolation abort is pre-existing in the seven harnesses that already had a start-of-run teardown, on a checkout where no install has ever run. It was invisible until this PR stopped discarding the output and is now a line in the run log; it is harmless there — a host with no `.env` has nothing to tear down — and the assertion that follows is unaffected.)

### A validator, since CI never reaches these paths

`scripts/validate-e2e-leftover-check.sh` runs in the `check` job and in `preflight/run-all.sh`. It sources the shipped library and drives it against a stub `docker`, so what is validated is the shipped code: that the nine suffixes still match `BootrootContainer::suffix` in `src/commands/container_name.rs`, that the resolution ignores the environment and follows `.env`, that the check is exact rather than a prefix match, that the failure carries the names and the removal command, that the cleanup status folds the way it should, and that every harness is wired to the checks. It also holds both halves of the library that a green run can never exercise: the label-scoped check is driven against the stub and its `com.docker.compose.project` filter recorded (a query that drifted to a `bootroot-*` name would still return nothing on the clean daemon a passing run leaves behind), every teardown is held to writing to the run log rather than merely away from `/dev/null`, and every `cleanup` to failing the run when its teardown fails.

The order and the ownership boundary are checked too, since neither is visible in the calls merely existing. The validator reads each harness's `main` for the relative position of the three calls, and walks every function that `main` can reach ahead of the assertion — transitively, so a teardown buried one call deep fails it as well — for anything that removes a container. It then holds every `cleanup` to guarding its teardown, and drives the shared cleanup shape to prove an unowned trap removes nothing and an owned one still folds a failed teardown and a leftover into the status.

The stub can now refuse a query the way an unreachable daemon refuses it, so every check is driven once against a daemon that cannot be asked: the startup assertion has to abort, the end-of-run report has to return non-zero, both have to say which query failed, and Docker's message has to survive into the output. The stub also rejects a container listing given no `--format`, so a check that went back to a substring-matching `--filter name=` fails there rather than quietly matching too much.

The run-scoped teardown is exercised the same way, on the code that runs: `teardown_instance` is `eval`'d out of the harness and driven against stubbed docker calls, one failing step at a time. Every resource kind is still attempted after one fails, the status still comes back non-zero, and that holds for a failed label query too — the case no assertion downstream can see, because a query that failed reports nothing left behind. Every call site is then held to acting on that status, and the explicit-instance precedence to being both implemented and handed over.

Three of the validator's own checks piped their input into a `grep -q`, which exits on its first match and SIGPIPEs whatever is still writing upstream; `set -o pipefail` then read the match as a failed search. Each failed in the direction that lets a violation through, and one was reproducible: a harness given `--instance-name` without wiring it to the checks passed. They read from a here-string now.

### The manual, in both languages

`docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md` gain a section on the check, because the first thing it does for most people is stop a run they expected to start. It names the nine containers, says why the check is by name rather than by project label, explains that the assertion comes ahead of the teardown and that a run stopping there leaves the host exactly as it found it — so the `docker rm -f` line is the operator's to run once they have established which of the two readings applies — and records that a daemon that cannot be asked fails the check rather than reading as a clean host.

## Test plan

Verified on this host:

- [x] With a co-located Compose-managed install up under project `bootroot` (a `bootroot-ca` container and a volume), `run-local-lifecycle.sh` fails before doing any work, naming it and the `docker rm -f` line, and leaves it **running** — where the same run at the previous commit removed it (`Container bootroot-ca Removed` in its run log) and carried on into `infra install`
- [x] With `bootroot-openbao` and `bootroot-ca` present, `run-local-lifecycle.sh` fails before doing any work, naming both and the `docker rm -f` line
- [x] It still fails under `COMPOSE_PROJECT_NAME=bootroot-e2e-unrelated-project-$$`, whose project label the leftovers do not carry — the check is not label-scoped
- [x] After `docker rm -f`, the same run proceeds past the check into `infra install`
- [x] A teardown made to fail (`COMPOSE_TEST_FILE` pointed at a missing file) reports `teardown failed; see .../run.log`, with Compose's own error in the log, instead of continuing silently
- [x] A mid-run failure whose `cleanup` then finds leftovers exits with the original failure's status and reason, not the leftover check's
- [x] A run on a clean host is unchanged in exit status and in what it does, compared against the same run on `main`
- [x] A container of that project that is **not** one of the nine (`bootroot-unrelated`) passes the check and is then removed by the teardown that follows it, which is what the teardown is still there for
- [x] Against the real `docker` CLI, not the stub: with two of the nine names planted at a recorded instance, the startup check names both and the `docker rm -f` line and the end-of-run report returns 1; with `DOCKER_HOST` pointed at a socket that does not exist, both halves of both checks fail with Docker's own connection error in the output, where the previous commit's `inspect` loop and `|| true` reported a clean host and a clean teardown
- [x] `scripts/validate-e2e-leftover-check.sh` passes, and mutating the library or a harness makes it fail: swallowing a failed container listing or a failed label query (both directions), resolving from the ambient `BOOTROOT_INSTANCE`, matching by prefix, putting the teardown back ahead of the assertion, dropping `mark_stack_owned`, making it a no-op, owning the stack from the start, removing a `cleanup`'s ownership guard, planting a `docker rm` in a function reached before the assertion, having `teardown_instance` return 0 regardless, having a failed step abort its sweep, letting a failed label query read as nothing to remove, dropping the `|| fail` or the `cleanup_status=1` from a `teardown_instance` call site, dropping the explicit instance name from the resolver or from either check, and giving a harness `--instance-name` without handing it to the checks (which the wired form then passes)
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, `ruff`, `biome`, `markdownlint-cli2`, `scripts/check-docs.sh`, `validate-deploy-compose.sh`, `validate-compose-instance-names.sh`, `validate-e2e-openssl-compat.sh`; `shellcheck -x` reports nothing new on any changed script

- [x] `run-two-instance-isolation.sh` passes with a default-identity `bootroot` stack up on the same host, and leaves it running — the co-located install was brought up with `infra install --openbao-host-port 18200 --stepca-host-port 19000 --postgres-host-port 15432 --http01-admin-host-port 18080`, which clears the held ports without touching the container names the check reads; the harness allocates its own ports, passed all 41 assertions, and left `bootroot-openbao`, `bootroot-postgres`, `bootroot-ca` and `bootroot-http01` running with `RestartCount` unchanged
- [x] `run-two-instance-isolation.sh` re-run on the new head with the propagating teardown: 41 assertions, exit 0, and `docker ps -a` / `docker volume ls` carry nothing of either instance afterwards — the teardown returning a status rather than swallowing one changes nothing about a run whose teardown works
- [x] The rotation harness's new start-of-run teardown runs rather than aborting on interpolation: on a checkout with no `.env`, `run-rotation-recovery.sh`'s run log carries no `POSTGRES_PASSWORD must be set`, and the assertion behind it reports a planted `bootroot-grafana`

Left to CI, because port 8200 on this host is held by an unrelated process and the harnesses that install at the default identity reach it through hard-coded `127.0.0.1:8200`, `:9000` and `:8080` URLs that a host-port override cannot redirect:

- [x] All 18 PR checks, the full Docker E2E matrix among them, pass on `bdc3a3d` — the current head. The matrix arms exercise the new listing query against a live daemon, and `Quality Check` runs the validator
- [x] `e2e-extended.yml`, which has no PR trigger and is dispatched by hand, passed on `bdc3a3d` as well
- [ ] `scripts/preflight/ci/e2e-matrix.sh --skip-hosts` reaching at least step 8 is the one item this host cannot produce, for the port reason above. The CI matrix runs the same harnesses; step 9 (`run-openbao-tls-reown.sh`) needs non-interactive `sudo -n` and was always CI's to verify


sehkone:
**[Reviewer Round 1]**

### Blocking: check for existing containers before tearing their project down

Every default-identity harness runs `compose_down || true` immediately before `assert_no_leftover_containers` (for example, [run-local-lifecycle.sh:1144](scripts/impl/run-local-lifecycle.sh#L1144)–[1145](scripts/impl/run-local-lifecycle.sh#L1145)). That `compose_down` uses the default `bootroot` Compose project when `COMPOSE_PROJECT_NAME` is absent ([run-local-lifecycle.sh:227](scripts/impl/run-local-lifecycle.sh#L227)) and is `down -v --remove-orphans`. A co-located default-identity installation is therefore removed, including its volumes, before the name-based assertion looks for `bootroot-*`; the assertion sees a clean daemon and the E2E run continues. This is precisely the real-install case the issue requires us to diagnose rather than treat as garbage, so the required failure message and safe manual-removal decision never occur. The three scripts that newly gained a start-of-run teardown have the same ordering as well.

Run the exact-name assertion before any destructive startup teardown. Also guard the EXIT cleanup (or install the trap after the preflight) until the harness has acquired/created its own stack: otherwise a failing preflight assertion still invokes `cleanup`, whose unconditional `compose_down` can delete that same co-located install on its way out. Please add a regression to the validator that models this ordering/ownership boundary; the current wiring check only proves the two calls exist, not that the assertion can observe the containers before either teardown runs.

sehkone:

sehkone:
**[Author Round 1]**

### Blocking: check for existing containers before tearing their project down — **Fixed**

Accepted in full; the finding is right and I reproduced it before changing anything.

With a Compose-managed install up under project `bootroot` (a `bootroot-ca` container and its volume), the previous head `f4a484e` removed it and continued. Its run log:

```
 Container bootroot-ca Stopping
 Container bootroot-ca Stopped
 Container bootroot-ca Removing
 Container bootroot-ca Removed
 Network bootroot_default Removing
 Network bootroot_default Removed
bootroot infra install: directories created
```

The assertion then saw a clean daemon and the run went on into `infra install`. Exactly as described: the real-install case the message asks the operator to establish could never be reported for it, because the message only ever fires for containers the teardown could not reach — which is why my earlier manual check passed. I had planted those containers with `docker run`, so they carried no project label, and that is the one shape the old ordering does report.

Worth stating plainly, because it decides the ordering: no check can spare a real install while removing a killed run's leftovers. Both carry `com.docker.compose.project=bootroot` and both are named `bootroot-*`. The issue's test plan wants a `kill -9`'d run's leftovers to fail the next run rather than be silently cleared, and that is the same containers under the same label — so the assertion has to come first, and the operator decides. What I had implemented cleared both.

What changed:

- **The assertion runs first**, in all ten default-identity harnesses — before the start-of-run teardown and before anything else that could remove a container. The teardown still follows it, for the volumes, networks and orphans the assertion does not look at, so the three harnesses that gained one keep it.
- **`cleanup` tears nothing down until the harness owns the stack.** `mark_stack_owned` sits between the assertion and the teardown, and `stack_owned` guards the `compose_down` and the leftover report in every `cleanup`. A startup assertion that aborts because a real install is here no longer has its own EXIT trap remove on the way out what it refused to touch on the way in.

I kept the trap installed where it was rather than moving it past the preflight, which was the alternative offered. Two of these harnesses do work before the assertion that their trap has to undo — `run-local-lifecycle.sh` and `run-remote-lifecycle.sh` call `configure_resolution_mode`, whose `/etc/hosts` entries `cleanup_hosts` removes — and installing the trap later would drop that. The ownership flag scopes the guard to the destructive half only, which is the part the boundary is about.

Verified on this host, against a co-located install brought up under project `bootroot`:

- The run fails at the startup check, names `bootroot-ca` and the `docker rm -f` line, and the container is still **running** afterwards. Its run log holds the fatal message and nothing else — `cleanup` invoked no teardown.
- The same run at `f4a484e` produced the log above.
- A container of that project that is *not* one of the nine (`bootroot-unrelated`) passes the check and is then removed by the teardown that follows, which is what that teardown is still there for.
- On a clean host the run reaches `infra install` unchanged (it fails there only on this host's occupied port 8200, which is a local limitation, not the branch's).

### Validator regression — **Fixed**

Agreed that the wiring check proved only that the two calls exist. `scripts/validate-e2e-leftover-check.sh` now models the boundary:

- It reads each harness's `main` for the relative position of `assert_no_leftover_containers`, `mark_stack_owned` and `compose_down`, and requires that order.
- It walks every function `main` can reach ahead of the assertion — transitively, over the script's own definitions — and fails if any of them removes a container. A teardown buried one call deep fails it too, not just a moved line. `trap` lines are excluded there deliberately: an EXIT trap firing before the assertion is what the ownership guard governs, and that is checked separately.
- It requires every `cleanup` to guard its teardown with `stack_owned`, and asserts the run-scoped harness carries no such guard, since its per-run token puts everything it touches out of reach of anyone else's install.
- It drives the shared cleanup shape behaviourally: an unowned trap invokes no teardown and no leftover check; an owned one does both and folds a failed teardown or a surviving container into the status without overwriting a run's own.

Each of those fails on the mutation it exists for — teardown moved back ahead of the assertion, `mark_stack_owned` dropped from `main` or made a no-op, the stack owned from the start, a `cleanup` guard removed, a `docker rm` planted in a function reached before the assertion — alongside the mutations the earlier checks already caught.

Docs updated in `docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md`: the ordering, why it is that way, and the fact that a run stopping at the assertion leaves the host as it found it. No `CHANGELOG.md` entry, per the issue.

The full Docker E2E matrix and the extended suite passed on `f4a484e`; since this changes what runs before the teardown, both need re-running on the new head, and the test plan in the PR body is unchecked for them again.

sehkone:
**[Reviewer Round 2]**

The Round 1 ordering and ownership boundary are resolved: each default-identity harness now checks before its startup teardown, and its guarded EXIT cleanup leaves a rejected pre-existing stack alone. The validator covers that ordering.

### Blocking: propagate a failed run-scoped teardown through cleanup

`run-two-instance-isolation.sh` still suppresses every operation in `teardown_instance` with `|| true` ([lines 914–927](scripts/impl/run-two-instance-isolation.sh#L914-L927)). Consequently both calls from `cleanup` ([lines 951–952](scripts/impl/run-two-instance-isolation.sh#L951-L952)) return success even when Compose or one of the Docker removal calls failed; `cleanup_status` is set only when a later label query finds a surviving container or volume. A failed daemon/query path can therefore leave a run that had otherwise passed exiting 0, with the actual error only buried in `run.log`.

The issue permits non-fatal *start-of-run* teardown, but requires teardown failures to be reported rather than silently continued past; the PR body also says all eleven cleanups fold a failed teardown into the status. Accumulate the best-effort failures inside `teardown_instance`, return a non-zero result after attempting every resource kind, and fold that result into `cleanup_status` while retaining the entry status. Extend the validator with an intentional `teardown_instance` failure, not just a surviving-resource result.

### Blocking: implement the required explicit instance-name precedence

`resolve_recorded_instance_name` documents the required `--instance-name` → recorded `.env` → default order, but it accepts only `compose_dir` and implements only the latter two ([leftovers.sh:140–168](scripts/impl/lib/leftovers.sh#L140-L168)). The validator likewise tests only `.env` and default resolution. That does not meet the issue's explicit requirement that the check resolve the same identity as `resolve_recorded_instance_name`, including an `--instance-name` when one is supplied.

Give the shared resolver an explicit-instance input and wire it from the install invocation, so it outranks the recorded value without consulting ambient `BOOTROOT_INSTANCE`; add a precedence case to the validator. This matters as soon as a default-identity harness gains an explicit instance, precisely the future change the issue calls out.

sehkone:

sehkone:
**[Author Round 2]**

Both items accepted. One of them turned up a third defect on the way in, which is fixed here too.

### Blocking: propagate a failed run-scoped teardown through cleanup — **Fixed**

You are right, and the PR body was claiming something that was not true of the eleventh script.

`teardown_instance` now accumulates rather than suppresses. Each step records its failure and the sweep carries on — a Compose `down` that errored is exactly when the per-resource removals behind it are worth running, so aborting on the first failure would be the wrong fix — and the accumulated status comes back once all four resource kinds have been attempted. The label queries count too: `docker ps -aq` and its siblings are now checked rather than left to fail into an empty loop, which is the path you called out, and the one the assertions downstream are blindest to. A query that could not reach the daemon returns no resources, and "no resources" is exactly what a clean teardown returns.

The two call sites act on it differently, deliberately:

- `cleanup` folds each teardown's failure into `cleanup_status` with a message naming which instance and pointing at the run log, and `exit_with_cleanup_status` keeps the entry status where the run had already failed. A teardown failure now fails a run that had otherwise passed.
- The mid-run `teardown-b` call fails the run outright through `fail`. There the teardown is the thing under test rather than a sweep on the way out, and `assert_no_project_leftovers` immediately after it would otherwise report on whatever the same daemon answered next.

On the validator, I did what you asked and then some: rather than assert against a reimplementation, it `eval`s `teardown_instance` out of the shipped harness and drives it against stubbed `instance_compose` and `docker`. Each step is failed in turn — the Compose `down`, `docker rm`, `docker volume rm`, `docker network rm`, and the container label query — and each case asserts both that the status comes back non-zero and that every later resource kind was still attempted. A separate check holds every `teardown_instance` call site to acting on the status, so a future call that ignores it fails the validator.

Verified live: `run-two-instance-isolation.sh` on the new head passes all 41 assertions, exits 0, and leaves no container or volume of either instance behind. Mutations confirm the coverage bites — `return 0` in place of the accumulated status, a failed step aborting the sweep, a failed label query reading as nothing to remove, and either call site dropping its handler are each caught.

### Blocking: implement the required explicit instance-name precedence — **Partially addressed**

Implemented in full; wired as an argument rather than plumbed through ten install invocations. Reasoning below.

`resolve_recorded_instance_name` now takes the explicit name as an optional second argument and returns it ahead of the recorded `.env` and the default. `assert_no_leftover_containers` and `report_leftover_containers` take it as an optional third and pass it down. Nothing reads the ambient `BOOTROOT_INSTANCE` or `COMPOSE_PROJECT_NAME` at any point — the validator sets both to poison values in every resolution case, including the new ones. Three precedence cases were added: explicit over recorded, explicit with no `.env` at all, and an empty explicit falling through to the recorded value. Both checks are also driven with an explicit name to prove the names they ask the daemon about follow it and not the `.env`.

What I did not do is add `--instance-name` plumbing to the ten harnesses' install invocations. None of them passes one, so a per-harness variable would be empty at every call site — dead plumbing in ten scripts, each of which installs differently and some of which install repeatedly, and none of which CI can cheaply prove I did not break. The real requirement behind your point is that the two cannot drift, and I have made that a check rather than a convention: if a default-identity harness's source ever passes `--instance-name` outside a comment, the validator requires every `assert_no_leftover_containers` and `report_leftover_containers` call in that script to be given a third argument, and fails the build otherwise. So the harness that first takes an instance name cannot land without wiring it, which is the failure mode you described, and the wiring lands with the change that needs it rather than ahead of it.

Verified by mutation: adding `--instance-name e2e` to `run-local-lifecycle.sh`'s install fails the validator, naming the unwired call; adding the third argument to both calls makes it pass again.

### Not raised, found while verifying the above — **Fixed**

The mutation that adds `--instance-name` to a harness was not caught on the first attempt, and the reason was a defect in the validator itself. Three of its checks piped their input into a `grep -q`. `-q` exits on its first match, which SIGPIPEs whatever is still writing upstream, and `set -o pipefail` then reports the pipeline as failed — so a match reads as no match, non-deterministically, depending on whether the upstream had already finished writing into the pipe buffer.

All three failed in the direction that lets a violation through: `calls_function` would stop the reachability walk at the function it was following, the destructive-command scan would report a chunk that removes nothing, and the `--instance-name` search would skip the very script it had matched. Only the last was reproducible on today's inputs; the other two are latent on a longer `main`. All three read from a here-string now.

### Documentation

No user-facing surface changed, so the docs site and README are untouched, and no `CHANGELOG.md` entry — test-harness only, as the issue directs. The reasoning for each change is in the comments at `scripts/impl/run-two-instance-isolation.sh`'s `teardown_instance` and in `scripts/impl/lib/leftovers.sh`'s resolver, and the PR body is updated to describe the teardown status propagation, the explicit-instance precedence and the validator's new coverage.

### Verification

- `scripts/validate-e2e-leftover-check.sh` passes, and every mutation listed above fails it
- `run-two-instance-isolation.sh`: 41 assertions, exit 0, nothing left behind
- `validate-deploy-compose.sh`, `validate-compose-instance-names.sh`, `validate-e2e-openssl-compat.sh` pass; `bash -n` and `shellcheck -x` report nothing new on the changed scripts
- All 18 PR checks passed on `3858884`, the head this round started from; the new commits are pushed, and `e2e-extended.yml` — which has no PR trigger — is dispatched on the new head

sehkone:
**[Reviewer Round 3]**

The two Round 2 findings are resolved: `teardown_instance` now accumulates and propagates failure through both call sites, and the shared resolver and checks implement explicit-instance precedence without reading ambient identity variables.

### Blocking: fail when the leftover query itself cannot run

`bootroot_leftover_containers` treats every failed `docker container inspect` as an absent container ([scripts/impl/lib/leftovers.sh:196-201](scripts/impl/lib/leftovers.sh#L196-L201)) and then unconditionally returns success. A stopped/unreachable Docker daemon, an authorization failure, or another inspect error therefore makes both `assert_no_leftover_containers` and `report_leftover_containers` report a clean host. The startup path then suppresses its subsequent `compose_down` failure with `|| true` and continues until a later, unrelated Docker operation fails—exactly the delayed and misleading failure this change is meant to replace. The inspection diagnostics are discarded as well.

Treat inability to perform the query as a fatal check/report failure with useful diagnostics, while retaining the normal absent-container case. Add a validator case for a failing inspection/query. The label-scoped helpers have the same `|| true` behavior at [leftovers.sh:292-298](scripts/impl/lib/leftovers.sh#L292-L298); those should likewise not turn an unavailable daemon into a successful end-of-run assertion.

sehkone:

sehkone:
**[Author Round 3]**

### Blocking: fail when the leftover query itself cannot run — **Fixed**

Accepted in full, both halves. Fixed in `bdc3a3d`.

The container half could not be repaired in place. `docker container inspect` answers "this container is absent" and "I could not ask" with the same non-zero status, and only the message on stderr separates them — so any per-name inspect reads an unreachable daemon as a clean host, whatever it does with the status. It is now one listing instead: `docker ps -a --format '{{.Names}}'`, matched here in full. That keeps the check exact — a `--filter name=` matches substrings, which is why it was per-name to begin with — while giving the query a status worth reading. A listing that fails now aborts `assert_no_leftover_containers` through `fail`, and makes `report_leftover_containers` return 1 so the trap folds it into the run's status rather than recording a clean teardown.

The label-scoped half lost both `|| true`s. `project_container_ids` and `project_volume_ids` now propagate the failure, `assert_no_project_leftovers` aborts on either, and `report_project_leftovers` returns 1 — still asking about volumes when the container query is the one that failed, since a run whose daemon went away wants to be told about both.

Docker's own diagnostics are not captured and re-printed: stderr is left alone, so the daemon's message lands in the harness's log immediately above the line naming which check could not run. Capturing it would mean a temporary file or a subshell-scoped variable that never reaches the caller — the message is more useful where Docker already put it.

**Validator cases added**, as asked. The stub `docker` can now refuse any of the three queries the way an unreachable daemon refuses it (a `Cannot connect to the Docker daemon` on stderr, non-zero, nothing on stdout), and each check is driven against it: the startup assertion must abort, the end-of-run report must return 1, both must say which query could not be run, and Docker's message must survive into the output. Each label kind is refused on its own, plus both together, so a report that gave up after the containers fails there. The stub also now rejects a container listing given no `--format`, which fails a check that ever went back to a substring-matching `--filter name=`.

Both directions were mutation-checked: restoring `2>/dev/null || true` on the container listing fails the new case, and restoring it on `project_container_ids` fails the label case.

Verified against the real `docker` CLI as well as the stub — with two of the nine names planted at a recorded instance the startup check names both and the removal line, and with `DOCKER_HOST` pointed at a socket that does not exist all four entry points fail carrying Docker's connection error, where the previous commit reported a clean host and a clean teardown.

### Documentation

`docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md` gain a paragraph on it, in the section describing the checks: a daemon that cannot be asked is not a clean host, and reading a failed query as "nothing found" is what would pass the startup assertion, slip past the start-of-run teardown's deliberately non-fatal `|| true`, and resurface as the mid-run failure these checks exist to replace. `mkdocs build --strict` and `markdownlint-cli2` pass. No `CHANGELOG.md` entry, per the issue — this is harness-only.

The PR description is updated with the new section, the validator's new coverage, and the test-plan entries above.

sehkone:
**[Reviewer Round 4]**

The Round 3 finding is resolved. `docker_container_names` now performs one full-name `docker ps -a --format` listing and propagates an unavailable-daemon failure, preserving Docker diagnostics; the exact-name matching remains in the library. The project-scoped container and volume queries likewise propagate failures, and the end-of-run reporters turn them into a non-zero cleanup status.

The validator covers the new unavailable-query cases for both halves, including each project query individually, and continues to exercise the shipped `teardown_instance` failure propagation. I found no remaining findings in the updated diff.

Review verdict: APPROVED.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Fail an E2E run that finds containers left behind
```

**Body**

```text
The E2E harness tore the compose stack down at the start of a run
and again on exit, and could tell you nothing about either.  Every
`compose_down` sent its output to /dev/null and swallowed the exit
status, so a teardown that removed nothing looked exactly like one
that removed everything.  Nothing survives SIGKILL either: a run
killed hard never reaches its cleanup, and the containers it leaves
behind are named after the install identity rather than the Compose
project, so they collide with the next run's on `up` whichever
project either run used.  That surfaced minutes later as a failure
with no apparent connection to its cause.

Teardown output now reaches the run log, and the status is the call
site's to act on: a start-of-run teardown may legitimately find
nothing to do, while one in cleanup fails the run.  Before doing any
work, each default-identity harness asserts that none of the nine
container names bootroot creates exists at the resolved identity,
naming what it found and the `docker rm -f` line that removes it.
The check is by exact name rather than by com.docker.compose.project:
a killed run's leftovers carry a project label no later run knows to
ask about, and the label the harnesses share with each other is a
co-located install's too.

The assertion runs ahead of the start-of-run teardown, because that
teardown is a `down -v --remove-orphans` at a project a real install
holds too; running it first would delete that install and leave the
assertion reading a daemon it had just cleaned.  Nothing tells such
an install apart from a killed run's leftovers, so the harness
removes neither and says so.  Only past the assertion does it own
the stack, and its EXIT trap tears nothing down before that point.

The identity is resolved the way `resolve_recorded_instance_name`
resolves it -- an explicit --instance-name, then BOOTROOT_INSTANCE as
recorded in the compose directory's .env, then the default -- and
never from the ambient environment, which gives the right answer
today only because every harness happens to install at the default
identity.

On the way out the same check runs again, folded into a cleanup
shape that keeps the status of a run that had already failed: a
harness leaving its own garbage behind is broken, but what ended the
run is the more useful reason to report.  A query that could not run
is not a clean host anywhere in this -- the container names come from
one `docker ps -a` listing whose status is read, and the label
queries no longer swallow theirs.

run-two-instance-isolation.sh had the label-scoped half of this
already; it moves to scripts/impl/lib/leftovers.sh and that script
calls the shared version.  run-baseline.sh, run-harness-smoke.sh and
run-rotation-recovery.sh gain the start-of-run teardown they never
had, the last of those supplying the variables Compose interpolates
before it will run a `down` at all.

CI never reaches these paths, so
scripts/validate-e2e-leftover-check.sh drives the shipped library
against a stub docker and holds every harness to the wiring.

Closes #845
```

PR #853, opened 2026-08-15
Report the DNS alias outcome in service add's summary (#830)

### LINKED ISSUE #830: Report the DNS alias outcome in service add's summary
## Problem

`service add` registers the HTTP-01 DNS aliases as its last act before printing its success summary (`src/commands/service.rs:550`), and the registration is best-effort by design. `apply_dns_aliases` (`src/commands/dns_alias.rs:101`) documents it plainly: it returns `Ok(())` "for all cases where the container remains connected to the network (including when aliases could not be applied)", and returns `Err` only when the container is left detached. A responder that is not running, a network it cannot resolve, and a failed disconnect all end in `Ok(())`.

That design is right and this issue does not propose changing it. Adding a service must not fail because the responder happens to be down; the operator can bring it up later and the aliases are replayed (`replay_dns_aliases`, called from `src/commands/infra.rs:310`).

What is missing is the report, and the gap is narrower than "the command says nothing". On the path where the aliases were attached, `apply_dns_aliases` already prints one stdout line per alias — `dns_alias_registered`, "bootroot service add: registered HTTP-01 DNS alias {value}" (`src/commands/dns_alias.rs:176`). On the paths where nothing was attached it prints a warning on stderr instead, and those warnings are good ones: `dns_alias_responder_not_running` says "WARNING: {container} container not running — skipping DNS alias registration. Run `bootroot infra up` first.", and `dns_alias_connect_recovered` says "WARNING: DNS alias registration failed but network connectivity was restored. Run `bootroot infra up` to apply aliases later. ({error})".

So the two runs are not literally indistinguishable — three aliases produce three stdout lines and none produce zero. **The problem is that failure is signalled only by an absence.** An absence is not something a script can assert on and not something a reader notices, and the success summary that follows is silent either way: the command then exits 0 and prints a summary that never mentions aliases. Nothing in the summary, and nothing in the exit status, states that the registration was skipped.

## Why now

Until recently two fallbacks stood behind this. `.github/workflows/ci.yml` re-created the responder with the aliases hard-coded in `docker-compose.test.yml`, and `scripts/preflight/ci/test-core.sh` wrote the same three names into step-ca's `/etc/hosts`. Both are now gone — #829 removed the first in `4d33ce6`, #826 removed the second in `ad27019` — both because `service add` already registers those names. Each removal is correct on its own, and both issues record this consequence in their Out of scope sections. `apply_dns_aliases` is now the only mechanism on either side.

A registration that attached nothing therefore produces a warning that is easy to scroll past, a zero exit status, and then a failure hundreds of lines later — an ACME challenge that cannot be answered, surfacing as a `verify` failure that names neither the alias nor the responder.

## Scope

Report the outcome where the command already reports its result. Keep the best-effort contract exactly as it is.

1. Have `apply_dns_aliases` return an outcome instead of `()`. Three states, because `register_dns_alias` short-circuits before it (`src/commands/dns_alias.rs:42`) and needs something to return:

   ```rust
   pub(crate) enum DnsAliasOutcome {
       /// Aliases were attached to the responder.
       Registered { count: usize },
       /// The alias set was empty, so nothing was attempted.
       NothingToRegister,
       /// Registration was attempted and attached nothing; a warning
       /// was already printed on stderr.
       Skipped,
   }
   ```

   `NothingToRegister` is unreachable from `service add` today — `instance_id` is mandatory there, resolved from the flag or prompted for (`src/commands/service/resolve.rs:127`), so a newly added service always contributes an alias — but `dns_alias_for_entry` returns `None` without one, so the short-circuit is live for a state file carrying entries that lack it.

   Do **not** mark the type `#[must_use]`. Item 5 below deliberately discards it at two call sites, and `-D warnings` turns `unused_must_use` on a discarded `expr?;` into a build failure.

2. Leave the existing `println!` and `eprintln!` in `apply_dns_aliases` exactly as they are — wording, position, and the conditions that emit them. This is a return value, not a new message and not a replacement for an old one. The per-alias `dns_alias_registered` lines in particular must stay: `apply_dns_aliases` is shared with `infra up` and `service remove`, so removing or reshaping them would change those commands' output, which item 5 forbids. The summary line added below is the aggregate; those lines remain the detail.

3. Propagate the outcome through the three public wrappers: `register_dns_alias`, `reconcile_dns_aliases`, `replay_dns_aliases` (`src/commands/dns_alias.rs:36`, `:59`, `:73`), each returning `Result<DnsAliasOutcome>`.

4. Print one line for it in the `service add` summary. The summary body is not in `print_service_add_apply_summary` (`src/commands/service.rs:637`), which only delegates — it is `print_service_add_summary` in `src/cli/output.rs:70`, driven by `ServiceAddSummaryOptions`. Add a field to that struct:

   - `dns_alias: Option<DnsAliasOutcome>`, printed immediately after the `service_summary_openbao_path` line, before the `applied` / `remote` / `note` blocks, so its position does not depend on which optional blocks are present.
   - The other four call sites (`src/commands/service.rs:348`, `:374`, `:415`, `:706`) pass `None` and their output stays byte-identical. Do not reuse the existing `note` field — all four of them already put real content in it.

   The line must be stable enough to grep: it names the count, and on the `Skipped` path it says none were registered and points at the warning above rather than repeating it.

5. Add two messages to `src/i18n/en.rs` and `src/i18n/ko.rs`, and wire them through `src/i18n.rs` and `src/i18n/service.rs`, following the nine `dns_alias_*` messages already there:

   - a count message, used for both `Registered { count }` and `NothingToRegister` (which reports zero — there is no warning above it to point at);
   - a skipped message, used only for `Skipped`, naming zero and pointing at the warning already on stderr.

6. Leave `infra up` and `service remove` printing what they print today. They call the other two wrappers and will receive the new return value; discarding it there keeps this change to one command's output. Note that `reconcile_dns_aliases` is called from a closure at `src/commands/service/remove.rs:184` whose result feeds `finalize_removal`, so discarding it there means a `.map(|_| ())` (or leaving `finalize_removal`'s bound at `Result<()>`), not simply ignoring a value.

## Out of scope

- Changing the exit status of `service add`, or turning any of the `Ok(())` paths into `Err`. The best-effort contract is deliberate, and a service whose responder is down must still be addable.
- Changing what gets attached, when, or how — no new retry, no new reconciliation trigger.
- Adding output to `infra up` or `service remove`, and changing the output of `service add`'s preview and remote-idempotent paths.
- The two issues that removed the fallbacks. This issue exists because of them; it does not modify them.

## Acceptance criteria

- `service add` prints a line reporting the alias outcome on both reachable paths: aliases attached, and none attached.
- With the responder stopped, `service add` still exits 0, still prints its success summary, and that summary now states that no aliases were registered.
- With the responder running, every alias the summary counts appears in `docker inspect` on the responder container. The count itself is smaller than what `docker inspect` lists: `apply_dns_aliases` always adds `RESPONDER_SERVICE_NAME` (`src/commands/dns_alias.rs:137`) and Docker adds the container's short ID, so an equality check on the two numbers is wrong.
- The `println!` and `eprintln!` calls in `src/commands/dns_alias.rs` are unchanged in wording and in the conditions that emit them — including the per-alias `dns_alias_registered` lines, which still appear once per alias on the success path alongside the new summary line.
- `src/i18n/en.rs` and `src/i18n/ko.rs` both carry the two new messages, and neither leaves a `{placeholder}` unsubstituted.
- `infra up`, `service remove`, and `service add --dry-run` / `--print-only` produce the same output as before.

## Test plan

1. With the stack up, `bootroot service add …`. Confirm the summary reports the aliases, and that every alias it counts appears in `docker inspect -f '{{json .NetworkSettings.Networks}}' bootroot-http01` (which also lists the `bootroot-http01` service alias and the container's short ID).
2. Stop `bootroot-http01`, add another service, and confirm: exit status 0, the existing `WARNING: … container not running …` on stderr, and a summary line stating no aliases were registered.
3. Bring the responder back with `bootroot infra up` and confirm the replay attaches the full set, and that `infra up`'s own output is unchanged.
4. `bootroot service remove` for one of the services and confirm its output is unchanged from before this issue.
5. `bootroot service add --dry-run` and confirm its summary is unchanged — the preview path never registers aliases and passes `None` for the new field.

sehkone:
## Summary

`service add` registered the HTTP-01 DNS aliases and then printed a success summary that never mentioned them, so a run that attached nothing was signalled only by the absence of the per-alias stdout lines — not something a script can assert on or a reader notices. This makes the outcome an explicit line in the summary while leaving the best-effort contract untouched.

- `apply_dns_aliases` now returns a `DnsAliasOutcome` (`Registered { count }`, `NothingToRegister`, `Skipped`) instead of `()`, propagated through `register_dns_alias`, `reconcile_dns_aliases`, and `replay_dns_aliases`. The type is deliberately not `#[must_use]`, since two call sites discard it. No `Ok` path became an `Err`, and no exit status changed.
- Every `println!` and `eprintln!` in `src/commands/dns_alias.rs` is unchanged in wording and in the conditions that emit it — the per-alias `dns_alias_registered` lines still appear once per alias, alongside the new aggregate line.
- `ServiceAddSummaryOptions` gains a `dns_alias: Option<DnsAliasOutcome>` field, printed immediately after the `service_summary_openbao_path` line so its position does not move with the optional blocks. The preview and remote-idempotent call sites pass `None` and their output is byte-identical.
- Two new messages in `src/i18n/en.rs` and `src/i18n/ko.rs`, wired through `src/i18n.rs` and `src/i18n/service.rs`: a count message shared by `Registered` and `NothingToRegister`, and a skipped message that reports zero and points at the warning already on stderr. The skipped line is the zero line plus that pointer, so one grep finds both reachable outcomes.
- `infra up` and `service remove` drop the new return value on purpose (the latter via `.map(|_| ())` inside the closure feeding `finalize_removal`), so their output is unchanged.
- `scripts/preflight/extra/cli-scenarios.sh` asserts the reported count on the second add, next to the existing `verify_dns_aliases` check of `docker inspect`. Docs (en/ko installation, troubleshooting, cli-examples) and the unreleased CHANGELOG entry for the alias feature describe the new line.

Closes #830

## Test plan

- [x] With the stack up, `bootroot service add …` prints the `- HTTP-01 DNS aliases registered: N` line, and every alias it counts appears in `docker inspect -f '{{json .NetworkSettings.Networks}}' bootroot-http01` (which also lists the `bootroot-http01` service alias, so the two numbers are not equal by design)
- [x] With `bootroot-http01` stopped, `service add` exits 0, still prints its success summary, emits the existing `WARNING: … container not running …` on stderr, and the summary states `0 (registration skipped; see the warning above)`
- [x] `bootroot infra up` afterwards replays the full alias set and its own output is unchanged
- [x] `bootroot service remove` output is unchanged from before this change
- [x] `bootroot service add --dry-run` / `--print-only` summaries are unchanged — the preview path passes `None` for the new field
- [x] Both locales render the count and neither leaves a `{placeholder}` unsubstituted (`BOOTROOT_LANG=ko` on the same add)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt -- --check --config group_imports=StdExternalCrate` are clean
- [x] `cargo test --bin bootroot` and `cargo test --test bootroot_service` pass
- [x] `./scripts/check-docs.sh` and `markdownlint-cli2` pass

### Verification

Static checks: clippy, rustfmt, `cargo test --bin bootroot` (1049 passed), `cargo test --test bootroot_service` (66 passed, including the three new alias-summary tests), `./scripts/check-docs.sh`, and markdownlint — all clean.

The Docker-backed items were run against a real stack, brought up on remapped host ports (`--openbao-host-port 28200 --stepca-host-port 29000 --http01-admin-host-port 28080 --postgres-host-port 25433`) because the defaults are held locally, followed by a full non-interactive `init`. Observed:

- First add reported `- HTTP-01 DNS aliases registered: 1`, the second `- HTTP-01 DNS aliases registered: 2`, each immediately after the `- OpenBao path:` line and ahead of the `Bootroot-managed:` block. `docker inspect` on the responder listed both counted FQDNs plus the `bootroot-http01` service alias — 3 entries against a reported count of 2, as designed.
- The two per-alias `bootroot service add: registered HTTP-01 DNS alias …` stdout lines still appeared alongside the new aggregate line.
- With the responder stopped, a third add exited 0, printed its summary with `- HTTP-01 DNS aliases registered: 0 (registration skipped; see the warning above)`, and emitted the unchanged `WARNING: bootroot-http01 container not running …` on stderr.
- `infra up` then replayed all three aliases (including the one the skipped run never attached) and its stdout was byte-identical to the pre-change binary's; stderr differed only in the order of Docker Compose's own concurrent progress lines.
- `service remove` output was byte-identical to the pre-change binary's on both the success path (which reconciles the alias set) and a permission-denied failure path.
- `--dry-run` and `--print-only` output was byte-identical to the pre-change binary's, with no alias line at all.
- Under `BOOTROOT_LANG=ko` the same paths rendered `- 등록된 HTTP-01 DNS 별칭: 4개` and `- 등록된 HTTP-01 DNS 별칭: 0개 (등록을 건너뛰었습니다; 위 경고를 확인하세요)`, with no unsubstituted `{placeholder}` in either.

Every "unchanged" comparison above is a diff against a binary built from `1cef1b8`, this branch's merge base, run against the same live stack.


sehkone:
**[Reviewer Round 1]**

I found one issue.

- **P2 — Do not diagnose every skipped registration as a stopped responder.** The new documentation says that `0 (registration skipped; see the warning above)` means the responder was not running ([docs/en/installation.md:595](https://github.com/aicers/bootroot/blob/sehkone/issue-830/docs/en/installation.md#L595), [docs/en/troubleshooting.md:219](https://github.com/aicers/bootroot/blob/sehkone/issue-830/docs/en/troubleshooting.md#L219), with the same claim in the Korean pages). That is narrower than the implementation: `DnsAliasOutcome::Skipped` is also returned when the responder network cannot be determined, disconnect fails, or the aliased reconnect fails and rollback restores connectivity ([src/commands/dns_alias.rs:135](https://github.com/aicers/bootroot/blob/sehkone/issue-830/src/commands/dns_alias.rs#L135), [src/commands/dns_alias.rs:144](https://github.com/aicers/bootroot/blob/sehkone/issue-830/src/commands/dns_alias.rs#L144), [src/commands/dns_alias.rs:165](https://github.com/aicers/bootroot/blob/sehkone/issue-830/src/commands/dns_alias.rs#L165)). In those cases the responder may be running, and telling the operator it was down sends diagnosis in the wrong direction. Please describe the skipped line as “no aliases were registered; consult the warning above for the specific cause,” in both locales; `infra up` remains the replay step once that cause is addressed.

sehkone:

sehkone:
**[Author Round 1]**

One item, accepted.

- **P2 — Do not diagnose every skipped registration as a stopped responder.** **Fixed** in `e27a9a7`. You are right that the docs were narrower than the code: `DnsAliasOutcome::Skipped` is returned from four places in `apply_dns_aliases` — responder container not found, network not determinable, disconnect failure, and a reconnect failure whose rollback restored connectivity — and only the first implies the responder was down. Both English pages and both Korean pages now say the zero means no aliases were registered and send the reader to the warning above it for the specific cause, with `infra up` as the replay step once that cause is addressed. The unreleased CHANGELOG entry carried the same narrowing ("when the responder could not be reached") and is corrected alongside them.

The summary line's own wording is untouched — `service_summary_dns_alias_skipped` already said `(registration skipped; see the warning above)` in both locales, naming no cause, so the fix was the prose that mis-explained it rather than the string. No code changed, so the alias behaviour, the counts, and every other command's output are as reviewed.

`./scripts/check-docs.sh` and `markdownlint-cli2` are clean on the result.

sehkone:
**[Reviewer Round 2]**

The prior P2 is resolved. Commit e27a9a7 updates both English and Korean installation/troubleshooting guidance, plus the changelog, to say that a skipped result means no aliases were registered and to consult the warning above for the cause. That matches every `DnsAliasOutcome::Skipped` path, including the recovered reconnect case, which is now covered by an integration test.

The updated diff satisfies the issue: the outcome is propagated without changing the best-effort error contract, the summary prints it at the required fixed position, existing detail/warning output and the other commands’ output remain unchanged, and the tests exercise the registered and skipped paths deterministically.

No remaining findings. Approved.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Report the DNS alias outcome in service add
```

**Body**

```text
`service add` registers the HTTP-01 DNS aliases as its last act, and
that registration is best-effort by design: a service whose responder
happens to be down must still be addable. But the only signal that
nothing was attached was the absence of the per-alias stdout lines —
an absence no script can assert on and no reader notices, followed by
a zero exit status and a summary that never mentions aliases at all.

The two fallbacks that used to cover a skipped registration are now
gone, so a challenge hostname that cannot be resolved surfaces
hundreds of lines later as a `verify` failure naming neither the alias
nor the responder.

Have `apply_dns_aliases` return what it did, and state it where the
command already states its result. The best-effort contract is
unchanged: no `Ok` path became an `Err`, no exit status moved, and
every existing `println!` and `eprintln!` is untouched in wording and
in the conditions that emit it — the per-alias lines stay the detail,
the new summary line is the aggregate. `infra up` and `service remove`
discard the new value so their output is as it was, and the preview
and remote-idempotent paths pass `None`.

The skipped line is the zero-count line plus a pointer at the warning
already on stderr, so one grep finds both reachable outcomes, and the
pointer rather than a named cause is what the line carries: an empty
alias set, an undeterminable network, a failed disconnect, and a
recovered reconnect all report the same thing, and only the warning
knows which.

Closes #830
```

PR #855, opened 2026-08-15
Secure unseal and EAB writes (#840)

### LINKED ISSUE #840: The unseal keys and the EAB credentials are world-readable while they are written
## Problem

`AGENTS.md`'s permissions rule is unambiguous:

> **Restrictive permissions at creation**: A file holding a secret gets its final permissions as it is created, never afterwards — `set_permissions` once the bytes are on disk leaves a window in which the file is world-readable.

Two secret-bearing writers do exactly what it forbids. Both write the bytes first and set the mode second.

**`save_unseal_keys`** (`src/commands/openbao_unseal.rs:19`):

```rust
tokio::fs::write(&path, contents)
    .await
    .with_context(|| messages.error_write_file_failed(&path.display().to_string()))?;
fs_util::set_key_permissions(&path).await?;
```

**`eab::write_key_file`** (`src/eab.rs:88`, the writer behind `write_eab_file`):

```rust
fs::write(path, next)
    .await
    .with_context(|| format!("Failed to write EAB file: {}", path.display()))?;
fs_util::set_key_permissions(path).await?;
```

`tokio::fs::write` and `fs::write` create at `0o666 & ~umask`. On a host with the common `0o022` the file is `0o644` — world-readable — from the moment the first byte lands until `set_key_permissions` runs. That window contains the whole write and two `await` points.

What is in the files decides how much this matters.

- **`unseal-keys.txt`** holds the OpenBao unseal keys. This is the one file in the tree that cannot be regenerated from anything else: lose it and the vault does not open; leak it and, with a quorum, someone else opens it. #805 named it for that reason when it deferred this work.
- **`eab.json`** holds the External Account Binding `kid` and `hmac`, which are ACME account credentials.

Nothing is failing and CI is green. The directory these land in is created `0o700` by `ensure_secrets_dir`, which narrows who can reach the path — but the rule is about the file, and a directory mode is not the guarantee the rule asks for.

## Also: neither write is atomic

Same two sites, second rule. Both truncate the destination in place rather than staging a temporary and renaming:

> **Atomic writes**: Write state, config, and other files that another process may read atomically — write to a temporary file in the same directory, then `fs::rename`. Never truncate-and-write in place.

A crash between the truncate and the last byte leaves a partial `unseal-keys.txt` — some keys, not all — which reads back as a shorter key list rather than as a failure. For a file whose whole purpose is to be recoverable later, a half-written version is worse than an absent one, because nothing about it announces that it is incomplete.

`save_unseal_keys` also performs no `sync_all` at all, so it fails the durability half too, as #805 noted.

## Scope

Both writers publish through the same shape, which `fs_util` already provides.

- Stage into a temporary in the destination's own directory, created with `KEY_FILE_MODE` at open time rather than chmod'ed afterwards.
- `sync_all` the temporary, then rename, then `fs_util::sync_parent_dir` — these are secrets that must survive a crash, and `sync_parent_dir` exists for exactly this after #805.
- `fs_util::atomic_write` is the async wrapper over `atomic_write_blocking`, which already does the create-with-mode, flush, rename and directory flush in one call. Prefer routing both sites through it to writing a third variant.

Since #841 landed, `atomic_write` is one of four spellings over `fs_util::publish_staged_blocking`, split on two axes: whether the containing directory is flushed (`atomic_write` does, `atomic_replace` does not) and whether a symlink at the final path component is resolved first (the `_through_symlink` variants do). Both writers here take `atomic_write` on the first axis — these are secrets that must survive a crash, as above — and the second axis is decided in the section below. Do not read the four names as interchangeable; picking the wrong one is silent. #848 further replaces the wrappers' bare mode argument with a `StagedMode` enum; under it both writers here take `StagedMode::Policy(KEY_FILE_MODE)`, since a secret's mode is a policy and never the umask's. Either landing order works — the decision this issue records is the same in both spellings.

`set_key_permissions` afterwards then has nothing left to do at these two sites. Do not leave it as a belt-and-braces call — a mode applied twice reads as a decision someone made rather than a leftover, and the next reader cannot tell which.

Note the `current == next` early return in `write_key_file` (`src/eab.rs:107`), which calls `set_key_permissions` on an unchanged file. That one is repairing a mode on a file this function did not write, which is a different thing from the window above; keep it, and say in a comment that is what it is for.

## Symlinked destinations

A truncating write opens with `O_TRUNC`, which follows a symlink at the final path component and delivers the bytes to its target. A rename replaces the name it is given, so it destroys the link instead and leaves the target holding the previous contents while the write reports success. Converting a writer therefore forces a choice, and #841 made it per file rather than globally: a destination the operator arranges resolves the link first, a credential at a path bootroot chose publishes at the name.

**Both files here take the bare rename — `atomic_write`, not `atomic_write_through_symlink`.** Both sit inside the secrets tree at a path bootroot derives from its own layout rather than one the operator names: `secrets/openbao/unseal-keys.txt`, and the `eab.json` beside each service's `secret_id`. There a link is a redirection vector rather than an operator convenience, and the reader reads the path bootroot handed it — which the rename leaves holding the current secret — so following one buys nothing and costs the guarantee.

This is stated here because the choice is invisible at the call site: both spellings compile, both publish by rename, and nothing fails until an operator or an attacker puts a link at one of these paths. Record the decision in a comment at each site, in the terms the converted credential writers already use, so the next reader does not have to re-derive it.

Note that the relocated `eab.json` — the one an operator moves into an agent-owned directory with a path override — is a different writer (`fs_util::write_owned_file_replace`) and is already staged and already refuses a symlink. It is not in scope here; this issue covers only the secrets-tree `eab.json` written by `eab::write_key_file`.

## Constraints

- Do not weaken `KEY_FILE_MODE` or change which directory these files live in.
- Do not resolve a symlink at either destination. See the section above; the bare `atomic_write` spelling is the decision, not an oversight to be corrected during implementation.
- Do not change what is written. The unseal-keys payload stays newline-joined with a trailing newline, since `read_unseal_keys` parses it back.
- `save_unseal_keys` is `async` and called from async paths; keep the blocking work inside the `spawn_blocking` that `fs_util::atomic_write` already carries rather than calling the blocking core directly. See #839 for the same rule being enforced elsewhere.

## Acceptance criteria

- Neither `save_unseal_keys` nor `eab::write_key_file` calls `set_permissions` on a file it has just written; both create at `KEY_FILE_MODE`
- Both publish by rename from a temporary in the same directory, never by truncating the destination
- Both flush the temporary before the rename and the containing directory after it
- Both publish at the path they are given rather than resolving a symlink there, and a comment at each site records that as the decision it is. A test seeds a symlink at each destination and asserts the link is replaced by a regular file holding the new secret, and that the link's former target is left untouched
- The four places #841 recorded these two as exclusions no longer do: the **How bootroot writes files** section of `docs/en/cli.md` and `docs/ko/cli.md`, the `publish_staged_blocking` rustdoc in `src/fs_util.rs`, and the `CHANGELOG.md` entry that names them as handled separately. Landing this issue is what makes those four statements false, and a reader who finds one of them after the fact cannot tell it is stale. If #841 has not been released when this lands, the changelog is corrected in place rather than gaining a second entry — no user saw the earlier form
- A test asserts the mode of a freshly written `unseal-keys.txt` and `eab.json` under a permissive umask (`0o022`), so a regression to create-then-chmod is caught. `src/commands/init/steps/orchestrator.rs` already has this shape for the root token file
- The errors each site reports are unchanged in message and context
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff, full CI green

## Not in this issue

The three remaining write sites #805 deferred that hold no secret — `StateFile::save`, `cert_group::write_cert_file`, and the two orchestrator writers. They fail the atomic-write rule the same way but carry no permissions window, so they are a separate piece of work with a different urgency. **#841 has since converted those, and every other production writer in the crate.** These two are what is left, which is why the symlink decision above can be stated as a settled rule rather than an open question: this issue is applying an existing convention to the last two sites, not choosing one.

The shape of that convention — a per-call-site choice between four spellings, with the decision recorded only in prose at each site — is itself worth revisiting, and #847 does that. Not here: adding two more sites under the existing rule is the smaller and safer change, and it does not get harder if #847 lands first. The two do not block each other in either order. If #847 lands first, these two writers are converted straight to `Destination::bootroot_owned` and the symlink section above reads as the reason for that constructor rather than for the bare spelling; the decision it records is the same either way.

sehkone:
## Summary

- Publish OpenBao unseal keys and secrets-tree EAB credentials through durable, restrictive-mode staged writes.
- Replace final-component symlinks rather than following them, with regression coverage for target preservation and umask-independent `0600` modes.
- Remove the stale exclusions from the CLI documentation, staged-write rustdoc, and changelog.

Although these writes replace a destination symlink rather than sending secret bytes through it, the shared staged publisher still derives the replacement file's uid/gid from the former link target. That broader ownership behavior is intentionally unchanged in this PR.

Closes #840

## Test plan

- [x] Verify newly written `unseal-keys.txt` and `eab.json` are created at `0600` under umask `022`.
- [x] Verify each destination replaces a final symlink with the new regular secret file and preserves the former target.
- [x] Verify unseal-key payload round-tripping and unchanged EAB-file permission repair.
- [x] Run formatting, Clippy, rustdoc, Markdown lint, documentation build, and security audit.
- [x] Run Docker E2E coverage: all non-sudo matrix scenarios and the extended suite locally with isolated ports and a host-specific non-loopback bind; local-hosts, remote-hosts, and re-own require passwordless sudo, and their CI jobs pass.


sehkone:
**[Reviewer Round 1]**

The approach is right and the two call sites read well. Both writers now go through `fs_util::atomic_write` with `StagedMode::Policy(KEY_FILE_MODE)` — the wrapper the issue names on both axes — the `set_key_permissions` that followed the bytes is gone, the error contexts at both sites are byte-identical to what they replaced, and the four exclusion statements #841 left behind (`docs/en/cli.md`, `docs/ko/cli.md`, the `publish_staged_blocking` rustdoc, the changelog entry) are removed rather than duplicated. The two symlink tests are the right tests: they seed a link, assert the link is gone and the destination is a regular file holding the new secret, and assert the former target still holds its old bytes — none of which passes on the pre-change truncating write. Points below, roughly in order of how much they matter.

### 1. A fifth statement the change makes false: the symlink group's history

The acceptance criteria named four places that assert these two files are exceptions. There is a fifth, and it is the one an operator is most likely to act on.

`docs/en/cli.md:193-199` puts "every credential" in the replace-the-link group — which now correctly includes `unseal-keys.txt` and the secrets-tree `eab.json` — and then says:

> Those files have been published by rename for several releases, so a link at one of those paths has never survived the command that creates the file

That is exactly false for the two files this PR just moved into that group. A truncating `write(2)` follows a final symlink, so until this commit a link at `secrets/openbao/unseal-keys.txt` or at a service's `eab.json` *did* survive `init --save-unseal-keys` / `bootroot openbao save-unseal-keys` / a fast-poll EAB apply, and the bytes went to its target. After this commit the link is destroyed and its target silently stops receiving updates. `docs/ko/cli.md:182-186` carries the same sentence ("해당 경로의 링크는 파일을 만드는 명령을 견딘 적이 없습니다") and needs the same correction.

`CHANGELOG.md:160-164` has the mirror-image gap. It enumerates the through-link set explicitly, then names the replace set as "an issued certificate, key or CA bundle path ... which is what the key file and the control node's own `agent.toml` have always done". A reader of the release notes who has symlinked their `unseal-keys.txt` elsewhere finds their file in neither list, and the behaviour change that would break their arrangement is unrecorded. Since #841 is still unreleased this is a correction in place, not a second entry — the same call the rest of this PR already makes.

I would not treat this as optional polish: it is the same class of stale statement the issue's fourth criterion is about, and it is the only user-visible behaviour change in the PR.

### 2. `write_key_file`'s new identical-content symlink branch (`src/eab.rs:105-114`)

The guard itself is a genuine catch and I am glad it is there — without it, an identical-payload call against a symlinked destination would `chmod 0600` the *link's target*, a file this writer neither created nor owns. Three things around it, though:

- **It is the one new branch with no test.** Both symlink tests write a different payload, so they enter through `current != next`. The path this PR actually added — `current == next` *and* the destination is a link — is never exercised. A test that seeds a symlink at `eab.json` pointing at a target already holding `serialize_eab_payload("kid-1", "hmac-1")` and then calls `write_eab_file(&path, "kid-1", "hmac-1")` would cover it and would pin the semantics in point two.
- **The return value now means something new.** That branch returns `Ok(true)` for a call in which no byte of the payload changed. `src/bin/bootroot-remote/io.rs:129` maps it to `ApplyStatus::Applied` and `src/commands/service/local_config.rs:354` consumes it too, so a bootstrap will report the EAB as applied when only the link was replaced. That is defensible — the file at that path did change — but `write_eab_file`'s rustdoc at `src/eab.rs:48-49` still says the opposite verbatim: "Returns `true` when the on-disk bytes changed and `false` when the file already held identical content." Either the doc gets the symlink case or the branch returns `false`; right now they disagree.
- **The comment explains the wrong half.** `// Repair a mode on an existing file this writer did not create.` documents the `set_key_permissions` call, which the issue asked for. Nothing documents why the symlink case skips the repair and falls through to a full republish, which is the part a reader will stop at.

### 3. The site comments claim more than `atomic_write` delivers

Both new comments say the path "must replace a final symlink, rather than follow a redirection an attacker could plant there." The content is not redirected — true. The ownership is. `publish_staged_blocking` under `StagedOwner::Destination` stats the destination with `std::fs::metadata` (`src/fs_util.rs:891`), which follows the link, and chowns the staged temporary to the *link target's* uid/gid. So a link at `secrets/openbao/unseal-keys.txt` pointing at a file owned by another uid yields a published `unseal-keys.txt` that is `0600` and owned by that uid — the unseal keys handed to whoever the link pointed at, while the write reports success.

The practical exposure is small: `ensure_secrets_dir` keeps the tree at `0700`, so planting the link needs write access there already, and the more realistic case is an operator who relocated the file deliberately and now silently loses it to the target's owner. And it is not this PR's bug — it is `publish_staged_blocking`'s, shared with every writer #841 converted, and the crate elsewhere refuses the situation outright for credentials (`create_owned_credential_noclobber`, `atomic_rewrite_owned_no_symlink`). But this PR is the change that puts the *unseal keys* behind that machinery and asserts in a comment that a planted link buys the attacker nothing, so it is worth either narrowing the claim or flagging the gap for the maintainer to schedule (`symlink_metadata` for that stat, or refusing a symlinked destination in the `StagedOwner::Destination` arm). Please don't open an issue for it yourself — note it in the PR and let the maintainer decide.

### 4. What the umask tests actually catch

Both new umask-`022` assertions would also pass on the code they replace: the old writers chmod'ed to `0600` after the bytes landed, so the *final* mode was already `0600` under any umask, and the window they close is transient and unobservable by `stat` afterwards. `write_eab_file_roundtrips_through_load_credentials` already asserted `mode == 0o600` before this PR for exactly that reason. What the child-process machinery buys going forward is a guard against a later hand swapping `StagedMode::Policy` for `PreserveOrUmask` — which is now the live regression and is worth guarding — so I would keep the tests and say that in the doc comment, rather than let the shape imply it catches create-then-chmod.

Related, `write_eab_file_roundtrips_through_load_credentials` now re-execs the whole test binary to run a roundtrip and a mode-repair assertion that have nothing to do with the umask. Splitting the umask assertion into its own `#[test]` and leaving the roundtrip as the plain `#[tokio::test]` it was would keep the child-process cost proportional to the one thing that needs it, and would leave the roundtrip readable.

### 5. Prose nits

- `docs/en/cli.md:138` — "every `AppRole` `role_id`/`secret_id` and `eab.json`, the OpenBao unseal keys, and the remote bootstrap artifact" reads as a list whose first `and` fights its commas. "every `AppRole` `role_id`/`secret_id`, `eab.json`, the OpenBao unseal keys, and the remote bootstrap artifact" scans cleanly; `docs/ko/cli.md:130` has the same shape.
- `CHANGELOG.md:145-147` and `docs/ko/cli.md:104` were edited without refilling, leaving a stub line mid-paragraph ("run reads back to resume flush the directory too," / "bootroot가 만드는 파일은 모두 스테이징 후 이름").
- Optional: the "final mode holds from the moment the file appears" bullet at `docs/en/cli.md:121-123` enumerates "CA password, recovery key, `secret_id` or responder HMAC". The unseal keys are the file that enumeration most wants now that they qualify.

Nothing above blocks the core change — the writer conversion itself looks correct and complete against the issue. Points 1 and 2 are what I would want fixed before merge.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed** — Updated both CLI manuals to say that the OpenBao unseal-keys and secrets-tree EAB files join the replace-the-link group in this release, rather than claiming every member has had that behavior for several releases. Corrected the unreleased changelog entry in place to list those credential paths, as required by the issue.

2. **Fixed** — Added an identical-content EAB symlink test. It verifies the link is replaced, the former target remains byte-for-byte unchanged, and its widened mode is not repaired. Updated `write_eab_file` rustdoc to define `true` as a replacement published at `path`, which covers this case, and documented why the identical-content symlink falls through to a republish.

3. **Partially addressed** — Narrowed both call-site comments to the guarantee the selected bare staged write provides: publish at the bootroot-managed pathname and replace a final symlink instead of redirecting the secret bytes. I did not change `StagedOwner::Destination`: its symlink-following ownership lookup is shared staged-publish behavior and changing it here would alter every converted writer, beyond this issue’s two call sites. Per the review, I have not opened a separate issue; the PR now makes the ownership limitation explicit for maintainer triage.

4. **Fixed** — Kept the umask tests as guards for `StagedMode::Policy`, clarified that they cannot observe the former create-then-chmod window after the fact, and split the EAB umask check from the ordinary async roundtrip/mode-repair test.

5. **Fixed** — Cleaned up the AppRole/EAB list punctuation, added unseal keys to the final-mode examples, repaired the changelog sentence, and reflowed the corresponding Korean manual text.

sehkone:
**[Reviewer Round 2]**

Round 1's five points are addressed. Taking them in order:

1. **Resolved.** `docs/en/cli.md:195-199` now says the two files "joined this group in the current release" instead of claiming everything in the group has behaved that way for several releases, `docs/ko/cli.md:183-186` carries the same correction, and `CHANGELOG.md:162-164` names the OpenBao unseal-keys and ACME EAB paths in the replace-the-link set. An operator who symlinked one of these can now find out from the release notes that the link stops surviving the write.
2. **Resolved.** `write_eab_file_replaces_an_identical_symlinked_destination` (`src/eab.rs:363-406`) covers the branch that had no test, and it asserts the right two things — the link is replaced and the former target keeps its widened `0644`, so a future change that reintroduces the chmod-the-target behaviour fails here. The rustdoc at `src/eab.rs:48-49` now defines `true` as "publishes a replacement at `path`", which is consistent with `ApplyStatus::Applied` at `src/bin/bootroot-remote/io.rs:129`, and `src/eab.rs:114-115` explains the fall-through.
3. **Accepting the pushback.** Not changing `StagedOwner::Destination`'s `std::fs::metadata` lookup here is the right call: it is shared behaviour across every writer #841 converted, and narrowing it inside a two-call-site change would be a much wider blast radius than the issue asked for. The call-site comments are now scoped to what the wrapper actually delivers. One request that is not a re-raise: the note currently lives only in your Round 1 reply. Please put a sentence in the PR body as well, so whoever merges sees that a symlinked destination still takes its uid/gid from the link's target without having to read the review thread.
4. **Resolved.** Both test doc comments now say what the assertion does and does not catch, and the EAB umask check is its own `#[test]` with the roundtrip left as a plain `#[tokio::test]` — which also picked up a real improvement, since the roundtrip now widens the mode to `0644` before the identical-payload call and asserts the repair back to `0600` (`src/eab.rs:300-305`). That is a stronger assertion than the one it replaced.
5. **Mostly resolved.** The list punctuation and the final-mode enumeration are fixed. Two leftovers under point 8 below.

The conversion itself I have no further concerns about. Three new points.

### 6. The PR breaks the umask-isolation convention its own tests inherit, and leaves three statements false

`umask_test_ran_in_child` is not just a helper; its rustdoc records a rule, and each of the two existing call sites restates it:

- `src/fs_util.rs:1077-1079` — "Making the umask-dependent test its own process is the only isolation that actually holds. Both bootroot test executables have exactly one such test, and each opens with..."
- `src/state.rs:397-400` — "Both umasks are exercised here, in sequence, for the same reason `fs_util` puts all three `StagedMode` arms in one test rather than three — one isolated process, every umask-dependent assertion in this binary inside it."
- `src/fs_util.rs:1707-1709` — "One test rather than three for the same reason — one isolated process, every umask-dependent publish in this crate sequenced through it."

`state` is declared in `src/main.rs:8` and `eab`/`fs_util` in `src/lib.rs`, so before this PR the split was exactly one per test executable: `fs_util::tests::staged_mode_arms_answer_the_create_and_the_rewrite` in the lib binary, `state::tests::save_creates_a_new_state_file_under_the_process_umask` in the `bootroot` binary. That is what those three sentences describe. After this PR the lib binary has two and the `bootroot` binary has two, and each additional one re-execs the whole test binary.

This is the same failure mode the issue's fourth acceptance criterion is about — a statement that was true until this commit and that a later reader cannot tell is stale — and it is the one class of problem this PR is otherwise scrupulous about. I do not think the extra child processes are wrong: a caller-level pin at the writer that has to hold `Policy(KEY_FILE_MODE)` reads better than folding an unseal-keys assertion into a test named for `state.json`, and `state.rs:385-390` is itself precedent for a per-writer pin. But then the rule has changed from "one per binary" to "one per umask-dependent writer", and all three sentences need to say so. Pick either direction; what does not work is adding the tests and leaving the prose describing the old arrangement.

This is the only item I would hold the merge for, and it is a three-place edit.

### 7. `write_secret_file` still has the hole you just closed in `write_key_file`

`src/bin/bootroot-remote/io.rs:100-103`:

```rust
if current == next {
    fs_util::set_key_permissions(path).await?;
    return Ok(ApplyStatus::Unchanged);
}
```

That is byte-for-byte the shape `src/eab.rs:105-116` replaced, in the sibling writer for the `secret_id` at `--secret-id-path`, four lines above the `atomic_write` call that already takes the bare rename. An identical-payload bootstrap re-run against a symlinked `--secret-id-path` chmods the link's *target* to `0600` — a file this writer neither created nor published to. Nothing to do in this PR: it is outside #840's two call sites, and its `Unchanged` return has a different meaning to defend than `write_key_file`'s did. Per Round 1's standing instruction, do not open an issue for it — I am flagging it here because this PR is what makes the two functions diverge, and the divergence now looks deliberate to the next reader when it is not.

### 8. Smaller things

- **The comment's premise is narrower than the writer.** `src/eab.rs:117` calls the destination "This bootroot-owned secrets-tree path", but `write_key_file` also serves `bootroot-remote bootstrap`, where the destination is `--eab-file-path` — an operator-named flag whose test default is `/tmp/eab.json` (`src/bin/bootroot-remote/agent_config.rs:665`). The *behaviour* is right and I am not asking for a change to it: it matches `write_secret_file` at the operator-named `--secret-id-path`, and the manual's rule is credential-versus-configuration, not who chose the path. It is the justification that does not hold for one of the three callers. The neighbouring comment at `src/bin/bootroot-remote/io.rs:104-115` states the reason without appealing to who owns the path; matching it would make the claim true everywhere the function runs.
- **A `--eab-file-path` clause in the manual.** `docs/en/cli.md:190-192` singles out `--agent-config-path` on `bootroot-remote bootstrap` as a link the operator arranged and therefore followed. `--eab-file-path` is the adjacent flag on the same command and now gets the opposite treatment. It is covered by "every credential" in the second bullet, so nothing is wrong — but the manual went out of its way to name the one flag on that command that follows links, and the reader most likely to have a link there is the one reading that sentence.
- **`CHANGELOG.md:145` is still unfilled.** "reads back to resume flush the directory too," sits at roughly two-thirds the width of the lines around it. Dropping "The" from "The files a run" shortened the sentence without refilling the paragraph.
- **`CHANGELOG.md:165-169`** enumerates the create-then-chmod fix as "including the step-ca CA password, the OpenBao recovery keys, the responder HMAC config and each `AppRole` `secret_id`". You added the unseal keys to the equivalent enumeration in `docs/en/cli.md:121`; the changelog list is where a reader who does not open the manual will look, and the unseal keys are the file in the tree that most wants naming.
- **`docs/ko/cli.md:114`** runs to 123 columns against a ~105 fill in the surrounding paragraph, from inserting "OpenBao 언실 키," without refilling the bullet.

PR hygiene is fine otherwise: `Closes #840` is right for a PR that meets every acceptance criterion, the test plan is a real checklist, and no GitHub issue was created for the deferred ownership question.

sehkone:

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed (already resolved)** — The manuals and changelog already distinguish the newly converted unseal-key and secrets-tree EAB paths from the older rename-published paths.

2. **Fixed (already resolved)** — The identical-content EAB symlink branch is covered, its return semantics are documented, and its fall-through is explained.

3. **Fixed** — Added the requested note to the PR description: replacing a final destination symlink prevents byte redirection, but the shared staged publisher still derives the replacement file's uid/gid from the former link target. That broader behavior remains unchanged here.

4. **Fixed (already resolved)** — The umask-test comments distinguish the policy-mode regression guard from the former unobservable create-then-chmod window, and the EAB umask test remains separate from the normal roundtrip test.

5. **Fixed (already resolved)** — The earlier punctuation and Korean/manual changelog prose corrections are retained.

6. **Fixed** — Updated all three stale umask-isolation statements. The helper now says every umask-dependent test re-executes its own test executable; the state and staged-mode comments now describe their writer-level grouping without claiming one isolated child covers every umask assertion in a binary or crate.

7. **Pushed back** — The identical-content symlink branch in `write_secret_file` is a valid observation, but that writer is outside #840's two named call sites and has different `ApplyStatus::Unchanged` semantics. Changing it would expand this PR beyond its issue contract, so I left it untouched and did not open a separate issue as requested.

8. **Fixed** — Generalized `write_key_file`'s comment so it is true for the operator-provided `--eab-file-path` caller, documented that adjacent flag's replace-the-link behavior in both manuals, reflowed the Korean text, and repaired the changelog wording while naming the unseal keys in its restrictive-mode list.

Validated with `cargo fmt -- --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings`, targeted EAB/unseal umask and symlink tests, `./scripts/check-docs.sh`, and Markdown lint.

sehkone:
**[Reviewer Round 3]**

Round 2's items 1–8 are all addressed or reasonably pushed back on. Taking the two that needed judgement:

- **6 — Resolved.** All three statements now describe the arrangement this PR actually creates. `src/fs_util.rs:1077-1078` drops the "both bootroot test executables have exactly one such test" claim for a rule that scales ("Each umask-dependent test re-executes its test executable"), `src/state.rs:397-398` and `src/fs_util.rs:1707-1709` no longer justify themselves by a one-child-per-binary invariant that four umask tests would break. That was the item I said I'd hold the merge for, and it is closed.
- **7 — Accepting the pushback.** Leaving `write_secret_file` (`src/bin/bootroot-remote/io.rs:100-103`) alone is the right call and matches what I asked for; I flagged it for the record, not for a fix. No issue was opened, as requested.

Item 3's note is in the PR body now, in the right terms — a reader sees that a replaced link still donates its target's uid/gid without opening the review thread.

Two things left, both small and mechanical, plus one flag.

### 9. A sixth statement the change makes false, in the crate this time

Round 1 found the fifth in the manuals. There is one more, and it is the most explicit of the lot — `src/commands/rotate/helpers.rs:56-61`, on `write_secret_file`:

> Applying that mode to the staged temporary also closes the window the `write` + `set_key_permissions` pair this replaced left open, in which the password sat world-readable at its final path. That is the same defect **#841's sibling issue tracks for `save_unseal_keys` and `eab::write_key_file`; those two are left to it**, but a site being re-plumbed for the torn-read fix anyway does not get to keep the window.

#841's sibling issue is #840, and this PR is it landing. "Those two are left to it" was true when it was written and is false the moment this merges: both writers are converted, three files away, in this same diff. A reader who finds that comment afterwards concludes the crate still carries two credential writers with an open create-then-chmod window and goes looking for them.

This is the same class as the issue's fourth acceptance criterion — a statement whose only defect is that landing this change falsified it, and which nothing else in the tree contradicts loudly enough to catch. The criterion enumerated four places because those were the four #841 left; it is the principle rather than the enumeration that matters, and Round 1's point 1 was already the fifth. Rewriting the clause to name the two as converted (or dropping the cross-reference, since the sentence stands without it) is a one-comment edit.

I checked the rest of the crate for the same shape: `grep` over the sources turns up no other comment asserting these two writers are unconverted, and the manuals, the `publish_staged_blocking` rustdoc and the changelog are clean. This is the last one.

### 10. Commit message hygiene

`8d08de2` and `47f163b` carry literal `\n` escapes rather than real newlines:

```
Make the identical-content symlink behavior explicit so a future\nchange does not repair the target file's mode.\n\nPart of #840
```

`AGENTS.md` names this failure specifically ("Do NOT rely on literal `\n` escapes for a commit body — they land in the subject line as characters, and git reports success"). Here they landed in the body rather than the subject, so nothing broke visibly, but the result is one unwrapped line with visible backslash-n where the 72-column body and the standalone `Part of #840` trailer should be. `47f163b`'s body is a single 130-column line for the same reason. `4a13fa8` and `10dadb9` are correct, so it is the last two that need an amend or a squash before merge — this is what lands in `main`'s history, and the review thread does not travel with it.

### 11. Leftover from item 8, and one flag

- **`CHANGELOG.md:146` is still short.** The edit reflowed lines 144-145 and pushed the ragged line down rather than through: `name survives a power loss and not merely a clean` is 51 columns against 67-72 on every line around it. One more pass over that paragraph closes it.
- **`src/fs_util.rs:1707-1708` parses badly.** "One test rather than three groups all [`StagedMode`] arms into one isolated run" — the reader takes "three groups" as a noun phrase and has to back up. The claim is right; it is the sentence that was just rewritten for clarity, so it is worth getting the second time. "Covers all three [`StagedMode`] arms in one isolated run rather than splitting them across three, keeping their shared create-and-rewrite assertions together."
- **Not this PR, no issue opened:** `write_key_file` calls `fs_util::ensure_secrets_dir(parent)` (`src/eab.rs:89-91`), which `chmod`s an *existing* parent to `0700` (`src/fs_util.rs:963-971`), not only a directory it creates. On the secrets-tree path that is correct and invisible. On `bootroot-remote bootstrap --eab-file-path`, the parent is whatever directory the operator named — a shared one gets narrowed to `0700` as a side effect of writing one credential into it. Predates this PR and is outside #840's scope; I raise it because this PR is what puts `--eab-file-path` in the manual (`docs/en/cli.md:192-194`), so it is the change that sends a reader to that flag.

Nothing here touches the conversion, which I have no remaining concerns about: both writers publish through `atomic_write` with `StagedMode::Policy(KEY_FILE_MODE)`, the error contexts are byte-identical, the symlink decision is recorded at both sites in terms that hold for all three callers of `write_key_file`, the four umask and symlink tests are the right tests and each fails on the code it replaced, and the identical-content-at-a-link branch is both covered and documented. Item 9 is the one I would fix before merge; 10 costs an amend.

sehkone:

sehkone:

sehkone:
**[Author Round 3]**

9. **Fixed** — Removed the stale cross-reference in write_secret_file that said save_unseal_keys and eab::write_key_file were still deferred. The comment now states the enduring security rationale without claiming any remaining exception.

10. **Fixed** — Squashed the two commits with literal newline escapes (8d08de2 and 47f163b) and this follow-up into a36ef7d. Its commit body uses real wrapped lines and a standalone Part of #840 trailer.

11. **Fixed** — Reflowed the changelog paragraph and rewrote the staged-mode test comment for clear parsing. **Pushed back** on changing ensure_secrets_dir: narrowing an operator-provided EAB-file parent predates this PR and is outside #840's two writer conversions, so I left it unchanged and did not open an issue.

Validation: formatting, Clippy, relevant EAB/unseal-key tests, docs build, Markdown lint, deploy smoke, and test-core pass. The E2E matrix passed all non-sudo scenarios through the TLS no-delta case using isolated ports and the host's assigned non-loopback address; the TLS re-own scenario correctly stopped because this machine lacks passwordless sudo, which CI supplies.

sehkone:
**[Reviewer Round 4]**

Round 3's items are closed. Taking them in order:

- **9 — Resolved.** `src/commands/rotate/helpers.rs:56-58` now reads "A site being re-plumbed for the torn-read fix does not get to keep that window", with the "#841's sibling issue tracks `save_unseal_keys` and `eab::write_key_file`; those two are left to it" clause gone. The sentence keeps the rationale it was there for and no longer claims two credential writers are still unconverted. Dropping the cross-reference rather than rewording it was the right of the two options.
- **10 — Resolved.** `a36ef7d` carries a real wrapped body and a standalone `Part of #840` trailer, and the two commits with literal `\n` escapes are gone from the branch. The three commits that remain are all clean.
- **11 — Resolved.** `CHANGELOG.md:143-152` reflows to 64-73 columns with no stub line left mid-paragraph, and `src/fs_util.rs:1707-1709` now parses on the first read ("Covers all three `StagedMode` arms in one isolated run rather than splitting them across three"). **Accepting the pushback** on `ensure_secrets_dir` narrowing an operator-named `--eab-file-path` parent — I raised it explicitly as not-this-PR, and leaving it is what I asked for. No issue was opened.

The conversion is unchanged from Round 3 and I still have no concerns about it. Three things this round, all in prose, two of them the same class the issue's fourth criterion is about.

### 12. `StagedDurability::RenameOnly` still cites `eab.json` as its example (`src/fs_util.rs:795-797`)

> `RenameOnly` — Rename and stop: a crash that loses the new directory entry leaves the previous file in place and costs a rewrite — a reissued certificate, **the next sync's `eab.json`** — rather than an outage.

After this PR, the only `eab.json` that reaches `StagedDurability` at all reaches it as `FlushDirectory`: `src/eab.rs:119` takes `atomic_write`, which is `StagedDurability::FlushDirectory`. The other `eab.json` — the relocated one — is written by `fs_util::write_owned_file_replace`, which is not a `publish_staged_blocking` caller and never constructs a `StagedDurability` at all, so it cannot be what this arm is illustrating either. I checked the other side too: no `atomic_replace*` call site in the crate writes an `eab.json`. So the enum's rename-only arm now names, as its one credential example, a file its sibling arm owns.

The citation was loose before this PR — it was borrowing `write_owned_file_replace`'s rationale at `src/fs_util.rs:232-235`, which is a real and correct decision but about a different writer. What this PR changes is that the borrowing is now actively contradicted three files away by its own call site. `src/fs_util.rs` is one of the four files the fourth acceptance criterion names; the PR corrected that file's `publish_staged_blocking` rustdoc and left this enum seventy lines above it. Dropping the `eab.json` clause, or swapping it for a rendered `ca.json` or a compose override, is the whole fix.

### 13. The manual's flushed list now claims `eab.json` flat, and that is false for the relocated copy

`docs/en/cli.md:138-139`:

> - every `AppRole` `role_id`/`secret_id`, `eab.json`, the OpenBao unseal keys, and the remote bootstrap artifact

under **"Flushed, so the published file survives a power loss"**. `docs/ko/cli.md:130-131` is the same line.

`eab.json` is the one entry in that list with two writers that disagree. The secrets-tree copy this PR converted is flushed. The copy an operator relocates into an agent-owned directory with a path override goes through `fs_util::write_owned_file_replace`, whose rustdoc at `src/fs_util.rs:232-235` says the flush is *deliberately* skipped:

> The containing directory is deliberately *not* flushed after the rename. `eab.json` is rewritten on every sync, so a crash that loses the new directory entry costs the next sync's rewrite and nothing else — not worth a disk round trip on every write.

The neighbouring entries do not have this problem: both relocated-credential writers for `role_id` and `secret_id` (`create_owned_credential_noclobber`, `atomic_rewrite_owned_no_symlink`) flush, so the flat claim holds for them.

Before this PR `eab.json` appeared in neither list, which was a gap. Adding it unqualified turns the gap into a statement that is wrong for exactly the operator most likely to be reading that section — the one who used the relocation documented at `docs/en/cli.md:1070`, and who is now told a file survives a power loss that is documented in the crate as not doing so. Qualifying the entry ("the `eab.json` beside each service's `secret_id`") and naming the relocated copy in the not-flushed list below would close it in both manuals.

### 14. Smaller things

- **`write_key_file`'s doc still claims it mirrors `write_secret_file`** (`src/eab.rs:85-87`): "mirroring the bootstrap `write_secret_file` behaviour so a first-sighting re-write from the fast-poll loop is byte-identical to the bootstrap writer". The byte claim still holds and the behaviour claim no longer does — Round 2's item 7 is precisely the divergence, and leaving `write_secret_file` alone was the right call, which makes the divergence deliberate rather than accidental. That is the reason to say it here: the next reader who notices `src/eab.rs:105-116` has no counterpart at `src/bin/bootroot-remote/io.rs:100-103` and reads this sentence as the contract will restore the symmetry by deleting the new branch. A clause naming the identical-content-at-a-link case as the one place the two differ is enough.
- **`publish_staged_blocking`'s own enumeration** (`src/fs_util.rs:806-813`) lists what goes through it and stops at the compose overrides. The two files this PR added are the newest members and are not in it. Nothing there is false — the sentence is "every writer ... goes through it", which the PR makes *more* true, and the list reads as illustrative. But the equivalent enumerations in `CHANGELOG.md:143-144` and `docs/en/cli.md:121` were both updated to name them, so this is the one list in the file the criterion names that was left behind. Optional.

Nothing above is in the writers. Both still publish through `atomic_write` with `StagedMode::Policy(KEY_FILE_MODE)` at the path they are given, the error contexts are byte-identical to what they replaced, and the four tests remain the right ones — each fails on the code it replaced, and the identical-content-at-a-link branch is covered by a test that pins both halves of its semantics (the link is replaced, the target's widened `0644` is not repaired). Item 13 is the one I would fix before merge, since it is user-facing and wrong rather than merely stale; 12 and 14 are one-clause edits alongside it.

PR hygiene is fine: `Closes #840` is right for a PR that meets every acceptance criterion, the ownership caveat is in the body where a merger will see it, the test plan is a real checklist, and no GitHub issue was created for any of the deferred items.

sehkone:

sehkone:
**[Author Round 4]**

12. **Fixed** — Corrected StagedDurability::RenameOnly to use regenerated ca.json rather than eab.json, which now uses durable FlushDirectory publishing. I also added the OpenBao unseal keys and EAB files to the publish_staged_blocking enumeration.

13. **Fixed** — Updated both CLI manuals to distinguish the flushed eab.json beside a service secret_id from a relocated eab.json, which deliberately skips the directory flush because the next sync rewrites it.

14. **Fixed** — Updated the write_key_file doc comment to state that ordinary first-sighting payloads match bootstrap write_secret_file, while identical content at a final symlink deliberately republishes to replace that link rather than repairing its target.

Validated formatting, Clippy, rustdoc, Markdown lint, the strict documentation build, and the full core suite. The Docker preflight passed its quality gate, deploy smoke, and non-hosts lifecycle and rotation coverage; hosts variants require unavailable passwordless sudo, while this host occupied fixed ports and lack of Linux docker0 prevent the remaining local E2E cases. CI will run those in its provisioned environment.

sehkone:
**[Reviewer Round 5]**

Round 4's items 12–14 are addressed. Taking them in order:

- **12 — Resolved.** `src/fs_util.rs:797-798` now reads "a reissued certificate or regenerated `ca.json`", so `StagedDurability::RenameOnly` no longer illustrates itself with a file its sibling arm owns. Adding the two files to the `publish_staged_blocking` enumeration (`src/fs_util.rs:810-811`) was the right call too — see item 15, which is partly *why* the plural "EAB files" there is accurate.
- **13 — Addressed, but the line landed on the wrong axis.** See item 15. The intent is right and the gap it was closing was real; the distinction the fix draws does not separate the two files.
- **14 — Resolved.** `src/eab.rs:85-90` names the identical-payload-at-a-link case as the deliberate exception, so the next reader who notices `src/eab.rs:108-119` has no counterpart at `src/bin/bootroot-remote/io.rs:100-103` will not read the divergence as an oversight. One nit below.

The conversion itself is unchanged and I still have no concerns about it: both writers publish through `atomic_write` with `StagedMode::Policy(KEY_FILE_MODE)` at the path they are given, the error contexts are byte-identical to what they replaced, and the four tests are the right ones.

### 15. The flushed/not-flushed split for `eab.json` is drawn by location, and the two files are not distinguishable by location

`docs/en/cli.md:138-140` and `docs/en/cli.md:147` (and `docs/ko/cli.md:130-131`, `:136`) now say:

> Flushed: ... the `eab.json` beside each service's `secret_id` ...
> Not flushed: ... a relocated `eab.json`, which the next sync rewrites

A relocated `eab.json` *is* the one beside the service's `secret_id`. `service_eab_file_path` (`src/commands/service.rs:48-53`) derives it as `secret_id_path.parent().join("eab.json")`, and `is_override` is just `secret_id_path_override.is_some()` (`src/commands/service/local_config.rs:37`) — an override `--secret-id-path` moves the `secret_id` and its `eab.json` together. So the two bullets name the same file by two descriptions, and an operator who relocated a service credential cannot tell which bullet is theirs.

What actually differs is the writer, and after this PR the relocated path has two:

- `bootroot service add` provisions it once through `fs_util::write_owned_file_replace` (`src/commands/service/local_config.rs:350`), which deliberately skips the directory flush.
- The running agent rewrites the *same path* on every EAB change through `bootroot::eab::write_eab_file` (`src/fast_poll.rs:1244`) — the writer this PR just converted to `atomic_write`, i.e. `FlushDirectory`. The agent gets that path from `--eab-file`, which `print_daemon_run_snippet` prints as `service_eab_file_path(secret_id_path)` (`src/cli/output.rs:386-392`), the relocated one included.

So the "not flushed" entry is falsified by this PR's own change for every write after the first: a relocated `eab.json` is unflushed only in `service add`'s initial provisioning, and flushed on every later sync rewrite. Before this PR neither eab.json write flushed anything, so this is new rather than inherited — the same class as the fourth acceptance criterion, and the reason it is worth another pass rather than leaving as-is.

This also corrects my own Round 4 framing: I wrote that the relocated copy "goes through `fs_util::write_owned_file_replace`", which is true of provisioning and incomplete for the sync path. The fix followed what I said rather than what the code does, so this one is on me to restate, not a moved goalpost.

Two ways out, either fine:

- Say it by writer: keep `eab.json` in the flushed list and add one clause — `service add`'s first write of a relocated `eab.json` skips the flush, every later rewrite by the agent takes it.
- Or drop the relocated entry from the not-flushed list and leave the flushed one unqualified, which is what the file's steady state is.

What does not work is the current pair, which reads as two files with two guarantees when it is one file with two writers.

Note this is also why `src/fs_util.rs:810-811`'s plural "the `OpenBao` unseal keys and EAB files" is right as written — both copies reach `publish_staged_blocking`, through different writers. That sentence and the manual currently disagree about how many EAB files the staged publisher touches.

### 16. `CHANGELOG.md:162-164` over-quantifies the replace-the-link group

> A link at an issued certificate, key or CA bundle path is replaced by the published file instead. **Every credential path** — including the OpenBao unseal-keys and ACME EAB files — now has that behavior, as does the control node's own `agent.toml`.

Not every credential path. A relocated `role_id`/`secret_id` **refuses** a link rather than replacing it: `create_owned_credential_noclobber` treats "a pre-existing regular file or a symlink planted at the target" as a hard error (`src/fs_util.rs:207-210`), and `atomic_rewrite_owned_no_symlink` rejects one outright at rotation time. The released entry at `CHANGELOG.md:1163-1174` already says exactly that — "created no-clobber and never follow a final-component symlink", "a symlink-rejecting writer" — so the unreleased entry now contradicts a shipped one three sections above it, and an operator with a link at a relocated `secret_id` is told the write replaces it when the command in fact fails.

This is new in this PR: the sentence replaced "which is what the key file and the control node's own `agent.toml` have always done and what the certificate beside them now matches", which claimed nothing about the override writers. Naming the group the sentence is actually about ("every credential bootroot publishes by rename", with the relocated `role_id`/`secret_id` refusal called out) fixes it. While there: "now has that behavior" reads as newness for the whole group, when only the two files this PR converted are new to it.

The manual's equivalent bullet (`docs/en/cli.md:197-203`, "every credential are published at the path itself") has the same shape, but that wording predates this PR and I am not asking for it here.

### 17. Nit

`src/eab.rs:87-90` — "A first-sighting fast-poll rewrite has the same payload as bootstrap `write_secret_file`. An identical payload reached through a final symlink is the deliberate exception". The second sentence's "exception" attaches to a sentence about payload bytes, so what it is an exception *to* is the byte claim rather than the behavioural one, and `write_secret_file`'s own identical-content branch is no longer mentioned. The intent came through and I would not hold merge for it; naming the divergence explicitly ("`write_secret_file` repairs the target's mode here; this writer republishes") would leave nothing to infer.

---

Nothing this round touches the writers, the tests, or the symlink decision, all of which I consider settled. Item 15 is the one I would fix before merge — it is user-facing, it is wrong in the steady state, and it is wrong because of a change in this diff. Item 16 is a one-sentence edit alongside it.

PR hygiene is fine: `Closes #840` is right for a PR that meets every acceptance criterion, the ownership caveat is in the body where a merger will see it, the test plan is a real checklist, and no GitHub issue was created for any deferred item.

sehkone:

sehkone:
**[Author Round 5]**

15. **Fixed** — The English and Korean manuals now describe the EAB durability split by writer, not location: the relocated `service add` provision intentionally skips the directory flush, while a later agent EAB update republishes and flushes that same path.

16. **Fixed** — Narrowed the changelog to the credentials bootroot publishes by rename. It records that the unseal-key and EAB paths newly join that group, while relocated `AppRole` `role_id` and `secret_id` paths reject a final link instead.

17. **Fixed** — Reworded the `write_key_file` comment to name the intentional identical-symlink divergence explicitly: bootstrap `write_secret_file` repairs the target mode; this writer republishes to replace the link.

Validation: `cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings`, focused EAB and OpenBao-unseal tests, Markdown lint, and `./scripts/check-docs.sh` all pass.

sehkone:
**[Reviewer Round 6]**

Round 5's items 15–17 are addressed. Taking them in order:

- **15 — Resolved.** The manual now splits the EAB durability by writer rather than by location: `eab.json` is back in the flushed list unqualified (`docs/en/cli.md:138-139`, `docs/ko/cli.md:130-131`), the "relocated `eab.json`" entry is gone from the not-flushed list, and `docs/en/cli.md:157-159` / `docs/ko/cli.md:147-149` carry the exception. I checked the claim rather than the wording: `write_local_eab_file` (`src/commands/service/local_config.rs:344-358`) branches on `is_override`, so the relocated copy is the only one `write_owned_file_replace` touches, `service add` is that function's sole caller in the crate, and `service update` never rewrites `eab.json` — so "its initial write" is exact, not an approximation. The agent's later rewrite of the same path goes through `eab::write_eab_file` → `atomic_write` → `FlushDirectory`, which is what the second sentence says. Both halves hold.
- **16 — Addressed, with one clause left over.** See item 18. The correction itself is right: the relocated `role_id`/`secret_id` refusal is now named, and it no longer claims a replace-the-link behaviour for writers that reject one.
- **17 — Resolved.** `src/eab.rs:87-90` now names the divergence on both sides ("`write_secret_file` repairs the target's mode while this writer republishes to replace the link"), so the identical-payload branch at `src/eab.rs:108-119` cannot be read as an accident by someone comparing it to `src/bin/bootroot-remote/io.rs:100-103`.

Two things left this round, both one-clause prose edits, neither of which I would hold the merge for.

### 18. `CHANGELOG.md:162-165` now reads as if the control node's `agent.toml` newly joined the group

> The OpenBao unseal-keys and ACME EAB files **now join** the credential paths bootroot publishes by rename, **as does** the control node's own `agent.toml`.

"As does" attaches to "now join", and the whole entry is in the `now`-register of `## [Unreleased]`, so a reader takes all three files as new to this behaviour. The sentence this replaced said the opposite about that one file — the key file and the control node's `agent.toml` "have always done" it — and the manual you wrote in the same commit keeps the distinction explicitly (`docs/en/cli.md:200-202`: "The OpenBao unseal-keys and secrets-tree EAB files joined this group in the current release; the other files have been published by rename for several releases"). So the changelog and the manual now disagree about which files are new to the group, and only two of the three are.

This is small, and it is the same class as item 16 rather than a separate problem: the fix narrowed the group correctly and lost the age distinction that was in the sentence before it. Splitting the "as does" clause off ("…publishes by rename, alongside the control node's own `agent.toml`, which has published that way for several releases") is the whole of it.

### 19. Nit: `write_owned_file_replace`'s rustdoc still frames the every-sync rewrite as its own

`src/fs_util.rs:226-236`:

> Used for a relocated `eab.json`, which is (re)written on every sync. […] The containing directory is deliberately *not* flushed after the rename. `eab.json` is rewritten on every sync, so a crash that loses the new directory entry costs the next sync's rewrite and nothing else.

The *rationale* survives this PR intact — the path is rewritten on every sync, so losing the provisioning write costs one sync and no more. What has changed is who performs those rewrites and with what guarantee: `write_owned_file_replace` has exactly one caller (`src/commands/service/local_config.rs:350`, `service add`'s provisioning), and every later write to that path is `eab::write_eab_file`, which now takes the flush. So the crate's rustdoc still conflates the two writers on that path in exactly the way item 15 just stopped the manual from doing, and the asymmetry it now hides — first write unflushed, every later write flushed — is created by this diff.

I am flagging it as the last of the five-round prose thread rather than as a defect: nothing there is false, and the framing predates this PR. `src/fs_util.rs` is one of the four files the fourth acceptance criterion names, though, so if the two lines are cheap to touch it is the right place for the same clause the manual got.

---

Nothing this round touches the conversion, which I consider settled and have re-checked once more against the issue rather than against my own earlier rounds:

- Both sites publish through `fs_util::atomic_write` with `StagedMode::Policy(fs_util::KEY_FILE_MODE)` — `StagedOwner::Destination` + `StagedDurability::FlushDirectory` — at the path they are given, with the symlink decision recorded at each site (`src/commands/openbao_unseal.rs:28-29`, `src/eab.rs:120-121`). No `set_permissions` follows either write; the one `set_key_permissions` left in `write_key_file` is the mode repair on a file it did not create, labelled as such.
- Error contexts are byte-identical to what they replaced at both sites, and neither payload changed (`keys.join("\n") + "\n"` still round-trips through `read_unseal_keys`).
- Relative destinations still work — `parent_dir` (`src/fs_util.rs:53-57`) falls back to `.`, so a relative `--eab-file-path` stages in the right directory rather than erroring, which a rename-based writer could easily have regressed.
- The four tests are the right ones. The two symlink tests pin both halves of the guarantee (link replaced by a regular file holding the new secret; former target untouched, including its widened `0644` in the identical-payload case), and the two umask tests go through `umask_test_ran_in_child`, which asserts `test result: ok. 1 passed` rather than the exit status — so a drifted test-name string fails loudly instead of vacuously passing.

PR hygiene is fine: `Closes #840` is right for a PR that meets every acceptance criterion, the ownership caveat (a replaced link still donates its former target's uid/gid to the published file) is in the body where a merger will see it, the test plan is a real checklist, and no GitHub issue was created for any of the items deferred across the six rounds.

sehkone:
[Review Verdict Round 6: APPROVED]

sehkone:
[Review Verdict Round 6: APPROVED]

sehkone:
**[Reviewer Unresolved Round 6]**

The review loop ended after Round 6's approval, so the two items I raised there were never answered. Both are prose only, both are one-clause edits, and neither blocked the approval — recording them so they are not lost with the thread.

### 18. `CHANGELOG.md:162-165` reads as if the control node's `agent.toml` newly joined the rename-published group

> instead. The OpenBao unseal-keys and ACME EAB files **now join** the credential paths bootroot publishes by rename, **as does** the control node's own `agent.toml`.

"As does" attaches to "now join", and the entry sits in the `now`-register of `## [Unreleased]`, so all three files read as new to the behaviour. Only two are. The sentence this replaced said the opposite about that one file ("have always done"), and `docs/en/cli.md:200-202`, written in the same commit, keeps the distinction: "The OpenBao unseal-keys and secrets-tree EAB files joined this group in the current release; the other files have been published by rename for several releases." The changelog and the manual now disagree about which files are new.

Fix: split the clause off — "…publishes by rename, alongside the control node's own `agent.toml`, which has published that way for several releases."

### 19. `src/fs_util.rs:226-236` still frames the every-sync rewrite as `write_owned_file_replace`'s own

> Used for a relocated `eab.json`, which is (re)written on every sync. […] The containing directory is deliberately *not* flushed after the rename. `eab.json` is rewritten on every sync, so a crash that loses the new directory entry costs the next sync's rewrite and nothing else.

The rationale survives this PR — the path is rewritten on every sync, so losing the provisioning write costs one sync and no more. What changed is who performs those rewrites: `write_owned_file_replace` has one caller (`src/commands/service/local_config.rs:350`, `service add`'s provisioning), and every later write to that path is `eab::write_eab_file`, which now takes the flush. The rustdoc still conflates the two writers on that path in the way item 15 stopped the manual from doing, and the asymmetry it hides — first write unflushed, every later write flushed — is created by this diff. `src/fs_util.rs` is one of the four files the fourth acceptance criterion names.

Fix: the same clause the manual got, naming `eab::write_eab_file` as the writer of the later rewrites.

---

Nothing else is outstanding. The conversion, the tests and the symlink decision are settled, and the items deferred in earlier rounds were deferred deliberately and with my agreement — `write_secret_file`'s identical-content branch (Round 2, item 7) and `ensure_secrets_dir` narrowing an operator-named `--eab-file-path` parent (Round 3, item 11) are both outside #840's two call sites, and no GitHub issue was opened for either.

sehkone:
## Suggested squash commit

**Title**

```text
Secure unseal and EAB writes
```

**Body**

```text
Write OpenBao unseal keys and secrets-tree EAB credentials at their
final 0600 mode before atomically publishing them. This eliminates the
temporary broad-read window and prevents readers from seeing partial
secrets after a crash.

Preserve bootroot-owned path semantics by replacing final symlinks.
Regression tests cover restrictive modes and preservation of link targets.

Closes #840
```

PR #860, opened 2026-08-15
Run the E2E run-scope validator in CI

### LINKED ISSUE #857: Run the E2E run-scope validator in CI
`scripts/validate-e2e-run-scope.sh` is the only guard the per-run E2E identity has, and nothing in CI runs it. It arrived with #846 (PR #854), which had to leave it out: that issue ruled out changing CI, and adding a step to the `check` job is a workflow change.

It runs today only from `scripts/preflight/run-all.sh`, which is a process control — `AGENTS.md` requires a preflight before pushing, and a control that depends on remembering is exactly the kind CI exists to replace.

## Why a green matrix is not a substitute

None of what the validator checks is observable from the E2E matrix. Each CI runner gets one run on an empty host, so a derivation that started colliding, a marker that stopped being written, or a sweep that began reaching past the runs it recorded would all leave every arm green. It surfaces instead as one developer's run deleting another's containers, or as a host filling with the leftovers of runs that were killed.

PR #854 is the evidence: all 18 checks passed on that tree, and every one of the validator's 45 assertions is about behaviour none of those 18 could see.

## Where it goes, and why not a job of its own

Into the `check` job (Quality Check) in `.github/workflows/ci.yml`, beside the four validators already there:

```yaml
      - name: Validate E2E Run Scope
        run: ./scripts/validate-e2e-run-scope.sh
```

after the existing `Validate E2E Leftover Check` step and before `Install cargo-audit`.

It costs **3.6 seconds**, measured. That is less than two of the validators already in that job — `validate-e2e-openssl-compat.sh` takes 9.9s and `validate-e2e-leftover-check.sh` 9.3s (`validate-compose-instance-names.sh` takes 0.3s).

`check` is on the critical path: both `test-core` and `test-docker-e2e-matrix` declare `needs: [check, changes]`, so every second there delays the whole matrix. On the last full run on `main` the job took 67s and the run took about 5m21s, so 3.6s is roughly 1.1% of wall clock.

A separate parallel job would remove even that, and is still the wrong trade: a fresh `ubuntu-latest` job spends 15-30s on runner acquisition and checkout to save 3.6s, and burns a runner doing it. Co-located in `check` the setup cost is zero, because the validator needs nothing that job has not already done.

Note that `check` has no docs-only gate — unlike `test-core` and `test-docker-e2e-matrix`, which carry `if: needs.changes.outputs.docs_only != 'true'` — so this step will also run on docs-only pushes. At 3.6s that is accepted rather than worked around; adding a gate for it would be more machinery than the thing it guards.

## What the step needs

Nothing beyond what the `check` job already provides: `bash`, a full `actions/checkout` (the validator greps `src/commands/compose_project.rs` to confirm the Rust test pinning the resolver ranking is still there), and at least one of `flock(1)`, `perl` or `python3` — `ubuntu-latest` has all three, and the validator exercises every one it finds. No Rust toolchain, no Docker, no cache.

It is safe to run beside anything: the validator redirects its marker directory and its hosts lock into its own `mktemp -d` (`BOOTROOT_E2E_RUN_MARKER_DIR`, `BOOTROOT_E2E_HOSTS_LOCK`), so it cannot touch the runner's `/etc/hosts` or any other job, and `docker` is a stub on `PATH` rather than the real daemon.

## The documentation the step contradicts

`docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md` both describe the validator as a local check that `scripts/preflight/run-all.sh` runs and no CI workflow does, and list it in their local-only extras table. The step makes both false, so each page moves it into the CI-equivalent table against its `check` step and rewrites the paragraph that calls it local. Nothing else in the manual changes.

## Deliberately not included

- **Making it a required check.** That is a branch-protection setting rather than a change to any file in this repository, and it needs admin rights on `aicers/bootroot`. This issue delivers the workflow step; flipping the required-check list is a maintainer action afterwards, and the issue is complete without it.
- **Changing what the validator asserts.** It passes on `main` today. This is about running it, not about extending it.
- **Moving or regrouping the four validators already in `check`.** They stay exactly where they are.
- **A `CHANGELOG.md` entry.** CI and test-harness only; a user of the last release cannot observe it.

## Test plan

- The step's command passes on a clean checkout of `main`: `./scripts/validate-e2e-run-scope.sh` exits 0 and reports its 45 `ok:` lines (one per assertion, on a non-root host).
- The step fails the job when what it guards is broken. Mutate one guarded behaviour — for example change the hosts lock's default away from `/etc/hosts` in `scripts/impl/lib/run-scope.sh`, or drop the pid-liveness filter on the holder label — and confirm the `check` job goes red with a message naming the guard. Revert.
- A full CI run completes, and the `check` job's duration is within a few seconds of what it was.
- The step runs on a docs-only push too, and passes there.

## Acceptance criteria

- [x] `.github/workflows/ci.yml` runs `./scripts/validate-e2e-run-scope.sh` in the `check` job, after `Validate E2E Leftover Check` and before `Install cargo-audit`
- [x] The step adds no toolchain setup, no service and no cache to that job
- [x] A break in the behaviour the validator guards turns the `check` job red
- [x] The four validators already in `check` are unchanged, and no other job is added or reordered
- [x] `check` job duration grows by no more than a few seconds
- [x] `docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md` no longer call the validator local-only, and list it against its `check` step
- [x] No `CHANGELOG.md` entry
- [x] Full CI passes

sehkone:
Adds `Validate E2E Run Scope` to the `check` job, after `Validate E2E Leftover Check` and before `Install cargo-audit`, and corrects the two documentation places that said the validator runs in no CI workflow.

`scripts/validate-e2e-run-scope.sh` was the per-run E2E identity's only guard, and nothing but `scripts/preflight/run-all.sh` invoked it — a control that depends on remembering. None of what it asserts is observable from the E2E matrix: each runner gets one run on an empty host, so a derivation that started colliding, a marker that stopped being written, or a sweep that began reaching past the runs it recorded would all leave every arm green.

The step adds no toolchain setup, no service and no cache. The validator redirects its marker directory and its hosts lock into its own `mktemp -d`, and stubs `docker` on `PATH`, so it cannot touch the runner's `/etc/hosts` or anything another job holds.

## Documentation

`docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md` both stated the validator "is a local check ... not run by any CI workflow", and listed it under local-only extras. Those sentences are now false, so each was rewritten to say where it runs and what it needs, and the entry moved from the local-only table into the CI-equivalent one.

No `CHANGELOG.md` entry: CI and test-harness only, so a user of the last release cannot observe it. No Rust code changed and no E2E code path is touched, so the Docker preflight scripts were not run locally — CI's matrix covers them.

## Test plan

- [x] `./scripts/validate-e2e-run-scope.sh` exits 0 on this tree, reporting 45 `ok:` lines, in 3.2s measured on a non-root host — and the same on a detached worktree of `origin/main` (exit 0, 45 `ok:`, 3.4s), which is the clean checkout the issue asks about.
- [x] A break in the guarded behaviour turns the step red, checked on two separate guards and reverted after each, leaving `git status` clean.
    - Hosts lock default in `scripts/impl/lib/run-scope.sh` changed from `/etc/hosts` to a path of the harness's own: exit 1, `FAIL: the default hosts lock is '/tmp/broken-lock' rather than /etc/hosts; a lock at a path of the harness's own can be swapped for another inode between the check and the open, and one under $TMPDIR is not even machine-wide`.
    - Pid-liveness filter dropped from `hosts_lock_label` (`run_pid_is_alive "$pid" || return 0` removed): exit 1, `FAIL: the refusal named a dead run's label as the holder: ... (pid 4194305, a run that was killed)`.
- [x] The `check` job's step order is `... Validate E2E Leftover Check`, `Validate E2E Run Scope`, `Install cargo-audit`, `Run Security Audit`, and the step declares only `run:` — no toolchain setup, service or cache.
- [x] The four validators already in `check` are unchanged, and no job is added or reordered — the workflow diff is three added lines.
- [x] `./scripts/check-docs.sh` passes (`mkdocs build --strict` plus the theme assertions); `markdownlint-cli2` reports 0 issues and `biome ci --error-on-warnings .` is clean.
- [x] No `CHANGELOG.md` entry.
- [x] Full CI passes on this pull request — all 18 checks green. The new step took **2s** on the runner (`Validate E2E Run Scope`, 23:19:01 -> 23:19:03), less than the `Validate E2E Leftover Check` step above it.
- [x] The step runs and passes on a docs-only push. No single run can show this from here — the branch changes `.github/workflows/**`, which the change filter counts as code — so it was checked against two real runs instead of argued from the workflow alone. That the `check` job runs in full on a docs-only push is observed on [run 30968767229](https://github.com/aicers/bootroot/actions/runs/30968767229) (PR #791, which touched only two files under `docs/`): `Change Filter` resolved `docs_only=true`, `Unit & CLI Smoke` and every E2E matrix arm skipped, and `Quality Check` still ran to success with each of its validator steps green. That the new step passes inside that job is observed on this pull request's own run. The step cannot distinguish the two cases: it declares no `if:`, and it reads `scripts/impl/lib/run-scope.sh` and `src/commands/compose_project.rs` out of the full checkout rather than anything about the diff.

Closes #857

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,9 @@ jobs:
       - name: Validate E2E Leftover Check
         run: ./scripts/validate-e2e-leftover-check.sh
 
+      - name: Validate E2E Run Scope
+        run: ./scripts/validate-e2e-run-scope.sh
+
       - name: Install cargo-audit
         run: cargo install cargo-audit || true
 
--- a/docs/en/e2e-ci.md
+++ b/docs/en/e2e-ci.md
@@ -278,8 +278,13 @@ world-writable directory is one a run's write would truncate wherever
 somebody else's link pointed.
 
 `scripts/validate-e2e-run-scope.sh` covers the derivation, the markers, the
-sweep and the `hosts` lock without Docker. It is a local check, run by
-`scripts/preflight/run-all.sh` and not by any CI workflow.
+sweep and the `hosts` lock without Docker, and runs in the `check` CI job. It
+needs nothing that job does not already have, and redirects its marker
+directory and its `hosts` lock into a temporary directory of its own, so it
+touches neither the runner's `/etc/hosts` nor anything another job holds.
+None of what it asserts is visible to the E2E matrix: each runner gets one
+run on an empty host, so a derivation that started colliding or a sweep that
+began reaching past its own runs would leave every arm green.
 
 ### Leftover containers fail a run before it starts
 
@@ -812,6 +817,7 @@ Or run individual scripts:
 | `./scripts/validate-compose-instance-names.sh` | `ci.yml` Validate Compose Instance Names |
 | `./scripts/validate-e2e-openssl-compat.sh` | `ci.yml` Validate E2E OpenSSL Compatibility |
 | `./scripts/validate-e2e-leftover-check.sh` | `ci.yml` Validate E2E Leftover Check |
+| `./scripts/validate-e2e-run-scope.sh` | `ci.yml` Validate E2E Run Scope |
 | `./scripts/preflight/ci/deploy-no-build-smoke.sh` | `ci.yml` Deploy Compose No-Build Smoke |
 | `./scripts/preflight/ci/test-core.sh` | `ci.yml` Unit & CLI Smoke |
 | `./scripts/preflight/ci/e2e-matrix.sh` | `ci.yml` Docker E2E Matrix |
@@ -825,7 +831,6 @@ Local-only extras (not in any CI workflow):
 
 | Script | Description |
 | --- | --- |
-| `./scripts/validate-e2e-run-scope.sh` | Run-scope derivation, markers, sweep, `hosts` lock |
 | `./scripts/preflight/extra/agent-scenarios.sh` | Agent scenario tests |
 | `./scripts/preflight/extra/cli-scenarios.sh` | CLI scenario tests |
 
--- a/docs/ko/e2e-ci.md
+++ b/docs/ko/e2e-ci.md
@@ -274,8 +274,13 @@ pid가 살아 있는 것은 건너뛰고, 죽은 것은 그 인스턴스의 컨�
 그것을 잘라 버리게 되는 경로이기 때문입니다.
 
 `scripts/validate-e2e-run-scope.sh`가 Docker 없이 이름 생성, 마커, 수거,
-`hosts` 잠금을 모두 확인합니다. 이 스크립트는 로컬 검사이며,
-`scripts/preflight/run-all.sh`가 실행하고 CI 워크플로에서는 실행되지 않습니다.
+`hosts` 잠금을 모두 확인하며, CI의 `check` 잡에서 실행됩니다. 그 잡이 이미
+갖춘 것 외에는 아무것도 필요하지 않고, 마커 디렉터리와 `hosts` 잠금을 자신이
+만든 임시 디렉터리로 돌려놓으므로 러너의 `/etc/hosts`도 다른 잡이 쥔 것도
+건드리지 않습니다. 이 스크립트가 확인하는 내용은 E2E 매트릭스에서는 보이지
+않습니다. 러너마다 빈 호스트에서 한 번씩만 실행하므로, 이름 생성이 충돌하기
+시작하거나 수거가 자기 실행 너머까지 손을 뻗기 시작해도 모든 갈래가 초록으로
+남습니다.
 
 ### 남은 컨테이너는 실행이 시작되기 전에 실패시킵니다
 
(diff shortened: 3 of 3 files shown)


sehkone:
**[Reviewer Round 1]**

No findings. The workflow change adds `Validate E2E Run Scope` in `.github/workflows/ci.yml` immediately after `Validate E2E Leftover Check` and before `Install cargo-audit`, exactly as the issue requires. It invokes the existing self-contained validator directly, without adding a job, toolchain setup, cache, or service.

The documentation updates in `docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md` are consistent with that behavior: they remove the stale local-only claim and map the script to the new `check` step. The diff is appropriately limited to the workflow and the two required documentation pages; the PR body also has the required issue linkage and a concrete test plan.

sehkone:

PR #861, opened 2026-08-16
Correct the record on staged publishing (#849)

### LINKED ISSUE #849: Correct the record on what staged publishing changed
## Problem

#841 replaced truncate-in-place with stage-then-rename across the crate. The manuals' **How bootroot writes files** section documents three consequences of that (no partial reads, final mode from the first moment, a new inode per write), but the conversion changed four more semantics that are recorded nowhere, and the changelog files the entire change under `### Fixed` although parts of it are behavior changes and one part is a security improvement. Nothing has shipped since 0.2.0, so every correction below is an edit in place.

The four undocumented semantic changes, each inherent to publishing by rename rather than a defect to fix:

1. **The permission that gates a write moved from the file to the directory.** A truncating write needed write permission on the *file*; a rename needs write permission on the *directory* and none on the file. Two setups flip behavior: a writable file in a read-only directory could be updated before and now fails at temp-file creation, and — the sharper one — a file an operator made read-only (`chmod 444`) as a write lock no longer blocks anything. The write succeeds by replacement. Directory permissions and `chattr +i` are the boundaries that still hold.
2. **Hard links detach.** A truncating write propagated new contents to every hard link of the inode; a rename installs a fresh inode at one name and every other link keeps the old contents. Symlinks got a per-writer decision and eleven review rounds; hard links got nothing and are simply not supported at bootroot-written paths.
3. **ACLs, xattrs and SELinux contexts are not preserved.** They lived on the inode, so truncation kept them; the staged temporary starts clean and only the mode bits (`& 0o7777`) are carried over. An ACL an operator attached to `state.json` or `.env` silently disappears at the next write. On SELinux hosts the fresh inode takes the directory's default labeling rather than a customized file label.
4. **A crash leaves a `.tmpXXXXXX` sibling, not a torn destination.** Strictly an improvement, but an operator finding an unfamiliar `0600` temp file beside `state.json` after a power loss should be able to look it up.

Two smaller items ride along because they touch the same code and the same section:

5. When a bare (non-`_through_symlink`) publish renames over a destination that is currently a symlink, the operator's link is destroyed silently. That is the decided behavior for credentials and for the cert triple, but decided is not the same as silent: nothing today distinguishes "replaced a regular file" from "replaced somebody's link" in the diagnostics.
6. `publish_staged_blocking`'s chown-failure context names the uid/gid it tried to preserve but not what to do about it.

### 1. Manual additions

Extend **How bootroot writes files** in `docs/en/cli.md` and `docs/ko/cli.md` with items 1–4 above, in the same register as the three consequences already listed: what changed, which operator arrangement it affects, and what the supported boundary is (directory permissions / `chattr +i` for write-locking; no hard links at bootroot-written paths; ACLs and custom labels do not survive a write). Add one sentence to `publish_staged_blocking`'s rustdoc stating it preserves mode bits only.

The warning added in §3 belongs in the same manuals, but in the symlink section rather than among items 1–4: that section is where an operator looks to find out what a link at one of these paths costs, and the changelog announces the warning, so manuals that said nothing about it would contradict the release notes.

### 2. Changelog restructure, in place

The single `### Fixed` mega-entry (currently `CHANGELOG.md:120` onward) splits by what a reader of the release notes needs:

- **`### Security`** — the permission-window closure: the CA password, recovery keys, responder HMAC and every `secret_id` reaching their final mode before the name appears. (Joins the `.env` `0600` entry #848 adds.)
- **`### Changed`** — the semantics an operator can observe and may have arranged around: the symlink answer at the issued cert/key/bundle and control-node `agent.toml` paths, the new-inode consequences (open fds, single-file bind mounts), and items 1–3 above.
- **`### Fixed`** — the torn-read elimination itself and the `reinit` cyclic-symlink preflight, which remain fixes.

No entry gains or keeps an issue/PR reference, and no new entry describes the interim forms — no released user saw them.

### 3. Warn when a publish replaces a symlink

In the bare staged-publish path, detect via `symlink_metadata` that the destination about to be renamed over is a symlink, and emit one `tracing::warn!` naming the destination and the target the link pointed to. The `_through_symlink` wrappers resolve before publishing, so the path they hand the core is never a link and they are exempt by construction — no flag needed.

The two halves sit on opposite sides of the rename. The destination is read *before* it, because that is the last moment the link is still there to read; the warning is emitted *after* the rename has succeeded, so a publish that fails there leaves the operator's link standing and says nothing about having replaced it.

Extract the detection as a small testable helper returning `Option<PathBuf>` (the link's target) and unit-test it on a link, a regular file, a missing destination, and a dangling link — a link whose target does not exist is one the publish destroys just the same, and `symlink_metadata` sees it without following it. Pin the log line itself as well, through the two spellings an operator actually reaches: a subscriber-capture test asserting that the bare publish warns exactly once naming both ends of the link, and that the `_through_symlink` publish over the same arrangement stays quiet while still delivering its bytes.

### 4. Chown-failure guidance

Append to the existing `Failed to preserve existing uid={} gid={} on {}` context what the failure means and the two ways out: the writer does not have privilege to re-own the staged file (needs `CAP_CHOWN`/root), so either run the command as the destination's owner or re-own the destination to the writing user. Message text only; the failure stays a failure.

## Constraints

- Lands **after** #848: both edit the same changelog entry, and this issue restructures the corrected text, not the current text. Neither blocks the other's implementation, only the merge order is fixed. Preferably after #840 as well: #840 deletes the two-writer exclusion statements from the same changelog entry and manual section this issue restructures. If #840 has not landed when this does, restructure around those statements with their wording untouched — #840 then deletes them from the restructured text.
- Documentation claims must match the code exactly as of this branch; `docs/ko/cli.md` mirrors `docs/en/cli.md` in content, not necessarily sentence-for-sentence.
- No behavior change beyond the one new warn line and the extended error text. In particular the symlink-replacement *behavior* is unchanged — #847 owns any restructuring of that axis, and the warn must survive #847's migration untouched (it lives in the core publish path, which #847 does not move).
- The warn fires on the bare path only; a destination resolved through a link must not produce it. The existing per-writer symlink tests (`rerender_publishes_agent_toml_at_its_own_name` and the wrapper-level set) must pass unchanged.

## Acceptance criteria

- Both manuals document items 1–4, and record the replaced-link warning in their symlink section; `./scripts/check-docs.sh` passes.
- The changelog carries the three sections as scoped, with the torn-read fix still under `### Fixed`, and no issue/PR references.
- Replacing a seeded symlink through a bare wrapper logs one warning naming the link and its former target; publishing through a `_through_symlink` wrapper over the same seeded link logs none. Helper unit-tested for all four destination states.
- The chown-failure message names the remediation; wording present in both the error and its test.
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` clean, full CI green.

## Test plan

- `cargo test` green with the new helper tests.
- `./scripts/check-docs.sh` for both manuals.
- `scripts/preflight/ci/check.sh` end to end. The E2E matrix is not required: the only behavior deltas are a log line and error text, on paths the unit tests pin.

## Not in this issue

- Changing which paths resolve a symlink and which replace one (#847), converting the two unstaged writers (#840), the create-mode corrections (#848, which this one lands after).
- Preserving ACLs/xattrs across a publish. Item 1's documentation states the limitation; implementing preservation is its own change with its own evidence to gather.

sehkone:
## Summary

Staged publishing replaced truncate-in-place across the crate, but four of the semantics it changed were recorded nowhere, and the changelog filed the whole conversion under `### Fixed` although parts of it are behaviour changes and one part is a security improvement. Nothing has shipped since 0.2.0, so every correction here is an edit in place.

### Manual additions

**How bootroot writes files** in `docs/en/cli.md` and `docs/ko/cli.md` grows from three consequences to seven, in the register of the three already there:

- **Hard links detach.** A truncating write reached every link of the inode; a rename installs a fresh inode at the one name it is given, so every other link keeps the old contents. Hard links are not supported at a bootroot-written path — use a symlink at one of the paths bootroot writes through, or copy after the command.
- **ACLs, extended attributes and SELinux labels do not survive a write.** They lived on the replaced inode and the staged temporary is created clean; only the permission bits cross over. A POSIX ACL on `state.json` or `.env` is gone at the next write, and on an SELinux host the new inode takes the containing directory's default labeling rather than a customized file label.
- **What gates a write moved from the file to its directory.** A rename needs write permission on the *directory* and none on the file, so a writable file in a read-only directory now fails at temp-file creation, and a `chmod 444` left as a write lock no longer blocks anything. Withdrawing write permission on the containing directory and `chattr +i` are the boundaries that still hold.
- **A crash leaves a `.tmpXXXXXX` sibling, not a torn destination.** The destination still holds the previous file; the leftover sits in the same directory carrying whatever mode the write had reached. Delete it — a write that merely *fails* removes its own temporary.

The symlink section, which is where an operator looks to find out what a link at one of these paths costs, records the new replaced-link warning, so the manuals no longer contradict the release notes. It also records where that warning is actually visible, stated by program rather than by enumerating paths: of the programs that write these files only `bootroot-agent` installs a log subscriber, so only a link it replaces while renewing on a service host is reported anywhere. Neither the `bootroot` CLI nor `bootroot-remote` installs one, so every replacement an operator command makes happens with nothing displayed, and the manuals tell an operator to check the path afterwards. The changelog carries the same qualification. `publish_staged_blocking`'s rustdoc states that mode bits (`& 0o7777`, under the two preserving arms of `StagedMode`) are the only thing that crosses onto the fresh inode.

### Changelog restructure

The `### Fixed` mega-entry splits in place, with no entry gaining or keeping an issue/PR reference:

- **`### Security`** — the permission-window closure: the CA password, recovery and unseal keys, responder HMAC and every `secret_id` reaching their final mode before the name exists. Sits beside the `.env` `0600` entry.
- **`### Changed`** — the new-inode consequences (open fds, single-file bind mounts) folded together with the hard-link and ACL/xattr/SELinux items, the directory-versus-file write gate, and the symlink answer at the cert/key/bundle, control-node `agent.toml` and credential paths.
- **`### Fixed`** — the torn-read elimination and its per-file flush decision, and the `reinit` cyclic-symlink preflight, both of which remain fixes.

### Warn when a publish replaces a symlink

`replaced_symlink_target` reads the destination with `symlink_metadata` and answers the link's target, or `None` for a regular file, a missing destination, or anything it cannot read without following. The two halves sit on opposite sides of the rename: the destination is read *before* it, because that is the last moment the link is still there to read, and the single `tracing::warn!` naming both ends is emitted *after* the rename succeeds, so a publish that fails there leaves the operator's link standing and says nothing about having replaced it.

The `_through_symlink` wrappers resolve before publishing, so the path they hand the core is the resolved target and never a link — exempt by construction, no flag. Behaviour is otherwise unchanged: which paths resolve a link and which replace one is untouched, and the warn lives in the core publish path.

Where the warning is *recorded* is a subscriber decision the CLI has never made — it installs none, so the event reaches no recorder — and making that decision here would change the output of every command, well past this issue's "no behaviour change beyond the one new warn line" constraint. The documentation states the limitation instead; closing it is a follow-up.

### Chown-failure guidance

`Failed to preserve existing uid={} gid={} on {}` now continues: re-owning the staged file needs `CAP_CHOWN` (run as root), so either run the command as the destination's owner or chown the destination to the user running it. Extracted into `chown_preserve_context` so the wording is pinned by a test. Message text only; the failure stays a failure.

## Test plan

- [x] `cargo test` green — 1,800+ tests across every target, including the new ones.
- [x] `replaced_symlink_target` unit-tested on all four destination states: a link, a regular file, a missing destination, and a dangling link (one whose target does not exist is destroyed just the same, and `symlink_metadata` sees it without following).
- [x] Subscriber-capture test pins the log line through the two spellings an operator reaches: the bare publish over a seeded link warns exactly once naming both ends, and the `_through_symlink` publish over the same arrangement stays quiet while still delivering its bytes.
- [x] `chown_preserve_context_names_the_remediation` asserts the uid/gid, the path, `CAP_CHOWN`, and both ways out.
- [x] The existing per-writer symlink tests, including `rerender_publishes_agent_toml_at_its_own_name` and the wrapper-level set in `fs_util`, pass unchanged.
- [x] `./scripts/check-docs.sh` — both manuals build under `--strict`, theme verified.
- [x] `cargo clippy --all-targets -- -D warnings` clean.
- [x] `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff.
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` clean.
- [x] `scripts/preflight/ci/check.sh` end to end.
- [x] Verified as an end user that a CLI publish over a seeded link (`bootroot openbao save-unseal-keys` against a symlinked `unseal-keys.txt`) does replace the link with a `0600` regular file, leaves the former target holding its old contents, and prints no warning even under `RUST_LOG=trace`, because the CLI installs no subscriber. The manuals and changelog were corrected to match.
- [x] Audited every non-test call site of the bare (link-replacing) publish helpers against the manuals' claim about where the warning is visible. The claim was wrong twice: the HTTP-01 responder calls no publish helper at all, so it can never emit this warning, and the `bootroot` CLI reaches far more link-replacing paths than the three the manuals named (the CA bundle, the step-ca password, the secrets-tree `role_id`/`secret_id`, the remote-bootstrap artifact, the rotation state), while `bootroot-remote` installs no subscriber either and writes credentials on a target host through the same publish. Both manuals and the changelog now state the rule by program instead.
- [x] Full CI green on the pushed branch.

The Docker E2E matrix is not required by the issue's test plan — the only behaviour deltas are a log line and error text, on paths the unit tests pin — but it runs on every PR here and passed, and `E2E Extended` was dispatched on this branch as well.

Closes #849


sehkone:
**[Reviewer Round 1]**

No blocking findings. I compared the diff with #849 and found the implementation aligned with the requested behavior and scope.

- `publish_staged_blocking` snapshots a final-component link before `persist` and emits the warning only after `persist` succeeds ([`src/fs_util.rs:1002`](src/fs_util.rs#L1002)); this avoids claiming a replacement when a failed rename leaves the operator link intact. The `_through_symlink` wrappers remain exempt through resolution rather than a new mode flag.
- The tests exercise the requested four helper states, including a dangling link, and capture the actual event for both bare and through-link publishing ([`src/fs_util.rs:1729`](src/fs_util.rs#L1729), [`src/fs_util.rs:1817`](src/fs_util.rs#L1817)). That would have caught both a missing warning and a warning emitted on the resolved path.
- The manual and changelog accurately separate the new-inode/security/fix consequences, place the warning in the symlink guidance, and explicitly describe the current subscriber limitation rather than implying CLI output that does not exist. The PR body has the required issue linkage and a concrete checked test plan.

No GitHub issue was created in this PR or its review thread.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Correct the record on staged publishing
```

**Body**

```text
Converting every write to stage-then-rename changed more than the
manuals and the changelog recorded. Four semantics moved with no
mention anywhere: the permission that gates a write moved from the
file to its directory, so a `chmod 444` left as a write lock stops
blocking anything; hard links detach, because a rename installs a
fresh inode at one name; ACLs, xattrs and SELinux labels do not
survive, since only the mode bits cross onto that inode; and a crash
now leaves a `.tmpXXXXXX` sibling an operator should be able to look
up. Each is inherent to publishing by rename rather than a defect to
fix, so each is documented in both manuals alongside the three
consequences already there.

The changelog filed the whole conversion under `### Fixed` although
only part of it is a fix. Nothing has shipped since 0.2.0, so the
entry is split in place rather than superseded: the permission-window
closure is `### Security`, the semantics an operator may have arranged
around are `### Changed`, and the torn-read elimination and the
`reinit` cyclic-symlink preflight stay `### Fixed`.

Replacing a symlink at a credential or cert path is decided
behaviour, but decided is not the same as silent. The bare publish now
reads the destination with `symlink_metadata` before the rename —
the last moment the link is there to read — and warns naming both ends
only once the rename has succeeded, so a publish that fails leaves the
link standing and claims nothing. The `_through_symlink` wrappers hand
the core a resolved target and are exempt by construction. Where that
warning is recorded is a subscriber decision the CLI has never made,
and making it here would change the output of every command; the
manuals and changelog state the limitation by program instead.

The chown-failure context named the ownership it was reproducing but
not what to do about it, and the failure is `EPERM` in every realistic
case, so it now names `CAP_CHOWN` and both ways out.

Closes #849
```

PR #862, opened 2026-08-16
Make the symlink decision a property of the destination (#847)

### LINKED ISSUE #847: Make the symlink decision a property of the destination
## Problem

#841 converted every production writer in the crate to publish by rename. That forced a decision at each one, because `O_TRUNC` follows a symlink at the final path component and `rename` replaces it: a destination the operator pointed elsewhere is either still written through, or silently destroyed and replaced by a regular file while the write reports success.

The decision was made correctly, per file. What it left behind is the way it is expressed. There are now four public spellings over one publisher:

|  | flushes the directory | rename only |
| --- | --- | --- |
| resolves a final symlink | `atomic_write_through_symlink` | `atomic_replace_through_symlink` |
| publishes at the name | `atomic_write` | `atomic_replace` |

with blocking halves of each, and 41 production call sites choosing among them — 20 taking a `_through_symlink` spelling, 21 taking a bare one. Both axes are invisible at the call site: every spelling compiles, every spelling publishes by rename, and every spelling returns `Result<()>`. Nothing distinguishes a correct choice from a wrong one until the failure reaches an operator, as a link they arranged replaced by a regular file, or a credential written through a link they did not arrange.

The rule itself is not the problem, and this issue does not propose changing which files follow a link. The rule is already stated in the rustdoc on `atomic_write_through_symlink`, and it is a good one: **a destination the operator names is written through the link; a path bootroot derives from its own tree layout is published at the name.** The problem is that the rule is applied by hand at 41 sites and enforced nowhere, so a writer added next year gets it right only if its author reads the right doc comment first.

This is the shape of defect the crate has hit before. #613 moved one `agent.toml` writer off a truncating write and left three others behind; #841's own review spent a round finding them. A convention that lives only in prose gets applied unevenly, and the unevenness is invisible until someone greps for it.

## Scope

Make the symlink axis a property of the destination rather than of the call site, so that a new writer cannot omit the decision and a wrong one is visible in review.

Introduce a `fs_util::Destination` newtype wrapping the path, with two constructors that name where the path came from:

- `Destination::operator_named(path)` — the path came from a CLI flag, a config value, or a file the operator arranges in bootroot's tree. Resolves a final symlink and publishes at the target.
- `Destination::bootroot_owned(path)` — the path is derived from the secrets tree or another layout bootroot controls. Publishes at the name.

The publisher takes a `Destination` instead of a `&Path` and reads the symlink policy off it. The durability axis stays exactly as it is — an argument on the call, unchanged in meaning, unchanged per site.

Then migrate all 41 production call sites, each keeping the answer it has today. This is a mechanical rewrite: a site currently calling a `_through_symlink` spelling becomes `Destination::operator_named`, a site calling a bare spelling becomes `Destination::bootroot_owned`. No file changes which side it is on, and no observable behaviour changes.

Two sites read the mapping off the destination rather than off the spelling, and are the one place it is not literal. `init`'s summary-JSON and root-token writers call `resolve_symlink_destination` by hand before they publish, because `tighten_existing_secret_file` has to narrow the same file the rename will land on, and then hand the resolved path to a bare spelling. The destination there is operator-named — it is `--summary-json` and `--root-token-output` — so it takes `Destination::operator_named`, which finds nothing left to follow and changes no behaviour. `bootroot_owned` would compile and publish identically while naming the wrong side, which is exactly what this issue exists to stop.

The four `_through_symlink` wrappers are then redundant and are removed; `atomic_write` and `atomic_replace` remain, taking a `Destination`.

Where a site's existing comment explains its symlink answer, shorten it to the part the constructor does not already say. The constructor name carries the answer; the comment should be left carrying only the reason, where the reason is not obvious.

## Constraints

- **No destination changes sides.** Every destination keeps the symlink behaviour #841 gave it. A migration that "fixes" a site it thinks is wrong is out of scope: leave that site exactly as it is, and record the observation and the reasoning behind it in the pull request description. Whether such a site should be corrected is a decision for a human reading that record, and it is not this run's to make or to schedule.
- Do not touch the durability axis. `atomic_write` still flushes the containing directory and `atomic_replace` still does not, and every site keeps the one it has.
- Do not touch the override credential writers — `create_owned_credential_noclobber`, `write_owned_file_replace`, `atomic_rewrite_owned_no_symlink`. They stage independently, refuse a symlink outright, and take ownership from the parent directory. They are a third answer that this newtype deliberately does not express, and folding them in would make one of the two callers wrong. Say so in the `Destination` rustdoc.
- Do not change what is written, the modes applied, or the errors reported at any site.
- `resolve_symlink_destination` stays public and stays where it is. `bootroot reinit`'s two output preflights call it directly to judge a target's mode before the destructive step, which is not a publish.
- #848 replaces these same wrappers' bare `mode: u32` with a `StagedMode` enum. The recommended order is #848 first, with this migration carrying the enum through unchanged; if this issue lands first, #848 applies its enum to the `Destination`-taking signatures instead. Either order, neither blocks the other, and neither changes the other's mapping. #848 landed first, so that is the branch taken.

## Acceptance criteria

The criteria below are the whole of it. This issue is satisfied when they hold and does not extend to any property they do not state.

- `fs_util::Destination` exists with exactly the two constructors above, and its rustdoc states the rule that decides between them and names the override credential writers as the third answer it does not cover
- `atomic_write`, `atomic_replace` and their blocking halves take a `Destination`; the four `_through_symlink` wrappers no longer exist
- Every production call site migrated, each on the same side of the symlink axis as before this issue. `git grep -n 'atomic_write\|atomic_replace' src/`, filtered to non-test code, shows no call passing a bare `&Path`
- A test asserts each constructor's behaviour once, at the primitive: `operator_named` delivers to a live link's target and leaves the link standing, `bootroot_owned` replaces the link at the same seeded path. The per-site symlink tests #841 added stay as they are — they pin the per-file answers this issue must not change
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` clean, full CI green including the Docker E2E matrix

## Test plan

- `cargo test` green, with no test changed except where a call site's signature moved
- `scripts/preflight/ci/check.sh` end to end
- `scripts/preflight/ci/e2e-matrix.sh`, since the migration touches `init`, rotation and `service` write paths even though none of them changes behaviour

## Not in this issue

- Which files follow a link. That is #841's decision and this issue preserves all 41 of it.
- The durability axis. Whether the two-spelling split between `atomic_write` and `atomic_replace` should also become a named property is the same kind of question, but it is a second migration over the same 41 sites and does not need to ride along with this one.
- `save_unseal_keys` and `eab::write_key_file`, the two writers that still truncate. #840 converts them, and states the symlink answer they take. If #840 has not landed when this starts, they are converted there and migrated here, in whichever order the two land — neither blocks the other, and each is a small merge conflict for the other at worst. #840 landed first and converted both, so they arrive here as bare-spelling call sites and are migrated with the rest — which is why the counts above are 41 and 21 rather than the 39 and 19 this issue was written against.

sehkone:
## Summary

Publishing by rename forced a symlink decision at every writer in the crate, and #841 made that decision correctly per file — but expressed it as four public spellings over one publisher, chosen by hand at 41 call sites. Every spelling compiles, every spelling publishes by rename, every spelling returns `Result<()>`, so nothing separated a correct choice from a wrong one until an operator found a link they arranged replaced by a regular file, or a credential written through a link they did not arrange.

This makes the answer a property of the destination instead:

- `fs_util::Destination::operator_named(path)` — a path from a CLI flag, a config value, or a file the operator arranges in bootroot's tree. Resolves a final symlink and publishes at the target.
- `fs_util::Destination::bootroot_owned(path)` — a path bootroot derived from the secrets tree or another layout it controls. Publishes at the name.

`atomic_write`, `atomic_replace` and their blocking halves now take a `Destination` and read the symlink policy off it; the four `_through_symlink` wrappers are gone. The durability axis is untouched — `atomic_write` still flushes the containing directory, `atomic_replace` still does not, and every site keeps the one it had. The `Destination` rustdoc states the rule that decides between the constructors and names the override credential writers (`create_owned_credential_noclobber`, `write_owned_file_replace`, `atomic_rewrite_owned_no_symlink`) as the third answer this type deliberately does not express, since they refuse a link outright and folding them in would make one of their two callers wrong. `resolve_symlink_destination` stays public and stays where it is, for `reinit`'s two output preflights.

All 41 production call sites are migrated, each on the same side of the symlink axis as before: 22 `operator_named` and 19 `bootroot_owned`. The 22 are the 20 former `_through_symlink` sites plus `init`'s summary-JSON and root-token writers, which resolve the link by hand before publishing so `tighten_existing_secret_file` can narrow the file the rename lands on — those destinations are `--summary-json` and `--root-token-output`, so `operator_named` names the right side and finds nothing left to follow. Where a site's comment explained its symlink answer, it now carries only the part the constructor name does not.

Test coverage follows the same move: the primitive-level tests that exercised the wrapper pair now exercise the constructors, asserting that `operator_named` delivers to a live link's target and leaves the link standing while `bootroot_owned` replaces the link at the same seeded path. The per-site symlink tests #841 added are unchanged — they pin the per-file answers this must not change.

No file changes which side it is on, no observable behaviour changes, and no destination was "fixed" along the way.

### Observation, no change made

The migration makes one asymmetry legible for the first time: the control node's `agent.toml` editor takes `bootroot_owned` (`src/commands/service.rs`), while `bootroot-remote bootstrap`'s writer of a file by the same name takes `operator_named` (`src/bin/bootroot-remote/agent_config.rs`). This is #841's decision and it is deliberate on both sides — on the control node `service::local_config` has created `agent.toml` by rename since #613, so a link there never survives the command that creates the file and the later editors have nothing to preserve; on a bootstrap target that writer *is* the creating writer, and the `O_TRUNC` write it replaced followed a final link, so an operator who pointed the destination elsewhere has a working installation today. Both comments now say so. Recorded here only because the two constructor names sit side by side in a way the four spellings did not, and a reader may notice it before reading either comment.

Closes #847

## Test plan

- [x] `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` clean
- [x] `cargo test` green, with no test changed except where a call site's signature moved
- [x] `scripts/preflight/ci/check.sh` end to end
- [x] `git grep -n 'atomic_write\|atomic_replace' src/`, filtered to non-test code, shows no call passing a bare `&Path`
- [x] No `_through_symlink` spelling remains anywhere in the crate
- [x] Every one of the 41 production sites verified against `origin/main` to be on the same side of the symlink axis as before
- [ ] `scripts/preflight/ci/e2e-matrix.sh` — partially runnable here. The two lifecycle scenarios that bind port 0 and pick free host ports (local no-hosts, remote no-hosts) run and pass locally. The `hosts` variants need `sudo -n`, which prompts for a password in this environment, and the rotation/recovery, reinit-recovery, step-ca SAN and two OpenBao TLS scenarios bind fixed loopback ports that an unrelated local OrbStack process already holds (`127.0.0.1:8080`, `127.0.0.1:8200`), so they abort at container start. All ten scenarios pass in CI's Docker E2E matrix.
- [x] Full CI green, including the Docker E2E matrix — 18/18 checks pass, including all ten Docker E2E scenarios

### DIFF
--- a/src/bin/bootroot-remote/agent_config.rs
+++ b/src/bin/bootroot-remote/agent_config.rs
@@ -209,20 +209,17 @@ pub(super) async fn apply_agent_config_updates(
         // against a stale responder HMAC or trust anchor with no signal
         // that a re-sync is needed.
         //
-        // A symlink at the destination is resolved first and the target
-        // is published, unlike the control plane's writers of the same
-        // file name. The rule there is that `service::local_config`
-        // creates `agent.toml` by rename, so a link at the path never
-        // survives the command that creates it and the later editors
-        // have nothing to preserve. On a bootstrap target this writer is
-        // the one that creates the file, and the write it replaces
-        // opened with `O_TRUNC`, which follows a final link — so an
-        // operator who pointed the destination at a config they keep
-        // elsewhere has a working installation today, and a bare rename
-        // would destroy the link, leave its target holding the previous
-        // profile, and report the item applied.
-        if let Err(err) = fs_util::atomic_write_through_symlink(
-            &args.agent_config_path,
+        // The destination is operator-named, unlike the control plane's
+        // writers of the same file name. The rule there is that
+        // `service::local_config` creates `agent.toml` by rename, so a
+        // link at the path never survives the command that creates it
+        // and the later editors have nothing to preserve. On a bootstrap
+        // target this writer is the one that creates the file, and the
+        // write it replaces opened with `O_TRUNC`, which follows a final
+        // link — so an operator who pointed the destination at a config
+        // they keep elsewhere has a working installation today.
+        if let Err(err) = fs_util::atomic_write(
+            fs_util::Destination::operator_named(&args.agent_config_path),
             with_profile.as_bytes(),
             fs_util::StagedMode::Policy(fs_util::KEY_FILE_MODE),
         )
--- a/src/bin/bootroot-remote/io.rs
+++ b/src/bin/bootroot-remote/io.rs
@@ -114,7 +114,7 @@ pub(super) async fn write_secret_file(path: &Path, contents: &str) -> Result<App
     // entry means re-running `service add` on the control plane, not
     // repeating the write here.
     fs_util::atomic_write(
-        path,
+        fs_util::Destination::bootroot_owned(path),
         next.as_bytes(),
         fs_util::StagedMode::Policy(fs_util::KEY_FILE_MODE),
     )
--- a/src/cert_group.rs
+++ b/src/cert_group.rs
@@ -417,11 +417,15 @@ pub async fn write_key_file(path: &Path, key_pem: &str, policy: CertGroupPolicy)
 /// the key writer has done since #593 and what the certificate and the
 /// bundle now do beside it — the point of the conversion was to remove
 /// the asymmetry between them, not to relocate it into how a link is
-/// treated. The configuration writers take
-/// [`fs_util::atomic_write_through_symlink`] instead, for destinations
+/// treated. That is [`fs_util::Destination::bootroot_owned`]'s answer,
+/// which these three would take were they not calling the publisher
+/// directly for their [`StagedOwner::PolicyGroup`] ownership; the
+/// configuration writers take
+/// [`fs_util::Destination::operator_named`] instead, for destinations
 /// no rename writer has ever owned.
 ///
-/// [`fs_util::atomic_write_through_symlink`]: crate::fs_util::atomic_write_through_symlink
+/// [`fs_util::Destination::bootroot_owned`]: crate::fs_util::Destination::bootroot_owned
+/// [`fs_util::Destination::operator_named`]: crate::fs_util::Destination::operator_named
 ///
 /// [`fs_util::publish_staged_blocking`]: crate::fs_util::publish_staged_blocking
 /// [`fs_util::atomic_write_blocking`]: crate::fs_util::atomic_write_blocking
--- a/src/commands/ca.rs
+++ b/src/commands/ca.rs
@@ -159,8 +159,8 @@ fn patch_ca_json_ctmpl(
 /// the previous document in place and costs the next render or a re-run
 /// of `bootroot ca update`, not an unrecoverable state.
 fn publish_ca_json(path: &Path, contents: &str, messages: &Messages) -> Result<()> {
-    fs_util::atomic_replace_through_symlink_blocking(
-        path,
+    fs_util::atomic_replace_blocking(
+        fs_util::Destination::operator_named(path),
         contents.as_bytes(),
         fs_util::StagedMode::PreserveOrUmask,
     )
--- a/src/commands/dotenv.rs
+++ b/src/commands/dotenv.rs
@@ -48,10 +48,10 @@ fn strip_quotes(value: &str) -> String {
 /// Writes a `.env` file from key-value pairs.
 ///
 /// Published by rename through
-/// [`fs_util::atomic_write_through_symlink_blocking`]. Two readers make
-/// a torn `.env` costly: `docker compose` interpolates it on every
-/// invocation, and bootroot itself reads it back to recover the instance
-/// name and the assigned host ports.
+/// [`fs_util::atomic_write_blocking`]. Two readers make a torn `.env`
+/// costly: `docker compose` interpolates it on every invocation, and
+/// bootroot itself reads it back to recover the instance name and the
+/// assigned host ports.
 ///
 /// It takes the directory flush for that second reader. The ports and
 /// the instance id here are the only record of which containers this
@@ -59,12 +59,11 @@ fn strip_quotes(value: &str) -> String {
 /// fresh ones and unable to find the stack it already started, which no
 /// re-run of `init` repairs.
 ///
-/// A symlinked `.env` is resolved before staging. Compose's own
-/// convention is to keep one `.env` beside the compose file, so pointing
-/// it at a shared file is a thing operators do, and the truncating write
-/// this replaced updated that shared file; a bare rename would replace
-/// the link and leave every other consumer of the target reading stale
-/// ports.
+/// The destination is [`fs_util::Destination::operator_named`]:
+/// compose's own convention is to keep one `.env` beside the compose
+/// file, so pointing it at a shared file is a thing operators do, and
+/// replacing the link would leave every other consumer of the target
+/// reading stale ports.
 pub(crate) fn write_dotenv(
     path: &Path,
     entries: &[(&str, &str)],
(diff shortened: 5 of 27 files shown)


sehkone:
**[Reviewer Round 1]**

- [P3] Start the two new public constructor docs with a third-person verb. `Destination::operator_named` (src/fs_util.rs:490) and `Destination::bootroot_owned` (src/fs_util.rs:518) currently begin with “A destination…”, but the repository’s rustdoc standard requires public function summaries to begin with verbs such as “Creates…”. Rewording these to “Creates a destination…” keeps the API documentation consistent; rustdoc itself will not catch this convention.

sehkone:

sehkone:
**[Author Round 1]**

- **[P3] Start the two new public constructor docs with a third-person verb — Fixed.** Both summaries opened with a noun phrase, which the repository's rustdoc standard reserves for types, traits, modules and constants; a function summary starts with a verb in the third-person singular. `Destination::operator_named` now opens "Creates a destination the operator named: …" and `Destination::bootroot_owned` "Creates a destination bootroot derived from a layout it owns: …", with the following sentences rewrapped and otherwise unchanged. The `Destination` struct doc itself keeps its noun phrase, which is what the standard asks of a type. Doc wording only — no signature, behaviour or call site moved.

Re-verified after the change: `cargo fmt -- --config group_imports=StdExternalCrate` leaves no diff, `cargo clippy --all-targets -- -D warnings` is clean, and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` is clean. No `cargo test` rerun was warranted for a comment-only edit; CI runs the full suite on the pushed commit. The PR description still describes the branch accurately, so it is unchanged.

sehkone:
**[Reviewer Round 2]**

No remaining findings. The Round 1 constructor-documentation item is resolved: both public constructor summaries now begin with “Creates…”. The updated implementation carries the policy through `Destination`, removes the four `_through_symlink` entry points, preserves the existing per-site mapping, and tests both constructors through both durability publishers.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Carry the symlink answer on the destination
```

**Body**

```text
Publishing by rename forced a symlink decision at every writer in the
crate. That decision was made correctly per file, but it was expressed
as four public spellings over one publisher, chosen by hand at 41 call
sites. Every spelling compiles, every spelling publishes by rename, and
every spelling returns Result<()>, so nothing separated a correct
choice from a wrong one until an operator found a link they arranged
replaced by a regular file, or a credential written through a link they
did not arrange. A convention that lives only in prose gets applied
unevenly, and the unevenness stays invisible until someone greps for
it.

Make the answer a property of the destination instead. Destination
carries the path together with where the path came from:
operator_named for a CLI flag, a config value, or a file the operator
arranges in bootroot's tree, which resolves a final symlink and
publishes at the target; bootroot_owned for a path derived from the
secrets tree or another layout bootroot controls, which publishes at
the name. atomic_write, atomic_replace and their blocking halves take a
Destination and read the policy off it, so a new writer cannot omit the
decision and a wrong one is visible in review. The four
_through_symlink wrappers are then redundant and are removed.

The durability axis is untouched, and no destination changes sides: all
41 sites keep the answer they had. The rustdoc names the override
credential writers as the third answer this type deliberately does not
express, since they refuse a link outright and folding them in would
make one of their two callers wrong.

Closes #847
```

PR #865, opened 2026-08-17
Serve step-ca metrics on the scraped port (#864)

### LINKED ISSUE #864: Enable step-ca metrics scraping
# Enable step-ca metrics scraping

## Context

`monitoring/prometheus.yml` already scrapes `step-ca:9102/metrics`, and both bundled Compose files expose port 9102 on the `step-ca` service. The pinned `smallstep/step-ca:0.30.2` process starts that endpoint only when the top-level `metricsAddress` in `ca.json` is set. It accepts the address through neither a command-line flag nor an environment variable, and the Compose command is fixed, so Compose cannot supply the missing configuration. Neither `step ca init` nor bootroot's current reconciliation writes the key; consequently, the declared Prometheus target is permanently down.

`ca.json` is co-owned by bootroot and the step-ca OpenBao Agent sidecar. `write_stepca_templates` reads the live file to build `secrets/templates/ca.json.ctmpl`, while `build_ca_json_template` explicitly stamps the bootroot-controlled `dnsNames` into the template rather than trusting that read. This matters because the sidecar can render its previous template between reconciliation and template generation; inheriting a metrics value from that file could bake the stale state into the new template and remove the listener again.

During `bootroot init`, `update_ca_json_with_backup` reports only `CaJsonUpdate::dns_names_changed`. `orchestrator.rs` uses that flag to gate both the step-ca OpenBao Agent restart and its on-disk reassertion, and, for an existing CA, the `docker compose restart step-ca` branch. A metrics-only repair of an existing installation therefore needs equivalent change tracking: unchanged `dnsNames` must not prevent the new template, sidecar restart, reconciliation, and step-ca restart from taking effect.

## Scope

Configure bootroot-managed step-ca configuration with `metricsAddress: ":9102"` so the existing scrape job becomes live. Set the fixed value explicitly in the direct `ca.json` reconciliation and explicitly inside `build_ca_json_template`, independent of the contents read from the on-disk file. Preserve unmodelled `ca.json` keys.

Extend `CaJsonUpdate` and the `stepca_dns_names_changed`-gated branches in `src/commands/init/steps/orchestrator.rs` so a changed `metricsAddress` triggers the same template/sidecar/`ca.json` reconciliation sequence and the step-ca restart. On an already-initialized installation with unchanged `dnsNames`, the next `bootroot init` must add the setting without re-running `step ca init` or replacing CA material, and it must remain present after at least one subsequent OpenBao Agent render.

Keep the existing Prometheus job and the existing internal-only `expose: 9102` declarations in `docker-compose.yml` and `docker-compose.deploy.yml`. The endpoint may be reachable by services on the Compose network but must not be published as a host port. Record the intentional unauthenticated, internal-network-only metrics exposure in source code next to the fixed setting.

Update `tests/monitoring_integration.rs`'s explanatory comment: production step-ca metrics are configured after init, but this test deliberately begins before `bootroot init` and therefore continues to assert the OpenBao target only. Add an Unreleased changelog entry that tells existing installations to run `bootroot init` to gain the step-ca target. The README already accurately states that Prometheus scrapes step-ca metrics, and neither manual makes a specific step-ca metrics claim, so no page documenting the endpoint needs an edit. The one manual change this does carry is consequential rather than about metrics: the E2E scope lists in `docs/en/e2e-ci.md` and `docs/ko/e2e-ci.md` describe `run-stepca-san.sh` as covering certificate SANs alone, and must name the metrics listener once that harness covers it too.

## Acceptance criteria

- [x] A fresh `bootroot init` writes exactly one top-level `metricsAddress` with the value `:9102` to `secrets/config/ca.json`, and `secrets/templates/ca.json.ctmpl` carries the same value even if the input `ca.json` lacks it.
- [x] On an already-initialized installation whose `dnsNames` are unchanged and whose `ca.json` lacks `metricsAddress`, `bootroot init` adds the setting, restarts the sidecar and step-ca as required, does not recreate the root or intermediate CA material, and leaves the setting present after the sidecar completes a later render.
- [x] A subsequent `bootroot init` is idempotent, and `bootroot reinit` preserves `metricsAddress` and the existing CA material.
- [x] With monitoring enabled after initialization, Prometheus reports the existing `step-ca` target as `up` and `http://step-ca:9102/metrics` returns Prometheus exposition data from within the Compose network.
- [x] Port 9102 is not published as a host port in either bundled Compose file and remains available to the Prometheus service.
- [x] Focused unit tests cover direct reconciliation, explicit template stamping, and idempotence for `metricsAddress`; an integration or E2E test verifies persistence after a real sidecar render.
- [x] The monitoring integration test continues to assert only OpenBao, still reports exactly one ignored test when invoked directly, and its comment accurately explains that step-ca metrics are configured by init but unavailable at this test's pre-init stage.
- [x] `CHANGELOG.md` has an Unreleased entry explaining that existing installations gain the working step-ca target after `bootroot init`.

## Constraints

Do not remove or rename the existing `step-ca` Prometheus job, alter OpenBao telemetry, publish port 9102 to the host, or weaken the Compose-network boundary. Do not make `metricsAddress` a user-selectable bind address in this issue: the fixed internal address must match the declared scrape target. Preserve the current atomic write, template snapshot, rollback, and sidecar-restart ordering for `ca.json` and `ca.json.ctmpl`.

## Out of scope

Reworking the monitoring integration test to initialize a CA before its monitoring checks, adding alerts or Grafana dashboards for step-ca metrics, changing other scrape jobs, and revisiting the monitoring stack's broader authentication policy are out of scope. The existing README statement is already correct, and no manual page documents this endpoint, so none needs a metrics edit — the E2E scope-list update named under Scope is the only manual change, and it follows from the harness gaining the coverage rather than from the endpoint itself.

## Test plan

- [x] Run focused `stepca_setup` tests for `ca.json` reconciliation and `.ctmpl` generation, including an input missing `metricsAddress`.
- [x] On an existing initialized Compose installation with unchanged `dnsNames`, run `bootroot init`, wait through an OpenBao Agent render interval, and verify `metricsAddress` remains in `ca.json` and step-ca exposes `/metrics` without CA fingerprint changes.
- [x] On a clean Compose installation, run `bootroot init`, enable monitoring, and query Prometheus's targets API until the `step-ca` job is `up`; fetch `http://step-ca:9102/metrics` from the Prometheus container.
- [x] Verify neither `docker-compose.yml` nor `docker-compose.deploy.yml` publishes host port 9102 while both still `expose` it, as an automated test beside the existing Compose assertions in `tests/docker_compose_security.rs` — a later edit that publishes the port should fail CI rather than pass unnoticed.
- [x] Run `bootroot init` again and run `bootroot reinit`; verify the setting persists and CA fingerprints do not change.
- [x] Run `cargo fmt -- --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --test monitoring_integration -- --include-ignored`, confirming the latter runs exactly one test.
- [x] Run `scripts/preflight/ci/e2e-matrix.sh` and `scripts/preflight/ci/e2e-extended.sh`.

## Dependencies

None.

## Pointers

Start with `src/commands/init/steps/stepca_setup.rs`: `update_ca_json_with_backup`, `build_ca_json_template`, `write_stepca_templates`, and the `CaJsonUpdate::dns_names_changed` result are the configuration ownership path. In `src/commands/init/steps/orchestrator.rs`, the `stepca_dns_names_changed` branches restart the step-ca OpenBao Agent, reassert `ca.json`, and restart step-ca; extend those gates for metrics changes. `src/commands/reinit.rs` calls the init flow and is the upgrade-path test entry point. The current target and network boundary are in `monitoring/prometheus.yml`, `docker-compose.yml`, and `docker-compose.deploy.yml`; the deliberately pre-init test is `tests/monitoring_integration.rs`.

AcoPiper:
## Summary

`monitoring/prometheus.yml` has always scraped `step-ca:9102`, and both bundled compose files expose the port — but step-ca starts that listener only when `ca.json` carries a top-level `metricsAddress`. Nothing wrote the key, and the pinned `smallstep/step-ca:0.30.2` accepts the address through no flag and no environment variable, so compose could not supply it either: the declared target was permanently down.

`bootroot init` now writes the fixed `:9102` into `secrets/config/ca.json` and stamps it explicitly into `secrets/templates/ca.json.ctmpl`, independent of what the co-owned file holds when the template is generated — the step-ca OpenBao Agent sidecar can render the previous template in between, and an inherited value would bake the stale state back in (the #733 hazard, applied to a second key).

`CaJsonUpdate` now reports the metrics change alongside `dns_names_changed`, and `run_init_inner`'s previously `dnsNames`-gated branches gate on `CaJsonUpdate::stepca_reload_required()` instead. That matters for the upgrade path: an installation predating the setting has an unchanged name set, so gating on `dnsNames` alone would regenerate nothing, leave the sidecar on the old template and leave step-ca without a listener.

The port stays `expose`d to the compose network and unpublished. The endpoint carries no authentication — step-ca offers none for it — so that boundary is its only protection, and the reasoning is recorded next to the constant in `stepca_setup.rs`.

Closes #864

## Changes

- `src/commands/init/steps/stepca_setup.rs` — `metricsAddress` constants (with the security note), `set_ca_json_metrics_address`, stamped into `build_ca_json_template`, written by `update_ca_json_with_backup`, re-asserted by `reconcile_ca_json_dns_names` (renamed `reconcile_ca_json_managed_keys`, since it now owns two keys). The two managed keys are set through one `set_ca_json_key` helper, so both insert into the parsed document and land exactly once whether or not they were there before; the re-assertion sets both before deciding its early return, so a drifted name set cannot short-circuit the metrics address. Unmodelled `ca.json` keys still survive, and the atomic write / snapshot / rollback / sidecar-restart ordering is unchanged.
- `src/commands/init/steps/orchestrator.rs` — the template, sidecar-restart, re-assertion and `docker compose restart step-ca` branches gate on `stepca_reload_required()`.
- `tests/docker_compose_security.rs` — 9102 is exposed and never published, in both bundled compose files. `services.step-ca` is read structurally and each mapping judged by the key it sits under, since the bare `- "9102"` that is the wanted entry under `expose:` is a publication to an ephemeral host port under `ports:`. The section reader classifies the key's value by the YAML form it is written in — block sequence, flow sequence on the key's own line, or a form whose entries live elsewhere (an alias, a scalar) — so a `ports:` it cannot enumerate fails the guard instead of reading as an absent section; an anchor alone still resolves to the block beneath it. Fixture tests pin the reader against each form and hold the short, long and flow publication syntaxes rejected along with an aliased `ports:`.
- `scripts/impl/run-stepca-san.sh` / `tests/docker_e2e_stepca_san.rs` — the harness that already drives a fresh install, an already-initialized repair and a full OpenBao Agent render interval now also asserts `metricsAddress` in `ca.json` and the template at each point, and fetches `http://step-ca:9102/metrics` from inside the compose network (deliberately not from the host, where a successful probe would mean the boundary had been widened). A new scenario C isolates the metrics gate: on the stack scenario B leaves behind it strips `metricsAddress`, restarts step-ca to confirm the endpoint goes silent, and re-runs `init` with `dnsNames` settled — so nothing but the metrics address moves — then asserts the listener is back. Scenario A additionally brings Prometheus up (by name, behind its profile, `--no-deps` so compose cannot recreate step-ca without the `--stepca-bind` override) and waits for the declared `step-ca` scrape job to report `up` — the endpoint answering a direct fetch and Prometheus resolving its target to it are two claims, and the second is the one an operator reads. The Rust wrapper asserts the dumped phase log, SAN dumps, metrics dumps and targets response, so a script that skipped a step cannot pass on its exit status alone.
- `scripts/impl/run-reinit-recovery.sh` — `reinit` preserves `metricsAddress` alongside the CA material.
- `tests/monitoring_integration.rs` — comment rewritten: step-ca metrics are configured by `init`, and this test deliberately runs before any `init`, so it still asserts the OpenBao target only.
- `docs/en/e2e-ci.md`, `docs/ko/e2e-ci.md` — the PR-critical scope list names the metrics listener now that `run-stepca-san.sh` covers it.
- `CHANGELOG.md` — Unreleased entry telling existing installations to run `bootroot init`.

## Test plan

Run locally against Docker (macOS, Docker Desktop):

- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate` and `cargo clippy --all-targets -- -D warnings` — clean.
- [x] `cargo test --bins` — including the `stepca_setup` tests for direct reconciliation, template stamping from a `ca.json` that lacks the key, metrics-only reload gating, clobbered-address repair and idempotence.
- [x] `cargo test --test docker_compose_security` — 9102 exposed and never published in both bundled compose files.
- [x] `cargo test --test monitoring_integration -- --include-ignored` — exactly 1 test, passed.
- [x] `scripts/preflight/ci/test-core.sh` — passed end to end (unit tests, monitoring integration, install, zero-config init, clean + reinstall, CLI init, three `service add`s, three `verify`s, CA health, agent issuance).
- [x] `scripts/impl/run-stepca-san.sh` and `cargo test --test docker_e2e_stepca_san -- --ignored` (with `STEPCA_BIND_HOST` set to a real interface, as `172.17.0.1` does not exist under Docker Desktop) — passed, including `scenario-b-assert-stable` after a full render interval.
- [x] `scripts/impl/run-reinit-recovery.sh` — all three scenarios passed; `ca.json` still carries `:9102` afterwards.
- [x] A scripted walk of the issue's test plan on a real compose install: fresh `init` writes exactly one `metricsAddress: ":9102"`; Prometheus reports the `step-ca` target `up` and the prometheus container fetches `http://step-ca:9102/metrics`; stripping the key from `ca.json` *and* the template and restarting reproduces the pre-fix state (no listener); a second `init` with unchanged `dnsNames` restores it with root and intermediate fingerprints unchanged, the target back `up`, and the setting still present after a 40s render interval; a third `init` and a `bootroot reinit` both preserve it and the CA material; `127.0.0.1:9102` is not reachable from the host.
- [x] Full `scripts/preflight/ci/e2e-matrix.sh` and `scripts/preflight/ci/e2e-extended.sh` sweeps — run on CI rather than on this host, see below.

## Sweeps run on CI

The full `e2e-matrix.sh` and `e2e-extended.sh` sweeps cannot run on this host: it has no passwordless `sudo` (hosts-mode steps, `run-openbao-tls-reown.sh`) and its hostname is not a DNS label (the extended suite's runner cases). Both are host limits rather than properties of this change, so the sweeps were run on CI against this branch instead.

- Matrix: every PR-critical Docker E2E job on this branch is green, including `stepca-san`, `reinit-recovery`, `rotation`, `two-instance`, both `local-*` and both `remote-*` lifecycles and both `openbao-tls-*` cases.
- Extended: [run 31993910365](https://github.com/aicers/bootroot/actions/runs/31993910365), dispatched against this branch, passed with all seven cases `pass` — `scale-contention`, `failure-recovery`, `runner-timer`, `runner-cron`, `ca-key-recovery`, `reinit-recovery`, `infra-lifecycle` — covering the three cases this host cannot run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,23 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Prometheus can now scrape step-ca. The bundled monitoring stack has
+  always declared a `step-ca` scrape target, but step-ca serves metrics
+  only when its `ca.json` carries a top-level `metricsAddress`, nothing
+  wrote that key, and the pinned image accepts the address through no
+  command-line flag and no environment variable — so the declared target
+  was permanently down. `bootroot init` now writes the address the
+  existing scrape job already points at, `:9102`, into both
+  `secrets/config/ca.json` and the OpenBao Agent template that file is
+  re-rendered from, so the listener survives the sidecar's next render.
+  An existing installation gains the working target by running
+  `bootroot init` again: the setting is added in place, without
+  re-running `step ca init` and without replacing the root or
+  intermediate CA material. The metrics port remains exposed to the
+  Compose network only and is still not published to the host, so the
+  endpoint — which carries no authentication, step-ca offering none for
+  it — stays reachable to the services on that network and to nothing
+  else.
 - Fixed the files bootroot writes being published by truncating the
   destination and writing over it, so a crash or a concurrent reader
   could see a half-written file at a name that is supposed to hold a
--- a/docs/en/e2e-ci.md
+++ b/docs/en/e2e-ci.md
@@ -72,7 +72,7 @@ PR-critical Docker test set validates:
 - remote-delivery E2E scenario (`hosts`)
 - rotation/recovery matrix (`secret_id,eab,responder_hmac,trust_sync`)
 - reinit recovery
-- step-ca certificate SANs
+- step-ca certificate SANs and the Prometheus metrics listener
 - OpenBao TLS transition with no compose delta
 - OpenBao TLS re-issuance after `secrets/` is re-owned
 - two co-located instances stay independent (install, containers, volumes,
--- a/docs/ko/e2e-ci.md
+++ b/docs/ko/e2e-ci.md
@@ -70,7 +70,7 @@ PR 필수 Docker 조합 검증은 다음을 검증합니다.
 - 원격 전달 E2E 시나리오 (`hosts`)
 - rotation/recovery matrix (`secret_id,eab,responder_hmac,trust_sync`)
 - reinit 복구
-- step-ca 인증서 SAN
+- step-ca 인증서 SAN 및 Prometheus 메트릭 리스너
 - compose 델타 없는 OpenBao TLS 전환
 - `secrets/` 소유자 변경 이후 OpenBao TLS 인증서 재발급
 - 같은 호스트의 두 인스턴스가 서로 독립적으로 유지되는지(설치,
--- a/scripts/impl/run-reinit-recovery.sh
+++ b/scripts/impl/run-reinit-recovery.sh
@@ -9,6 +9,8 @@ set -euo pipefail
 #
 #   - step-ca root/intermediate fingerprint unchanged before vs. after
 #   - secrets/password.txt not overwritten
+#   - step-ca's `metricsAddress` survives, so the Prometheus target
+#     `init` configured is still live after a reinit (#864)
 #   - non-loopback OpenBao bind survives (compose override survives,
 #     intent persists in rewritten state.json, post-reinit OpenBao
 #     listens on the bind, the second init pass reaches it)
@@ -367,6 +369,27 @@ assert_post_reinit_contracts() {
   # 2. password.txt unchanged (byte-for-byte).
   if ! cmp -s "$dir/password.txt" "$SECRETS_DIR/password.txt"; then
     fail "[$label] secrets/password.txt was overwritten across reinit"
+  fi
+
+  # 2b. step-ca's metrics listener survives (#864).  `reinit` keeps
+  # `config/ca.json` and re-runs the init flow over it, so the address
+  # `monitoring/prometheus.yml` scrapes has to come back out the other
+  # side; a reinit that dropped it would take the declared Prometheus
+  # target down with no other assertion here noticing.  Retried for the
+  # same reason `run-stepca-san.sh` retries: `ca.json` is co-owned with
+  # the OpenBao Agent sidecar, whose render can land between reinit
+  # returning and this read.
+  local ca_json="$SECRETS_DIR/config/ca.json"
+  local metrics_address="" attempt
+  for attempt in $(seq 1 10); do
+    if [ -f "$ca_json" ]; then
+      metrics_address="$(jq -r '.metricsAddress // empty' "$ca_json" 2>/dev/null || true)"
+      [ "$metrics_address" = ":9102" ] && break
+    fi
+    sleep 2
+  done
+  if [ "$metrics_address" != ":9102" ]; then
+    fail "[$label] post-reinit ca.json metricsAddress='$metrics_address' expected ':9102'"
   fi
 
   # 3. Non-loopback bind survives.
--- a/scripts/impl/run-stepca-san.sh
+++ b/scripts/impl/run-stepca-san.sh
@@ -21,8 +21,13 @@ set -euo pipefail
 #   scenario-b  already-initialized path — `init` on loopback first,
 #               then `infra install --stepca-bind <ip>:9000` and a
 #               second `init`
+#   scenario-c  metrics-only upgrade — on the stack scenario B leaves,
+#               `metricsAddress` is stripped and step-ca restarted
+#               without it, then `init` runs with `dnsNames` already
+#               settled, so nothing but the metrics address moves
 #
-# Both assert the same two contracts against the running step-ca:
+# Scenarios A and B assert the same three contracts against the running
+# step-ca:
 #
 #   - the presented certificate carries `IP Address:<ip>` (an IP SAN,
 #     not a `DNS:` entry, which would not satisfy verification of a
@@ -42,6 +47,12 @@ set -euo pipefail
 # the root and intermediate fingerprints are unchanged across the
 # second `init`, and the CA bundle captured *before* it still validates
 # the chain afterwards.
+#
+# Scenario C isolates the metrics gate.  A and B both move `dnsNames`,
+# so neither can tell an `init` that carried the reload sequence for the
+# name set from one that carried it for the metrics address; C removes
+# the name change and asserts the listener still comes back — with the
+# CA material, and the chain a consumer already trusts, untouched.
 
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -181,6 +192,26 @@ compose_stop() {
 
 run_bootroot() {
   ( cd "$WORKSPACE_DIR" && "$BOOTROOT_BIN" "$@" )
+}
+
+# Prints the step-ca OpenBao Agent sidecar's container name.
+#
+# Named directly rather than driven through `compose`: the sidecar is
+# declared in the agent override file `init` writes under `secrets/`,
+# which this harness's compose file set does not include, so compose
+# does not know the service exists.
+#
+# The name is the recorded instance plus the suffix, resolved through
+# the same helper the leftover check uses — not `COMPOSE_PROJECT_NAME`,
+# which the Rust wrapper sets to a value that is not an instance name,
+# and not a hard-coded `bootroot`, which would quietly name the wrong
+# container the day this harness takes an `--instance-name`.  Resolved
+# per call because `.env` is written by `init`, not by the bring-up.
+stepca_agent_container() {
+  local instance
+  instance="$(resolve_recorded_instance_name "$(dirname "$COMPOSE_FILE")")" \
+    || fail "could not resolve the recorded instance name from $(dirname "$COMPOSE_FILE")/.env"
+  printf '%s-openbao-agent-stepca\n' "$instance"
 }
 
 # Finding *an* `openssl` on PATH is not enough. macOS ships LibreSSL as
@@ -486,6 +517,40 @@ assert_stepca_metrics_endpoint() {
   fail "[$label] step-ca served no Prometheus metrics at http://step-ca:9102/metrics from inside the compose network (see $out)"
 }
 
+# Asserts step-ca serves nothing on 9102 — the state an installation
+# predating the setting is in, and the baseline scenario C's upgrade is
+# measured against.
+#
+# step-ca is waited for first, on its ACME port.  Without that, a fetch
+# failing because the container is still coming back up reads exactly
+# like one failing because the listener is gone, and the baseline would
+# pass on a step-ca that was simply not there yet — which would make the
+# "it came back" assertion afterwards meaningless.  Past the wait the
+# process is serving, and 9102 answering or not is a statement about the
+# configuration alone.
+#
+# Then polls rather than probing once, because a container that has not
+# finished stopping still answers on the old listener.  The listener
+# exists or it does not, so the first empty fetch is conclusive and only
+# one that never goes away runs the loop out.
+assert_stepca_metrics_endpoint_absent() {
+  local label="$1"
+  local out="$ARTIFACT_DIR/stepca-metrics-${label}.txt"
+  local attempt
+  wait_for_stepca_tls "${STEPCA_BIND_HOST}:${STEPCA_BIND_PORT}"
+  for attempt in $(seq 1 "$STEPCA_READY_ATTEMPTS"); do
+    if ! compose exec -T openbao wget -qO- "http://step-ca:9102/metrics" \
+        >"$out" 2>>"$RUN_LOG"; then
+      return 0
+    fi
+    if ! grep -q '^# HELP ' "$out"; then
+      return 0
+    fi
+    sleep "$STEPCA_READY_DELAY_SECS"
+  done
+  fail "[$label] step-ca still served Prometheus metrics on 9102 after ca.json was stripped of metricsAddress (see $out)"
+}
+
 # Asserts Prometheus itself reports the `step-ca` scrape job healthy.
 #
 # The endpoint check above proves step-ca answers; this proves the target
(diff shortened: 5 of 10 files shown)


AcoPiper:
**[Reviewer Round 1]**

[P1] Make the metrics-port guard inspect the `step-ca.ports` section rather than exempting every bare `"9102"` entry. In `tests/docker_compose_security.rs:65-68`, a future change such as `expose: ["9102"]` plus `ports: ["9102"]` passes: the required `expose` snippet is still present and the newly added `- "9102"` is explicitly excluded from the failure condition. Compose treats that short syntax as a host publication, exposing this unauthenticated endpoint on all host interfaces. That misses the issue requirement that any later 9102 publication fail CI. Parse the Compose document (and handle short and long port syntax), then assert that `services.step-ca.ports` has no target/published port 9102; retain a separate assertion for the required `step-ca.expose` entry.

AcoPiper:

AcoPiper:
**[Author Round 1]**

**[P1] Metrics-port guard exempted every bare `"9102"` entry — Fixed.**

Confirmed and fixed. The guard scanned the whole file and excluded `- "9102"` wherever it appeared, because that is exactly what the required `expose:` entry looks like. Under `ports:` the same short entry is a publication: Compose gives container port 9102 an ephemeral host port on every interface, so `expose: ["9102"]` plus `ports: ["9102"]` put the unauthenticated endpoint on the host and the test still passed.

`tests/docker_compose_security.rs` now reads `services.step-ca` structurally and judges each mapping by the key it sits under:

- `service_key_block(compose, "step-ca", key)` walks the document by indentation — `services:` at column 0, service names at 2, keys at 4 — and returns the entries nested under that one key, skipping blanks and comments. A neighbouring service publishing 9102 no longer influences the step-ca verdict, and a bare `- "9102"` is now read as an exposure or a publication according to where it sits.
- The exposure assertion is separate, as you asked: `step-ca.expose` must carry an entry equal to `9102`.
- The publication assertion rejects the port appearing anywhere under `step-ca.ports`. That covers all three shapes in one rule — `- "9102:9102"` and `- "127.0.0.1:9102:9102"`, the short `- "9102"`, and the long `- target: 9102` mapping, which carries no `host:container` shape for a syntax-aware split to key on.

One deliberate departure from the suggestion: no YAML parser. The crate has no YAML dependency (`config` is built with the `toml` feature only), and adding one for a test that reads two files this repository owns is more than the check needs — so the reader is ~25 lines of indentation walking rather than `serde_yaml`. The risk that buys is a silent mis-parse: a reader that found the wrong section, or none, would make the guard vacuous rather than red. Three fixture tests close that off — `service_key_block_reads_the_named_section_only` pins the returned entries for `ports`, `expose` and an absent key against a document whose `openbao` service publishes 9102, and `short_form_publication_of_the_metrics_port_is_rejected` / `long_form_publication_of_the_metrics_port_is_rejected` are `#[should_panic]` on exactly the two cases the previous guard let through.

`cargo test --test docker_compose_security` — 13 passed. `cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings` and `cargo test --bins` (74 passed) are clean. No documentation or changelog impact: the change is confined to a test, and the compose files themselves are unchanged.

The PR description now describes the section-scoped guard.

AcoPiper:
**[Reviewer Round 2]**

[P1] Reject inline and aliased `ports` forms instead of treating an unrecognised `ports` value as an empty section. The Round 1 change correctly distinguishes the normal block forms, including bare and long-form mappings. However, `service_key_block` only recognizes a key when the entire trimmed line equals `ports:` ([tests/docker_compose_security.rs:74](https://github.com/aicers/bootroot/blob/AcoPiper/issue-864/tests/docker_compose_security.rs#L74)). A valid future Compose edit such as `ports: ["9102"]` on `step-ca`, while retaining the existing block `expose: - "9102"`, makes that function return `None`; `unwrap_or_default()` at [line 111](https://github.com/aicers/bootroot/blob/AcoPiper/issue-864/tests/docker_compose_security.rs#L111) then lets the host publication pass. An alias (for example `ports: *stepca_ports`) has the same failure mode. That leaves the unauthenticated listener publishable without CI noticing, contrary to the issue's explicit "later edit ... should fail CI" requirement and the test's claim to reject every syntax. Parse the Compose YAML, or make the section reader reject/inspect inline and alias-valued `ports` keys; add a fixture for the form chosen.

AcoPiper:

AcoPiper:
**[Author Round 2]**

**[P1] An unrecognised `ports:` value was read as an empty section — Fixed.**

Confirmed. `service_key_block` recognized a key only when the whole trimmed line was `ports:`, so `ports: ["9102"]` or `ports: *shared_ports` on `step-ca` returned `None`, and `unwrap_or_default()` cleared the section without ever looking at an entry. The exposure block being intact meant the rest of the guard stayed green, so the publication landed with CI quiet — exactly the "later edit should fail CI" case the issue asks the test to catch.

I took the second of the two options you offered — teach the section reader the forms — rather than parsing the YAML. The crate has no YAML dependency at all, and adding one for a test guard runs into the dependency rule; the only obvious candidate, `serde_yaml`, is deprecated and unmaintained, which is a poor thing to make a security assertion depend on. The reader now classifies the key's value by the form it is written in:

- `SectionValue::Block` — the entries are the lines nested under the key, as before.
- `SectionValue::Flow` — the key's own line holds a flow sequence; the entries are split out of it, so `ports: ["9102"]` is judged like any other publication.
- `SectionValue::Unreadable` — anything else sharing the key's line: an alias, a flow mapping, a scalar. The entries are not there to read, so `assert_metrics_port_exposed_never_published` panics on it instead of clearing the section. A guard that cannot see the entries cannot clear them of `9102`, and the failure names the form and asks for a list or a reader that knows it.

One form is deliberately *not* rejected: a bare anchor, `ports: &shared_ports`, declares a name for a value whose entries are still on the lines below, so it reads as a `Block`. Rejecting it would fail CI on an edit that publishes nothing. An anchor with an inline value (`ports: &p ["9102"]`) is not bare and falls through to the flow/unreadable split.

`expose:` is put through the same classification, so a flow-form exposure keeps working rather than reporting the port missing.

Fixtures added, per your request:

- `service_key_value_classifies_the_forms_that_share_the_keys_line` — pins `Flow`, `Unreadable` and the bare-anchor-still-a-block case directly on the reader.
- `flow_sequence_publication_of_the_metrics_port_is_rejected` — `ports: ["9102"]` with the `expose:` block retained, i.e. the exact edit from your report, now panics with "must not publish step-ca's metrics port".
- `aliased_ports_are_rejected` — `ports: *shared_ports` panics with "ports: in a form this guard cannot read".

**Unprompted, while in the same function.** The service scan kept `in_service` set after `services:` ended, so an indent-4 key under a later top-level mapping (`volumes:`, an `x-` extension) could answer for the service. It cannot bite today — neither bundled file has a `ports:` or `expose:` at that position — but it is the same class of "the reader found the wrong section" mistake the fixtures exist to pin, so the flag is now cleared at the document's top level and set only under `services:`.

Verified: `cargo test --test docker_compose_security` — 16 passed (the two new `should_panic` fixtures included); `cargo test --bins` — 74 passed; `cargo fmt -- --check --config group_imports=StdExternalCrate` and `cargo clippy --all-targets -- -D warnings` clean. The change is confined to `tests/docker_compose_security.rs`, so nothing user-facing moved and the CHANGELOG entry stands; the PR description's bullet for this file has been updated to describe the classification and the new fixtures.

AcoPiper:
**[Reviewer Round 3]**

Round 2 P1 is resolved. `service_key_value` now distinguishes block values, flow sequences, and unreadable inline forms; `ports: ["9102"]` is inspected and rejected, while aliases and other values the guard cannot enumerate fail closed. The new fixtures cover both cases, so the metrics listener cannot be silently published through the gap identified in Round 2.

The updated change satisfies the issue: `metricsAddress: ":9102"` is explicitly managed in both `ca.json` reconciliation and the generated sidecar template, metrics-only changes trigger the required sidecar/template/step-ca reload sequence, and the E2E scenario proves the existing-install upgrade and persistence after a sidecar render without replacing CA material. The port-boundary guard, Prometheus assertion, documentation, and changelog are all consistent with that behavior.

No remaining findings.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Serve step-ca metrics on the scraped port
```

**Body**

```text
`monitoring/prometheus.yml` has always scraped `step-ca:9102` and both
bundled compose files expose the port, but step-ca starts that listener
only when `ca.json` carries a top-level `metricsAddress`.  Nothing
wrote the key, and the pinned image accepts the address through no flag
and no environment variable, so compose could not supply it either: the
declared target was permanently down.

`init` now writes the fixed `:9102` into `secrets/config/ca.json` and
stamps it explicitly into `secrets/templates/ca.json.ctmpl` rather than
inheriting whatever the co-owned file holds when the template is
generated — the step-ca OpenBao Agent sidecar can render the previous
template in between, and an inherited value would bake the stale state
back in.

`CaJsonUpdate` reports the metrics change alongside `dns_names_changed`
and the previously `dnsNames`-gated branches gate on
`stepca_reload_required()` instead.  That is what the upgrade path
needs: an installation predating the setting has an unchanged name set,
so gating on `dnsNames` alone would regenerate no template, leave the
sidecar on the old one and leave step-ca without a listener.

The port stays exposed to the compose network and unpublished.  The
endpoint carries no authentication — step-ca offers none for it — so
that boundary is its only protection, and a compose guard now fails on
any `ports:` entry naming 9102, including a section it cannot
enumerate.

Closes #864
```

PR #869, opened 2026-08-17
Split the namespace key from the SAN label with a required registration_id (#756)

### LINKED ISSUE #756: Split the namespace key from the SAN label with a required `registration_id`
# Split the namespace key from the SAN label with a required `registration_id`

## Context

Today `service_name` does two incompatible jobs. It is the second label of the issued certificate's single SAN — `<instance_id>.<service_name>.<hostname>.<domain>`, composed by `src/config.rs::profile_domain` and consumed by `src/acme/flow.rs::build_csr_params` — so it wants to be the component's plain keyword (`piglet`, `roxyd`). It is *also* the sole key of every per-service namespace bootroot owns, which demands deployment-wide uniqueness. Those two demands are compatible only while a component is installed exactly once deployment-wide, which is why each host's roxyd has to register under a host-qualified name and ends up with the host in its SAN twice.

Modules are installed per host and are meant to run several instances on one host, so no single string can satisfy both jobs. This issue splits the second job into a new field.

## Scope

Introduce `registration_id` as a first-class, **required** field and move every namespace key onto it, leaving `service_name` as the SAN label only.

### The validator

- Add a path-safe validator to `src/input_validation.rs` for `registration_id`: lowercase alphanumeric and hyphen, must start and end alphanumeric, non-empty, **maximum 131 octets** (the structural maximum of `<63-octet host>-<63-octet component>-<3-digit instance>`). Add a `ValidationError::InvalidRegistrationId` variant, because no existing variant expresses "path-safe but over-long" precisely. The `match` arms over `ValidationError` in `src/commands/service/resolve.rs:362` and `src/bin/bootroot-remote/validation.rs:25` onward are exhaustive with no `_` arm, so the compiler will point at every mapper that must handle the new variant — handle each one rather than adding a catch-all. **This validator is a shared contract, not a CLI-local check.** The later registrar derivation library applies this same function to the `registration_id` it derives from parts, so keep it in `src/input_validation.rs` beside the other validators — do not inline it into the `service add` resolve path. **Declare it `pub`, and the new `ValidationError` variant with it — this is one of the places the repository's `pub(crate)`-unless-forced default does not apply, and getting it wrong makes the shared-contract rule unenforceable.** `input_validation` is a `pub mod` of the **library**, and its existing consumers reach it across crate boundaries as `bootroot::input_validation::…` — `src/commands/service/resolve.rs` in the `bootroot` binary and `src/bin/bootroot-remote/validation.rs` in a second binary both do, which is why `validate_dns_label` and `ValidationError` are already `pub`. A `pub(crate)` validator would be invisible to both of the call sites this issue itself has to update, and to the derivation library, whose consumer wraps `src/commands/` and therefore also sits outside the library crate. The next author would then either widen it under compile errors or, worse, write the second copy this rule exists to prevent. This issue validates keys that arrive as explicit input and the derivation library validates keys it composes, but there must be exactly **one** function deciding which keys are legal: a second copy would drift on octets-versus-characters or on whether the charset is checked alongside the length, and the two paths would then disagree about the same key.
- Do **not** widen `validate_dns_label`. `service_name`, `hostname` and every `domain` label must stay ≤63 because they are DNS labels in the X.509 CN/SAN (`config.rs::profile_domain` → `acme/flow.rs::build_csr_params`) and in the HTTP-01 Docker network alias (`src/commands/dns_alias.rs`). Repoint only the per-field wrappers that guard the *namespace key*.

### State and namespaces

- Add `registration_id: String` to `ServiceEntry` (`src/state.rs:140`) as a required field — **not** `Option<String>`, **not** `#[serde(default)]`, and **not** defaulting to `service_name`.
- Key `StateFile::services` (`src/state.rs:74`) on `registration_id`, and repoint `require_service_entry` (`src/commands/service/remove.rs:202`) and every other registry lookup accordingly.
- Move all of the following off `service_name` and onto `registration_id`: the AppRole and policy names (`service_role_name` / `service_policy_name`, `src/commands/service/approle.rs:103`/`:107`); the paths inside the generated policy body (`build_service_policy`, `approle.rs:83`); the KV namespace `bootroot/services/<key>/…` (`SERVICE_KV_BASE` consumers in `src/trust_bootstrap.rs`, `src/fast_poll.rs:379-406`, `src/commands/service/secrets.rs`, `src/commands/trust.rs:166`, `src/commands/rotate/approle.rs`, `src/commands/rotate/ca.rs`, `src/commands/rotate/responder_hmac.rs`, `src/commands/service/remove.rs:239`, `src/bin/bootroot-remote/io.rs:152`, `src/bin/bootroot-remote/apply_secret_id.rs:100`); the agent-config managed block markers (`upsert_managed_profile`, `src/commands/service/local_config.rs:442`; `upsert_managed_profile_block` in `src/bin/bootroot-remote/agent_config.rs`; and the marker helpers in `src/trust_bootstrap.rs`); the per-service fast-poll state filename (`state_path_basename` and `default_state_path_for`, `src/commands/service/local_config.rs:265`/`:285`, plus `default_state_path_for` in `src/bin/bootroot-remote/agent_config.rs:404`); the default remote cert/key filenames (`src/bin/bootroot-remote/agent_config.rs:261`/`:265`, the only site that derives a cert or key filename from a registration — `bootroot service add` takes `--cert-path`/`--key-path` explicitly and derives nothing, and `src/commands/init/paths.rs` uses `service_name` only as a Docker Compose top-level service key and for the fixed HTTP-01 responder TLS paths, neither of which is a registration namespace); and the remote-bootstrap artifact directory (`src/commands/service/remote_bootstrap.rs`).
- `state_path_basename` carries a `debug_assert!` and a doc comment asserting its input "is a validated DNS label". Update both to name the new path-safe rule, since the input is now a `registration_id` that may exceed 63 octets.
- Leave `default_fast_poll_state_path` (`src/config/defaults.rs:93`) alone. It returns the fixed constant `bootroot-agent-state.json` (`defaults.rs:26`) and never embedded a service name, so it is not part of this migration.
- SAN composition is unchanged: `profile_domain` keeps reading `instance_id`, `service_name`, `hostname` and `domain`. `registration_id` is **not** a certificate field and must not appear in any SAN.
- The remote-bootstrap artifact carries `registration_id` as a required field alongside `service_name`, since the remote host has no other source for the key its namespaces derive from. A new required field is breaking under the artifact's own compatibility contract, so `schema_version` goes 4 → 5 and `bootroot-remote` accepts 5 only (`MIN_SUPPORTED_SCHEMA_VERSION` == `MAX_SUPPORTED_SCHEMA_VERSION` == 5). Per the no-compatibility constraint below there is no arm for version 4.
- `bootroot-remote` takes the key as an explicit flag on both verbs that name a registration: `bootstrap --registration-id … --service-name …` (both, since it emits a `[[profiles]]` block and so needs the SAN label too) and `apply-secret-id --registration-id …`. Each validates it through the same shared `validate_registration_id`, so the two ends of the artifact cannot disagree about which keys are legal.

### Agent config

- Add `registration_id: String` to `DaemonProfileSettings` (`src/config.rs:124`) as a **required** key with no `#[serde(default)]`, declared exactly the way `service_name` is today. A `[[profiles]]` block that omits `registration_id` is an invalid config and must fail at deserialization, before any validation runs.
- `bootroot-agent` builds its fast-poll KV paths and its per-profile state filename from `registration_id`, not `service_name`.
- Extend `validate_profile` (`src/config/validation.rs:242`, which today enforces only non-empty + ASCII on `profiles.service_name`) to validate `registration_id` with the new ≤131 path-safe rule, and to route `profiles.service_name` through the existing DNS-label rule so an over-long or non-label SAN component is rejected at config load rather than at CSR time.
- Every emitter of a `[[profiles]]` block writes `registration_id`: `agent.toml.example`, `agent.toml.compose`, `src/commands/service/local_config.rs` (the render inputs at `:192` and `upsert_managed_profile` at `:442`), and `src/bin/bootroot-remote/agent_config.rs`.

### CLI surface

- `bootroot service add` takes a `--registration-id` flag. Follow the existing idiom for the other required inputs: declare it as `Option<String>` on `ServiceAddArgs` (`src/cli/args.rs:1333`), and in `resolve_service_add_args` (`src/commands/service/resolve.rs:63`) resolve it exactly like `service_name` — validate the flag when present, otherwise `prompt.prompt_with_validation` for it — so it is required in effect while a bare `service add` stays interactive. Validate with the new rule and persist it on `ServiceEntry`. Do **not** derive it from `--service-name`/`--hostname`/`--instance-id`; see Constraints.
- Add the new prompt label and invalid-value message to the i18n tables: the struct fields in `src/i18n.rs`, the accessors in `src/i18n/service.rs`, and the strings in both `src/i18n/en.rs` and `src/i18n/ko.rs`.
- The flags that *look up* an existing registration now name a `registration_id`, so rename them from `--service-name` to `--registration-id`: `ServiceInfoArgs` (`src/cli/args.rs:1475`), `ServiceUpdateArgs` (`:1482`), `ServiceRemoveArgs` (`:1559`), `VerifyArgs` (`:1610`), `RotateAppRoleSecretIdArgs` (`:659`) and `RotateForceReissueArgs` (`:734`). Per the no-compatibility constraint below, add **no** `--service-name` alias: leaving that spelling in place to mean "the registry key" would re-create the exact conflation this issue removes.
- Update `resolve_verify_service_name` (`src/commands/verify.rs:216`) and the `conflicts_with_all` / `ArgGroup` member lists that name `"service_name"` as a string (`src/cli/args.rs:656`, `:704`, `:717`), since clap resolves those against the field name.

### Fixtures

- Update every fixture and test that constructs a `ServiceEntry`, an `agent.toml`, or a state file — including `tests/bootroot_service.rs`, `tests/bootroot_remote.rs`, `tests/bootroot_rotate.rs`, `tests/bootroot_verify.rs`, and the docker E2E scripts under `tests/e2e/` and `scripts/`.

## Acceptance criteria

- [x] `ServiceEntry` carries a required `registration_id`; deserializing a state file that omits it fails, and there is no `Option`, no `service_name` fallback, and no legacy compatibility arm anywhere in the codebase.
- [x] Every namespace listed in Scope is derived from `registration_id`: registry key, AppRole name, policy name, policy-body KV paths, KV namespace, agent-config managed-block markers, per-service fast-poll state filename, default remote cert/key filenames, and the remote-bootstrap artifact directory.
- [x] The SAN's service label is the plain `service_name`; no SAN anywhere contains a `registration_id`.
- [x] Two registrations of the same component on one host — e.g. `registration_id` `h1-piglet-001` and `h1-piglet-002`, both with `service_name = "piglet"` and `hostname = "h1"` — coexist with distinct registry entries, AppRoles, policies, KV namespaces, state filenames and cert/key paths, and SANs that differ only in the instance label.
- [x] The one-key shape covers the no-instance cases explicitly: a one-per-deployment singleton registers as `review` and a one-per-host component as `h1-roxyd`, each written out literally in a test.
- [x] For the one-per-deployment singletons (`review`, `aice-web-next`, `aimer`) the derived OpenBao paths, AppRole names and filenames are **byte-identical** to today's, since their `registration_id` is still `<component>`.
- [x] The new validator accepts a 131-octet key and rejects a 132-octet one; `validate_dns_label` still rejects anything over 63 octets, and `service_name` / `hostname` / `domain` labels still go through it.
- [x] The validator and its `ValidationError` variant are `pub` items of `src/input_validation.rs`, callable as `bootroot::input_validation::…` from outside the library crate; a reviewer can confirm both binary-side call sites this issue updates reach the same function, so the sibling derivation library has one implementation to consume rather than a reason to write a second.
- [x] `bootroot service add --registration-id h1-piglet-001 --service-name piglet …` registers under that key with no prompting; omitting the flag prompts for it; an invalid value is rejected with the new message in both locales.
- [x] An `agent.toml` whose `[[profiles]]` block omits `registration_id` fails to load with a deserialization error naming the missing field, and one whose `registration_id` violates the path-safe rule fails `validate_profile`. Neither reaches the polling loop.
- [x] Every emitter of a `[[profiles]]` block produces a loadable profile: the blocks written by `service add`, by `upsert_managed_profile`, by the remote-bootstrap agent-config writer, and by the shipped `agent.toml.example` and `agent.toml.compose` templates each carry `registration_id`, and each round-trips through config deserialization and `validate_profile` without error.
- [x] `validate_profile` rejects a `profiles.service_name` that is not a ≤63-octet DNS label.
- [x] `bootroot-agent` polls the namespace named by the profile's `registration_id`, not its `service_name`.
- [x] `service info`, `service update`, `service remove`, `verify`, `rotate approle-secret-id` and `rotate force-reissue` accept `--registration-id`, and none of them still accepts `--service-name` as a lookup key.

## Constraints

- The product is pre-release. There is **no migration and no compatibility shim**: registrations are re-created by re-installing. Do not write a converter, do not accept an old-shaped state file or an old-shaped `agent.toml`, and do not add a "if absent, fall back to `service_name`" arm anywhere.
- Because there is no compatibility arm, an agent and the `agent.toml` written for it move together at install. This forces **no** agent-first rollout ordering and no version negotiation.
- Do **not** implement the `registration_id` derivation rule in this issue — not in the CLI, not anywhere. The rule (per multiplicity class, with zero-padded instance numbers) is owned by ecosystem RFC-A §4 and is implemented exactly once, in a later issue's registrar config-and-derivation library (which the mint verb then calls rather than re-deriving). Deriving it in `service add` would require the component's multiplicity class, which the CLI has no source for, and would create a second derivation site that library later has to reconcile with. Here the key arrives as an explicit input and is only validated.
- Keep the existing visibility discipline (`pub(crate)` / `pub(super)` over `pub`) and the project's no-`unwrap()`-in-production rule — with the stated exception of the shared validator and its error variant, which are `pub` because their consumers live outside the library crate, exactly as the neighbouring validators already are.
- `service_name` must remain constrained to ≤63 octets permanently; widening the shared DNS-label validator would silently widen the certificate's own name.

## Out of scope

- The registrar verbs, the daemon endpoint, the audit record, and any credential work.
- Allocating instance numbers. v1 pins the instance to `001`; this issue delivers only the key *shape*, which is what would be expensive to add later.
- Changing SAN composition or anything in the ACME/EAB issuance path.

## Test plan

- [x] Unit tests for the new validator: accepted charset, boundary at 131 octets, rejection of uppercase, underscores, leading/trailing hyphens, non-ASCII, and empty.
- [x] Unit test asserting `build_service_policy` emits KV paths under the `registration_id`, and that the reissue path is still the only rule carrying `create`/`update` (extend the existing assertions in `src/commands/service/approle.rs`).
- [x] Serde test asserting a `ServiceEntry` JSON without `registration_id` fails to deserialize.
- [x] Config tests: an `agent.toml` whose `[[profiles]]` block omits `registration_id` fails to load, and `validate_profile` rejects both an over-131-octet `registration_id` and an over-63-octet `service_name`.
- [x] CLI parse tests in `src/cli/args.rs` covering `--registration-id` on `service add` and on each renamed lookup command, plus one asserting `--service-name` is rejected where it was removed.
- [x] Integration test in `tests/bootroot_service.rs` registering two instances of one component on one host and asserting all derived artifacts are distinct and the two SANs differ only in the instance label.
- [x] Integration test asserting the singleton and one-per-host shapes (`review`, `h1-roxyd`) produce the expected literal strings, and that singleton paths are byte-identical to the pre-change values.
- [x] Agent-side test asserting a profile's fast-poll KV paths and state filename derive from `registration_id`.
- [x] Test asserting `detect_state_path_collisions` (`src/bin/bootroot-remote/agent_config.rs`, warned on at `src/bin/bootroot-remote/bootstrap.rs:392`) reports no collision for two co-located registrations of one component, since their state paths now differ by `registration_id`.
- [x] The docker E2E lifecycle suites (`tests/docker_e2e_local_lifecycle.rs`, `tests/docker_e2e_remote_lifecycle.rs`, `tests/docker_e2e_two_instance_isolation.rs`) pass end to end after the rename.

## Dependencies

None — this is the foundation and is independent of all confinement work. Part of the registrar-surface umbrella.

## Pointers

- `src/state.rs:140` (`ServiceEntry`), `:201` (`ServiceRoleEntry`), `:74` (`StateFile::services`)
- `src/commands/service/approle.rs:83` (`build_service_policy`), `:103`/`:107` (name derivation)
- `src/commands/service.rs:26` (`SERVICE_ROLE_PREFIX`), `src/trust_bootstrap.rs:7` (`SERVICE_KV_BASE`)
- `src/fast_poll.rs:379-406` (KV path builders); `src/commands/service/local_config.rs:265`/`:285` (per-service state filename derivation)
- `src/config.rs:103` (`profile_domain`), `:124` (`DaemonProfileSettings`); `src/config/validation.rs:242` (`validate_profile`)
- `src/input_validation.rs:8` (`ValidationError`), `:21` (`validate_dns_label`), `src/bin/bootroot-remote/validation.rs`
- `src/cli/args.rs:1333` (`ServiceAddArgs`), `:1475`/`:1482`/`:1559`/`:1610`/`:659`/`:734` (lookup-key flags)
- `src/commands/service/resolve.rs:63` (`resolve_service_add_args`), `src/commands/verify.rs:216` (`resolve_verify_service_name`)
- `src/i18n.rs`, `src/i18n/service.rs`, `src/i18n/en.rs`, `src/i18n/ko.rs` (new prompt and error strings)
- `src/commands/service/remove.rs`, `src/commands/service/remote_bootstrap.rs`, `src/bin/bootroot-remote/agent_config.rs`
- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §2 and §5.5

AcoPiper:
## Summary

`service_name` was doing two incompatible jobs: it was the second label of the issued certificate's SAN (`<instance_id>.<service_name>.<hostname>.<domain>`), so it wanted to be the component's plain keyword, and it was simultaneously the sole key of every namespace bootroot owns, which demands deployment-wide uniqueness. Those demands only coexist while a component is installed exactly once across the deployment — which is why each host's roxyd had to register under a host-qualified name and ended up with the host in its SAN twice.

This PR introduces `registration_id` as a required, first-class field and moves the namespace job onto it, leaving `service_name` as the SAN label and nothing else.

**The validator.** `pub fn validate_registration_id` lives in `src/input_validation.rs` beside the existing validators, with a new `pub ValidationError::InvalidRegistrationId` variant. It is path-safe rather than DNS-label-safe and bounded at 131 octets — the widest key `<63-octet host>-<63-octet component>-<3-digit instance>` can compose. Both binaries reach it across the crate boundary (`src/commands/service/resolve.rs`, `src/bin/bootroot-remote/validation.rs`), which is why it and its error variant are `pub`: the later registrar derivation library consumes the same function, so a key composed by derivation and a key typed by an operator cannot disagree about legality. `validate_dns_label` is untouched — widening it would silently widen the certificate's own name.

**Namespaces.** `registration_id` now names the `state.json` entry key, the AppRole and policy names, the paths inside the generated policy body, the `bootroot/services/<key>/…` KV subtree, the managed `agent.toml` block markers, the per-registration fast-poll state filename and credential directory, the default remote cert/key filenames, and the remote-bootstrap artifact directory. `profile_domain` is unchanged, so no SAN anywhere contains a `registration_id`.

**Agent config.** `DaemonProfileSettings` gains a required `registration_id` with no `#[serde(default)]` — a `[[profiles]]` block that omits it fails at deserialization. `validate_profile` validates it under the new rule and additionally routes `profiles.service_name` and `profiles.hostname` through the DNS-label rule, so an over-long or dotted SAN component is rejected at config load rather than at CSR time. Every emitter writes the key: `agent.toml.example`, `agent.toml.compose`, `service add`'s local-config render, `upsert_managed_profile`, and the remote-bootstrap agent-config writer.

**CLI.** `service add` takes `--registration-id`, resolved exactly like `service_name` (validate when present, prompt otherwise), so it is required in effect while a bare `service add` stays interactive. It is never derived from the other flags — the derivation rule needs the component's multiplicity class, which the CLI has no source for, and is owned by a later issue's registrar library. The six commands that *look up* an existing registration (`service info`, `service update`, `service remove`, `verify`, `rotate approle-secret-id`, `rotate force-reissue`) rename their flag to `--registration-id` with no `--service-name` alias.

**No compatibility.** The product is pre-release: there is no migration, no converter, and no fallback arm anywhere. A state entry, an `agent.toml` profile or a bootstrap artifact missing the key fails to load. The artifact `schema_version` goes 4 → 5 and `bootroot-remote` accepts 5 only.

Closes #756

## Test plan

- [x] Validator unit tests: accepted charset, the 131-octet boundary (131 accepted, 132 rejected), and rejection of uppercase, underscores, leading/trailing hyphens, non-ASCII and empty
- [x] `validate_dns_label` still rejects anything over 63 octets, and `service_name` / `hostname` / `domain` still go through it
- [x] `build_service_policy` emits KV paths under the `registration_id`, and the reissue path is still the only rule carrying `create`/`update`
- [x] A `ServiceEntry` JSON without `registration_id` fails to deserialize
- [x] An `agent.toml` whose `[[profiles]]` block omits `registration_id` fails to load, and `validate_profile` rejects an over-131-octet `registration_id`, an over-63-octet or dotted `service_name`, and an over-63-octet or dotted `hostname`
- [x] CLI parse tests cover `--registration-id` on `service add` and on each renamed lookup command, and `--service-name` is rejected where it was removed
- [x] Two registrations of one component on one host (`h1-piglet-001` / `h1-piglet-002`, both `service_name = "piglet"`, `hostname = "h1"`) yield distinct registry entries, AppRoles, policies, KV namespaces, state filenames and cert/key paths, with SANs differing only in the instance label
- [x] The singleton (`review`) and one-per-host (`h1-roxyd`) shapes produce the expected literal strings, and singleton-derived paths, AppRole names and filenames are byte-identical to the pre-change values
- [x] A profile's fast-poll KV paths and state filename derive from `registration_id`, not `service_name`
- [x] `detect_state_path_collisions` reports no collision for two co-located registrations of one component
- [x] `bootroot service add --registration-id h1-piglet-001 --service-name piglet …` registers with no prompting; omitting the flag prompts; an invalid value is rejected with the new message in both locales
- [x] `cargo clippy --all-targets --all-features -- -D warnings` is clean and `cargo fmt -- --config group_imports=StdExternalCrate --check` passes
- [x] `mkdocs build --strict` passes via `./scripts/check-docs.sh`, and `markdownlint-cli2` is clean
- [x] The docker E2E lifecycle suites (`docker_e2e_local_lifecycle`, `docker_e2e_remote_lifecycle`, `docker_e2e_two_instance_isolation`) pass end to end on CI after the rename


AcoPiper:
**[Reviewer Round 1]**

- **[P1] Keep the changelog release-shaped.** `CHANGELOG.md:7` adds an `[Unreleased]` section (and a `0.3.0...HEAD` reference) to a file whose contract explicitly forbids an unreleased heading: release tooling finds the heading matching the tag and requires it to exist when the release is cut. Do not open an unreleased section here; add this material under the actual release heading when that release is cut.

- **[P2] Validate `profiles.hostname` with the DNS-label validator too.** `src/config/validation.rs:264` still accepts any non-empty ASCII hostname, so a hand-written profile with a 64-character hostname or `hostname = "edge.proxy"` passes config validation and reaches SAN construction. The issue and PR test plan both retain the 63-octet single-DNS-label rule for hostname, and the stated reason for the new `service_name` check applies equally here: reject an invalid SAN component at config load rather than later in issuance. Route it through `validate_dns_label` and add the corresponding config test.

AcoPiper:

AcoPiper:
**[Author Round 1]**

- **[P1] Keep the changelog release-shaped.** — **Pushed back.**

  The diff does not add an `[Unreleased]` section to a released file; it opens one on the first change to land *after* a release, which is what the changelog rule prescribes. `0.3.0` was cut on `origin/main` at `0ad08ec`'s ancestor on 2026-08-17: the heading `## [Unreleased]` became `## [0.3.0] - 2026-08-17`, its link reference became `compare/0.2.0...0.3.0`, and the file that shipped carries no unreleased heading. The rule's next clause is the case this branch is in — "and the next change to land opens a new one." Adding `## [Unreleased]` above `## [0.3.0]` and `[Unreleased]: compare/0.3.0...HEAD` at the bottom is exactly that reopening, and it is the whole of the CHANGELOG diff against `origin/main`.

  The failure mode the rule names is a *different* one: an **empty** section left behind *at release time*, where `.github/workflows/release.yml` greps `^## \[$VERSION\]|^## $VERSION` for the tag and finds nothing after the tag is already pushed. That grep is anchored on the version string, so a populated `[Unreleased]` heading sitting above the released one cannot match it and cannot affect note extraction. What the section becomes at the next cut is the released heading for that version — so this material does land "under the actual release heading when that release is cut", by being renamed there rather than by being withheld until someone remembers it.

  Withholding it is also the outcome the changelog contract argues against on its own terms: `0.3.0` is released and its users have `--service-name` as the lookup flag, a `state.json` without `registration_id` and an `agent.toml` without it. Every one of those breaks for them here, which is precisely "what changed for a user of the last release". Leaving it unwritten until a future release cut means the breaking change is only in git history at the moment a reader needs it in the release notes.

- **[P2] Validate `profiles.hostname` with the DNS-label validator too.** — **Fixed** in `ab92269`.

  Correct, and the gap was one-sided rather than deliberate: every *writer* of a `[[profiles]]` block already holds the hostname to the label rule — `src/commands/service/resolve.rs:376` for `service add` and `src/bin/bootroot-remote/validation.rs:22` for `bootstrap --profile-hostname` both call `validate_dns_label` — so only a hand-edited `agent.toml` could reach `validate_profile` with `hostname = "edge.proxy"` or a 64-octet value, and it was waved through on non-empty + ASCII. The reasoning I used for the new `service_name` check applies unchanged: `hostname` is the SAN's third label in `profile_domain`, so an unusable value belongs to config load, not to a rejected ACME order. RFC 0001 §2 already states the rule ("`hostname` is validated as a single DNS label"); this makes the config path keep it.

  - `src/config/validation.rs`: `profiles.hostname` now goes through `validate_dns_label`, replacing the non-empty + ASCII pair, with `ValidationError::Empty` still mapped to `profiles.hostname must not be empty` so the empty-value message is unchanged.
  - `src/config.rs`: `validate_rejects_non_dns_label_hostname` covers the 64-octet value, the dotted `edge.proxy` value, and the empty value.
  - Docs and changelog: `docs/en/configuration.md` and `docs/ko/configuration.md` now state the ≤63-octet single-label rule for both `service_name` and `hostname` in the profiles section, and the `CHANGELOG.md` entry names both fields rather than only `service_name`.

  No fixture moved: every `hostname` in `agent.toml.example`, `agent.toml.compose`, the `src/` test configs, `tests/bootroot_agent_hardening.rs` and `scripts/preflight/extra/agent-scenarios.sh` is already a plain label (`edge-node-01`, `bootroot-agent`, `node-201`).

`cargo clippy --all-targets --all-features -- -D warnings` is clean, `cargo fmt -- --config group_imports=StdExternalCrate --check` passes, the full `cargo test` run is green (445 lib + 1066 bin + integration suites, 0 failures), `./scripts/check-docs.sh` builds with `--strict`, and `markdownlint-cli2` reports 0 issues. The PR description is updated to name `profiles.hostname` alongside `profiles.service_name`.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Split the namespace key from the SAN label
```

**Body**

```text
`service_name` was doing two incompatible jobs. It was the second
label of the issued certificate's SAN, so it wanted to be the
component's plain keyword, and it was simultaneously the sole key of
every namespace bootroot owns, which demands deployment-wide
uniqueness. Those demands coexist only while a component is installed
exactly once across the deployment, which is why each host's roxyd had
to register under a host-qualified name and ended up with the host in
its SAN twice. Modules are installed per host and run several
instances on one host, so no single string can satisfy both jobs.

`registration_id` now takes the namespace job, leaving `service_name`
as the SAN label and nothing else. It names the state entry, the
AppRole and policy, the paths inside the generated policy body, the
`bootroot/services/<key>` KV subtree, the managed `agent.toml` block
markers, the per-registration fast-poll state filename and credential
directory, the default remote cert/key filenames, and the
remote-bootstrap artifact directory. SAN composition is untouched, so
no certificate carries a `registration_id`.

The validator is `pub` in `src/input_validation.rs`, path-safe rather
than DNS-label-safe and bounded at 131 octets — the widest key
`<63-octet host>-<63-octet component>-<3-digit instance>` can compose.
Both binaries reach it across the crate boundary, and the later
registrar derivation library consumes the same function, so a key
composed by derivation and a key typed by an operator cannot disagree
about legality. `validate_dns_label` is deliberately not widened:
doing so would silently widen the certificate's own name.

The product is pre-release, so there is no migration and no
compatibility arm. A state entry, an `agent.toml` profile or a
bootstrap artifact missing the key fails to load; the artifact
`schema_version` goes 4 to 5 and `bootroot-remote` accepts 5 only.
Registrations are re-created by re-installing.

Closes #756
```

PR #872, opened 2026-08-18
Read the rendered registrar config and derive identities (#757)

### LINKED ISSUE #757: Read the bootler-rendered registrar config and derive registration identities
# Read the bootler-rendered registrar config and derive registration identities

## Context

The registrar verbs added in a sibling issue must not take a composed identity name from the wire, and must not take the deployment domain from the wire either. A caller that could send a composed name could bind one host's name to another host's identity; a caller that could send a domain could mint certificates under a name suffix of its choosing. So the verbs take only the identity's *parts* and read everything deployment-wide from a file the provisioning tool renders onto the bootroot host at install.

This issue builds that reader and the derivation library, with no transport and no OpenBao writes. It is deliberately separated from the verbs so the derivation rule — which two repositories must agree on exactly, or every enrollment in the fleet fails — lands as a small, densely tested unit.

## Scope

- Implement the **registrar config** as **one TOML file**, in the exact shape given below, and document it. The format is settled, not open: this repository's entire configuration surface is TOML (`agent.toml.example`, `responder.toml.example`, the `config` crate built with only the `toml` feature, `src/toml_util.rs`), and a single file makes the render atomic — with three files, or with a format the provisioning tool picks independently, a partial render could leave `domain` present while the component table is missing, which is exactly the fail-open the shape check exists to close. Because this file is the cross-repository contract with the provisioning tool, implement it verbatim rather than re-deriving an equivalent shape.

```toml
fingerprint = "3d0c1f…64 lowercase hex…9ab2"
# Rendered by the provisioning tool (bootler) onto BOTH the
# bootroot/registrar host and the REView host at install, at the same
# absolute path on each. bootroot only reads this file; nothing in this
# repository writes it in production.
#
# The line above is the fingerprint and is NOT part of the digested body.
schema_version = 1
domain = "trusted.domain"

# One table per component, keyed by package-id. A component absent from
# this file is refused, never defaulted. `bootroot` has no entry: it is
# the issuer and never enrolls through the registrar.
[components.review]
multiplicity = "one-per-deployment"   # | "one-per-host" | "many-per-host"
cert_group = 3000                     # numeric gid; omit the key to mean
                                      # "registrations carry no cert-group policy"
reload = { kind = "docker-restart", target = "review" }

[components.roxyd]
multiplicity = "one-per-host"
# cert_group omitted on purpose
reload = { kind = "systemd", target = "roxyd.service" }

[components.piglet]
multiplicity = "many-per-host"
cert_group = 3001
reload = { kind = "docker-restart", target = "piglet" }
```

**This shape is settled jointly with the writer, and the writer is authoritative.** The file is rendered by `aicers/bootler` issue #200, which owns the path, the envelope, the field set, the ordering and the digest rule; where this issue and that one disagree, **#200 wins and this issue is corrected**. Do not change the shape unilaterally — a change here that is not also a change there breaks every enrollment in the deployment, and it breaks it at deploy time rather than in either repository's tests.

**The two varying fields mirror the types this repository already applies to a service certificate**, verified against `origin/main` `491eeef`:

- `cert_group` is a **bare numeric gid**, or the key is omitted. That is `cert_group_gid: Option<u32>` (`src/commands/service/resolve.rs:53`, `src/state.rs:183`) — the same policy `src/cert_group.rs` applies at issuance and at every rotation. Do not model it as a tagged table.
- `reload` is an inline table `{ kind, target }` whose `kind` is exactly one of `sighup`, `systemd`, `docker-restart`, `none`, matching `ReloadStyle` (`src/cli/args.rs:346-356`) — SIGHUP to a named process, reload a systemd unit, restart a Docker container, or no post-renew hook. `none` carries no `target`; the other three require one. **`ReloadStrategy` in `src/state.rs` is NOT this type** — it is the *infrastructure*-certificate rotation strategy (`InfraCertEntry::reload_strategy`, `src/state.rs:56`) and its `container_restart`/`container_signal` vocabulary must not be used here. An earlier draft of this issue took its shape from that enum; that was the wrong type.

- Model `multiplicity` and `reload.kind` as `enum`s over those exact kebab-case strings, and keep them as those `enum`s in the validated config — neither field is ever *stored* as a `String`. An unrecognised value is a load failure, not an unknown-variant fallback, and that failure carries the offending value verbatim and the component whose entry held it, in typed fields (`UnknownMultiplicity` and `UnknownReloadKind` in the error set below). Resolving the string into the `enum` therefore happens outside serde, in the step that turns the parsed body into the validated config: a serde error can only carry the offending value inside a message, so deserializing the field directly into the `enum` would forfeit exactly the typed refusal this issue requires elsewhere. Reading the field as a string in the transient wire struct and converting it immediately is what satisfies both requirements; carrying a `String` into the validated model is what this rule forbids. Declare the default file path as a file-top constant and have the loader take the path as a parameter so tests never touch the real one. **The path is `/etc/clumit-security/provisioning.toml`** — the writer's, not this repository's. It is **not** under `/etc/bootroot/`, deliberately: the same file is rendered to the REView host too (the upload-time safe-set check runs there), and that host has no bootroot and no `/etc/bootroot`. One product-namespaced path is what lets both readers use one constant. `clumit-security` is the Security product's namespace, the same one bootler already stamps into its managed `/etc/hosts` block; bootroot participates in no other product, so the constant is exact rather than a guess.

- **Validate the fingerprint before parsing the body, and refuse an unimplemented `schema_version`.** These are the file's two integrity gates and both are load-bearing here, not the writer's business alone:
  - The first line is exactly `fingerprint = "<64 lowercase hex>"` followed by one `\n`. Split there, SHA-256 the remaining bytes **verbatim as written**, compare, and only then parse. A truncated file can still parse as valid TOML with entries missing, and under the no-entry rule those components are then refused **individually** — a silent per-component enrollment outage indistinguishable from a component that was never provisioned. The digest is what turns that into one loud failure. Do not re-serialize to compute it; the digested region is a byte range.
  - `schema_version` is a `u32` compatibility gate. A version this build does not implement is refused with its own typed variant — never best-effort parsed, and never defaulted when absent.
  - Both refusals are distinct typed variants in the error enum below, so an operator can tell a truncated file from one written by a newer provisioning tool.
- Ship the example file above and a schema description under `docs/`, and keep them honest with a test that loads the shipped example through the production loader.
- Implement a loader that reads the file at verb-invocation time (not cached indefinitely — a re-render must take effect), validates it, and returns typed values. A **missing or unreadable file is a hard failure**, never a default: minting under a guessed domain would issue a certificate no peer will ever verify.
- **Fix the `component` / `service_name` relationship before anything else uses either.** There is exactly one keyword here, not two: the verb's wire input is `(service_name, host, instance, delivery_mode, wrap_ttl, spec)` with **no separate `component` field on the wire**, and RFC §5.1/§5.5 define `service_name` as "the component's plain keyword" (`piglet`, `roxyd`) — a single DNS label that is simultaneously the SAN label. So the wire `service_name`, and only it, does all three jobs: it **selects the config entry** (`[components.<service_name>]`, hence the multiplicity class and the safe-set), it is the **`<component>` segment** of the derived `registration_id`, and it is the **`<service>` segment** of the SAN. Do not add a second selector.
- **Refuse a spec whose `component` disagrees with the wire `service_name`.** The requested `ServiceRegistration` carries its own `component` field; if it is present and not equal to the wire `service_name`, refuse with the **spec-identity disagreement** variant and derive nothing. This is the check that stops a request from having its safe-set validated against one component's rendered entry while its `registration_id` and DNS identity are derived for a different service label — the two would otherwise be selected by different strings. Compare before the safe-set check, so the refusal is unambiguous.
- **Apply the same rule to the spec's `service_name`, and say so explicitly rather than leaving it unmentioned.** The bootler `ServiceRegistration` carries **four** fields — `component`, `service_name`, `reload` and `cert_group` (RFC §5.1) — and the rendered safe-set validates only the two that vary per component, so `service_name` inside the spec is covered by neither the safe-set comparison nor the `component` check as written. Left unstated, a request whose spec carries a matching `cert_group`/`reload` and a *conflicting* `spec.service_name` is neither refused nor deliberately ignored, which is precisely the ambiguity the `component` check exists to remove. So: if `spec.service_name` is present and not equal to the wire `service_name`, refuse with the **same** variant — the spec-identity disagreement, with the offending field recorded — before the safe-set check and before any derivation; if it is absent or equal, proceed. It is **never** a selector, never a segment of the derived `registration_id`, and never a source of the SAN — the wire `service_name` alone does those three jobs, and a composed name arriving inside the spec is exactly the caller-supplied composed name the design removed from the wire (RFC §5.1).
- **Implement against the post-split contract: the spec's identity fields carry the plain keyword, never a composed name.** Under the `registration_id` split the composed name is derived on this side, so a `spec.service_name` of `roxyd-h1` against a wire `service_name` of `roxyd` is a disagreement and is refused like every other — do **not** relax the equality check to accommodate the pre-split `roxyd-<host>` shape, and do not add a compatibility path that strips a host suffix. Cover the composed-name case in tests so the strict behaviour is pinned. Whether the provisioning tool's renderer actually emits the post-split shape is **not verifiable from this repository and is not part of this issue** (see Out of scope); the strict rule is correct either way, because the alternative is a library that quietly accepts a name the design removed from the wire.
- Implement `registration_id` derivation per ecosystem RFC-A §4, from `(service_name, host, instance)` — where `service_name` is the component keyword above — plus the component's multiplicity class: `<component>` for one-per-deployment, `<host>-<component>` for one-per-host, `<host>-<component>-<instance>` for many-per-host, with `instance` rendered as three zero-padded digits. This is the **only** implementation of that rule in this repository.
- Implement SAN composition for a registration from `(instance, service_name, host, domain)`, reusing the existing `<instance>.<service>.<host>.<domain>` shape, with `domain` taken only from the rendered file. **Take the four parts as arguments rather than a profile or a service entry, and pin agreement with the composer that already exists.** The repository composes this name in two places today and neither takes bare parts — `config::profile_domain` formats it from a `Settings` plus a `DaemonProfileSettings`, and `verify.rs::expected_dns_name` from a `ServiceEntry` — so this is a third site by necessity, not by choice, and three `format!` calls that must agree is exactly how a fleet-wide SAN mismatch is introduced. Do **not** refactor the other two: `profile_domain` sits on the ACME issuance path, which a sibling issue explicitly walls off. Instead assert equality in a test — for the same four values, this issue's composer and `profile_domain` must produce byte-identical output — so a later edit to either surfaces here rather than as certificates whose name the verification path rejects. `expected_dns_name` is private to the binary crate and cannot be called from this test; name it in the documentation as the third site that must stay in step. The SAN's `<service>` segment is the same wire `service_name` that selected the config entry — never the derived `registration_id`, which is a namespace key and not a certificate field.
- **The SAN always has four segments, including for a request that correctly omits `instance`.** A one-per-host or one-per-deployment component has no instance dimension, so its request carries no `instance` — but the SAN shape has no optional segment and this repository's certificates are uniformly four-part (`src/commands/verify.rs::expected_dns_name` errors outright on a `ServiceEntry` with no `instance_id`). In that case the `<instance>` segment is the literal `001`, per RFC §5.1 ("a component whose class has no instance dimension takes the default `001`"). A supplied `instance` is rendered three digits zero-padded into the same segment.
- **The `001` default is SAN-only and must never reach the `registration_id` derivation.** A one-per-host component's id stays `<host>-<component>`, never `<host>-<component>-001`; a one-per-deployment component's stays `<component>`. Defaulting before the derivation arm is selected would collapse the two-part/three-part distinction the identity-shape check exists to enforce, and would hand a wrongly-omitted-`instance` module the exact phantom identity that check refuses. Apply the default only when composing the SAN string, after the id has been derived.
- **`host` is required on every request and validated as a DNS label regardless of multiplicity class.** The one-per-deployment derivation arm ignores it — the id is just `<component>` — but the SAN's third segment still needs it, so a missing or non-label `host` is refused for a one-per-deployment component exactly as it is for the other two. "Not used by the id arm" is not "optional".
- Implement the **identity-shape check**: refuse a registration whose `instance` presence contradicts the component's multiplicity class — `instance` present for a one-per-host or one-per-deployment component, or absent for a many-per-host one. A component with **no entry** in the config is refused too, never defaulted; fail-open would let a module whose `instance` was wrongly omitted mint a valid-but-phantom two-part identity that a later correct request then duplicates.
- Implement **spec safe-set validation** against this data model: each component has **exactly one** allowed spec, and a requested spec is inside the safe-set only when **both** `cert_group` and `reload` equal that one rendered spec. It is not a per-field allow-list and not a list of allowed pairs. Independent per-field sets would admit cross-products no component ever declares — an allowed `cert_group` paired with a `reload` it never co-occurs with. A list of pairs is ruled out by RFC §5.1, which states the spec is **host-agnostic** (the same `cert_group`/`reload` for every host of a component) and makes the registration spec a `version-invariant` property of the package-id — a property that only means something if the component has one canonical shape for an uploaded template to be compared against. An omitted `cert_group` in the rendered spec means the component's registrations must carry none; a request that supplies one is outside the safe-set. Comparison is **literal**: there is no template language, no placeholder, and no per-instance parameterization of the rendered `reload`. A component's rendered spec is instance-independent and every registration of that component — including each instance of a many-per-host one — must present it byte-for-byte. If a component ever genuinely needs a per-instance container name, that is a coordinated format change in the provisioning repository, not something this validator pattern-matches around; until then it must refuse, because a matcher that tolerates a varying segment silently widens what one rendered line authorises. The request spec shape is the ecosystem's `ServiceRegistration` — `component`, `service_name`, `reload`, `cert_group`. There is **no privilege field and none is being added**; bootroot derives every service's authority from the fixed derived policy, so no component differs from another on a privilege dimension.
- Define the typed error set these checks produce, as a `thiserror` enum, and treat it as **the** definition of these refusals for the whole repository: `ServiceInstanceMismatch` (the `instance`-presence contradiction), a **distinct** variant for a component **absent** from the config, `ServiceSpecOutsideSafeSet`, the **spec-identity disagreement**, the label/format failures for `service_name`, `host` and the derived key, and the two file-integrity failures — a **fingerprint mismatch** (carrying the declared and the computed digest) and an **unsupported `schema_version`** (carrying the found version and the one this build implements). **The spec-identity disagreement is ONE variant covering both identity-restating fields** — `spec.component` and `spec.service_name` — because they are one refusal semantically: the spec names an identity other than the one the wire selected. It **carries which field disagreed** (and both, when both do), so the audit record and an operator can tell them apart behind the single variant; do not declare two variants, and do not collapse the field into a free-form message string. Downstream issues wrap these variants; none of them redeclares one, so add here every refusal these checks can produce rather than leaving a caller to invent a parallel name for it. The implemented enum is `RegistrarError`, and beyond the refusals listed above it declares the loader-side ones those checks also produce: `ConfigUnreadable`, `FingerprintLineMalformed`, `ConfigMalformed`, `UnknownMultiplicity`, `UnknownReloadKind`, `InvalidDomain`, `InvalidReloadTarget` (a `reload` missing the `target` its `kind` requires, carrying an empty one, or carrying one its `kind` forbids) and `InvalidComponentKey` (a `[components.<key>]` no registration could ever use, because it is not a single DNS label so no wire `service_name` could select it, or not path-safe so no id derived from it could pass). The last two are refused at load rather than left to fail every one of that component's enrollments individually, which is the same silent per-component outage the digest gate exists to turn into one loud failure.
- **The component-absent refusal is its own variant even though it shares a wire identifier with `ServiceInstanceMismatch`.** The endpoint maps both onto that one caller-facing identifier deliberately, and the audit record stores the **internal** variant — so a single variant here would erase, in the one artifact kept for forensic review, the difference between a caller sending the wrong `instance` shape for a known component and a caller probing component names that do not exist. Collapsing is the serialization boundary's job; this library hands it two things to collapse.
- **Do not model a conflict with a previously stored spec here.** This library has no access to stored state and cannot know what a registration was minted with; comparing a request against a stored spec is the verb layer's step and produces a different refusal (`ServiceSpecConflict`). Keeping the two apart matters because they mean different things to an operator: a safe-set refusal says the request was never permissible for this component, while a stored-spec conflict says it was permissible but disagrees with what already exists.
- Validate `service_name` and `host` as single DNS labels before anything else, and enforce the path-safe ≤131-octet bound on the derived `registration_id` **explicitly** rather than inferring it from input lengths, since the inputs are what a caller controls. **Apply the validator the `registration_id` split issue adds to `src/input_validation.rs` — the one raising `ValidationError::InvalidRegistrationId` — to the derived string; do not write a second bound or charset check here.** That issue owns the rule for keys arriving as explicit CLI input and this one owns it for keys derived from parts, but it is **one** rule: two checks would drift on exactly the details that decide legality (octets versus characters, whether the charset is enforced alongside the length), and the fleet would then disagree about which keys are legal depending on how a registration happened to be created.
- **Expose these as separately callable steps, not as one `resolve(request)` entry point, because the verb layer must interleave its own checks between them.** This is a contract with the sibling verbs issue and not an implementation preference. That issue's order is fixed as a correctness requirement — validate labels, resolve the multiplicity class, run the identity-shape check, **derive**, take the per-identity mutex, run the durable-binding **collision check**, then compare the spec — and two of its properties break under a fused call:
  - **A pre-derivation refusal must carry no `registration_id`.** The label checks, the multiplicity-class resolution (including the component-absent refusal) and the identity-shape check therefore have to be reachable **without** deriving anything, so that a refusal on that arm has no id to report and none is computed as a side effect.
  - **The collision check runs between derivation and safe-set validation.** It reads durable state this library has no access to, and it must precede the spec comparison: the rendered spec is host-agnostic, so two hosts of one component present the *identical* spec, and a flow that validated the spec first would match and re-wrap the first host's identity for the second. So safe-set validation must be callable on its own, after the caller has derived and consulted its binding.
  Shape the surface accordingly: a loaded-config handle carrying the domain and the component table, with the label validation, the class resolution and shape check, the derivation, the SAN composition and the safe-set validation each invocable against it. A convenience wrapper that runs them end to end is fine to add for tests, but it must not be the only way in — and nothing in this library may read or require registration state, which is what makes the collision check the caller's step rather than a missing one here.
- Ship a **reusable test fixture builder** for the rendered registrar config — a test-support constructor that writes a valid config (domain, multiplicity classes, safe-sets) into a caller-supplied directory, with overrides for the shapes a test needs. Nothing in this repository renders this file in production; it comes from the provisioning tool in another repository. Without a shared fixture, the verb, endpoint and acceptance-suite issues will each hand-write their own rendered file, and three hand-written approximations of an artifact this repo never produces will drift from each other and from the real one. Export it (a `pub(crate)` test-support module, or a `#[cfg(any(test, feature = "test-support"))]` item) so siblings consume it rather than copy it, and document it in the example-file documentation as the canonical shape. As implemented it is `bootroot::registrar::fixture::RegistrarConfigFixture`, compiled unconditionally rather than behind `cfg(test)` or a `test-support` feature: the siblings that must share it are not all in one compilation unit — unit tests in the library, unit tests in the binary crate (which links the library built *without* `cfg(test)`), and integration tests under `tests/` (which cannot enable a feature on a package they are part of) — so either gate would leave exactly the callers it exists for writing their own.

## Acceptance criteria

- [x] Deriving `("piglet", host "h1", instance 1)` for a many-per-host component yields `h1-piglet-001`; `("roxyd", host "h1", no instance)` for a one-per-host component yields `h1-roxyd`; `("review", host "h1", no instance)` for a one-per-deployment component yields `review` — the host is supplied and validated, it simply does not appear in the id. These three literals are a **contract with the `registration_id` split issue**, which asserts the same three strings as namespace keys in its own fixtures — `review`, `h1-roxyd`, and the `h1-piglet-001` shape. A derivation that emitted a different string for any of them would key a minted registration on a namespace nothing else in the tree produces, so a test here must assert them literally rather than by round-tripping the derivation against itself.
- [x] The loader accepts the documented TOML shape above, and the example file shipped under `docs/` loads through the production loader in a test — so the documented contract and the parser cannot drift apart.
- [x] An unrecognised `multiplicity` or `reload.kind` string is a typed load failure carrying the offending value; neither field is stored as a bare `String`. `reload.kind` accepts exactly `sighup`, `systemd`, `docker-restart` and `none`, and `cert_group` is read as a bare `u32` or absent — the shape the writer renders, mirroring `ReloadStyle` and `cert_group_gid`.
- [x] A missing, unreadable, or malformed registrar config file makes every derivation call fail with a distinct typed error — never a default domain, never an assumed multiplicity class.
- [x] The digest is validated **before** the body is parsed: a file whose body has been truncated at a table boundary — still valid TOML, entries missing — is refused with the fingerprint-mismatch variant rather than loading with fewer components. A test truncates a valid file exactly there and asserts the refusal is the digest failure, not a per-component refusal.
- [x] The declared fingerprint is reachable only from a successfully validated load, so no caller can report a digest it did not check.
- [x] A `schema_version` this build does not implement is refused with its own variant carrying the found and supported versions; an absent `schema_version` is refused, not defaulted.
- [x] The loader's default path constant is `/etc/clumit-security/provisioning.toml`, and a reviewer can confirm the loader still takes the path as a parameter so no test reads the real one.
- [x] A `service_name` or `host` that is not a single DNS label is refused before any derivation happens, and the refusal carries the raw offending string for the caller to record.
- [x] The identity-shape check refuses both mismatch directions with `ServiceInstanceMismatch`, refuses a component absent from the config with a **distinct** variant of its own, and accepts the two matching shapes; a test asserts the two refusals are not the same variant.
- [x] A request that correctly omits `instance` composes a four-segment SAN with the literal `001` in the instance position — `roxyd` on `h1` yields `001.roxyd.h1.<domain>`, `review` on `h1` yields `001.review.h1.<domain>` — while the same requests derive `h1-roxyd` and `review`, with no `001` in either id.
- [x] A supplied `instance` is rendered three digits zero-padded in both the SAN and the id for a many-per-host component: `piglet` instance 1 on `h1` yields SAN `001.piglet.h1.<domain>` and id `h1-piglet-001`.
- [x] A missing or non-DNS-label `host` is refused for a one-per-deployment component, not silently accepted because that arm's id ignores it.
- [x] A derived id longer than 131 octets is refused explicitly, with a test that constructs inputs at and just past the structural maximum, and the refusal comes from the shared `registration_id` validator rather than a length check written locally in this library.
- [x] Spec safe-set validation accepts only a spec equal to the component's single rendered spec in **both** `cert_group` and `reload`, and refuses one with an attacker-chosen `cert_group` or `reload`, with the distinct typed error `ServiceSpecOutsideSafeSet`.
- [x] For a component whose rendered spec omits `cert_group`, a request supplying any `cert_group` is refused with `ServiceSpecOutsideSafeSet`, and a request omitting it is accepted.
- [x] A request whose spec `component` differs from the wire `service_name` is refused with the spec-identity disagreement variant, before the safe-set check and before any derivation, including the case where the spec `component` names a different component that *is* present in the config.
- [x] The same holds for the spec's `service_name`: a request whose `spec.service_name` differs from the wire `service_name` is refused with that same variant even when its `cert_group` and `reload` match the safe-set exactly, and a request whose `spec.service_name` is absent or equal is accepted. A test asserts that no derivation, no `registration_id` and no SAN segment is ever taken from `spec.service_name`.
- [x] The variant records **which** identity field disagreed: a test asserts a `spec.component`-only disagreement, a `spec.service_name`-only disagreement and a both-disagree request are distinguishable behind the single variant, so the audit record does not lose which field the caller got wrong.
- [x] One and the same wire `service_name` value selects the config entry, supplies the `<component>` segment of the derived `registration_id`, and supplies the `<service>` segment of the SAN; a test asserts all three for one request rather than each in isolation.
- [x] This issue's SAN composer takes the four parts and agrees with the existing one: a test asserts its output is byte-identical to `config::profile_domain`'s for the same `instance`, `service_name`, `host` and `domain`, and a reviewer can confirm neither `profile_domain` nor the ACME issuance path was modified to achieve that.
- [x] A rendered `reload` containing a placeholder-looking token (for example `{instance}`) is treated as a literal string, not expanded — so a request whose `reload` carries the expanded form is refused with `ServiceSpecOutsideSafeSet`.
- [x] No error variant in this library refers to a stored or previously applied spec; nothing here reads registration state.
- [x] The derivation rule appears in exactly one place in the repository; nothing else composes a `registration_id`.
- [x] The library's steps are individually callable in the order the verb layer needs: a test runs the label checks, the class resolution and the identity-shape check **without** deriving — asserting no `registration_id` is produced on a refusal there — and a second test derives, then calls safe-set validation as a separate step, asserting the two are not fused and that neither requires registration state. A surface reachable only through a single end-to-end call does not satisfy this.
- [x] A test fixture builder for the rendered config is available to sibling test code, produces a config the loader accepts, and allows a test to override the domain, a component's multiplicity class and a component's safe-set.

## Constraints

- Pure library plus file I/O. No OpenBao calls, no listener, no daemon wiring in this issue.
- The on-disk format is fixed by the shape documented above, and its owner is `aicers/bootler` #200. Do not substitute JSON, YAML, or a multi-file layout, do not move the path, and do not rename or re-nest a field: the provisioning tool renders against this contract, a unilateral change here breaks every install, and it breaks it at deploy time — both repositories' tests stay green. A shape change is a coordinated edit to #200 first and this issue second, never the reverse.
- Do not paraphrase or "improve" the derivation rule: a second implementation that disagreed by one character would fail every enrollment in the fleet, and only after deployment. Implement exactly the three arms above with three-digit zero-padded instances.
- The safe-set reference is the **locally rendered file**, never a signed package. The registrar is the bootroot-co-located roxyd; it never receives the signed package and cannot read the review-host module store, so it has no signed template to compare against.
- The safe-set has no runtime-update channel; it is fixed at install. Do not add one.
- Do not introduce a placeholder, glob, regex or any other template syntax into the rendered spec. Safe-set comparison is equality on parsed values and nothing else.
- Use `thiserror` for the error enum and keep it `pub(crate)` unless a compiler error forces wider visibility.
- No `unwrap()` in production code; no `[]` indexing into slices or vectors.

## Out of scope

- Any OpenBao write, the durable host binding, and the mint/deregister control flow.
- The transport, the socket, and mTLS.
- Rendering the config file — that is the provisioning tool's job in another repository. This issue only consumes it and ships the documented shape plus an example.
- Confirming what that renderer actually emits into the spec's identity fields, or reconciling a divergence with the other repository. This issue implements the strict post-split rule and tests the composed-name refusal; the cross-repository check is coordination outside this leaf and is not a deliverable here.

## Test plan

- [x] Table-driven unit tests over all three derivation arms, including hyphenated component and host names.
- [x] Unit tests for every hard-failure path: file missing, file unreadable, malformed content, unknown component, missing domain, unrecognised `multiplicity`, unrecognised `reload.kind`, absent or unsupported `schema_version`, a first line that is not the exact fingerprint form, and a body whose digest does not match.
- [x] A truncation test: take a valid file, cut the body at a `[components.*]` boundary so it still parses as TOML with entries missing, and assert the load fails on the fingerprint rather than succeeding with a short component table.
- [x] A test that loads the example file shipped under `docs/` through the production loader and asserts its components resolve.
- [x] Unit tests for both `ServiceInstanceMismatch` directions, for the both-matching cases, and for the component-absent refusal asserting it carries its own variant rather than reusing `ServiceInstanceMismatch`.
- [x] Unit test asserting a known colliding pair derives the same id — components `web` and `aimer-web`, with `web` on host `h1-aimer` and `aimer-web` on host `h1`, both yielding `h1-aimer-web-001` — documenting that the derivation is not injective in general and that a uniqueness check elsewhere is therefore required.
- [x] Unit tests pairing id and SAN for one request across all three multiplicity classes, asserting the `001` default appears in the SAN and never in the id — the regression that would catch the default being applied before arm selection.
- [x] A unit test that a one-per-deployment request with a missing or malformed `host` is refused.
- [x] Unit tests for the ≤131 bound at the boundary.
- [x] Unit tests for safe-set accept and reject, including a spec that differs only in `cert_group`, one that differs only in `reload.kind`, one that differs only in `reload.target`, and the omitted-`cert_group` component in both directions.
- [x] Unit tests for the `component` / `service_name` agreement check: equal values pass; a spec `component` naming an unrelated string is refused; a spec `component` naming a *different configured* component is refused rather than validated against that other component's safe-set.
- [x] Unit tests for the `spec.service_name` agreement check on the same three shapes, plus the absent case, plus a spec carrying a *composed* name (`roxyd-h1`) against wire `service_name` `roxyd` — refused, with the derived id and SAN unaffected because neither reads the field.
- [x] A test that, for a single many-per-host request, asserts the selected config entry, the derived `registration_id` and the composed SAN all key off the same `service_name` — the regression that would catch a second selector being introduced.
- [x] Composer-agreement test asserting this issue's SAN string equals `config::profile_domain`'s for the same four values, across all three multiplicity classes including the `001`-defaulted ones.
- [x] Tests use `tempfile::tempdir()` for the rendered config; no fixed paths.
- [x] The fixture builder is itself exercised: a test builds a config with it, loads it through the production loader, and asserts each override takes effect.

## Dependencies

Depends on the `registration_id` split issue (it produces the key this derives, and the path-safe `registration_id` validator this consumes). Part of the registrar-surface umbrella.

**The file's shape is no longer an assumption — it is reconciled with the writer.** The path, the envelope (`fingerprint`, `schema_version`, `domain`), the flat `[components.<package-id>]` table and the two varying fields are settled jointly with `aicers/bootler` #200, and both bodies now carry the same shape with #200 named as authoritative. That reconciliation also corrected two things this issue had wrong: the file lives at the writer's product-namespaced path rather than under `/etc/bootroot/`, and `reload` follows `ReloadStyle` rather than the infrastructure-certificate `ReloadStrategy` an earlier draft copied.

**One assumption remains, and it is about the spec's contents rather than the file:** the provisioning tool renders the post-split spec, with plain keywords in `component` and `service_name`. If that assumption is wrong, the failure is loud but late and belongs to whoever coordinates the two repositories, not to this issue's implementer — the safe-set compares only `cert_group` and `reload`, so composed identity fields are invisible to the loader and to every test written against the fixture builder here, and would surface at deploy time as *every* enrollment of that component refused with the spec-identity disagreement, indistinguishable from a caller attacking the check. Flagged here so the coordination happens; nothing in this issue's scope turns on it. The wire-contract issue is where it may get answered without anyone leaving this repository: it transcribes the spec's field set from the source at a pinned revision and is asked to record what that document says about these two fields' shape, or to record that it says nothing. Read its reference before assuming either way.

## Pointers

- `src/input_validation.rs` (`validate_dns_label`, `validate_numeric_instance_id`; plus the path-safe `registration_id` validator and `ValidationError::InvalidRegistrationId` that the split issue adds there — consume it, do not re-add it)
- `src/config.rs:103` (`profile_domain`) for the SAN shape, `src/commands/verify.rs::expected_dns_name` for the verification-side re-derivation
- `src/config.rs:255` (`Config::builder`) and `src/toml_util.rs` for the repo's TOML loading idiom; `agent.toml.example` for the documented-example convention
- `src/cli/args.rs:346-356` (`ReloadStyle`: `Sighup` / `Systemd` / `DockerRestart` / `None`) and `src/commands/service/resolve.rs:53` + `src/state.rs:183` (`cert_group_gid: Option<u32>`), with `src/cert_group.rs` for what the gid policy does — these are the two varying fields' real types on the **service** path. `src/state.rs:21` (`ReloadStrategy`) is the **infrastructure**-certificate strategy and is deliberately **not** referenced by this file's shape.
- `aicers/bootler` issue [#200](https://github.com/aicers/bootler/issues/200) — the writer. It owns the path, the envelope, the field set, the ordering and the digest rule; this issue mirrors them and is corrected when the two disagree.
- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §5.1 and §5.5

AcoPiper:
Closes #757.

Adds `bootroot::registrar`: the reader for the bootler-rendered provisioning file, and the registration-identity derivation the registrar verbs will run against it. Pure library plus file I/O — no OpenBao calls, no listener, no daemon wiring, and nothing here reads registration state.

## The file

`/etc/clumit-security/provisioning.toml`, implemented verbatim from the shape settled with `aicers/bootler` #200. One TOML file, envelope of `fingerprint` / `schema_version` / `domain`, one `[components.<package-id>]` table per component carrying `multiplicity`, an optional bare-`u32` `cert_group`, and a `{ kind, target }` `reload` whose `kind` mirrors `ReloadStyle` (not the infrastructure-certificate `ReloadStrategy`). `multiplicity` and `reload.kind` are enums; an unrecognised value is a typed load failure carrying the offending string, and neither is stored as a `String`.

Both integrity gates run before anything downstream:

- The first line must be exactly `fingerprint = "<64 lowercase hex>"` followed by one `\n`, and the SHA-256 of the remaining bytes — verbatim as written, as a byte range, never re-serialized — must equal it. A body truncated at a `[components.*]` boundary still parses as valid TOML with entries missing, and those components would then be refused individually; the digest turns that into one loud failure. A test cuts a valid file exactly there and asserts the refusal is the digest one.
- `schema_version` is refused when it is not this build's, with its own variant carrying both versions, and refused when absent rather than defaulted.

The declared fingerprint is reachable only through `RegistrarConfig::load`, so no caller can report a digest it did not check. A missing or unreadable file is a hard failure. The loader takes the path as a parameter, so no test reads the real one.

Two shapes are refused at load rather than left to fail every one of a component's enrollments individually — the same silent per-component outage the digest gate exists to close. A `reload` whose `target` contradicts its `kind` (missing, empty, or present under `none`) is one; a `[components.<key>]` no registration could ever use is the other, since the key is spent twice and must satisfy both the DNS-label rule that lets a wire `service_name` select it and the `registration_id` rule that lets an id derived from it pass.

## The derivation

One rule, one place. `<component>` / `<host>-<component>` / `<host>-<component>-<instance>` by multiplicity class, instances three digits zero-padded. `("piglet", h1, 1)` → `h1-piglet-001`, `("roxyd", h1, none)` → `h1-roxyd`, `("review", h1, none)` → `review`, asserted as literals rather than round-tripped. The ≤131-octet bound is enforced on the derived string by the shared `input_validation::validate_registration_id`, not by a second check written here.

The SAN takes the four parts and always has four segments: a request that correctly omits `instance` takes the literal `001` there, and that default never reaches the id. A test asserts this composer's output is byte-identical to `config::profile_domain`'s for the same four values across all three classes; neither `profile_domain` nor the ACME issuance path was touched. `commands::verify::expected_dns_name` and `commands::dns_alias::dns_alias_for_entry` are private to the binary crate and cannot be called from that test, so both are named in the documentation's table of composition sites that must stay in step. `RegistrarConfig::san_for` supplies the domain from the loaded file, so a verb cannot compose under one that arrived on the wire.

One keyword does all three jobs — the wire `service_name` selects the config entry, is the `<component>` segment of the id and the `<service>` segment of the SAN. There is no second selector, and a test asserts all three for one request.

## Refusals

`RegistrarError` is the definition of these refusals for the repository. `ServiceInstanceMismatch` and the component-absent refusal are **distinct** variants even though the endpoint will map both onto one caller-facing identifier: the audit record stores the internal one, and collapsing them there would erase the difference between a wrong `instance` shape and a caller probing names that do not exist. The spec-identity disagreement is **one** variant covering `spec.component` and `spec.service_name`, carrying which field disagreed (and both, when both do), in a typed field rather than a message string. Nothing refers to a stored or previously applied spec — that is the verb layer's `ServiceSpecConflict`.

Safe-set validation is equality on both `cert_group` and `reload` against the component's single rendered spec: not a per-field allow-list, not a list of pairs, no template language. A rendered `piglet-{instance}` is a literal, and a request presenting `piglet-001` is refused.

## Separately callable steps

Label validation, class resolution, the shape check, derivation, SAN composition and safe-set validation are each invocable on their own, because the verb layer interleaves its per-identity mutex and durable-binding collision check between the derivation and the spec comparison. Two tests pin it: one runs the pre-derivation arm and asserts no `registration_id` is produced on a refusal there, another derives and then validates the spec as a separate call. `resolve_end_to_end` exists for tests and is documented as not for the verb layer.

## Docs and fixture

`docs/reference/registrar-provisioning-config.md` documents the schema, and `docs/reference/provisioning.toml.example` is the shipped example with a real digest of its own body. One test loads the example through the production loader; another asserts the documentation still quotes it verbatim. Both sit outside the mirrored `docs/en/` + `docs/ko/` pair on the precedent of `registrar-wire-contract.md`.

`RegistrarConfigFixture` renders the file for tests, with overrides for the domain, a component's multiplicity class, a component's safe-set, the schema version, the fingerprint and a raw entry for the values the loader must refuse.

## Deviations from the issue, stated

- **The fingerprint line is first, and the example's commentary follows it.** The issue's example block puts its comment header above `fingerprint`, while the same block's own comment — and the digest rule — say the first line *is* the fingerprint. The rule is the load-bearing half, so the shipped example carries the fingerprint on line 1 and every comment verbatim after it.
- **The library is `pub`, not `pub(crate)`.** Nothing inside the library crate calls it yet, so `pub(crate)` items fail the dead-code lint under `-D warnings`; and the verb, endpoint and acceptance-suite code that will consume it lives in the binary crate and under `tests/`, neither of which can see a `pub(crate)` item here. The fixture builder is exported unconditionally for the same reason: `#[cfg(any(test, feature = "test-support"))]` is invisible to both of those consumers, which are exactly the siblings it exists to keep from hand-writing their own. It contains no panics and no production code path reaches it.
- **`multiplicity` and `reload.kind` are converted from strings by a private parse layer** rather than by `#[derive(Deserialize)]` on the enums, because the refusal has to carry the offending value in a typed field and a serde error can only carry it inside a message. Neither field is *stored* as a string; the public model holds enums.

## Test plan

- [x] All three derivation arms yield the contract literals: `h1-piglet-001`, `h1-roxyd`, `review` — asserted as literals, including hyphenated component and host names.
- [x] The example shipped under `docs/` loads through the production loader and its components resolve, so the documented contract and the parser cannot drift.
- [x] An unrecognised `multiplicity` or `reload.kind` is a typed load failure carrying the offending value; all four of `sighup`, `systemd`, `docker-restart`, `none` are accepted; `cert_group` reads as a bare `u32` or absent.
- [x] A missing, unreadable or malformed file fails with a distinct typed error — never a default domain, never an assumed class.
- [x] A body truncated at a `[components.*]` boundary — still valid TOML, entries missing — is refused with the fingerprint mismatch, not a per-component refusal.
- [x] The declared fingerprint is reachable only from a validated load.
- [x] An unsupported `schema_version` is refused with its own variant carrying both versions; an absent one is refused, not defaulted.
- [x] A first line that is not the exact fingerprint form, and a digest that does not match the body, are each refused.
- [x] A `reload` whose `target` contradicts its `kind`, and a `[components.<key>]` no registration could use, are refused at load.
- [x] The default path constant is `/etc/clumit-security/provisioning.toml`, and the loader still takes the path as a parameter.
- [x] A `service_name` or `host` that is not a single DNS label is refused before any derivation, carrying the raw offending string.
- [x] The identity-shape check refuses both mismatch directions with `ServiceInstanceMismatch`, refuses an absent component with a distinct variant, and a test asserts the two are not the same variant.
- [x] A request omitting `instance` composes `001.roxyd.h1.<domain>` and `001.review.h1.<domain>` while deriving `h1-roxyd` and `review` — no `001` in either id.
- [x] A supplied `instance` renders zero-padded in both: SAN `001.piglet.h1.<domain>`, id `h1-piglet-001`.
- [x] A one-per-deployment request with a missing or malformed `host` is refused.
- [x] A derived id at and just past 131 octets is refused by the shared `validate_registration_id`, not a local length check.
- [x] Safe-set validation accepts only the single rendered spec and refuses one differing in `cert_group` alone, `reload.kind` alone, or `reload.target` alone, with `ServiceSpecOutsideSafeSet`.
- [x] For a component rendering no `cert_group`, a request supplying one is refused and a request omitting it is accepted.
- [x] A `spec.component` disagreement is refused before the safe-set check and before any derivation, including when it names a different *configured* component.
- [x] A `spec.service_name` disagreement is refused with the same variant even when `cert_group` and `reload` match exactly; absent or equal is accepted; a composed `roxyd-h1` against wire `roxyd` is refused, and no derivation or SAN segment reads the field.
- [x] The variant records which field disagreed — component-only, service-name-only and both-disagree are distinguishable behind the single variant.
- [x] One wire `service_name` selects the entry, supplies the id's `<component>` and the SAN's `<service>`, asserted together for one request.
- [x] The SAN composer's output is byte-identical to `config::profile_domain`'s across all three classes, and neither `profile_domain` nor the ACME issuance path was modified.
- [x] A rendered `{instance}`-looking token is a literal: a request carrying the expanded form is refused with `ServiceSpecOutsideSafeSet`.
- [x] No error variant refers to a stored or previously applied spec, and nothing here reads registration state.
- [x] The derivation rule appears in exactly one place in the repository.
- [x] The steps are individually callable: one test runs the pre-derivation arm and asserts no id is produced on a refusal, another derives and then validates the spec separately.
- [x] The colliding pair (`web` on `h1-aimer`, `aimer-web` on `h1`) both derive `h1-aimer-web-001`, documenting that the derivation is not injective.
- [x] The fixture builder produces a config the loader accepts, and each override takes effect.
- [x] Tests use `tempfile::tempdir()`; no fixed paths.

## Testing

- `cargo test` — full suite green across all targets, including 45 tests in `registrar::tests`.
- `scripts/preflight/ci/check.sh` — `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, rustdoc with `-D warnings`, ruff, biome, markdownlint, `scripts/check-docs.sh` (`mkdocs build --strict` plus the theme verification), and `cargo audit` (one pre-existing allowed warning for the unmaintained `rustls-pemfile`).
- The Docker E2E preflights (`e2e-matrix.sh`, `e2e-extended.sh`) were not run. This change adds a library nothing calls yet plus two documentation files; it touches no Docker lifecycle, no E2E script and no code path those suites exercise.

### DIFF
--- a/docs/reference/provisioning.toml.example
+++ b/docs/reference/provisioning.toml.example
@@ -1,0 +1,28 @@
+fingerprint = "62299a244c4b3833c4cb3a36b4d3da763bd87f664c7d31845f48e53be4170f3e"
+# Rendered by the provisioning tool (bootler) onto BOTH the
+# bootroot/registrar host and the REView host at install, at the same
+# absolute path on each. bootroot only reads this file; nothing in this
+# repository writes it in production.
+#
+# The first line is the fingerprint and is NOT part of the digested body.
+schema_version = 1
+domain = "trusted.domain"
+
+# One table per component, keyed by package-id. A component absent from
+# this file is refused, never defaulted. `bootroot` has no entry: it is
+# the issuer and never enrolls through the registrar.
+[components.review]
+multiplicity = "one-per-deployment"   # | "one-per-host" | "many-per-host"
+cert_group = 3000                     # numeric gid; omit the key to mean
+                                      # "registrations carry no cert-group policy"
+reload = { kind = "docker-restart", target = "review" }
+
+[components.roxyd]
+multiplicity = "one-per-host"
+# cert_group omitted on purpose
+reload = { kind = "systemd", target = "roxyd.service" }
+
+[components.piglet]
+multiplicity = "many-per-host"
+cert_group = 3001
+reload = { kind = "docker-restart", target = "piglet" }
--- a/docs/reference/registrar-provisioning-config.md
+++ b/docs/reference/registrar-provisioning-config.md
@@ -78,10 +78,19 @@ under a guessed domain would issue a certificate no peer will ever verify.
 
 ## 3. Component entries
 
-`[components.<package-id>]`, one per component. A component **absent** from
-this file is refused, never defaulted — fail-open would let a module whose
-`instance` was wrongly omitted mint a valid-but-phantom two-part identity that
-a later correct request then duplicates. `bootroot` has no entry: it is the
+`[components.<package-id>]`, one per component. The key is spent twice — a
+wire `service_name` selects the entry, and the same value is the
+`<component>` segment of every `registration_id` derived for it — so it must
+be both a single DNS label and path-safe: lowercase letters, digits and
+hyphens, starting and ending alphanumeric, at most 63 octets. A key that is
+not is `RegistrarError::InvalidComponentKey` at load, rather than an entry
+that loads and then refuses every one of its enrollments at the derivation
+step.
+
+A component **absent** from this file is refused, never defaulted — fail-open
+would let a module whose `instance` was wrongly omitted mint a
+valid-but-phantom two-part identity that a later correct request then
+duplicates. `bootroot` has no entry: it is the
 issuer and never enrolls through the registrar.
 
 | Key | Type | Required | Meaning |
--- a/docs/reference/registrar-provisioning-config.md
+++ b/docs/reference/registrar-provisioning-config.md
@@ -1,0 +1,263 @@
+<!-- markdownlint-configure-file {
+  "MD013": { "tables": false, "code_blocks": false }
+} -->
+
+# Registrar provisioning config (`provisioning.toml`)
+
+**Writer rule.** This repository *reads* this file and never writes it in
+production. It is rendered by the provisioning tool, `aicers/bootler`, whose
+issue [#200](https://github.com/aicers/bootler/issues/200) owns the path, the
+envelope, the field set, the ordering and the digest rule. Where this file and
+that issue disagree, **#200 wins and this file is corrected.** A change made
+here that is not also made there breaks every enrollment in the deployment, and
+breaks it at deploy time rather than in either repository's tests.
+
+Like `registrar-wire-contract.md`, this reference sits outside the mirrored
+`docs/en/` + `docs/ko/` operator pair: it is a cross-repository contract rather
+than operator documentation, so `mkdocs.yml` lists it in neither locale's nav
+and there is no `docs/ko/` counterpart. Its path is stable and is what a test
+reads it by.
+
+## 1. Path
+
+| Item | Value |
+| --- | --- |
+| Absolute path | `/etc/clumit-security/provisioning.toml` |
+| Constant | `bootroot::registrar::DEFAULT_CONFIG_PATH` |
+| Rendered onto | the bootroot/registrar host **and** the REView host, at the same path on each |
+
+The path is the writer's product-namespaced one, deliberately **not** under
+`/etc/bootroot/`. The same file is rendered onto the REView host — the
+upload-time safe-set check runs there — and that host has no bootroot and no
+`/etc/bootroot`. One product-namespaced path is what lets both readers use one
+constant. `clumit-security` is the Security product's namespace, the same one
+bootler stamps into its managed `/etc/hosts` block.
+
+`RegistrarConfig::load` takes the path as a parameter, so no test reads the
+real one.
+
+## 2. Envelope
+
+One TOML file, not three and not a format the writer picks independently. A
+single file makes the render atomic: with a partial render, `domain` could be
+present while the component table was missing, which is exactly the fail-open
+the digest gate exists to close.
+
+| Key | Type | Required | Meaning |
+| --- | --- | --- | --- |
+| `fingerprint` | 64 lowercase hex characters | yes | SHA-256 of the file's body. **Must be the first line**, exactly `fingerprint = "<64 lowercase hex>"` followed by one `\n`. Not part of the digested body. |
+| `schema_version` | `u32` | yes | Compatibility gate. This build implements **1**. Never defaulted when absent. |
+| `domain` | dot-separated DNS name | yes | The deployment-wide domain, the SAN's fourth segment. Read only from this file, never from the wire. |
+| `components` | table of tables | yes in practice | One entry per component, keyed by package-id. |
+
+### 2.1 The two integrity gates
+
+Both are load-bearing on this side, not the writer's business alone.
+
+1. **The digest is validated before the body is parsed.** Split at the first
+   `\n`, SHA-256 the remaining bytes **verbatim as written**, compare with the
+   declared value, and only then parse. The digested region is a byte range;
+   nothing is re-serialized to compute it. A truncated file can still parse as
+   valid TOML with entries missing, and under the no-entry rule those
+   components would then be refused *individually* — a silent per-component
+   enrollment outage indistinguishable from a component that was never
+   provisioned. The digest turns that into one loud failure,
+   `RegistrarError::FingerprintMismatch`, carrying both the declared and the
+   computed digest.
+2. **An unimplemented `schema_version` is refused** with
+   `RegistrarError::UnsupportedSchemaVersion`, carrying the found version and
+   the one this build implements. It is never best-effort parsed, and never
+   defaulted when absent.
+
+The declared fingerprint is reachable only through a successfully validated
+load (`RegistrarConfig::fingerprint`), so no caller can report a digest it did
+not check.
+
+A missing or unreadable file is a **hard failure**, never a default: minting
+under a guessed domain would issue a certificate no peer will ever verify.
+
+## 3. Component entries
+
+`[components.<package-id>]`, one per component. A component **absent** from
+this file is refused, never defaulted — fail-open would let a module whose
+`instance` was wrongly omitted mint a valid-but-phantom two-part identity that
+a later correct request then duplicates. `bootroot` has no entry: it is the
+issuer and never enrolls through the registrar.
+
+| Key | Type | Required | Meaning |
+| --- | --- | --- | --- |
+| `multiplicity` | enum, see §3.1 | yes | How many times the component may be installed. Selects the `registration_id` derivation arm. |
+| `cert_group` | bare `u32` | no | Numeric gid. **Omitting the key means "registrations carry no cert-group policy"** — not "any gid is allowed". |
+| `reload` | inline table, see §3.2 | yes | The post-renew hook. |
+
+`cert_group` mirrors `cert_group_gid: Option<u32>`, the type this repository
+already applies to a service certificate (`src/commands/service/resolve.rs`,
+`src/state.rs`), and it is the policy `src/cert_group.rs` applies at issuance
+and at every rotation. It is a bare number, never a tagged table.
+
+### 3.1 `multiplicity`
+
+| Value | `registration_id` arm |
+| --- | --- |
+| `one-per-deployment` | `<component>` |
+| `one-per-host` | `<host>-<component>` |
+| `many-per-host` | `<host>-<component>-<instance>` |
+
+An unrecognised value is a typed load failure carrying the offending string
+(`RegistrarError::UnknownMultiplicity`), never an unknown-variant fallback and
+never stored as a bare string.
+
+### 3.2 `reload`
+
+An inline table `{ kind, target }`, mirroring `ReloadStyle`
+(`src/cli/args.rs`) — the type this repository already applies to a **service**
+certificate.
+
+| `kind` | Meaning | `target` |
+| --- | --- | --- |
+| `sighup` | Send `SIGHUP` to a process by name | required |
+| `systemd` | Reload a systemd unit | required |
+| `docker-restart` | Restart a Docker container | required |
+| `none` | No post-renew hook | must be absent |
+
+`ReloadStrategy` in `src/state.rs` is **not** this type. It is the
+*infrastructure*-certificate rotation strategy, and its
+`container_restart` / `container_signal` vocabulary must not appear here.
+
+An unrecognised `kind` is `RegistrarError::UnknownReloadKind`, carrying the
+offending string.
+
+`target` is compared **literally**. There is no template language, no
+placeholder, no glob, no regex and no per-instance parameterization: a rendered
+`target` of `piglet-{instance}` is the literal string `piglet-{instance}`, and
+a request presenting the expanded `piglet-001` is outside the safe-set. A
+component's rendered spec is instance-independent and every registration of
+that component — including each instance of a many-per-host one — must present
+it byte-for-byte. A component that genuinely needs a per-instance container
+name is a coordinated format change in the provisioning repository, not
+something this validator pattern-matches around.
+
+## 4. The safe-set
+
+A component has **exactly one** allowed spec: its rendered `cert_group` and
+`reload`. A requested spec is inside the safe-set only when **both** are equal
+to it.
+
+It is not a per-field allow-list — independent per-field sets would admit
+cross-products no component ever declares, an allowed `cert_group` paired with
+a `reload` it never co-occurs with. It is not a list of allowed pairs either:
+RFC-F §5.1 makes the spec **host-agnostic**, the same `cert_group`/`reload` for
+every host of a component, and a `version-invariant` property of the package-id
+— which only means something if the component has one canonical shape for an
+uploaded template to be compared against.
+
+Where a component's rendered spec **omits** `cert_group`, its registrations
+must carry none: a request supplying any gid is refused with
+`RegistrarError::ServiceSpecOutsideSafeSet`.
+
+The safe-set reference is this locally rendered file, never a signed package —
+the registrar is the bootroot-co-located roxyd, which never receives the signed
+package and cannot read the review-host module store. It is fixed at install
+and has no runtime-update channel.
+
+## 5. What the file does **not** decide
+
+| Not here | Where it lives |
+| --- | --- |
+| A privilege field | Does not exist. bootroot derives every service's authority from the fixed derived policy, so no component differs from another on a privilege dimension. |
+| The composed identity name | Derived on this side from the wire's parts (§6). The wire carries no composed name. |
+| A conflict with a previously stored spec | The verb layer's `ServiceSpecConflict`. Nothing in this library reads registration state. |
+
+## 6. What this repository derives from it
+
+`bootroot::registrar::identity` owns the derivation. There is exactly one
+keyword: the wire `service_name` selects the config entry, supplies the
+`<component>` segment of the derived `registration_id`, and supplies the
+`<service>` segment of the SAN. There is no second selector, and
+`spec.service_name` is never any of those three.
+
+| Output | Rule |
+| --- | --- |
+| `registration_id` | The multiplicity arm in §3.1, with `instance` rendered three digits zero-padded. |
+| SAN | `<instance>.<service>.<host>.<domain>`, always four segments. A request that correctly omits `instance` takes the literal `001` in the first segment. |
+
+The `001` default is **SAN-only**. A one-per-host component's id stays
+`<host>-<component>` and a one-per-deployment component's stays `<component>`;
+the default is applied only when composing the SAN, after the id has been
+derived.
+
+`host` is required on every request and validated as a DNS label regardless of
+class. The one-per-deployment arm ignores it, but the SAN's third segment still
+needs it, so "not used by the id arm" is not "optional".
+
+### 6.1 The three SAN composition sites
+
+This repository composes `<instance>.<service>.<host>.<domain>` in **three**
+places, and they must agree byte for byte or the fleet gets certificates whose
+name the verification path rejects.
+
+| Site | Takes | Kept in step by |
+| --- | --- | --- |
+| `bootroot::registrar::identity::compose_san` | the four parts | this table, and the test below |
+| `bootroot::config::profile_domain` | a `Settings` plus a `DaemonProfileSettings` | a test asserting byte-identical output for the same four values |
+| `commands::verify::expected_dns_name` | a `ServiceEntry` | this table alone — it is private to the binary crate and no test here can call it |
+
+`RegistrarConfig::san_for` is the same composer with the domain taken from
+the loaded file, and is what a verb should call: it makes composing under a
+wire-supplied domain structurally impossible.
+
+## 7. The example
+
+`docs/reference/provisioning.toml.example` is the shipped example, and a test
+loads it through the production loader so the documented contract and the
+parser cannot drift apart. Its `fingerprint` is a real digest of its own
+body. The block below is that file verbatim, and a test asserts the two have
+not drifted.
+
+```toml
+fingerprint = "62299a244c4b3833c4cb3a36b4d3da763bd87f664c7d31845f48e53be4170f3e"
+# Rendered by the provisioning tool (bootler) onto BOTH the
+# bootroot/registrar host and the REView host at install, at the same
+# absolute path on each. bootroot only reads this file; nothing in this
+# repository writes it in production.
+#
+# The first line is the fingerprint and is NOT part of the digested body.
+schema_version = 1
+domain = "trusted.domain"
+
+# One table per component, keyed by package-id. A component absent from
+# this file is refused, never defaulted. `bootroot` has no entry: it is
+# the issuer and never enrolls through the registrar.
+[components.review]
+multiplicity = "one-per-deployment"   # | "one-per-host" | "many-per-host"
+cert_group = 3000                     # numeric gid; omit the key to mean
+                                      # "registrations carry no cert-group policy"
+reload = { kind = "docker-restart", target = "review" }
+
+[components.roxyd]
+multiplicity = "one-per-host"
+# cert_group omitted on purpose
+reload = { kind = "systemd", target = "roxyd.service" }
+
+[components.piglet]
+multiplicity = "many-per-host"
+cert_group = 3001
+reload = { kind = "docker-restart", target = "piglet" }
+```
+
+## 8. The test fixture
+
+`bootroot::registrar::fixture::RegistrarConfigFixture` is the canonical way to
+fabricate this file in a test. Nothing in this repository renders it in
+production, so without a shared fixture the verb, endpoint and acceptance-suite
+tests would each hand-write their own approximation of an artifact this
+repository never produces, and the three would drift from each other and from
+the real one.
+
+It renders the three components above under the domain `trusted.domain`, and a
+test overrides only what it varies: `with_domain`, `with_multiplicity`,
+`with_spec`, `with_component`, `without_component`, `with_schema_version`,
+`with_fingerprint` (for a deliberate digest mismatch) and `with_raw_component`
+(for a `multiplicity` or `reload.kind` the loader must refuse).
+`write_to(dir)` places it as `provisioning.toml` inside a caller-supplied
+directory and returns the path.
--- a/docs/reference/registrar-provisioning-config.md
+++ b/docs/reference/registrar-provisioning-config.md
@@ -190,9 +190,9 @@ derived.
 class. The one-per-deployment arm ignores it, but the SAN's third segment still
 needs it, so "not used by the id arm" is not "optional".
 
-### 6.1 The three SAN composition sites
+### 6.1 The SAN composition sites
 
-This repository composes `<instance>.<service>.<host>.<domain>` in **three**
+This repository composes `<instance>.<service>.<host>.<domain>` in **four**
 places, and they must agree byte for byte or the fleet gets certificates whose
 name the verification path rejects.
 
@@ -201,6 +201,11 @@ name the verification path rejects.
 | `bootroot::registrar::identity::compose_san` | the four parts | this table, and the test below |
 | `bootroot::config::profile_domain` | a `Settings` plus a `DaemonProfileSettings` | a test asserting byte-identical output for the same four values |
 | `commands::verify::expected_dns_name` | a `ServiceEntry` | this table alone — it is private to the binary crate and no test here can call it |
+| `commands::dns_alias::dns_alias_for_entry` | a `ServiceEntry` | this table alone, for the same reason |
+
+The last one composes the HTTP-01 DNS alias rather than the certificate SAN,
+but from the same four values and in the same shape, so a change to the shape
+that misses it desynchronises the alias from the name the certificate carries.
 
 `RegistrarConfig::san_for` is the same composer with the domain taken from
 the loaded file, and is what a verb should call: it makes composing under a
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,7 @@ pub mod kv_payload;
 pub mod locale;
 pub mod openbao;
 pub mod profile;
+pub mod registrar;
 pub mod tls;
 pub mod toml_util;
 pub mod trust_bootstrap;
--- a/src/registrar.rs
+++ b/src/registrar.rs
@@ -1,0 +1,40 @@
+//! The registrar's provisioning config and the registration identities
+//! derived from it.
+//!
+//! Two repositories must agree on the derivation rule exactly, or every
+//! enrollment in the fleet fails — and it fails after deployment, with
+//! both repositories' tests green. So the rule lives here as one small,
+//! densely tested unit, separate from the verbs that call it:
+//!
+//! - [`config`] reads the file the provisioning tool renders onto this
+//!   host, validates its digest and schema version, and yields the
+//!   deployment `domain` plus one entry per component.
+//! - [`identity`] derives the `registration_id`, composes the SAN,
+//!   checks the identity shape and validates a requested spec against
+//!   the component's rendered safe-set — as separately callable steps,
+//!   because the verb layer interleaves its own checks between them.
+//! - [`error`] declares every refusal the two produce.
+//! - [`fixture`] builds a rendered config for tests.
+//!
+//! Nothing here reads or requires registration state. That is what makes
+//! the durable-binding collision check the caller's step rather than a
+//! missing one here, and it is why no variant in [`error`] refers to a
+//! stored or previously applied spec.
+
+pub mod config;
+pub mod error;
+pub mod fixture;
+pub mod identity;
+
+#[cfg(test)]
+mod tests;
+
+pub use config::{
+    CONFIG_FILE_NAME, ComponentEntry, DEFAULT_CONFIG_PATH, Multiplicity, RegistrarConfig,
+    RegistrationSpec, ReloadKind, ReloadSpec, SUPPORTED_SCHEMA_VERSION,
+};
+pub use error::{RegistrarError, SpecIdentityField};
+pub use identity::{
+    DerivedIdentity, RequestedSpec, check_instance_shape, check_spec_identity, compose_san,
+    derive_registration_id, validate_request_labels,
+};
--- a/src/registrar/config.rs
+++ b/src/registrar/config.rs
@@ -539,7 +539,20 @@ fn resolve_component(name: &str, raw: RawComponent) -> Result<ComponentEntry, Re
             value: raw.reload.kind.clone(),
         })?;
     let target = raw.reload.target;
-    if kind.takes_target() != target.as_ref().is_some_and(|value| !value.is_empty()) {
+    // `none` must carry no `target` key at all, not merely an empty one:
+    // an empty string is still `Some`, and storing it would break the
+    // "`target` is `None` exactly when `kind` is `None`" invariant that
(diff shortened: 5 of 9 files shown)


AcoPiper:
**[Reviewer Round 1]**

No findings. The implementation matches the issue’s cross-repository contract: it verifies the exact first-line fingerprint before parsing, rejects unsupported schema versions and unusable component/reload shapes at load time, and keeps the wire `service_name` as the sole selector and identity/SAN segment.

I also checked the failure ordering that matters to the upcoming verb layer: label and spec-identity refusals occur before derivation, the component/instance checks are separately callable, and safe-set validation remains after derivation so callers can interleave the durable-binding collision check. The tests cover those boundaries as well as the SAN-only `001` default, malformed/truncated config cases, safe-set equality, and the documented example.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Read the registrar config and derive identities
```

**Body**

```text
The registrar verbs must not take a composed identity name or the
deployment domain from the wire: a caller that could send either
would bind one host's name to another host's identity, or mint
certificates under a name suffix of its choosing. So the verbs read
everything deployment-wide from a file the provisioning tool renders
onto the host, and this lands that reader plus the derivation rule
ahead of the verbs that call it.

The rule is a cross-repository contract — bootler renders the file
and both sides must agree exactly, or every enrollment in the fleet
fails, and it fails at deploy time with both repositories' tests
green. That is why it lands as one small, densely tested unit
rather than inside the verb layer.

Two integrity gates run before anything downstream reads the file.
The declared fingerprint is checked against the digest of the body
as written, so a truncated render fails once and loudly instead of
refusing each missing component individually; and a schema_version
this build does not implement is refused rather than best-effort
parsed. A reload target contradicting its kind, and a component key
no registration could ever select, are refused at load for the same
reason.

The steps are separately callable rather than one resolve() call,
because the verb layer must interleave its per-identity mutex and
its durable-binding collision check between the derivation and the
spec comparison.

The library is pub rather than pub(crate), and the fixture builder
is compiled unconditionally: the siblings that must share them are
the binary crate and tests/, neither of which can see a pub(crate)
item or enable a test-support feature here.

Closes #757.
```

PR #873, opened 2026-08-18
Add the registrar client and endpoint identities (#760)

### LINKED ISSUE #760: Issue a distinguishable registrar client identity and pin the endpoint's server identity
# Issue a distinguishable registrar client identity and pin the endpoint's server identity

## Context

bootroot issues exactly one certificate shape today. `build_csr_params` (`src/acme/flow.rs:24`) builds a CSR whose `subject_alt_names` is a single `rcgen::SanType::DnsName` equal to `<instance_id>.<service_name>.<hostname>.<domain>` (`src/config.rs::profile_domain:103`), repeats that same string as the CN, and sets **no** extended key usages — a server-shaped leaf. A forthcoming host-local endpoint authenticates its caller by certificate, and that caller's identity must be distinguishable from an ordinary service leaf, so it cannot come from an ordinary `service add`.

The other direction needs an answer too. A TLS handshake over a unix-domain socket has **no hostname** to match a server certificate against. Left unspecified, that gets implemented as "accept any certificate" — and the operation the endpoint serves returns a CA anchor that becomes a newly onboarded host's trust root, so anything that can impersonate the endpoint injects a CA of its choosing.

Three facts about this tree shape the contract below, and are why the contract is written out here rather than left to the implementer to invent:

- bootroot already has exactly one pinning idiom, and it pins **trust anchors**, not leaves. `trust.trusted_ca_sha256` (`src/config.rs:177`) is a list of lowercase-hex SHA-256 digests over a certificate's **DER** (`src/tls.rs::sha256_hex:406`, `ca_bundle_fingerprints:398`), validated as exactly 64 ASCII hex characters and compared case-insensitively (`src/kv_payload.rs::parse_fingerprints:186` and `:72`). `PinnedCertVerifier` (`src/tls.rs:307`) already implements pin-based server verification, and its direct-pin arm **rejects any presented certificate that is not a CA** (`validate_direct_pin_certificate:428`).
- Every issuance generates a **fresh key pair** (`rcgen::KeyPair::generate()`, `src/acme/flow.rs:362`), so neither a leaf-certificate digest nor an SPKI digest survives a renewal.
- The `--expect-sha` / `--expect-trust` idiom the RFC points at is **roxyd's CLI surface in another repository; it exists nowhere in this tree**. Do not go looking for it — the in-repo analog is the `trusted_ca_sha256` list above, and this issue follows that.

This issue delivers both certificate-side primitives ahead of the endpoint that consumes them, so neither side has to invent the other's artifact. It adds no listener, no wire protocol and no authorization logic.

### 1. The registrar client identity (specified here, not left open)

The client identity reuses today's DNS SAN composition and distinguishes itself by a **reserved service label**, so no new SAN type, no new name grammar and no change to `profile_domain` / `expected_dns_name` composition are needed:

- **Client identity name** (the registrar's leaf): `<instance>.bootroot-registrar.<host>.<domain>` — for the v1 single registrar per bootroot host, `instance` is `001` and `host` is that host's single DNS label. Example, with `domain = example.internal`: `001.bootroot-registrar.h1.example.internal`.
- **Endpoint server identity name** (the daemon's leaf, needed here because the pinning helper must check it): `<instance>.bootroot-registrar-endpoint.<host>.<domain>` — same rules. Example: `001.bootroot-registrar-endpoint.h1.example.internal`.
- **`<domain>` is a suffix, not a label.** `domain` is the configured `network.domain`, and `validate_domain_name` (`src/input_validation.rs:35`) accepts **any number** of labels — `example.internal` is two, `corp.example.internal` is three, a bare `internal` is one. So these names are **three fixed leading labels followed by the configured domain suffix**, and their total label count is `3 + <label count of the configured domain>`. Never hard-code a total label count, and never treat the domain as a single trailing label; both are wrong for every real deployment (`src/config.rs:605` asserts `001.edge-proxy.edge-node-01.example.internal` for a two-label domain).
- Both names carry **exactly one** DNS SAN and no SAN of any other type, and repeat the SAN string as the CN, exactly like today's leaves.
- Define both reserved labels as `const` items at the top of the file that owns them, not as inline literals.
- **Put the two labels, the `bootroot-` prefix constant and the recognition predicate in the library crate as `pub` items exported from `src/lib.rs`, because their consumers sit on both sides of the crate boundary.** `service add` is in the **binary** crate — `src/main.rs` declares `mod commands;`, so `src/commands/service.rs` reaches the library only through `bootroot::…` paths — while the endpoint's client verifier and the internal privileged credential's leaf-name choice are library-side. No narrower visibility serves both: a `pub(crate)` item in the library is invisible to the guard, and a constant defined next to the guard in `src/commands/` is invisible to the verifier. Getting this wrong is not a style question but the failure this bullet exists to prevent — the sibling that cannot reach the constant writes its own copy, and two lists of reserved names drift apart. This is the project's stated exception to preferring `pub(crate)`: a library public API with consumers outside the crate, as the `registration_id` validator also takes.

Concretely, this is how the registrar client identity differs from the registrar host's own ordinary service certificate: roxyd's service leaf on that host is `001.roxyd.<host>.<domain>` (its plain component keyword in the second label — the `registration_id` split keys namespaces elsewhere and does not change the SAN label). The registrar client identity differs in the **second label only**, and that label is one an ordinary `service add` cannot produce, because:

- **`service add` gains a reserved-prefix guard**: a `service_name` whose ASCII-lowercased form starts with `bootroot-` is refused with a typed validation error, alongside the existing DNS-label check (`src/input_validation.rs::validate_dns_label:21`, which accepts mixed case, hence the case-insensitive comparison). `bootroot-` is already bootroot's own namespace prefix (`bootroot-service-`, `bootroot-runtime-`), no component keyword a consumer registers uses it, and the product is pre-release, so no migration arm is needed. One *shipped fixture* does sit under the prefix and has to move with this change: the agent's own self-test registration carries `service_name = "bootroot-agent"` in `.github/workflows/ci.yml`, `scripts/preflight/ci/test-core.sh`, `scripts/preflight/extra/agent-scenarios.sh` and `agent.toml.compose`, and the SAN it produces is a responder alias in `docker-compose.test.yml`. Rename that SAN label to `agent-selftest` and leave its `registration_id`, `hostname` and certificate paths alone — only the second label is reserved. Without this guard the recognition predicate below is not sound — anyone able to call `service add` could mint the registrar's name.
- **This is the effort's single `service add` reserved-name rule, and it reserves the whole `bootroot-` namespace rather than the two labels above.** Export the prefix as a `const` and the check as one shared predicate, so every bootroot-internal identity this effort or a later one introduces — including the internal privileged credential's own leaf — is covered by this guard without adding a second one. **No sibling issue adds another `service add` guard**; a sibling that needs a name protected picks it under this prefix and asserts the existing refusal. A second guard would mean two lists of reserved names drifting apart, and the one that is not consulted is the hole.

**Extended key usage — what is asserted, and where.** CSR construction gains an additive option to request `clientAuth` (`rcgen::ExtendedKeyUsagePurpose::ClientAuth`), and the client path uses it. Keep the two levels apart, because they are testable in different places and only one of them is under this repo's control:

- **On the CSR (`rcgen::CertificateParams`) — asserted.** This is what `build_csr_params` produces and what this issue actually changes, so it is unit-testable with no CA in the loop: the client path's params carry `ClientAuth` in `extended_key_usages` and the reserved-label SAN; the existing service path's params carry the unchanged SAN and an **empty** `extended_key_usages`. These are hard acceptance criteria.
- **On the issued certificate — SAN asserted, EKU only observed and documented.** The issued leaf's EKU set is chosen by step-ca's certificate template, not by this repo: there is no `extKeyUsage` string and no x509 template anywhere in the tree, so issuance runs on step-ca's defaults and an ordinary service leaf may well come back carrying `clientAuth` too. Assert the **SAN** on an issued certificate; for EKU, **observe** what the issuance path actually returns for both shapes and record that observation in the docs deliverable. Do not write a test that asserts a particular issued EKU set — it would be asserting an upstream default this repo does not pin, and it would fail the day that default moves.

**The predicate must not rely on EKU at all**, for exactly the reason above: an attribute every leaf may carry cannot discriminate. Recognition is by name.

### 2. The recognition predicate (specified here)

Expose one reusable predicate that a server-side verifier calls, so the endpoint consumes it rather than re-deriving the rule. Given a presented end-entity certificate and the locally configured `domain`, it accepts **only** when all of the following hold, and otherwise returns a typed rejection:

1. the certificate carries **exactly one** SAN, of type DNS (parse with `x509_parser`, as `src/commands/verify.rs:252` already does);
2. the DNS name **ends with the configured `domain`**, matched on a label boundary — i.e. the name equals `<something>.<domain>` — never a bare string suffix, so `evil-example.internal` must not match the domain `example.internal`;
3. stripping that domain suffix leaves **exactly three** labels (equivalently: the name's total label count is `3 + <label count of domain>`, derived from the configured value, never hard-coded);
4. of those three, the **second** equals `bootroot-registrar`;
5. the **first** is a valid instance label and the **third** a valid DNS label (`validate_dns_label`).

All label and suffix comparisons are ASCII-case-insensitive, since DNS labels are case-insensitive and `validate_dns_label` admits mixed case.

The CN is **not** consulted — recognition is SAN-only, even though today's leaves mirror the SAN into the CN. On success the predicate returns the parsed parts (`instance`, `host`, `domain`) so the endpoint can scope authorization by host; **which** recognized identity is authorized at **which** operation is the endpoint's decision, not this issue's. The predicate performs no chain building and no time validation — it is a name rule applied to an already-verified peer certificate.

### 3. The endpoint server pin file (specified here)

**What is pinned: trust anchors, plus an exact expected SAN on the presented leaf.** Not the endpoint leaf's DER and not its SPKI — both change at every daemon renewal (fresh key pair per issuance, above) while the writer of this file is the provisioning tool, runs once at install, and lives in another repository, so a leaf-shaped pin would go stale with nothing to refresh it. Anchor pinning also matches the one pinning idiom this tree already has and lets the existing `PinnedCertVerifier` semantics be reused rather than a second scheme grown beside them. Pinning the anchor alone would admit any leaf that CA ever issued, so the anchor pin is paired with a mandatory check that the presented end-entity certificate's SAN is exactly the endpoint server identity name above — which the reserved-prefix guard makes unmintable through `service add`.

**File.** Basename `registrar-endpoint-anchors.sha256`, defined as a `const` here.

**What this issue fixes about the location, and what it deliberately does not.** bootroot has **no fixed certificate directory**: every certificate and key path is configured per profile (`Paths { cert, key }`, `src/config.rs:112`), and this issue neither provisions nor places the registrar client certificate. There is therefore no `/etc/bootroot/certs`-style convention to name here, and inventing an absolute directory would collide with whatever the issue that provisions that certificate settles on. This issue owns the basename, the format, and a **derivation rule** — not an absolute path:

- the pinning helper takes the pin file path as an **explicit argument**. It performs no path discovery, reads no configuration and has no built-in default location;
- beside it, a small pure function derives the conventional path from a **given** registrar client certificate path: that path's parent directory joined with the basename `const`. It also takes its input as an argument and reads no configuration;
- so nothing implemented here depends on knowing where the registrar client certificate finally lives. The absolute placement is settled by the issue that provisions that certificate; the documentation deliverable below records the basename and this beside-the-client-certificate rule, and says plainly that the absolute directory follows the configured client-certificate path rather than being fixed here.

**Format.** UTF-8 text, LF-separated lines:

- each significant line is exactly 64 ASCII hex characters — the SHA-256 of one trust anchor certificate's **DER**, i.e. `tls::sha256_hex(cert.as_ref())`, the same value `ca_bundle_fingerprints` produces;
- leading and trailing ASCII whitespace on a line is trimmed before interpretation;
- blank lines, and lines whose first non-whitespace character is `#`, are ignored;
- uppercase hex is accepted and normalized with `to_ascii_lowercase` on read, matching `kv_payload.rs:72`; entries are written lowercase;
- order is irrelevant and duplicates collapse;
- **multiple entries are required to be supported** — CA rotation (`RotationMode::IntermediateOnly` / `Full`, `src/commands/trust.rs:21`) means old and new anchors must be pinnable simultaneously;
- at least one valid entry must be present, and any non-ignored line that is not 64 hex characters rejects the **whole file** — no partial acceptance.

Worked example:

```
# bootrootd registrar endpoint — pinned trust anchors (SHA-256 of certificate DER)
# written by the provisioning tool at install; one anchor per line
3b1f0c9a5d2e47b8c6a1f0e93d7b425c8e6a09f31d4b7c25e8a06f93b1d4c72e
# second entry present only during a CA rotation
9d4c72e1b0a538f6c2e94b7d051a63f8c4b29e0d75a1f386c0b9e24d7f513a6b
```

### 4. The pinning helper (specified here)

A reusable helper that a client builds its TLS configuration from, given the pin file path and the expected endpoint SAN. It accepts a handshake only when the presented end-entity certificate chains to — or, in the direct-pin arm, **is** — an anchor whose DER SHA-256 is in the pin set, that anchor is CA-capable and time-valid, **and** the end-entity's single DNS SAN equals the expected endpoint identity name. Each of missing file, unreadable file, no valid entry, no anchor match, and SAN mismatch is a **distinguishable typed refusal**. There is no permissive arm: nothing falls back to trusting whatever was presented, and no code path disables verification because `AF_UNIX` supplies no hostname.

**The direct-pin arm, spelled out — it is inherited, not a deployment shape.** `PinnedCertVerifier` accepts two ways (`verify_server_cert`, `src/tls.rs:342`): the presented leaf chains to a pinned anchor, or the presented certificate is *itself* a pinned entry, in which case `validate_direct_pin_certificate` (`src/tls.rs:428`) additionally demands it be **CA-capable and time-valid**. The endpoint's server leaf is CA-issued, so a correctly provisioned deployment always takes the **chained** arm; the direct-pin arm exists here only because the helper reuses `PinnedCertVerifier` unmodified, which the constraints below require in preference to growing a second verification path. Keep it, and pin its consequence down explicitly rather than leaving it to be discovered:

- **the exact-SAN check is unconditional** and is applied to the presented end-entity certificate under either arm;
- therefore a directly pinned certificate must be CA-capable, time-valid **and** carry the endpoint identity name as its single DNS SAN. A pinned, CA-capable, time-valid certificate carrying any other SAN is refused with the SAN-mismatch refusal — it is not accepted just because its digest is in the pin set;
- this combination is an odd one and is not expected in any real deployment. It is specified and tested only so the reuse has a defined behaviour, not because anything is meant to be provisioned this way.

### 5. Documentation

Document the client identity name (stating that `<domain>` is the configured multi-label suffix, not a single label), the reserved-prefix guard, the recognition rule, the pin file's basename, its beside-the-client-certificate derivation rule (stating that bootroot configures certificate paths per profile, so this issue fixes no absolute directory and the pin file's directory follows the configured registrar client certificate path), its format and worked example, the writer-side contract (the provisioning tool writes the file at install; it is not written by this repo), and the observed EKU result from scope item 1, under `docs/`.

## Acceptance criteria

- [x] **CSR params, client path:** requesting a client identity yields `rcgen::CertificateParams` whose single DNS SAN is `<instance>.bootroot-registrar.<host>.<domain>` (the domain appearing as a suffix of whatever label count is configured) and whose `extended_key_usages` contains `ClientAuth`; a unit test asserts both directly on the params, with no CA involved.
- [x] **CSR params, existing service path unchanged:** an ordinary `service add` still yields exactly today's single DNS SAN and an **empty** `extended_key_usages`; a unit test asserts the existing params are unchanged for every current caller.
- [x] **Issued certificate:** a client certificate can be issued and its parsed single DNS SAN equals the expected reserved-label name. The issued **EKU set is not asserted** — it is observed for both the client and the ordinary service shape and the observation is written into the docs deliverable, because step-ca's template, not this repo, decides it.
- [x] `service add` refuses a `service_name` whose ASCII-lowercased form begins with `bootroot-`, with a typed error; ordinary names such as `roxyd`, `piglet` and `edge-proxy` are unaffected.
- [x] The recognition predicate accepts the registrar name under configured domains of **differing label counts** (at least a one-label and a two-label domain, e.g. `001.bootroot-registrar.h1.internal` and `001.bootroot-registrar.h1.example.internal`), deriving the expected label count from the configured domain rather than a constant.
- [x] The predicate rejects, each distinguishably: an ordinary service leaf (`001.roxyd.h1.<domain>`) including one issued for the registrar host's own service; a name with the right reserved label but the wrong number of labels before the domain suffix; a name whose suffix is not the configured domain, including the non-boundary near-miss (domain `example.internal` must not accept `001.bootroot-registrar.h1.evil-example.internal`); a certificate carrying more than one SAN; and a certificate whose only SAN is not a DNS name. The `bootroot-registrar` match is ASCII-case-insensitive.
- [x] The predicate returns the parsed `instance`, `host` and `domain` on success, and does not itself decide authorization.
- [x] The pin file parser accepts the documented format — 64-hex lines, `#` comments, blank lines, surrounding whitespace, uppercase hex, duplicates, multiple anchors — and rejects a file containing any malformed non-ignored line, and a file with no valid entry.
- [x] The pinning helper accepts a server whose chain reaches a pinned anchor **and** whose SAN equals the expected endpoint name, and refuses — with distinguishable errors, none of them a fallback to trusting the certificate — on: anchor mismatch, SAN mismatch, expired or not-yet-valid anchor, non-CA anchor in the direct-pin arm, unreadable pin file, and missing pin file.
- [x] The pinning helper takes the pin file path as an explicit argument, with no built-in default and no configuration lookup; a separate pure function derives `<parent directory of a given registrar client certificate path>/registrar-endpoint-anchors.sha256`, and a unit test asserts that derivation for two unrelated input paths (no filesystem access needed).
- [x] In the direct-pin arm the exact-SAN check still applies: a pinned, CA-capable, time-valid certificate whose single DNS SAN is not the endpoint identity name is refused with the SAN-mismatch refusal, while one that does carry that SAN is accepted.
- [x] The pin file's basename and format, the beside-the-client-certificate derivation rule, the client identity name, the endpoint identity name and the writer-side contract are documented with the worked example. The doc states that the provisioning tool writes the file at install, and that the file's absolute directory follows the configured registrar client certificate path and is not fixed by this issue.

## Constraints

- Do not widen the ordinary service leaf. The client identity is an additive issuance shape, not a change to the existing one; existing certificates and their verification must be unaffected.
- Do not write a permissive custom certificate verifier. "No hostname available over `AF_UNIX`" is answered by anchor pinning plus the exact-SAN check, not by disabling verification.
- This issue defines and **consumes** the pin file; it does not implement the writer, which belongs to the provisioning tool in another repository. The format above is the shared contract, so it must be documented exactly as implemented.
- Reuse `src/tls.rs` — `sha256_hex`, `ca_bundle_fingerprints`, and the `PinnedCertVerifier` machinery — rather than adding a second fingerprint format, a second verification path, or a second TLS stack. That reuse is why the direct-pin arm survives; do not special-case it away, and do not weaken the exact-SAN check to accommodate it.
- Do not invent an absolute on-disk directory for the pin file, and do not give the helper a default path or a configuration lookup. What is fixed here is the basename, the format and the derivation from a supplied client-certificate path; certificate paths in this tree are per-profile configuration, not a baked-in layout.
- Both reserved labels must remain valid single DNS labels under `validate_dns_label`, and all label comparisons are ASCII-case-insensitive.
- No `unwrap()` in production code; no `[]` indexing into slices or vectors; `thiserror` for new error types; `pub(crate)` visibility unless a compiler error forces wider; constants at the top of the file.

## Out of scope

- Any listener, socket, wire protocol, or endpoint authorization scoping — which identities are accepted at which operations belongs to the endpoint issues.
- Renewing either certificate, issuing them from the daemon, and provisioning any OpenBao credential.
- Writing the pin file, and any change to ACME/EAB issuance for ordinary services.
- Fixing the absolute on-disk location of the registrar client certificate, and hence of the pin file beside it — that follows the certificate's provisioning.

## Test plan

- [x] CSR-params tests (no CA in the loop): the client path's params carry the reserved-label SAN and `ClientAuth`; the existing service path's params carry today's SAN and an empty `extended_key_usages`.
- [x] Issued-certificate test: the issued client leaf's parsed single DNS SAN is the reserved-label name. No assertion on the issued EKU set; the observed sets for both shapes go into the docs.
- [x] Fixture-drift test: no workflow, preflight script or shipped example config registers a `service_name` under the reserved prefix. A reserved fixture is otherwise caught only by a Docker E2E job, as a `service add` that suddenly exits non-zero, so this runs in `cargo test`.
- [x] Guard test: `service add` refuses `bootroot-registrar`, `BOOTROOT-Registrar` and `bootroot-anything`, and accepts `roxyd` / `piglet` / `edge-proxy`. The prefix constant and the predicate are reachable from sibling code, so no second guard has to be written — specifically, both are `pub` items named through `bootroot::…`, the binary-crate guard in `src/commands/service.rs` calls them that way, and a reviewer can confirm neither is `pub(crate)` nor defined inside the binary crate where a library-side consumer could not reach it.
- [x] Predicate tests, run against configured domains of at least two different label counts: accepted registrar identity; rejected ordinary service leaf, registrar-host's own service leaf, wrong label count before the domain suffix, non-boundary suffix near-miss (`evil-example.internal` vs domain `example.internal`), multi-SAN certificate, non-DNS SAN; mixed-case reserved label accepted; parsed parts returned.
- [x] Pin-file parser tests: valid single and multi-anchor files; comments, blank lines, surrounding whitespace and uppercase hex accepted; a 63-hex line, a non-hex line and an entries-free file each rejected.
- [x] Pin-path derivation test: the conventional path is the supplied client certificate path's parent directory joined with `registrar-endpoint-anchors.sha256`, asserted for two unrelated input paths; the helper itself is only ever handed an explicit path.
- [x] Pinning tests, using locally generated certificates: anchor match plus SAN match accepted; anchor mismatch, SAN mismatch, expired anchor, not-yet-valid anchor, non-CA direct pin, unreadable file and missing file each refused with a distinguishable error.
- [x] Direct-pin edge-case tests: a pinned, CA-capable, time-valid certificate whose single DNS SAN is the endpoint identity name is accepted; the same certificate shape carrying a different SAN is refused with the SAN-mismatch refusal.
- [x] Tests use `tempfile::tempdir()` for certificate and pin-file material; no fixed paths.

## Dependencies

None — independent of the namespace-key and verb work, and deliberately ahead of the endpoint that consumes both primitives. Part of the registrar-surface umbrella.

## Pointers

- `src/acme/flow.rs::build_csr_params:24` (single DNS SAN, CN mirror, no EKU requested) and `:362` (fresh `rcgen::KeyPair` per issuance)
- `src/config.rs::profile_domain:103` and `src/commands/verify.rs::expected_dns_name:376` (the three-leading-labels-plus-domain-suffix composition), `src/config.rs:605` (the test asserting `001.edge-proxy.edge-node-01.example.internal`, i.e. a multi-label domain), `src/input_validation.rs::validate_domain_name:35` (any label count) and `src/commands/verify.rs:252` (reading DNS SANs with `x509_parser`)
- `src/tls.rs`: `sha256_hex:406`, `ca_bundle_fingerprints:398`, `PinnedCertVerifier:307`, `check_direct_pin:326`, `validate_direct_pin_certificate:428` (the existing anchor-pinning semantics to reuse; note every rustls config in the tree is `with_no_client_auth()` today)
- `src/kv_payload.rs::parse_fingerprints:186` (the 64-hex validation and case-insensitive comparison this file format mirrors)
- `src/input_validation.rs::validate_dns_label:21` (accepts mixed case) and `src/commands/service.rs` (where the reserved-prefix guard belongs)
- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §4 and §5.3

AcoPiper:
Closes #760.

Delivers both certificate-side primitives the host-local registrar endpoint needs, ahead of the endpoint itself. No listener, wire protocol or authorization logic comes with them.

## The two identities and the guard that protects them

`src/registrar.rs` (new, `pub mod registrar` from `src/lib.rs`) owns the reserved labels, the prefix constant and the recognition predicate. They are `pub` because their consumers sit on both sides of the crate boundary: `service add` is in the binary crate and reaches them through `bootroot::registrar::…`, while the endpoint's verifier is library-side.

- Client identity: `<instance>.bootroot-registrar.<host>.<domain>`. Endpoint server identity: `<instance>.bootroot-registrar-endpoint.<host>.<domain>`. `<domain>` is the configured suffix of whatever label count it carries — nothing hard-codes a total label count, and the tests run under one-, two- and three-label domains.
- `service add` refuses a `--service-name` whose lowercased form starts with `bootroot-`, with its own typed error rather than the DNS-label one. The whole prefix is reserved rather than the two labels in use, so a later bootroot-internal identity is covered by this one guard. The check lands in `src/commands/service/resolve.rs::validate_service_name`, the single choke point every `service add` path (flag, prompt, and the final `validate_service_add`) already goes through for the DNS-label rule — the issue points at `src/commands/service.rs`, and this is that module's validation file rather than a second guard beside it.
- `bootroot-remote bootstrap --service-name` calls the same shared predicate. It is the second way an operator-supplied name becomes the SAN's second label: the flag lands in `[[profiles]].service_name`, and a run without `--artifact` takes it straight from the command line rather than from an artifact `service add` already vetted. One predicate rather than two lists that drift.
- `recognize_registrar_client` accepts only a certificate with exactly one DNS SAN, ending in the configured domain **on a label boundary**, with exactly three labels before it, the second being `bootroot-registrar`, the first numeric and the third a valid DNS label — all ASCII-case-insensitive, CN never consulted. It returns the parsed `instance`, `host` and `domain` and decides no authorization. Every failure is its own variant.

`build_registrar_client_csr_params` is the additive CSR shape: the same single DNS SAN and CN mirror, plus `ClientAuth`. The service path is refactored through a shared helper and is byte-for-byte what it was, with an empty `extended_key_usages` — asserted for three different profile shapes.

The one fixture that sat under the reserved prefix moved with the guard: the agent's self-test registration carried `service_name = "bootroot-agent"` in CI, the preflight scripts and the shipped compose profile, and its SAN label is now `agent-selftest`. Its `registration_id`, `hostname` and certificate paths are unchanged, since only the second label is reserved. A `cargo test` regression scan keeps the next such fixture from being caught only by a Docker E2E job twenty minutes in.

## The pin file and the pinning helper

`src/registrar/endpoint_pin.rs` owns the basename `registrar-endpoint-anchors.sha256`, the format (64-hex lines, `#` comments, blank lines, surrounding whitespace, uppercase hex, duplicates collapsing, multiple anchors required, whole-file rejection on any malformed line), the parser, and the verifier.

The verifier wraps `tls::PinnedCertVerifier` with its behaviour unmodified — the only change to `src/tls.rs` is widening `build_pinned_verifier` and `install_crypto_provider` to `pub(crate)` — and adds the unconditional exact-SAN check on the presented leaf, applied **first** so a SAN mismatch is reported as one instead of being masked by the chain failure it would also cause. It takes the pin-file path as an explicit argument and reads no configuration; `anchor_pin_path_for_client_certificate` derives the conventional path from a supplied client-certificate path with no filesystem access.

One consequence worth calling out, because it is a design decision the issue's inputs force rather than an obvious reading: **the trust anchors are taken from the certificates the peer presents**, keeping only those whose DER SHA-256 is pinned. The helper's only inputs are the pin-file path and the expected SAN, and a pin file carries digests and no certificate material — and the endpoint's caller is precisely the party that does *not* yet have the CA bundle, since fetching that anchor is what the endpoint is for. So a pin has to name a certificate the endpoint actually sends: in a bootroot deployment, the CA bundle fingerprints the server presents alongside its leaf, which is exactly what `tls::ca_bundle_fingerprints` returns. The docs state this as the writer-side contract.

The inherited direct-pin arm is kept and pinned down: a pinned, CA-capable, time-valid certificate is accepted only if its single DNS SAN is the endpoint identity name, and refused with the SAN-mismatch refusal otherwise.

## The observed EKU

Recorded in `docs/reference/registrar-client-identity.md` — which also carries the two identity names, the guard, the recognition rule and the pin-file contract with its worked example — from three runs against `smallstep/step-ca:0.30.2` (the image `docker-compose.yml` pins) with its default configuration:

| Shape | How | Issued EKU |
| --- | --- | --- |
| ordinary service | bootroot's own ACME path, reading `edge-proxy.crt` / `web-app.crt` out of a `run-local-lifecycle.sh` workspace | `TLS Web Server Authentication, TLS Web Client Authentication` |
| ordinary service | `step ca sign` of this repo's CSR, ACME and JWK provisioners | `Server Authentication, Client Authentication` |
| registrar client | `step ca sign` of this repo's CSR | `Server Authentication, Client Authentication` |

step-ca's default leaf template ignores the EKU the CSR requests — the service CSR requests none and gets both; the client CSR requests `clientAuth` and gets both. So an ordinary service leaf is `clientAuth`-capable too, which is why recognition is a name rule that consults no EKU, and why nothing here asserts an issued EKU set.

`docs/en/cli.md` and `docs/ko/cli.md` note the refusal under both `--service-name` flags, and `CHANGELOG.md` records it under `### Changed`.

## What the guard does and does not prove

The guard covers the two CLI routes that let an operator name a service — `service add` and `bootroot-remote bootstrap` — which is the issue's scope, and the issue is explicit that no sibling adds a third. Worth stating plainly all the same: a host that already holds a valid registration can edit its own `agent.toml`'s `service_name` and order any name step-ca's ACME provisioner will validate, so the CLI guards are not by themselves a proof that no `bootroot-registrar` leaf can exist. Closing that would mean constraining ACME/EAB issuance, which this issue puts out of scope.

## Dependency note

`rcgen` gains the `x509-parser` feature under `[dev-dependencies]`, so the issuance tests can parse a CSR back and sign it the way a CA does. That feature adds nothing to `Cargo.lock`: it turns on `x509-parser` 0.18 with `verify`, the exact dependency and feature this crate already declares.

The one `Cargo.lock` change on this branch is unrelated to the work above. RUSTSEC-2026-0258 was published on 2026-08-17 against `h2` 0.4.15 — a peer can make it buffer empty HTTP/2 DATA frames without bound — and `h2` is pinned transitively here through `hyper`, `reqwest` and `poem`, so `cargo audit` in the `Quality Check` job went red on every commit of this branch. Bumped to 0.4.16, which is a patch release with no API to adapt to, and recorded under `### Security` in `[Unreleased]` following the `rustls-webpki` entry in 0.3.0. The bump is hand-applied to the two `h2` lines rather than taken from `cargo update`, which also rewrote a dozen unrelated `windows-sys` dependency edges; `cargo metadata --locked` accepts the result.

## Test plan

- [x] CSR params, client path: the params carry the reserved-label SAN and `ClientAuth`, asserted directly on the params with no CA involved.
- [x] CSR params, service path unchanged: today's single DNS SAN and an **empty** `extended_key_usages`, asserted for every current caller shape.
- [x] Issued certificate: the issued client leaf's parsed single DNS SAN is the reserved-label name; no assertion on the issued EKU set.
- [x] Guard: `service add` refuses `bootroot-registrar`, `BOOTROOT-Registrar` and `bootroot-anything`, and accepts `roxyd` / `piglet` / `edge-proxy`; `bootroot-remote bootstrap --service-name` refuses the same through the same predicate.
- [x] Reachability: the prefix constant and the predicate are `pub` items in the library named through `bootroot::…` from the binary crate, so no second guard has to be written.
- [x] Fixture drift: a `cargo test` scan finds no workflow, preflight script or shipped example config registering a `service_name` under the reserved prefix, and reads the flag across a line continuation, an `=` join and a TOML assignment.
- [x] Predicate, under domains of one, two and three labels: accepted registrar identity; rejected ordinary service leaf, the registrar host's own service leaf, wrong label count before the domain suffix, the non-boundary near-miss (`evil-example.internal` against domain `example.internal`), a multi-SAN certificate and a non-DNS SAN — each distinguishably. Mixed-case reserved label accepted; parsed parts returned.
- [x] Pin-file parser: single and multi-anchor files, comments, blank lines, surrounding whitespace, uppercase hex and duplicates accepted; a 63-hex line, a non-hex line and an entries-free file each reject the whole file.
- [x] Pin-path derivation: the parent directory of a supplied client-certificate path joined with `registrar-endpoint-anchors.sha256`, asserted for two unrelated inputs with no filesystem access.
- [x] Pinning: anchor match plus SAN match accepted, including through an in-memory handshake; anchor mismatch, SAN mismatch, expired anchor, not-yet-valid anchor, non-CA direct pin, unreadable file and missing file each refused with a distinguishable error and no permissive fallback.
- [x] Direct-pin edge cases: a pinned, CA-capable, time-valid certificate carrying the endpoint identity name is accepted; the same shape carrying another SAN is refused with the SAN-mismatch refusal.
- [x] Tests use `tempfile::tempdir()` for certificate and pin-file material; no fixed paths.
- [x] `cargo test --all-features`: 1918 passing, 0 failing.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo fmt --check -- --config group_imports=StdExternalCrate` clean.
- [x] `scripts/check-docs.sh`: `mkdocs build --strict` green, theme assets verified.
- [x] `scripts/preflight/ci/e2e-matrix.sh` and `e2e-extended.sh` — **not runnable on the author's host** (no passwordless `sudo`, so the matrix aborts at the OpenBao TLS re-own step; the extended suite needs a DNS-label hostname this machine does not have), so both were confirmed on CI instead. All ten Docker E2E matrix scenarios pass on the head commit, and a dispatched `E2E Extended` run over this branch ([32112778197](https://github.com/aicers/bootroot/actions/runs/32112778197)) reports `overall_status: pass` with all seven cases green — `scale-contention`, `failure-recovery`, `runner-timer`, `runner-cron`, `ca-key-recovery`, `reinit-recovery`, `infra-lifecycle` — which carries the renamed `agent-selftest` fixture through the Docker lifecycle.
- [x] `run-local-lifecycle.sh` `verify-after-ca-key` phase: it fails on the author's machine with `Untrusted CA fingerprint` for `web-app`, reproducible on `origin/main` and unrelated to this branch. Confirmed on CI, where `Docker E2E (local-hosts)` and `(local-no-hosts)` run the script end to end on the head commit and both pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -343,7 +343,7 @@ jobs:
 
           cargo run --bin bootroot -- service add \
             --registration-id bootroot-agent \
-            --service-name bootroot-agent \
+            --service-name agent-selftest \
             --hostname bootroot-agent \
             --domain trusted.domain \
             --agent-config "$(pwd)/tmp/agent-bootroot-agent.toml" \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,12 @@ service host. For a component installed once per deployment, reusing its
 old `service_name` as the key reproduces every previous path and name
 byte for byte.
 
+### Security
+
+- Bumped `h2` from 0.4.15 to 0.4.16 to address RUSTSEC-2026-0258
+  (unbounded buffering of empty HTTP/2 DATA frames, which lets a peer
+  drive a connection's memory use without bound).
+
 ## [0.3.0] - 2026-08-17
 
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- `bootroot service add` and `bootroot-remote bootstrap` refuse a
+  `--service-name` whose lowercased form starts with `bootroot-`, with
+  its own message rather than the DNS-label one. The prefix is reserved
+  for bootroot's own certificate identities, so a name a bootroot
+  component treats as one of its own cannot be obtained by registering
+  an ordinary service — including through a bootstrap run that supplies
+  the name on the command line instead of through an artifact. Ordinary
+  component keywords are unaffected, and so is the bare name `bootroot` —
+  the reserved prefix includes the hyphen. The `agent.toml.compose` smoke
+  profile that shipped with `service_name = "bootroot-agent"` now uses
+  `agent-selftest`; its `registration_id`, `hostname` and certificate
+  paths are unchanged, since only the SAN's second label is reserved.
 - A registration is now keyed by a new required `registration_id`,
   separate from `service_name`. `service_name` used to do two
   incompatible jobs: it was the certificate SAN's second label, so it
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -720,9 +720,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.15"
+version = "0.4.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6cb093c84e8bd9b188d4c4a8cb6579fc016968d14c99882163cd3ff402a4f155"
+checksum = "a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27"
 dependencies = [
  "atomic-waker",
  "bytes",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,12 @@ webpki-root-certs = "1"
 x509-parser = { version = "0.18", features = ["verify"] }
 
 [dev-dependencies]
+# `x509-parser` lets `rcgen` parse a CSR back and sign it, which is how
+# the issuance tests take a `build_csr_params` CSR through a local CA and
+# read the SAN off the leaf. It pulls in nothing new: the feature turns
+# on `x509-parser` 0.18 with `verify`, the exact dependency and feature
+# this crate already declares above.
+rcgen = { version = "0.14", features = ["x509-parser"] }
 tokio = { version = "1", features = ["test-util"] }
 wiremock = "0.6"
 
--- a/agent.toml.compose
+++ b/agent.toml.compose
@@ -31,7 +31,11 @@ backoff_secs = [5, 10, 30]
 
 [[profiles]]
 registration_id = "bootroot-agent"
-service_name = "bootroot-agent"
+# The SAN's second label may not sit under the reserved `bootroot-`
+# prefix, which names bootroot's own certificate identities and which
+# `service add` refuses. `registration_id` and `hostname` are not the
+# SAN label and keep naming the agent.
+service_name = "agent-selftest"
 instance_id = "001"
 hostname = "bootroot-agent"
 
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -3,6 +3,6 @@ services:
     networks:
       default:
         aliases:
-          - 001.bootroot-agent.bootroot-agent.trusted.domain
+          - 001.agent-selftest.bootroot-agent.trusted.domain
           - 001.edge-proxy.edge-node-01.trusted.domain
           - 001.web-app.web-01.trusted.domain
--- a/docs/en/cli.md
+++ b/docs/en/cli.md
@@ -1126,6 +1126,12 @@ Input priority is **CLI flags > environment variables > prompts/defaults**.
 - `--service-name`: service name identifier, used as the SAN's second label
   - Must be a single DNS label: letters, digits, and hyphens only, max 63
     characters, no dots or underscores
+  - Must not start with `bootroot-` (compared case-insensitively). That
+    prefix is reserved for bootroot's own certificate identities, so that
+    a name a bootroot component recognizes as one of its own cannot be
+    minted through `service add`. Ordinary component keywords such as
+    `roxyd`, `piglet` and `edge-proxy` are unaffected, as is the bare name
+    `bootroot`.
 - `--delivery-mode`: delivery mode (`local-file` or `remote-bootstrap`).
   Note: `remote-bootstrap` is a mode value, not an executable binary, and the
   executable used for this mode is `bootroot-remote`.
@@ -2696,6 +2702,12 @@ Key inputs:
     letter or digit, max 131 characters
 - `--service-name`: required unless `--artifact` is provided.
   - Must follow the same single-label DNS rule as `bootroot service add`
+  - Must not start with `bootroot-` either (compared case-insensitively),
+    for the same reason and with its own error. This flag becomes
+    `[[profiles]].service_name`, hence the second label of the SAN the
+    agent orders, and a run without `--artifact` takes it straight from
+    the command line rather than from an artifact `service add` already
+    vetted.
 - `--role-id-path`, `--secret-id-path`, `--eab-file-path`: required
   unless `--artifact` is provided.
 - `--agent-config-path`: required unless `--artifact` is provided.
--- a/docs/ko/cli.md
+++ b/docs/ko/cli.md
@@ -1099,6 +1099,11 @@ post-renew 훅이 컨테이너를 리로드합니다
 - `--service-name`: SAN의 두 번째 label로 쓰이는 서비스 이름 식별자
   - 단일 DNS label이어야 합니다. 영문자/숫자/하이픈만 허용하며, 최대 63자,
     점(`.`)과 밑줄(`_`)은 허용되지 않습니다.
+  - `bootroot-`로 시작할 수 없습니다(대소문자 구분 없이 비교). 이 접두사는
+    bootroot 자체 인증서 identity 전용으로 예약되어 있으며, bootroot 컴포넌트가
+    자기 것으로 인식하는 이름을 `service add`로 발급받지 못하게 막습니다.
+    `roxyd`, `piglet`, `edge-proxy` 같은 일반 컴포넌트 키워드와 `bootroot`
+    자체는 영향을 받지 않습니다.
 - `--delivery-mode`: 전달 모드 (`local-file` 또는 `remote-bootstrap`).
   참고: `remote-bootstrap`은 실행 파일이 아니라 모드 값이며, 이 모드에서
   사용하는 실행 파일은 `bootroot-remote`입니다.
@@ -2559,6 +2564,11 @@ fast-poll 루프로 trust와 `secret_id`를 최신 상태로 유지하며,
     하고, 최대 131자입니다.
 - `--service-name`: `--artifact` 미지정 시 필수.
   - `bootroot service add`와 같은 단일 DNS label 규칙을 따릅니다.
+  - 마찬가지로 `bootroot-`로 시작할 수 없으며(대소문자 구분 없이 비교),
+    같은 이유로 별도의 오류 메시지를 냅니다. 이 값은
+    `[[profiles]].service_name`이 되어 agent가 요청하는 SAN의 두 번째
+    label이 되며, `--artifact` 없이 실행하면 `service add`가 이미 검증한
+    artifact가 아니라 명령줄 값이 그대로 쓰입니다.
 - `--role-id-path`, `--secret-id-path`, `--eab-file-path`:
   `--artifact` 미지정 시 필수.
 - `--agent-config-path`: `--artifact` 미지정 시 필수.
--- a/docs/reference/registrar-client-identity.md
+++ b/docs/reference/registrar-client-identity.md
@@ -1,0 +1,268 @@
+<!-- markdownlint-configure-file {
+  "MD013": { "tables": false, "code_blocks": false }
+} -->
+
+# Registrar client identity, endpoint identity, and the endpoint pin file
+
+This file is the contract between bootroot and the provisioning tool that
+installs a registrar. bootroot **consumes** the pin file described here; it does
+not write it. It is checked in outside the mirrored `docs/en/` + `docs/ko/`
+operator pair, on the precedent of `docs/rfcs/` and
+`docs/reference/registrar-wire-contract.md`: it is a cross-repository contract
+rather than operator documentation, so `mkdocs.yml` lists it in neither locale's
+nav and there is no `docs/ko/` counterpart.
+
+## 1. The two reserved identities
+
+bootroot issues one certificate shape for ordinary services: a single DNS SAN
+`<instance>.<service_name>.<host>.<domain>`, repeated as the common name, with no
+extended key usage requested (`src/acme/flow.rs::build_csr_params`). The
+registrar surface needs two names that an ordinary `bootroot service add` cannot
+produce. They reuse that composition exactly and differ in the **second label
+only**:
+
+| Identity | Name | Defined as |
+| --- | --- | --- |
+| Registrar client — the leaf the registrar authenticates to the host-local endpoint with | `<instance>.bootroot-registrar.<host>.<domain>` | `bootroot::registrar::REGISTRAR_CLIENT_LABEL` |
+| Endpoint server — the leaf the daemon's endpoint presents, and the name a client pins | `<instance>.bootroot-registrar-endpoint.<host>.<domain>` | `bootroot::registrar::REGISTRAR_ENDPOINT_LABEL` |
+
+For the v1 single registrar per bootroot host, `<instance>` is `001` and
+`<host>` is that host's single DNS label. With `domain = example.internal` the
+two names are:
+
+```text
+001.bootroot-registrar.h1.example.internal
+001.bootroot-registrar-endpoint.h1.example.internal
+```
+
+**`<domain>` is a suffix, not a label.** It is the configured `network.domain`,
+and `validate_domain_name` accepts any number of labels — `internal` is one,
+`example.internal` is two, `corp.example.internal` is three. Both names are
+therefore *three fixed leading labels followed by the configured domain suffix*,
+and their total label count is `3 + <label count of the configured domain>`.
+Never hard-code a total label count and never treat the domain as a single
+trailing label.
+
+Both names carry **exactly one** SAN, of type DNS, and no SAN of any other type,
+and repeat the SAN string as the common name — exactly like today's service
+leaves.
+
+How this differs from an ordinary service leaf on the same host: roxyd's leaf
+there is `001.roxyd.h1.example.internal`, its plain component keyword in the
+second label. The `registration_id` split keys bootroot's namespaces and does
+not change the SAN label. The registrar client identity differs in the second
+label alone.
+
+## 2. The reserved-prefix guard on operator-supplied service names
+
+`bootroot service add` refuses a `--service-name` whose ASCII-lowercased form
+starts with **`bootroot-`** (`bootroot::registrar::RESERVED_SERVICE_NAME_PREFIX`),
+alongside the existing DNS-label check, with its own error rather than the
+DNS-label one. The comparison is case-insensitive because `validate_dns_label`
+admits mixed case and DNS labels compare case-insensitively, so
+`BOOTROOT-Registrar` is refused too.
+
+`bootroot-remote bootstrap --service-name` is held to the same rule, through the
+same predicate. That flag becomes `[[profiles]].service_name` in the `agent.toml`
+the bootstrap writes, and therefore the second label of the SAN the agent orders;
+a run without `--artifact` takes it straight from the command line rather than
+from an artifact `service add` already vetted, so it is a second way into the
+same name. Both call
+`bootroot::registrar::is_reserved_service_name` — one predicate, so there is no
+second list of reserved names to drift.
+
+Ordinary component keywords — `roxyd`, `piglet`, `edge-proxy` — are unaffected,
+as is the bare name `bootroot`: the prefix includes the hyphen.
+
+**Where the guard is deliberately not applied.** It sits on the two operator
+inputs that create a registration, not on `[[profiles]].service_name` as the
+daemon loads it: `src/config/validation.rs` still holds that field to the
+DNS-label rule alone. Every writer of a `[[profiles]]` block is one of the two
+guarded commands, so only a hand-edited `agent.toml` can reach the daemon with a
+reserved label — and that config belongs to the host's own root, which already
+holds the agent's credentials. Guarding it there would also stand in the way of
+bootroot's own identities, which are ordered under this very prefix. So the
+guard's claim is precise: a name under the prefix cannot be obtained by
+*registering a service*. It is not a claim that no ACME order for such a name can
+ever be placed by a host against itself, and an endpoint that scopes
+authorization by the `host` label recognition returns (§3) is the layer that
+decides what a recognized identity may then do.
+
+This guard is what makes recognition (§3) sound. Without it anyone able to call
+`service add` could mint the registrar's own name. The **whole `bootroot-`
+namespace** is reserved rather than the two labels above, so a later
+bootroot-internal identity picks a name under the same prefix and is covered by
+this one guard — there is deliberately no second list of reserved names that
+could drift from it. `bootroot-` was already bootroot's own namespace prefix
+(`bootroot-service-`, `bootroot-runtime-`), and the product is pre-release, so
+nothing needs migrating. One fixture did sit under it: the agent's own
+self-test registration in CI and in the `agent.toml.compose` smoke profile
+registered `service_name = "bootroot-agent"`. Its SAN label is now
+`agent-selftest`; its `registration_id`, `hostname` and certificate paths are
+unchanged, since only the second label is reserved.
+`tests/reserved_service_name_fixtures.rs` keeps that from drifting back —
+a workflow, script or example config that registers a reserved name fails
+`cargo test` rather than a Docker E2E job.
+
+## 3. The recognition rule
+
+`bootroot::registrar::recognize_registrar_client(end_entity_der, domain)` is the
+single implementation a server-side verifier calls. Given an **already-verified**
+peer certificate and the locally configured `domain`, it accepts only when all
+of the following hold:
+
+1. the certificate carries **exactly one** SAN, of type DNS;
+2. that name ends with the configured `domain` **on a label boundary** — the name
+   equals `<something>.<domain>`, never a bare string suffix, so the domain
+   `example.internal` does not accept `…​.evil-example.internal`;
+3. stripping the domain suffix leaves **exactly three** labels (equivalently:
+   the total label count is `3 + <label count of domain>`, derived from the
+   configured value);
+4. of those three, the **second** is `bootroot-registrar`;
+5. the **first** is a numeric instance label and the **third** a valid DNS label.
+
+Every label and suffix comparison is ASCII-case-insensitive. The common name is
+**not** consulted: today's leaves mirror the SAN into the CN, but recognition is
+SAN-only.
+
+On success the rule returns the parsed `instance`, `host` and `domain`, each
+ASCII-lowercased, so the endpoint can scope authorization by host. **Which**
+recognized identity may invoke **which** operation is the endpoint's decision,
+not this rule's. Every failure is its own typed rejection
+(`RegistrarIdentityError`).
+
+The rule performs no chain building and no time validation — it is a name rule
+applied to a peer certificate something else has already verified. It also
+**never consults the extended key usage**, for the reason recorded in §5: every
+leaf this CA issues carries the same EKU set, so that attribute cannot
+discriminate. Recognition is by name.
+
+## 4. The endpoint pin file
+
+### 4.1 What is pinned
+
+**Trust anchors, plus an exact expected SAN on the presented leaf.** Not the
+endpoint leaf's DER and not its SPKI: every bootroot issuance generates a fresh
+key pair, so a leaf-shaped pin goes stale at the next daemon renewal, while the
+writer of this file runs once at install and lives in another repository. Anchor
+pinning is also the one pinning idiom this tree already has
+(`trust.trusted_ca_sha256`, `tls::ca_bundle_fingerprints`), so the existing
+`PinnedCertVerifier` semantics are reused rather than a second scheme grown
+beside them.
+
+Pinning the anchor alone would admit any leaf that CA ever issued, so the anchor
+pin is paired with a **mandatory** check that the presented end-entity
+certificate's single DNS SAN is exactly the endpoint server identity name from
+§1 — a name the §2 guard makes unmintable through `service add`.
+
+### 4.2 Where the file lives
+
+bootroot has **no fixed certificate directory**: every certificate and key path
+is configured per profile (`[[profiles]].paths`). This contract therefore fixes
+the **basename** and a **derivation rule**, and no absolute directory:
+
+- basename: **`registrar-endpoint-anchors.sha256`**
+  (`bootroot::registrar::endpoint_pin::REGISTRAR_ENDPOINT_ANCHORS_FILE`);
+- location: the **parent directory of the registrar client certificate**, joined
+  with that basename —
+  `bootroot::registrar::endpoint_pin::anchor_pin_path_for_client_certificate`
+  computes exactly this and touches no filesystem and no configuration.
+
+So the file's absolute directory **follows the configured registrar client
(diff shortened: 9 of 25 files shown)


AcoPiper:
**[Reviewer Round 1]**

No findings. The implementation matches #760’s contract: the additive registrar CSR carries the reserved-label single DNS SAN and `ClientAuth` without widening ordinary service CSRs; the shared, case-insensitive `bootroot-` guard protects both service-add routes; and registrar recognition is SAN-only with a label-boundary domain check.

The endpoint verifier is also appropriately fail-closed: it parses the explicit pin file, selects only presented pinned CA-capable and time-valid anchors, reuses the existing pinning verifier for chain/direct-pin handling, and enforces the exact endpoint SAN independently of the caller’s dialed `ServerName`. The tests cover the meaningful negative cases, including multi-label domains, malformed/multi/non-DNS SANs, pin-file parsing, chain failures, and direct-pin behavior.

The PR body has the required issue linkage and a checked test plan; the thread contains no author-created follow-up issue.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Add the registrar client and endpoint identities
```

**Body**

```text
The host-local registrar endpoint has to tell its caller apart from an
ordinary service, and that caller has to know which server it reached
over a socket that supplies no hostname. Both are certificate-side
questions, so they are settled ahead of the endpoint itself rather
than left for whichever side is built first to invent.

Recognition is a name rule and consults no extended key usage:
step-ca's default template hands every leaf it issues both server and
client authentication, so an attribute every leaf carries cannot
discriminate. The registrar reuses today's SAN composition and takes a
reserved second label, `bootroot-registrar`, with the endpoint on
`bootroot-registrar-endpoint`; the domain stays a suffix of whatever
label count it was configured with, never a single trailing label.
`service add` and `bootroot-remote bootstrap` refuse a `--service-name`
anywhere under the `bootroot-` prefix through one shared predicate,
which is what makes those names unmintable. Reserving the whole prefix
rather than the two labels in use means a later internal identity is
covered by the same guard instead of a second list that drifts from
it. The one shipped fixture sitting under the prefix, the agent's
self-test registration, moves to `agent-selftest`, and a `cargo test`
scan keeps the next one from being caught only by a Docker E2E job.

The endpoint is pinned the way this tree already pins: trust anchors
by DER SHA-256, reusing `PinnedCertVerifier` rather than growing a
second verification path, plus an unconditional check that the
presented leaf's single DNS SAN is the endpoint identity name. Neither
half stands alone — an anchor pin by itself admits every leaf that CA
ever issued, and a leaf or SPKI pin would go stale at the next
renewal, with the provisioning tool that wrote it long gone. The
helper takes the pin file path as an explicit argument and fixes no
absolute directory, because certificate paths here are per-profile
configuration; only the basename and the beside-the-certificate
derivation rule are fixed.

Closes #760.
```

AcoPiper:
Rebased onto `main` at 3af222d ("Read the registrar config and derive identities"). The branch is now a single commit on top of that.

**Conflict resolved:** one add/add conflict in `src/registrar.rs`. Both sides created the file from scratch: `main` as the provisioning-config and derivation library (`config`/`error`/`fixture`/`identity` submodules, a file-based `#[cfg(test)] mod tests;`, and the re-export block), this branch as the certificate-identity module (`endpoint_pin` submodule, the reserved-prefix guard, the two identity composers, `single_dns_san`, and the client recognition rule).

The two are complementary rather than competing, so the resolution keeps both in full:

- Union of the module declarations, with `pub mod endpoint_pin;` added alongside `main`'s four.
- `main`'s `pub use` re-exports kept verbatim.
- This branch's items kept inline in `registrar.rs`, which is where every caller already reaches them from — `src/commands/service/resolve.rs`, `src/bin/bootroot-remote/validation.rs`, `src/acme/flow.rs` and `tests/reserved_service_name_fixtures.rs` all use `bootroot::registrar::…`/`crate::registrar::…` paths, so no path changed.
- The module doc is merged into one block: `main`'s description of the derivation library, then this branch's description of the two reserved identities.
- The only rename: this branch's inline `#[cfg(test)] mod tests` became `mod recognition_tests`, because `main` now occupies the name `tests` with the file-based `src/registrar/tests.rs`. Both modules compile and run — `registrar::tests` 45 passed, `registrar::recognition_tests` 14 passed, `registrar::endpoint_pin` 27 passed.

No other file conflicted; `src/lib.rs` merged clean (both commits declare `pub mod registrar;`).

**Verification after the rebase:** `cargo build --all-targets`, `cargo fmt --check` (with `group_imports=StdExternalCrate`), `cargo clippy --all-targets -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc`, `markdownlint-cli2`, `ruff format --check` / `ruff check`, `biome ci`, and `./scripts/check-docs.sh` (`mkdocs build --strict`) all pass. The full `cargo test` suite is green: ~2,000 tests, 0 failures.

One caveat: `cargo audit` reports RUSTSEC-2026-0258 (`h2` 0.4.15, unbounded empty DATA frames). It is pre-existing and not from this branch — `Cargo.lock` is byte-identical to `main`'s and `main` resolves the same `h2` version.

Docker E2E (`test-core.sh`, `e2e-matrix.sh`, `e2e-extended.sh`) was not run locally; those are left to CI.

PR #874, opened 2026-08-18
Add the registrar mint and deregister verbs (#758)

### LINKED ISSUE #758: Implement the restricted registrar mint and deregister verbs
# Implement the restricted registrar mint and deregister verbs

## Context

Registrar configuration already supplies component multiplicity and a rendered safe-set, while registrar identity helpers separately validate labels and spec identity, validate instance shape, derive a registration id and SAN, and validate the safe-set. The service AppRole path currently creates only the derived bootroot-service-<registration_id> role and policy. The registrar must offer a transport-free, restricted control plane: it derives identities from request parts, never accepts caller-authored OpenBao authority, and preserves a durable host/spec binding.

The bootroot- service-name namespace is reserved by bootroot::registrar::is_reserved_service_name with case-insensitive prefix semantics. Registrar verbs must enforce that predicate before configuration lookup so an operator configuration cannot cause ordinary service issuance to mint a registrar client identity.

## Scope

Add a crate-private registrar verbs module under src/registrar/ and document it in src/registrar.rs. Its construction dependencies are a privileged OpenBaoClient, KV mount, loaded RegistrarConfig, fixed trusted SecretIdOptions, role-level AppRole TTL configuration, and a bounded registrar wrap-TTL policy that supplies a maximum and validation rules. It has no binary-crate types, state.json access, prompts, authentication resolution, delivery mode, localized Messages, or caller-controlled credentials.

Both verbs accept an opaque caller identity plus service_name, host, and optional numeric instance; mint additionally accepts the requested spec and wire Duration wrap_ttl. Carry the caller identity unchanged in outcomes, but never parse it, authenticate it, select an OpenBao client from it, or derive identity from it. Neither verb accepts a composed name, registration id, domain, policy body, role definition, policy list, or credential.

Use this shared order:

1. Under a process-wide tokio::sync::Mutex, validate service_name and host as labels and reject is_reserved_service_name service names. For mint only, check spec identity. For both mint and deregister, resolve component multiplicity from the rendered configuration and check instance shape before derivation. Component absence and instance-shape mismatch are pre-derivation refusals for both verbs. Release this guard before any await.
2. Derive registration id and SAN with the existing registrar helpers. Treat derivation as fallible even after label validation; a derivation error has no registration id, does not acquire a per-id lock, and performs no OpenBao or binding operation.
3. Acquire the tokio::sync per-registration-id lock and complete binding and OpenBao work. Mint and deregister share the lock. Store weak per-id lock entries in a mutex-protected map; an acquisition upgrades or inserts an Arc while holding that map guard, and cleanup removes only the identical dead Weak entry while holding it. This reclaims entries without racing acquisition or retaining an unbounded caller-controlled map.

Extract a shared, client-injected provisioning helper and a client-injected teardown helper from the binary-only service modules. The provisioning helper creates or converges only the derived policy, AppRole, and role id; it never creates or returns a secret id. It accepts explicit role-level token_ttl and secret_id_ttl rather than importing binary-only constants; the CLI passes its current TOKEN_TTL and SECRET_ID_TTL, and the registrar supplies its fixed construction configuration. Issuance remains at each caller's boundary: when a CLI wrap TTL is requested, the CLI preserves its existing create_secret_id_wrapped wrap-and-immediate-unwrap delivery, while registrar mint separately invokes OpenBaoClient::create_secret_id_wrap_only and retains the WrapInfo token for the remote caller. No registrar path may use create_secret_id_wrapped, because it returns the unwrapped secret_id.

Before any OpenBao call, validate the requested wire Duration under the construction-supplied policy: refuse zero, negative, and values that cannot be represented as an OpenBao duration string. For a valid request, grant min(requested, policy maximum), convert that granted duration to the OpenBao string, and invoke create_secret_id_wrap_only with it. Compute expires_at as time::OffsetDateTime by parsing WrapInfo.creation_time as RFC3339 and adding the WrapInfo.ttl OpenBao actually reports with a checked conversion to time::Duration. Successful mint outcomes carry the SAN derived alongside the registration id, role_id, a redacted wrapped secret token, and expires_at; expires_at is the granted deadline after both the registrar and any OpenBao-side clamp, never the requested wrap_ttl echoed back. Hold the wrapped token in a private redacting newtype, not a bare String, and provide a hand-written Debug implementation that prints <redacted>; no outcome Debug implementation may reveal it. The successful mint outcome, not the newtype alone, provides a consuming pub(crate) accessor that moves the token to the endpoint only at its serialization boundary, so no caller needs to clone or inspect the secret.

Declare pub(crate) const REGISTRAR_BINDING_KV_SUFFIX: &str = "registrar_binding". At bootroot/services/<registration_id>/registrar_binding, store a versioned, verb-local binding record; do not add serde derives to RegistrationSpec, ReloadSpec, or ReloadKind. Its schema-version-1 JSON fields are schema_version, host, state, requested_spec, and applied_spec. requested_spec and applied_spec use the local shape {"cert_group": <u32 or null>, "reload": {"kind": <"sighup" | "systemd" | "docker-restart" | "none">, "target": <string or null>}}. The local reload-kind enum is serde kebab-case, the lifecycle enum is serde snake_case (creating or active), and decoding rejects unknown fields. Convert explicitly between these local types and the registrar types using ReloadKind::as_str()/FromStr. A creating record has requested_spec and applied_spec: null; transitioning it to active sets applied_spec to that same requested value. Compare a matching-host creating re-drive against requested_spec and an active re-mint against applied_spec. Reject an unknown schema version as a fail-closed unavailable error without changing data. The binding is the durable authority for host collision, host mismatch, and re-mint decisions.

Extend src/openbao.rs with a public absent-only KV v2 write API, returning anyhow::Result<KvCreateIfAbsent>. Its public KvCreateIfAbsent enum distinguishes Created(Option<u64>) from AlreadyExists; all transport, protocol, and non-CAS API failures remain anyhow errors. The request body is {"options":{"cas":0},"data":...}. Classify a response as AlreadyExists only when OpenBao returns the documented CAS-mismatch 400 response containing "check-and-set parameter did not match the current version"; preserve every other 400 as an error. Unit-test the request envelope, successful result, exact mismatch classification, and a different 400 error with the existing canned-response Wiremock pattern; Wiremock never maintains OpenBao state. The verb claims a new binding with this API rather than write_kv or write_kv_with_version. The creating-to-active transition updates a path that already carries a version, which an absent-only create cannot express, so that one write stays an ordinary KV v2 write under the per-id lock.

After derivation and while holding the per-id lock, inspect any binding before safe-set validation. A different stored host is a collision before either rendered-safe-set or stored-spec comparison; a same-host different spec is a conflict. Only after those binding checks validate the requested spec against the rendered safe-set. A same-host matching active binding reuses role and policy and returns fresh wrap-only material. When no binding exists, claim a creating binding with the absent-only KV v2 CAS API. On AlreadyExists, reread the binding and repeat the same host-first order. The CAS claimant, or a matching-host re-drive of a creating binding, idempotently converges role and policy, changes the binding to active, and only then issues a credential. If convergence fails or is interrupted after the creating claim wins, retain that creating binding and return the failure; never roll it back or delete it on the graceful-failure path. The matching host recovers through the same mint re-drive or through deregister's teardown-before-unbind sequence, while every other host remains refused. This keeps every partial role, policy, or service-material state covered by its durable binding rather than manufacturing unbound orphans; no credential may be issued before an active binding is persisted. CAS is mandatory because other processes can write the OpenBao namespace; an in-process mutex does not provide a cross-process collision guarantee.

Keep the test-backend boundary explicit. Fast unit tests cover control flow that refuses before a stateful OpenBao interaction. Every outcome depending on prior OpenBao writes is an ignored #[cfg(test)] module inside src/registrar/verbs/, not an integration test under tests/, so it can construct the crate-private service and assert crate-private outcomes. A new scripts/impl/run-registrar-verbs-e2e.sh Docker scenario brings up live OpenBao using the existing lifecycle conventions and invokes cargo test --lib registrar::verbs::tests -- --ignored. The script supplies BOOTROOT_REGISTRAR_TEST_OPENBAO_URL, BOOTROOT_REGISTRAR_TEST_OPENBAO_TOKEN, and BOOTROOT_REGISTRAR_TEST_KV_MOUNT to that child command; tests read but never mutate those variables. Add this scenario to scripts/preflight/ci/e2e-matrix.sh and the test-docker-e2e-matrix CI job, which is the required gate for the ignored library tests. Do not build a stateful in-process protocol fake.

The shared teardown helper derives role and policy names from registration_id and accepts explicit KV suffixes. It attempts every supplied KV path, AppRole, and policy even if one deletion fails, and returns a typed per-resource outcome of removed, already absent, or failed plus an aggregate-success predicate. The CLI adapter preserves its existing per-resource reporting and retains its state.json entry on aggregate failure; it must not short-circuit. Registrar deregister deletes registrar_binding only after aggregate success.

Use the public trust_bootstrap constants SERVICE_EAB_KV_SUFFIX, SERVICE_RESPONDER_HMAC_KV_SUFFIX, SERVICE_TRUST_KV_SUFFIX, SERVICE_SECRET_ID_KV_SUFFIX, and SERVICE_REISSUE_KV_SUFFIX for all service-material paths, replacing remove.rs's duplicate private strings. Registrar deregister always supplies those five suffixes and excludes registrar_binding. The CLI retains its state-derived suffix set and intentionally continues not to delete reissue, preserving current CLI behavior; that leaves its existing reissue inheritance hazard in place. Registrar deregister removes reissue to prevent that hazard for registrar-managed identities.

With a matching active binding, deregister performs full teardown and then deletes the binding. A matching creating binding uses the same teardown and deletion sequence as the operational recovery path, and returns identity removed because it removes the durable registration claim whether or not role material was ever completed. A different host is a typed host-mismatch refusal for either lifecycle state and changes nothing. With no binding, it returns an idempotent already-absent outcome but still sweeps the full five-suffix set. A deregister request whose component was removed from the rendered configuration cannot resolve multiplicity or derive an id, so it is refused pre-derivation and leaves its binding/material intact until that configuration entry is restored. The accepted hazard is explicit: a manually deleted/restored binding or an identity created through the direct CLI can leave live material sweepable by an absent-binding registrar deregister request for the same derived id. Do not add an intent marker or second state store.

Define crate-private typed results and errors. Wrap individually matchable registrar errors rather than redeclaring them, and add only verb-owned errors: reserved service name, registration-id collision, stored-spec conflict, host mismatch, unusable requested wrap TTL, and unavailable/transient failure. The wrap-TTL refusal is its own arm rather than an unavailable failure: it is a permanent property of the request, so an unchanged retry fails identically. Keep safe-set refusal, stored-spec conflict, spec-identity disagreement, component absence, and instance-shape mismatch distinct internally even when the wire contract groups them. Generate a request id at invocation entry. Every outcome carries request id, unchanged caller identity, producing arm, and registration id only after derivation; successful outcomes distinguish first mint, idempotent re-mint, identity removed, and already absent.

## Acceptance criteria

- [x] Mint creates or reuses only the derived bootroot-service-<registration_id> role and policy through the shared provisioning helper, which returns role_id but never creates a secret ID; reading the role shows exactly token_policies == ["bootroot-service-<registration_id>"].
- [x] SecretIdOptions and role-level TTLs are fixed construction dependencies; no mint request selects secret-id TTL, use count, metadata, token-bound CIDRs, token TTL, or role secret-id TTL.
- [x] Registrar mint uses create_secret_id_wrap_only, never create_secret_id_wrapped. Before an OpenBao call it refuses zero, negative, and OpenBao-duration-unrepresentable requested wrap_ttl values; otherwise it grants min(requested, its construction-supplied maximum). It returns a fresh wrapped token whose outcome has a redacted Debug representation, a consuming crate-private accessor on the outcome itself, and an expires_at: time::OffsetDateTime calculated from the actual WrapInfo creation time and TTL with RFC3339 parsing and checked duration conversion, including any OpenBao-side clamp.
- [x] Reserved names are refused case-insensitively before component lookup or derivation. Invalid labels, absent component, and instance-shape mismatch are separately matchable pre-derivation refusals for both verbs; spec-identity disagreement is an additional mint-only pre-derivation refusal.
- [x] A 63-octet host plus 63-octet component with instance 1000, and an uppercase host with a lowercase configured component, both fail derivation with no per-id lock or OpenBao/binding action.
- [x] The durable version-1 binding is stored at bootroot/services/<registration_id>/registrar_binding with the specified verb-local JSON shape and reload-kind spellings, survives restart, is excluded from teardown, is active before material is returned, and rejects an unknown schema version or unknown record field without mutation. A failed or interrupted post-CAS convergence retains its creating binding rather than rolling it back, so matching-host deregister removes it only after aggregate teardown while another host is refused.
- [x] The non-injective pair web on h1-aimer and aimer-web on h1, both instance 001, derives one id; a second host is refused before spec comparison, including across processes simulated by two verb service instances sharing OpenBao.
- [x] src/openbao.rs exposes the documented public absent-only CAS API with a Created-versus-AlreadyExists result, exact CAS-mismatch classification, and non-CAS 400 propagation. New binding creation uses it, so competing creators cannot both receive material. A same-host re-drive of a creating binding converges safely after a prior convergence failure, while a different host cannot take it over.
- [x] Same-host, matching-spec re-mint returns fresh wrapped material without replacing the role or policy; safe-set refusal and stored-spec conflict are distinct no-change outcomes.
- [x] Matching-host deregister tears down role, policy, and all five service-material suffixes before deleting the binding; wrong-host deregister changes neither.
- [x] Teardown attempts every resource and reports per-resource outcomes. A registrar binding is retained after any teardown failure, while already-absent role, policy, and material remain successful idempotent cases.
- [x] Absent-binding deregister sweeps planted orphan role, policy, secret_id, reissue, and every other service-material suffix, returns already absent, and leaves no binding.
- [x] Shared process locks use tokio::sync primitives without holding the global lock across await, reclaim dead per-id entries safely, and serialize mint plus deregister for one id.
- [x] Canned-response Wiremock tests cover only client request/response and pre-stateful refusal control flow. Durable bindings, CAS, re-mint, wrong-host refusal, absent-binding sweep, teardown-before-unbind ordering, and both concurrency criteria run in ignored src/registrar/verbs/ library tests against live OpenBao. scripts/impl/run-registrar-verbs-e2e.sh supplies their read-only connection environment and runs them through scripts/preflight/ci/e2e-matrix.sh and the test-docker-e2e-matrix CI job; no stateful protocol fake is introduced.
- [x] CLI service remove and registrar deregister share teardown behavior while retaining their intentionally different suffix sets and CLI reporting/state-retention behavior.
- [x] Production code passes cargo fmt -- --check --config group_imports=StdExternalCrate and cargo clippy --all-targets -- -D warnings.

## Constraints

Use src/registrar/config.rs and src/registrar/identity.rs as the only implementation source for request-label validation, spec-identity validation, configuration lookup, instance shape, registration-id/SAN derivation, safe-set validation, and existing registrar error variants. Do not call RegistrarConfig::resolve_end_to_end: after derivation, acquire the per-id lock and perform binding host/collision checks before rendered-safe-set validation. Stored-spec comparison remains after the host comparison.

Move only reusable policy/AppRole provisioning (policy, AppRole, and role ID only) and teardown operations into library code. Credential issuance stays at the caller boundary: the CLI keeps its existing raw-secret flow, while registrar mint alone owns wrap-only issuance. run_service_remove remains a CLI adapter because it constructs authenticated clients, reads state, prompts, and handles local artifacts; it must plan and confirm as today before invoking the shared teardown.

The privileged OpenBao client, fixed SecretIdOptions, role-level TTLs, and bounded wrap-TTL policy are construction dependencies. The policy supplies the registrar maximum and the pre-OpenBao validation rules; it is not selected by a mint request. Their provisioning and renewal are outside this issue. Bindings are never in the shared teardown suffix set, and only registrar deregister deletes one after successful aggregate teardown. The direct-CLI/lost-binding sweep hazard must be documented rather than masked.

The checked-in registrar wire reference controls serialization and wire identifiers. This issue supplies typed in-process outcomes only, including time::OffsetDateTime expires_at and the redacting token's crate-private serialization accessor; it neither serializes nor maps them. Do not add chrono: the existing time dependency is sufficient. Use thiserror for new verb errors, keep the CAS client's transport/protocol failures as anyhow errors, avoid unwrap in production, and use the narrowest visibility required.

## Out of scope

The Unix socket, listener, peer authentication or mTLS, request/response codec, caller-facing identifiers, and serialization are endpoint work. Privileged credential provisioning or renewal, certificate issuance, audit persistence, rate limiting, retries, a second binding store, and a caller-controlled credential path are also out of scope.

## Test plan

- [x] Use registrar fixtures and tempfile::tempdir() for on-disk state.
- [x] Keep pre-derivation control-flow tests in the fast unit tier, using the existing canned-response Wiremock pattern where an OpenBao client is needed: reserved-name and label refusal, absent component, and instance-shape mismatch all occur before derivation; mint additionally refuses spec-identity disagreement there. Test both named derivation failures and that neither performs OpenBao or per-id-lock work.
- [x] Keep local, pre-OpenBao behavior in fast unit tests: test the binding record's version-1 JSON round trip for every reload kind, reject unknown record fields and unknown schema versions, and cover zero/negative/unrepresentable wrap-TTL refusal plus registrar-side clamping.
- [x] In ignored src/registrar/verbs/ library tests, run every write-dependent case against live OpenBao: first mint, fresh same-host re-mint, exact role policy, fixed secret-id options and role TTLs, wrap-only delivery, OpenBao-side clamp reflected in granted expiry, safe-set refusal, stored-spec conflict, restart-persistent collision, host mismatch, matching-host removal, and each absent-binding outcome. The tests read only the scenario-supplied OpenBao URL, token, and KV mount.
- [x] In the ignored library Docker tier, test two service instances sharing one live OpenBao backend for CAS collision, including the CAS API's exact conflict path, and concurrent mint plus deregister for a single identity. Force a post-CAS role/policy convergence failure and verify its creating binding is retained, a matching-host mint re-drive or deregister recovers it, and no rollback manufactures an unbound partial state.
- [x] In the ignored library Docker tier, plant orphan role, policy, and all five service records without a binding; verify a sweep removes them, including reissue and secret_id. Re-mint after registrar deregister and verify no reissue record is inherited.
- [x] In the ignored library Docker tier, delete a bound identity's role and policy before deregistering and verify the binding is still removed after the aggregate idempotent teardown. Remove the component from the rendered configuration and verify deregister is a pre-derivation component-absent refusal that leaves its binding and material intact.
- [x] Pass a distinctive opaque caller identity through both verbs and verify it is preserved in outcomes but affects neither identity derivation nor OpenBao requests.
- [x] Do not implement a stateful OpenBao fake. Wiremock remains canned-response only; the dedicated registrar-verbs Docker scenario uses the local container network for every stateful OpenBao case, and test-docker-e2e-matrix executes its ignored library tests in CI.

## Dependencies

Depends on the delivered registrar configuration and identity helpers. Part of #790.

## Pointers

- src/registrar/config.rs, src/registrar/identity.rs, and src/registrar/error.rs
- src/registrar.rs::is_reserved_service_name
- src/commands/service/approle.rs and src/commands/service/remove.rs
- src/trust_bootstrap.rs and src/openbao.rs
- tests/support/docker_harness.rs, tests/docker_e2e_local_lifecycle.rs, scripts/preflight/ci/e2e-matrix.sh, and .github/workflows/ci.yml
- docs/reference/registrar-wire-contract.md

AcoPiper:
## What this delivers

A transport-free registrar control plane: `src/registrar/verbs.rs` and its siblings, crate-private, with no socket, codec, peer authentication or serialization. The endpoint owns all of that; this is the decision procedure alone.

Both verbs take an opaque caller identity plus the identity's parts — `service_name`, `host`, an optional `instance`, and for mint the requested spec and a `time::Duration` `wrap_ttl`. No composed name, no `registration_id`, no domain, no policy body, no role definition, no credential. The privileged `OpenBaoClient`, KV mount, loaded `RegistrarConfig`, fixed `SecretIdOptions`, role-level TTLs and the bounded wrap-TTL policy are construction dependencies, so no request reaches any of them. The caller identity is carried into every outcome unchanged and used for nothing else.

Closes #758

### The fixed order

1. **Pre-derivation**, under a process-wide `tokio::sync::Mutex` released before any `.await`: label validation, the case-insensitive `is_reserved_service_name` guard *before* the component lookup, mint's spec-identity check, multiplicity resolution and the instance-shape check. Mint also validates its `wrap_ttl` here, so every refusal on this arm carries no `registration_id`.
2. **Derivation** of the `registration_id` and the SAN, treated as fallible even after the labels validated. A derivation failure takes no per-id lock and performs no OpenBao or binding operation.
3. **Per-id work** under the `registration_id`'s own `tokio::sync::Mutex`, shared by mint and deregister.

Inside stage 3 the binding is consulted before the safe-set: a different stored host is a collision, a same-host different spec is a conflict, and only then is the requested spec compared against the rendered safe-set. The rendered spec is host-agnostic, so a safe-set-first flow would match for a second host and re-wrap the first host's identity for it.

### The durable binding

`REGISTRAR_BINDING_KV_SUFFIX = "registrar_binding"`, at `bootroot/services/<registration_id>/registrar_binding`. Schema-version-1 JSON with `schema_version`, `host`, `state` (snake_case `creating`/`active`), `requested_spec` and `applied_spec` in the verb-local `{cert_group, reload: {kind, target}}` shape, reload kinds kebab-case. Decoding reads the version on its own before the body, rejects an unknown version fail-closed without touching the data, and rejects an unknown field. No serde derives were added to `RegistrationSpec`, `ReloadSpec` or `ReloadKind`; conversion is explicit both ways and a test pins each local spelling against `ReloadKind::as_str` / `FromStr`.

New bindings are claimed with a new public absent-only KV v2 API, `OpenBaoClient::create_kv_if_absent` → `KvCreateIfAbsent::{Created(Option<u64>), AlreadyExists}`, sending `{"options":{"cas":0},"data":...}`. Only the documented CAS-mismatch 400 becomes `AlreadyExists`; every other 400 and every transport/protocol failure stays an `anyhow` error. CAS is mandatory because the per-id mutex is a *process* lock and another process can write the same namespace.

A claim that wins is never rolled back. If convergence fails or is interrupted, the `creating` binding is retained and the failure returned — that is what keeps whatever role material got created covered by a durable claim rather than manufacturing an unbound orphan. The matching host recovers through a mint re-drive or through deregister's teardown-before-unbind sequence; every other host stays refused. No credential is issued before an active binding is persisted.

### Shared provisioning and teardown

`src/service_material.rs` holds the derived role/policy names, the policy body, `provision_service_role` (policy, AppRole and `role_id` only — it never creates or returns a `secret_id`) and `teardown_service_material`, which attempts every resource, never short-circuits, and returns per-resource outcomes plus an aggregate-success predicate. Role-level TTLs arrive as an explicit `ServiceRoleTtls`, so the CLI keeps `init`'s `TOKEN_TTL`/`SECRET_ID_TTL` and the registrar supplies its construction values.

Issuance stays at each caller's boundary. `service add` keeps its `create_secret_id_wrapped` wrap-and-immediate-unwrap delivery; registrar mint alone calls `create_secret_id_wrap_only` and retains the `WrapInfo` token. `expires_at` is `OffsetDateTime`, computed by parsing `WrapInfo.creation_time` as RFC 3339 and adding the TTL OpenBao actually reported with a checked conversion, so an OpenBao-side clamp is reflected as well as the registrar's own. The token lives in a private newtype whose hand-written `Debug` prints `<redacted>` and which derives neither `Clone` nor `PartialEq`; the consuming accessor is on the successful outcome, not the newtype, so the token is moved out exactly once at the serialization boundary.

`run_service_remove` remains a CLI adapter: it plans, prompts and reports exactly as before, keeps its state-derived suffix set and its intentional omission of `reissue`, and retains its `state.json` entry on aggregate failure. `remove.rs`'s duplicate private suffix strings are gone in favour of the public `trust_bootstrap` constants. Registrar deregister always sweeps all five material suffixes — `reissue` included, which is what keeps that inheritance hazard off registrar-managed identities — and never the binding, which it deletes separately after aggregate success.

### The accepted hazard, documented rather than masked

A deregister with no binding still sweeps the full material set and returns already-absent. That is how an identity whose binding was lost gets cleaned up, and it means a manually deleted/restored binding, or an identity created through the direct CLI, can leave live material sweepable by a registrar deregister for the same derived id. No intent marker and no second state store were added; the hazard is stated in the module header.

## Testing

`cargo test`: 570 lib + 1066 bin + every integration target, all green. `cargo fmt --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings` and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` are clean; `markdownlint-cli2` reports 0 issues.

**Fast tier** (22 tests): reserved-name refusal proven to run before the component lookup — the fixture declares an ordinary `bootroot-decoy` component and it is still unmintable; invalid labels, absent component and instance-shape mismatch for both verbs; spec-identity disagreement for mint; both named derivation failures (63-octet host + 63-octet component with instance 1000, and an uppercase host against a lowercase component) asserted to make zero OpenBao requests and take zero per-id locks; the binding record's version-1 JSON for every reload kind, unknown field and unknown schema version; zero, negative, sub-second and past-`time.Duration`-range wrap-TTL refusal (a whole-second request larger than 9_223_372_036s has no `OpenBao` spelling, so it is refused rather than clamped) and registrar-side clamping; the redacted `Debug`; canned-response Wiremock coverage of the CAS API's envelope, success, exact mismatch classification and non-CAS 400; and the per-id lock map, where a released guard reclaims its entry, a second holder of one id blocks on the *same* lock until the first releases rather than being handed a private mutex, and a reclaimed id is lockable again.

**Ignored tier** (15 tests) in `src/registrar/verbs/tests.rs`, run by the new `scripts/impl/run-registrar-verbs-e2e.sh` against a live OpenBao container on a free loopback port at the image tag `docker-compose.yml` pins. Verified locally against real OpenBao — 15 passed, 0 failed. They cover first mint with `token_policies == ["bootroot-service-<id>"]` and the fixed role TTLs and secret-id options, wrap-only delivery proven by the test itself performing the one and only unwrap, registrar clamping in the granted expiry, fresh same-host re-mint without replacing the role, the non-injective pair refused across two verb service instances sharing one OpenBao, distinct safe-set and stored-spec no-change refusals, a forced post-CAS convergence failure (a scoped token with KV but no policy privileges) whose `creating` binding is retained, refuses a different host and is recovered by a matching-host re-drive, wrong-host deregister changing nothing, teardown-before-unbind across all five suffixes with no `reissue` inherited by a re-mint, a deregister whose sweep fails in aggregate (a token that can delete the KV material but cannot see the AppRole or policy) asserted to attempt every resource, report the partial outcome and retain both the binding and the role it could not remove, the absent-binding orphan sweep, an already-absent teardown still unbinding, the component-removed pre-derivation refusal leaving binding and material intact, two instances racing one id producing exactly one claimant through the CAS conflict path, and a concurrent mint plus deregister asserted never to leave a half-existing identity. The scenario is wired into `scripts/preflight/ci/e2e-matrix.sh` and the `test-docker-e2e-matrix` CI job as `registrar-verbs`, and documented on the manual's CI page in both languages; the connection details reach the child `cargo test` through the environment and the tests only read them. The scenario's log collection filters the `-dev` root token out of the uploaded container log.

No stateful OpenBao fake was introduced — one would be a second implementation of the CAS semantics the claim depends on, and a bug in the first would agree with it.

## Deviations worth a reviewer's attention

- **A fourth `wrap_ttl` refusal arm.** The issue names zero, negative and "values that cannot be represented as an `OpenBao` duration string". That last one is two distinct reasons, and `WrapTtlRefusal` splits them: a sub-second remainder has no whole-second spelling, and a whole-second value above 9_223_372_036s overflows the Go `time.Duration` `OpenBao` parses `<n>s` into. Both are refused before any `OpenBao` call rather than clamped — clamping is for durations that could have been granted — and they are separate arms because the reason a caller cannot have the value differs.
- **The `creating` record carries its `requested_spec`.** The issue says "A `creating` record has `requested_spec` and `applied_spec`: null", but the next sentences — compare a creating re-drive *against `requested_spec`*, and activation sets `applied_spec` "to that same requested value" — are vacuous unless the claim records it. The claim writes `requested_spec`, leaves `applied_spec` null, and decoding still accepts a `creating` record whose `requested_spec` is null (nothing to disagree with, so the re-drive proceeds), so both readings decode.
- **The per-id lock map's own guard is a `std::sync::Mutex`.** The locks themselves are `tokio::sync::Mutex`, as required, and the map guard is held across no `.await` at all. It is a std mutex because reclaiming a dead `Weak` has to happen in `Drop`, which runs when a verb's future is cancelled as well as when it returns — an explicit async reclaim would leak an entry per cancellation. The reclaim removes only the identical, now-dead entry while holding the guard.
- **No `CHANGELOG.md` entry.** Nothing here is observable to someone running the last release: the verbs are crate-private with no endpoint, and the CLI refactor preserves `service remove`'s reporting, suffix set and state-retention behaviour exactly. This matches the precedent set by the registrar config and identity work, which added none either.

## Test plan

- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate` and `cargo clippy --all-targets -- -D warnings` are clean.
- [x] Fast tier: reserved names are refused case-insensitively before the component lookup; invalid labels, absent component and instance-shape mismatch are separately matchable pre-derivation refusals for both verbs, and spec-identity disagreement for mint.
- [x] Fast tier: both named derivation failures (63-octet host + 63-octet component at instance 1000; uppercase host against a lowercase component) perform no OpenBao request and take no per-id lock.
- [x] Fast tier: the version-1 binding record round-trips for every reload kind and rejects an unknown field and an unknown schema version without mutation.
- [x] Fast tier: zero, negative and OpenBao-unrepresentable `wrap_ttl` — sub-second remainders and whole-second values past Go's `time.Duration` range — are refused before any OpenBao call, and a valid request is clamped to the construction maximum.
- [x] Fast tier: Wiremock canned responses pin the CAS request envelope, `Created`, exact `AlreadyExists` classification and a non-CAS 400 staying an error.
- [x] Fast tier: the per-id lock map reclaims a released entry, shares one lock between two holders of the same id, and re-acquires a reclaimed id.
- [x] `scripts/impl/run-registrar-verbs-e2e.sh` runs green: 15 ignored library tests against live OpenBao.
- [x] Ignored tier: first mint reads back `token_policies == ["bootroot-service-<registration_id>"]`, the fixed role TTLs and secret-id options, and wrap-only material whose sole unwrap is the test's own; `expires_at` reflects an OpenBao-side clamp.
- [x] Ignored tier: same-host matching-spec re-mint returns fresh material without replacing role or policy; safe-set refusal and stored-spec conflict are distinct no-change outcomes.
- [x] Ignored tier: the non-injective pair derives one id and the second host is refused before spec comparison, across two verb service instances sharing one OpenBao.
- [x] Ignored tier: a forced post-CAS convergence failure retains its `creating` binding, refuses a different host, and is recovered by a matching-host re-drive or deregister.
- [x] Ignored tier: matching-host deregister tears down role, policy and all five material suffixes before deleting the binding; wrong-host deregister changes nothing.
- [x] Ignored tier: a teardown that fails in aggregate still attempts every resource, reports per-resource outcomes and retains the binding.
- [x] Ignored tier: absent-binding deregister sweeps planted orphan role, policy and all five records including `reissue` and `secret_id`, returns already-absent, and leaves no binding; a re-mint inherits no `reissue`.
- [x] Ignored tier: two instances racing one id produce exactly one claimant through the CAS conflict path, and concurrent mint plus deregister never leaves a half-existing identity.
- [x] Ignored tier: a component removed from the rendered configuration makes deregister a pre-derivation refusal that leaves binding and material intact.
- [x] A distinctive opaque caller identity is preserved in outcomes and affects neither derivation nor any OpenBao request.
- [x] `service remove` retains its planning, prompting, per-resource reporting, state-derived suffix set and `state.json` retention on aggregate failure.
- [x] CI: `test-docker-e2e-matrix` runs the `registrar-verbs` scenario, and `scripts/preflight/ci/e2e-matrix.sh` runs it locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -470,6 +470,10 @@ jobs:
             script: run-two-instance-isolation.sh
             resolution: ""
             artifact: ci-two-instance
+          - label: registrar-verbs
+            script: run-registrar-verbs-e2e.sh
+            resolution: ""
+            artifact: ci-registrar-verbs
     steps:
       - uses: actions/checkout@v7
 
@@ -486,7 +490,7 @@ jobs:
         run: cargo build --bin bootroot --bin bootroot-remote --bin bootroot-agent
 
       - name: Run E2E scenario (lifecycle)
-        if: matrix.scenario.label != 'rotation' && matrix.scenario.label != 'reinit-recovery' && matrix.scenario.label != 'stepca-san' && matrix.scenario.label != 'openbao-tls-no-delta' && matrix.scenario.label != 'openbao-tls-reown' && matrix.scenario.label != 'two-instance'
+        if: matrix.scenario.label != 'rotation' && matrix.scenario.label != 'reinit-recovery' && matrix.scenario.label != 'stepca-san' && matrix.scenario.label != 'openbao-tls-no-delta' && matrix.scenario.label != 'openbao-tls-reown' && matrix.scenario.label != 'two-instance' && matrix.scenario.label != 'registrar-verbs'
         run: |
           ARTIFACT_DIR="$(pwd)/tmp/e2e/${{ matrix.scenario.artifact }}-${GITHUB_RUN_ID}"
           # Local arms run the daemon-only model: every service agent is
@@ -639,6 +643,24 @@ jobs:
             for f in "$ARTIFACT_DIR"/snapshots/*.diff; do
               [ -f "$f" ] && { echo "--- $(basename "$f") ---"; cat "$f"; } || true
             done
+            exit 1
+          }
+
+      - name: Run E2E scenario (registrar verbs)
+        if: matrix.scenario.label == 'registrar-verbs'
+        # This scenario stands its own single OpenBao up on a free
+        # loopback port, so it needs no compose project, no secrets
+        # wiring and none of the bootroot binaries. It is the gate for
+        # the ignored `registrar::verbs::tests` library tests.
+        run: |
+          ARTIFACT_DIR="$(pwd)/tmp/e2e/${{ matrix.scenario.artifact }}-${GITHUB_RUN_ID}"
+          ARTIFACT_DIR="$ARTIFACT_DIR" \
+          ./scripts/impl/run-registrar-verbs-e2e.sh || {
+            echo "registrar-verbs failed"
+            [ -f "$ARTIFACT_DIR/phases.log" ] && cat "$ARTIFACT_DIR/phases.log" || true
+            [ -f "$ARTIFACT_DIR/run.log" ] && tail -n 200 "$ARTIFACT_DIR/run.log" || true
+            [ -f "$ARTIFACT_DIR/cargo-test.log" ] && tail -n 200 "$ARTIFACT_DIR/cargo-test.log" || true
+            [ -f "$ARTIFACT_DIR/openbao.log" ] && tail -n 200 "$ARTIFACT_DIR/openbao.log" || true
             exit 1
           }
 
--- a/docs/en/e2e-ci.md
+++ b/docs/en/e2e-ci.md
@@ -77,6 +77,9 @@ PR-critical Docker test set validates:
 - OpenBao TLS re-issuance after `secrets/` is re-owned
 - two co-located instances stay independent (install, containers, volumes,
   published ports and HTTP-01 responder DNS aliases)
+- the registrar's mint and deregister verbs against a live OpenBao (durable
+  bindings, the KV v2 compare-and-set claim, re-mint, wrong-host refusal, the
+  absent-binding sweep and both concurrency properties)
 
 Primary scripts:
 
@@ -88,6 +91,7 @@ Primary scripts:
 - `scripts/impl/run-openbao-tls-no-delta.sh`
 - `scripts/impl/run-openbao-tls-reown.sh`
 - `scripts/impl/run-two-instance-isolation.sh`
+- `scripts/impl/run-registrar-verbs-e2e.sh`
 
 Three of those scripts are handed no project name at all, and derive their
 own instead. `run-two-instance-isolation.sh` installs two instances into two
@@ -99,6 +103,13 @@ All three pick run-scoped `--instance-name` values and free host ports
 themselves, and every teardown and leftover check is scoped to those exact
 names — they are safe to run on a host that already has a default `bootroot`
 install.
+
+`run-registrar-verbs-e2e.sh` is outside that discussion entirely: it stands a
+single OpenBao container up on a free loopback port and needs no compose
+project, no secrets wiring and none of the bootroot binaries. It is the gate
+for the `#[ignore]`d `registrar::verbs::tests` library tests, which it runs
+with the container's URL, token and KV mount passed in on the child process's
+environment.
 
 ### Two lifecycle runs can share a host
 
--- a/docs/en/e2e-ci.md
+++ b/docs/en/e2e-ci.md
@@ -13,7 +13,7 @@ PR-critical CI (`.github/workflows/ci.yml`) runs:
 
 - `test-core`: unit/integration smoke path
 - `test-docker-e2e-matrix`: Docker E2E test set for end-to-end flow +
-  rotation/recovery (10 scenarios run in parallel via matrix strategy)
+  rotation/recovery (11 scenarios run in parallel via matrix strategy)
 
 Extended E2E (`.github/workflows/e2e-extended.yml`) runs separately:
 
--- a/docs/ko/e2e-ci.md
+++ b/docs/ko/e2e-ci.md
@@ -75,6 +75,9 @@ PR 필수 Docker 조합 검증은 다음을 검증합니다.
 - `secrets/` 소유자 변경 이후 OpenBao TLS 인증서 재발급
 - 같은 호스트의 두 인스턴스가 서로 독립적으로 유지되는지(설치,
   컨테이너, 볼륨, 게시 포트, HTTP-01 리스폰더 DNS 별칭)
+- 실제 OpenBao를 상대로 한 registrar의 mint/deregister 동사(영속 바인딩,
+  KV v2 compare-and-set 클레임, 재발급, 호스트 불일치 거부, 바인딩 없는
+  정리, 두 가지 동시성 속성)
 
 주요 스크립트:
 
@@ -86,6 +89,7 @@ PR 필수 Docker 조합 검증은 다음을 검증합니다.
 - `scripts/impl/run-openbao-tls-no-delta.sh`
 - `scripts/impl/run-openbao-tls-reown.sh`
 - `scripts/impl/run-two-instance-isolation.sh`
+- `scripts/impl/run-registrar-verbs-e2e.sh`
 
 위 스크립트 가운데 셋은 프로젝트 이름을 전혀 전달받지 **않고** 스스로
 만들어 씁니다. `run-two-instance-isolation.sh`는 basename이 같은
@@ -97,6 +101,12 @@ PR 필수 Docker 조합 검증은 다음을 검증합니다.
 스스로 만들고 비어 있는 호스트 포트도 직접 고르며, 모든 정리와 잔여물
 확인이 그 이름들로만 한정되므로 기본 `bootroot` 설치본이 이미 있는
 호스트에서도 안전하게 실행할 수 있습니다.
+
+`run-registrar-verbs-e2e.sh`는 이 논의 바깥에 있습니다. 비어 있는 루프백
+포트에 OpenBao 컨테이너 하나만 띄우므로 compose 프로젝트도, secrets 배선도,
+bootroot 바이너리도 전혀 필요하지 않습니다. 이 스크립트는 `#[ignore]`가 붙은
+`registrar::verbs::tests` 라이브러리 테스트의 게이트이며, 컨테이너의 URL과
+토큰, KV 마운트를 자식 프로세스의 환경 변수로 넘겨 그 테스트들을 실행합니다.
 
 ### 라이프사이클 실행 두 개는 한 호스트를 공유할 수 있습니다
 
--- a/docs/ko/e2e-ci.md
+++ b/docs/ko/e2e-ci.md
@@ -13,7 +13,7 @@ PR 필수 CI(`.github/workflows/ci.yml`)는 다음을 실행합니다.
 
 - `test-core`: 단위/통합 스모크 경로
 - `test-docker-e2e-matrix`: 전체 흐름 + 회전/복구 Docker E2E 조합 검증
-  (10개 시나리오가 matrix 전략으로 병렬 실행)
+  (11개 시나리오가 matrix 전략으로 병렬 실행)
 
 확장 E2E(`.github/workflows/e2e-extended.yml`)는 별도 실행됩니다.
 
--- a/scripts/impl/run-registrar-verbs-e2e.sh
+++ b/scripts/impl/run-registrar-verbs-e2e.sh
@@ -1,0 +1,166 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Docker-backed scenario for the registrar mint/deregister verbs (#758).
+#
+# The verbs' write-dependent behaviour — durable bindings, the KV v2
+# compare-and-set claim, re-mint, wrong-host refusal, the absent-binding
+# sweep, teardown-before-unbind ordering, and both concurrency properties
+# — cannot be asserted against a canned-response mock without building a
+# second implementation of OpenBao's CAS semantics, which is exactly the
+# thing a bug in the first one would then agree with.  So they live as
+# `#[ignore]`d library tests in `src/registrar/verbs/tests.rs` and this
+# scenario is what runs them:
+#
+#   1. stand a single OpenBao container up on a free loopback port, at
+#      the image tag `docker-compose.yml` already pins;
+#   2. enable the AppRole auth method the verbs provision into;
+#   3. hand the child `cargo test` the URL, the token and the KV mount
+#      through the environment, and run the ignored tests.
+#
+# The connection details travel as environment variables on the child
+# process only.  The tests read them and never write them: mutating the
+# process environment from a test is unsound in Rust 2024, and there is
+# no reason to here.
+#
+# This is a self-contained OpenBao rather than the full compose stack.
+# The verbs touch no step-ca, no PostgreSQL and no responder, and a
+# scenario that stood them up anyway would be slower and would fail for
+# reasons that have nothing to do with what it is testing.
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd "$ROOT_DIR"
+
+RUN_ID="${GITHUB_RUN_ID:-local-$(date +%s)-$$}"
+ARTIFACT_DIR="${ARTIFACT_DIR:-$ROOT_DIR/tmp/e2e/registrar-verbs-${RUN_ID}}"
+mkdir -p "$ARTIFACT_DIR"
+ARTIFACT_DIR="$(cd "$ARTIFACT_DIR" && pwd)"
+
+PHASE_LOG="$ARTIFACT_DIR/phases.log"
+RUN_LOG="$ARTIFACT_DIR/run.log"
+TEST_LOG="$ARTIFACT_DIR/cargo-test.log"
+OPENBAO_LOG="$ARTIFACT_DIR/openbao.log"
+
+CURRENT_PHASE="startup"
+CONTAINER_NAME="bootroot-registrar-verbs-${RUN_ID}"
+KV_MOUNT="${KV_MOUNT:-secret}"
+READY_ATTEMPTS="${READY_ATTEMPTS:-60}"
+READY_DELAY_SECS="${READY_DELAY_SECS:-1}"
+
+log_phase() {
+  CURRENT_PHASE="$1"
+  printf '{"ts":"%s","phase":"%s"}\n' "$(date -u +%Y-%m-%dT%H:%M:%SZ)" "$1" >>"$PHASE_LOG"
+  echo "[registrar-verbs][$1]"
+}
+
+fail() {
+  printf '[fatal][%s] %s\n' "$CURRENT_PHASE" "$1" >>"$RUN_LOG" 2>/dev/null || true
+  echo "[registrar-verbs][${CURRENT_PHASE}] $1" >&2
+  exit 1
+}
+
+# shellcheck source=lib/ports.sh
+. "$SCRIPT_DIR/lib/ports.sh"
+
+cleanup() {
+  # Best effort, but the logs are collected first: a container removed
+  # before its output was read leaves a failure with nothing to read.
+  if docker inspect "$CONTAINER_NAME" >/dev/null 2>&1; then
+    docker logs "$CONTAINER_NAME" >>"$OPENBAO_LOG" 2>&1 || true
+    docker rm -f "$CONTAINER_NAME" >/dev/null 2>&1 || true
+  fi
+}
+trap cleanup EXIT
+
+command -v docker >/dev/null 2>&1 || fail "docker is required"
+command -v curl >/dev/null 2>&1 || fail "curl is required"
+command -v cargo >/dev/null 2>&1 || fail "cargo is required"
+
+# One pinned image for the whole repository: read it out of the compose
+# file rather than restating it here, so a bump lands in both places at
+# once.
+OPENBAO_IMAGE="$(awk '/^  openbao:/{found=1} found && /image: openbao\//{print $2; exit}' \
+  "$ROOT_DIR/docker-compose.yml")"
+[ -n "$OPENBAO_IMAGE" ] || fail "could not read the OpenBao image tag from docker-compose.yml"
+
+log_phase "allocate-port"
+pick_free_port
+OPENBAO_HOST_PORT="$PICKED_PORT"
+OPENBAO_URL="http://127.0.0.1:${OPENBAO_HOST_PORT}"
+
+# A throwaway root token for a throwaway container published on loopback
+# only.  Drawn from the system CSPRNG all the same, because a predictable
+# one would be a habit worth not forming.
+if command -v openssl >/dev/null 2>&1; then
+  OPENBAO_TOKEN="root-$(openssl rand -hex 16)"
+else
+  OPENBAO_TOKEN="root-$(head -c 16 /dev/urandom | od -An -tx1 | tr -d ' \n')"
+fi
+[ -n "${OPENBAO_TOKEN#root-}" ] || fail "could not generate a dev root token"
+
+log_phase "openbao-up"
+echo "[registrar-verbs] image=$OPENBAO_IMAGE port=$OPENBAO_HOST_PORT" >>"$RUN_LOG"
+docker rm -f "$CONTAINER_NAME" >/dev/null 2>&1 || true
+docker run -d \
+  --name "$CONTAINER_NAME" \
+  -p "127.0.0.1:${OPENBAO_HOST_PORT}:8200" \
+  -e "BAO_DEV_ROOT_TOKEN_ID=${OPENBAO_TOKEN}" \
+  -e "BAO_DEV_LISTEN_ADDRESS=0.0.0.0:8200" \
+  "$OPENBAO_IMAGE" server -dev >>"$RUN_LOG" 2>&1 ||
+  fail "could not start $OPENBAO_IMAGE"
+
+log_phase "openbao-ready"
+ready=0
+for _ in $(seq 1 "$READY_ATTEMPTS"); do
+  if curl -fsS "${OPENBAO_URL}/v1/sys/health" >/dev/null 2>&1; then
+    ready=1
+    break
+  fi
+  sleep "$READY_DELAY_SECS"
+done
+[ "$ready" -eq 1 ] || {
+  docker logs "$CONTAINER_NAME" >>"$OPENBAO_LOG" 2>&1 || true
+  fail "OpenBao did not become ready on ${OPENBAO_URL}"
+}
+
+# Dev mode mounts KV v2 at `secret/` already; AppRole is not enabled by
+# default and the verbs provision roles into it.  Enabling it twice is a
+# 400, so an already-enabled mount is not a failure.
+log_phase "enable-approle"
+enable_status="$(curl -sS -o "$ARTIFACT_DIR/enable-approle.json" -w '%{http_code}' \
+  -X POST \
+  -H "X-Vault-Token: ${OPENBAO_TOKEN}" \
+  -d '{"type":"approle"}' \
+  "${OPENBAO_URL}/v1/sys/auth/approle" || true)"
+case "$enable_status" in
+  2*) ;;
+  400)
+    grep -q "path is already in use" "$ARTIFACT_DIR/enable-approle.json" ||
+      fail "could not enable the AppRole auth method (HTTP $enable_status)"
+    ;;
+  *) fail "could not enable the AppRole auth method (HTTP $enable_status)" ;;
+esac
+
+# Assert the mount the tests are told to use really is KV v2, so a
+# misconfiguration surfaces here rather than as a puzzling 400 inside a
+# compare-and-set assertion.
+log_phase "verify-kv-mount"
+curl -fsS -H "X-Vault-Token: ${OPENBAO_TOKEN}" \
+  "${OPENBAO_URL}/v1/sys/mounts/${KV_MOUNT}" >"$ARTIFACT_DIR/kv-mount.json" 2>>"$RUN_LOG" ||
+  fail "KV mount ${KV_MOUNT} is not present"
+grep -q '"version": *"2"' "$ARTIFACT_DIR/kv-mount.json" ||
+  fail "KV mount ${KV_MOUNT} is not version 2"
+
+log_phase "run-ignored-library-tests"
+set +e
+BOOTROOT_REGISTRAR_TEST_OPENBAO_URL="$OPENBAO_URL" \
+BOOTROOT_REGISTRAR_TEST_OPENBAO_TOKEN="$OPENBAO_TOKEN" \
+BOOTROOT_REGISTRAR_TEST_KV_MOUNT="$KV_MOUNT" \
+  cargo test --lib registrar::verbs::tests -- --ignored 2>&1 | tee "$TEST_LOG"
+status="${PIPESTATUS[0]}"
+set -e
+[ "$status" -eq 0 ] || fail "the ignored registrar verb tests failed (see $TEST_LOG)"
+
+log_phase "done"
+echo "[registrar-verbs] artifacts: $ARTIFACT_DIR"
--- a/scripts/impl/run-registrar-verbs-e2e.sh
+++ b/scripts/impl/run-registrar-verbs-e2e.sh
@@ -63,11 +63,21 @@ fail() {
 # shellcheck source=lib/ports.sh
 . "$SCRIPT_DIR/lib/ports.sh"
 
+# Collects the container's output, with the dev root token redacted.
+# `server -dev` prints "Root Token: <value>" on startup, and this log is
+# uploaded as a CI artifact; the token dies with the container a moment
+# later, but a live credential does not belong in a build artifact on the
+# way there.
+collect_openbao_log() {
+  docker logs "$CONTAINER_NAME" 2>&1 |
+    sed "s|${OPENBAO_TOKEN}|<redacted>|g" >>"$OPENBAO_LOG" || true
+}
+
 cleanup() {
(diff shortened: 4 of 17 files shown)


AcoPiper:
**[Reviewer Round 1]**

- `src/registrar/verbs/wrap_ttl.rs:125` — `openbao_duration` treats every positive whole-second `time::Duration` as representable, but OpenBao parses the `Ns` value as a Go `time.Duration`, whose largest whole-second value is `9_223_372_036s`. For example, `Duration::seconds(9_223_372_037)` is accepted here (and then clamped to the configured maximum), so `mint` provisions/uses the identity instead of returning the required pre-OpenBao `InvalidWrapTtl` refusal. The issue explicitly requires refusal of values that cannot be represented as an OpenBao duration string even when the requested value would otherwise be clamped. Bound the whole-second value to OpenBao's duration range before applying the maximum, and add this overflow case to the fast-tier no-request test.

AcoPiper:

AcoPiper:
**[Author Round 1]**

### `src/registrar/verbs/wrap_ttl.rs:125` — a whole-second `wrap_ttl` past OpenBao's duration range is clamped instead of refused

**Fixed.** The finding is correct, including the bound. `OpenBao` hands an `Ns` value to Go's `time.ParseDuration`, whose result is a nanosecond count in an `i64`, so the largest whole-second value that has a spelling is `i64::MAX / 1_000_000_000` = `9_223_372_036s`. `openbao_duration` only rejected a sub-second remainder, so `Duration::seconds(9_223_372_037)` rendered a string OpenBao cannot parse, passed validation, and — being larger than the maximum — was then clamped and the identity provisioned. That inverts the required order: the issue asks for a pre-OpenBao refusal of every value with no OpenBao duration spelling, and the clamp is for durations that could have been granted in the first place, not a way to make an unrepresentable request succeed under a lifetime nobody asked for.

What changed:

- `openbao_duration` now bounds the whole-second count against a new `MAX_OPENBAO_WHOLE_SECONDS: i64 = i64::MAX / 1_000_000_000`, before `grant` compares the request with the maximum. The bound check is inside the shared renderer, so `WrapTtlPolicy::new` holds a construction-supplied maximum to the same rule — a policy whose ceiling could not be spelled is refused at construction rather than at the first request that clamps to it.
- The refusal is a new `WrapTtlRefusal::ExceedsOpenBaoRange` arm rather than reusing `NotWholeSeconds`. The reason a caller cannot have the value is different — the request *is* a whole number of seconds — and folding the two together would make them indistinguishable in an outcome. It still surfaces as the single `VerbError::InvalidWrapTtl` the wire contract groups these under, so nothing outside the module changes shape.
- The now-unreachable `u64::try_from` fallback is gone. It mapped an impossible negative-to-unsigned failure onto `NotWholeSeconds`, and with a real upper bound in place the value is provably in range, so the string is formatted from the `i64` directly and there is no dead error path pretending to be a reachable one.

Tests, as requested, in the fast tier:

- `mint_refuses_an_unusable_wrap_ttl_before_touching_openbao` gains the `Duration::seconds(9_223_372_037)` case, so the overflow is covered by the same `assert_untouched` that proves the other three make no OpenBao request and take no per-id lock — this is the assertion the finding named, since the whole point is that the identity is *not* provisioned.
- `wrap_ttl_policy_refuses_zero_negative_and_unrepresentable_requests` covers the overflow directly, and adds the boundary on the other side: `Duration::seconds(9_223_372_036)` is representable, so it clamps to the maximum like any ordinary over-long request rather than being caught by the new check.
- `wrap_ttl_policy_holds_its_own_maximum_to_the_same_rules` covers an out-of-range policy maximum.

Docs: the module header and `grant`'s `# Errors` now state that validation runs *before* the comparison with the maximum and why, and the new variant's doc records where the `9_223_372_036s` bound comes from. No user-facing documentation or `CHANGELOG.md` entry applies — the verbs remain crate-private with no endpoint, so none of this is observable to someone running the last release. The PR description's fast-tier and test-plan lines now name both unrepresentable forms instead of only the sub-second one.

### Verification

`cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings` and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` are clean. `cargo test --lib`: 570 passed, 0 failed. `scripts/impl/run-registrar-verbs-e2e.sh` against live OpenBao: 15 passed, 0 failed — including `a_request_over_the_maximum_is_clamped_in_the_granted_expiry`, which confirms the tightened validation did not turn an ordinary clamp into a refusal.

AcoPiper:
**[Reviewer Round 2]**

Round 1 is resolved: `src/registrar/verbs/wrap_ttl.rs` now rejects whole-second durations beyond Go `time.Duration`'s `9_223_372_036s` limit before it considers the policy maximum, and `mint_refuses_an_unusable_wrap_ttl_before_touching_openbao` covers that path with the no-request assertion.

- `docs/en/e2e-ci.md:16` and `docs/ko/e2e-ci.md:16` — both pages still say `test-docker-e2e-matrix` runs 10 scenarios, but `.github/workflows/ci.yml` now defines 11 matrix entries after adding `registrar-verbs`. Please update the English and Korean counts so the CI overview matches the workflow users actually see.

AcoPiper:

AcoPiper:
**[Author Round 2]**

### `docs/en/e2e-ci.md:16` and `docs/ko/e2e-ci.md:16` — the CI overview still announces 10 matrix scenarios

**Fixed.** The finding is correct. `main` had ten `scenario:` entries under `test-docker-e2e-matrix` and both pages said ten; this branch adds `registrar-verbs` as the eleventh and updated everything on the page *except* the count. The rest of the CI overview already describes the new scenario — `scripts/impl/run-registrar-verbs-e2e.sh` is listed and explained in both languages — so the count was the single line contradicting the workflow a reader can open and check for themselves, which is the worst place for a stale number to sit.

Both pages now read eleven. I also swept the repository for any other place the matrix size is stated in prose (`10 scenario` / `10개 시나리오` across Markdown, shell and workflow files) and there is none — the count lived only on these two lines.

No other change was needed for this: `scripts/preflight/ci/e2e-matrix.sh` and `.github/workflows/ci.yml` already carry the scenario, and no production code is touched by this round.

### Verification

`markdownlint-cli2` reports 0 issues. `./scripts/check-docs.sh` passes — the vendored theme verifies, `mkdocs build --strict` completes, and the theme assets reach `site/`. The Rust tree is unchanged from Round 1, where `cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items`, `cargo test --lib` (570 passed) and `scripts/impl/run-registrar-verbs-e2e.sh` (15 passed) were all green.

The PR description needed no edit: it never quoted a scenario count, and its claim that the scenario is "wired into `scripts/preflight/ci/e2e-matrix.sh` and the `test-docker-e2e-matrix` CI job as `registrar-verbs`, and documented on the manual's CI page in both languages" is now true without the caveat this finding identified.

AcoPiper:
**[Reviewer Round 3]**

Round 2 is resolved: `docs/en/e2e-ci.md:16` and `docs/ko/e2e-ci.md:16` now state 11 matrix scenarios, matching the 11 `test-docker-e2e-matrix` entries in `.github/workflows/ci.yml`.

I found no remaining or new findings. The updated diff preserves the required claim → converge → activate → issue ordering, keeps host/spec checks in the required order, and retains bindings on partial teardown. The wrap-TTL range fix and its no-OpenBao-request coverage are present.

Approved.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Add the registrar mint and deregister verbs
```

**Body**

```text
The registrar needs a control plane it can be trusted with: one that
derives every name from the request's parts and the locally rendered
config, never accepts caller-authored OpenBao authority, and keeps a
durable record of which host owns which derived identity.

The derivation is not injective — component names and host labels both
admit hyphens — so "is this id taken, and by whom" cannot be answered
from a request. A versioned binding record at
bootroot/services/<registration_id>/registrar_binding answers it, and
it has to survive a restart to answer it at all. The binding is
consulted before the rendered safe-set, because the rendered spec is
host-agnostic: a safe-set-first flow would match for a second host and
re-wrap the first host's identity for it.

The claim is a KV v2 compare-and-set rather than a lock, because the
per-id mutex is a process lock and another process can write the same
namespace. Once a claim wins it is never rolled back: a creating
binding whose convergence failed is what keeps whatever role material
did get created covered by a durable claim instead of stranded as an
orphan, and the matching host recovers through a mint re-drive or
through deregister.

Issuance stays at each caller's boundary. The policy/AppRole
provisioning and the teardown move into library code the CLI and the
registrar share, but the CLI keeps its raw-secret delivery and its
narrower state-derived suffix set, while registrar mint alone uses
create_secret_id_wrap_only — the wrapped variant unwraps immediately
and hands back the raw secret_id, which is the one value no registrar
path may hold. The wrapped token it does hold is moved out once, at
the serialization boundary: the newtype prints <redacted> and derives
neither Clone nor PartialEq, so no caller can keep a copy beside the
one it sent or compare one byte at a time.

A requested wrap TTL is refused before any OpenBao call whenever it
has no OpenBao spelling at all — zero, negative, a sub-second
remainder, or a whole-second value past the Go time.Duration range
OpenBao parses it into. Clamping is for durations that could have been
granted; clamping one that could not be written down would provision
an identity under a lifetime nobody asked for.

Everything whose outcome depends on a prior OpenBao write is an
ignored library test rather than an integration test, so it can
construct the crate-private service, and a dedicated Docker scenario
is its gate. A stateful OpenBao fake would be a second implementation
of the compare-and-set semantics the claim depends on, and a bug in
the first one would agree with it.

Part of #790
Closes #758
```

Issue #788, opened 2026-08-04
Rate-limit both registrar verbs with two token buckets


sehkone:
# Rate-limit both registrar verbs with two token buckets

## Summary

Once a record is required for **every** registrar verb invocation, including refusals, the record store becomes the resource an attacker exhausts — and because the OpenBao file audit device on the same host is mandatory and OpenBao fails requests it cannot audit, a full disk stops far more than enrollment. This group bounds how fast one client identity may drive the two verbs. **It is a grouping issue and carries no work of its own.**

**The seam between the two implementation children is mechanism versus evidence, and the third child writes no code at all.** The **limiter-mechanism child** builds the limiter: two per-`(client identity, verb)` token buckets — a pre-derivation refusal bucket and a post-derivation admission bucket — both charged *before* the invocation's intent write, with a monotonic refill, a capped bucket map, four integer configuration keys and a written sizing formula, and a typed retryable throttle returned only where the outcome is genuinely undetermined. It suppresses records and emits an event; it writes nothing durable and puts nothing on the wire. The **evidence child** turns that suppressed traffic into evidence and tells the caller: it implements the sink as a coalescing window closed on the daemon's existing tick over a bounded map, writes one counted record per window per key as a third variant inside the audit store's existing types, amends both of the record reader's classification rules for it, supplies the `match` arm that produces the throttle's already-settled wire identifier, and publishes the limited-invocation counters as a health member.

They are strictly ordered — decision child, then limiter mechanism, then evidence — and the interface between the two implementation children is three things the **limiter-mechanism child** defines and exports: the two-value bucket enum, the limited-invocation sink trait, and the typed retryable outcome. That seam is deliberate — it lets the limiter be implemented, tested and reviewed before any record shape exists.

## Children

- [x] #785 Settle the limiter's bucket shape against the RFC — Detection *(RFC amendment transcribing two rulings, no code)*
- [x] #786 Bound both registrar verbs with two token buckets — Detection
- [x] #787 Record limited registrar invocations and report the throttle — Detection

## Shared background

**Suppression is both the point and the hazard.** The limiter works by *not* writing the records a flood would otherwise produce — which is exactly what would make the flood invisible, since the mechanism that would have recorded it is the mechanism being skipped. The answer is one counted record per window per key rather than one per invocation: a sustained flood costs a bounded number of records while the fact of it, who drove it and which bucket ran dry all survive. Neither child may drift back toward per-invocation cost, and neither may drop the fact entirely.

[...]

PR #876, opened 2026-08-19
Share registrar verb locks across the process (#875)

### LINKED ISSUE #875: Share registrar verb locks across the process
# Share registrar verb locks across the process

## Context

The registrar verb layer validates request parts, derives a registration identity, and performs binding and OpenBao work under a per-registration-id lock. RegistrarVerbs::new currently constructs an IdLocks map for each service instance. Two instances in one process can therefore mint and deregister the same derived identity while holding different locks, so their binding and OpenBao work can interleave. This issue corrects that settled lock-ownership defect.

OpenBao's absent-only KV CAS remains the ownership primitive across processes. It cannot provide the required in-process serialization between mint and deregister, and an in-process lock cannot replace its cross-process guarantee.

Pre-derivation reads only immutable per-instance configuration and calls pure validation functions. It has no shared mutable state to guard, so independent validations may proceed concurrently. Fallible derivation still completes before acquisition of the per-registration-id lock.

## Scope

Make the IdLocks map a private module-owned std::sync::LazyLock in src/registrar/verbs.rs. Every RegistrarVerbs instance must obtain its per-id lock through one private instance helper that mint and deregister also use; that helper returns the shared map. Do not add lock state to RegistrarVerbsConfig or make sharing a caller convention. Remove id_locks from RegistrarVerbs.

Provide a narrow cfg(test) accessor that delegates to that same private helper. It is for the fast test to acquire a lock through each RegistrarVerbs instance, not for production callers; it must not be public. The receiver on the production helper is deliberate: it makes the test's cross-instance assertion exercise each service's lock-selection path. Add a local #[allow(clippy::unused_self)] with an adjacent comment stating that reason; do not follow the lint's associated-function suggestion. The test must hold a unique id through the first instance and show acquisition through the second is pending until release with the existing yield-and-atomic pattern. With an implementation that still returns each instance's current map, this test must fail its blocking assertion.

Remove entry_lock entirely. Rewrite the module documentation and pre_derivation item documentation to state that pre-derivation is stateless and unsynchronized, and that the only in-process serialization boundary is the shared per-registration-id lock. Do not add an entry-stage mutex or a test asserting entry-stage serialization. A future entry-stage feature that introduces shared state, such as the rate limiter owned by #788, must bring its own synchronization rather than reintroducing a blanket entry lock. Retain the validation and derivation order: fallible derivation completes before per-id acquisition. The map's std::sync::Mutex is never held across an await; the per-id Tokio mutex remains held across binding and OpenBao awaits so mint and deregister for one identity cannot interleave.

Retain weak-entry reclamation: acquiring a per-id lock upgrades or inserts its Arc while the map is locked, and releasing a guard removes only the identical dead Weak entry while the map is locked. IdLocks remains independently constructible for its three focused reclamation tests; those tests use local maps, not the process-wide static. The verb inputs, outcomes, durable binding format, and OpenBao operations remain unchanged.

Remove tracked_lock_count entirely, along with its assertion in assert_untouched. A global map count is shared mutable test-binary state and is not a sound pre-derivation assertion. Keep assert_untouched's no-OpenBao-request assertion.

Because the fast tier reuses fixed identity labels, a process-wide map can harmlessly queue unrelated tests that use the same id. Do not add timing or global-count assertions to that tier; the new contention test must use a unique id.

Convert the ignored live-OpenBao mint/deregister concurrency test from one shared RegistrarVerbs instance to two separately constructed instances for the same active identity. Load the fixture twice and keep both TempDir guards alive for the whole test body. Its order-independent assertion remains that the final state is either complete and active or fully removed, never partial.

Update the existing two-instance claimant test's stale documentation only. Its assertions remain valuable because concurrent requests still produce exactly one winner, but with shared in-process locking the second claimant observes the persisted binding and receives the existing host-collision refusal rather than exercising a CAS mismatch. The comment must state that the verb layer's KvCreateIfAbsent::AlreadyExists reread branch is only exercised by another process and is not covered by this suite. Keep the canned-response tests in src/openbao.rs for the separate client-level distinction between OpenBao's documented CAS mismatch and other HTTP 400 responses.

## Acceptance criteria

- [x] A private module-owned LazyLock supplies one weak per-registration-id map to every RegistrarVerbs instance, and neither RegistrarVerbs nor RegistrarVerbsConfig owns per-id lock state.
- [x] Mint and deregister obtain their per-id lock through one private instance helper that returns the shared map. A cfg(test) accessor delegates to that helper and is not public. The helper has a local, commented allow for clippy::unused_self because its receiver is required for the cross-instance test; it is not refactored into an associated function.
- [x] A fast test acquires a unique id through one RegistrarVerbs instance and proves acquisition through a second instance remains pending until release. The test fails its blocking assertion if the helper instead returns each instance's own map.
- [x] entry_lock is removed. The module header and pre_derivation item documentation state that the pre-derivation stage is stateless and unsynchronized, and that the shared per-registration-id lock is the only in-process serialization boundary. Fallible derivation completes before per-id acquisition.
- [x] No entry-stage mutex or test asserting entry-stage serialization remains. A future shared entry-stage feature supplies its own synchronization.
- [x] tracked_lock_count and the global map-count assertion in assert_untouched are deleted; pre-derivation refusals still prove that they make no OpenBao request.
- [x] IdLocks remains independently constructible, and its existing local-map reclamation tests continue to assert entry creation, contention, cleanup, and reacquisition without observing the process-wide static.
- [x] Per-id acquisition still upgrades or inserts an Arc while holding the map mutex, cleanup removes only the identical dead Weak entry while holding that mutex, and the map mutex is never held across an await.
- [x] The per-id Tokio mutex remains held through binding and OpenBao work.
- [x] Ignored live-OpenBao mint/deregister coverage uses separate instances with two live fixture directories and, independent of ordering, proves the final state is wholly active or wholly absent.
- [x] The two-instance claimant test retains its exactly-one-winner assertions and documents that a true cross-process AlreadyExists reread is not exercised there.
- [x] The existing canned-response tests continue to distinguish OpenBao's documented CAS mismatch from other HTTP 400 responses at the client boundary.

## Constraints

Do not weaken, replace, or move the absent-only OpenBao CAS claim: it remains the cross-process ownership primitive. Do not change request or outcome types, binding schema, OpenBao API, credential-issuance behavior, identity derivation, endpoint transport, audit behavior, or rate limiting. Use the standard library and existing Tokio synchronization; do not add a dependency solely to initialize the static. Preserve safe weak-entry reclamation for caller-controlled registration ids. The test accessor is cfg(test), delegates to the production lock-selection helper, and exposes no production API. The production helper retains its receiver and uses one local, commented allow for clippy::unused_self; do not suppress that lint more broadly. Pre-derivation remains unsynchronized; a future feature that adds shared entry-stage state must synchronize only that state.

## Out of scope

Cross-process locking or a multi-process concurrency harness; new canned-response infrastructure for verb-level CAS races; a blanket entry-stage lock or entry-stage serialization test; registrar endpoint, mTLS, client, credential-provisioning, audit, and rate-limit work; and any redesign of registrar derivation or durable bindings.

## Test plan

- [x] Add the fast cross-instance contention test with a unique id and the existing yield-and-atomic waiter pattern; use each instance's cfg(test) accessor rather than naming the static directly.
- [x] Delete tracked_lock_count and remove its assertion from assert_untouched; retain the no-request assertion for every pre-derivation refusal. Rename derivation_failures_take_no_lock_and_no_openbao_work, whose name states a lock claim the assertion left to it no longer makes.
- [x] Keep a_released_id_lock_reclaims_its_map_entry, a_contended_id_lock_is_shared_and_survives_the_first_release, and a_reclaimed_id_is_re_acquirable on independently created local IdLocks values.
- [x] Convert the ignored live-OpenBao mint/deregister test to separate instances, keeping both fixture directories alive, and retain its complete-or-absent final-state assertion.
- [x] Retain the two-instance claimant test's exactly-one-winner assertions and rewrite its stale CAS-path comment.
- [x] Run cargo clippy --all-targets -- -D warnings, the ignored registrar verb library tests through scripts/impl/run-registrar-verbs-e2e.sh, and the registrar-verbs Docker E2E matrix scenario.

## Dependencies

Part of #790.

## Pointers

- src/registrar/verbs.rs: module lock-order documentation, RegistrarVerbs::new, the pre_derivation item docs, IdLocks, and IdGuard
- src/registrar/verbs/tests.rs: assert_untouched, the three IdLocks reclamation tests, two_instances_racing_one_id_produce_exactly_one_claimant, and a_concurrent_mint_and_deregister_never_interleave
- Cargo.toml and .github/workflows/ci.yml: the pedantic lint setting and the clippy invocation
- src/openbao.rs: absent-only KV CAS response-classification tests
- scripts/impl/run-registrar-verbs-e2e.sh and scripts/preflight/ci/e2e-matrix.sh

AcoPiper:
Two `RegistrarVerbs` in one process derived the same `registration_id` and each took a per-id lock from its own map, so one instance's mint could interleave with another's deregister over the same binding and `OpenBao` material. The map is now a private module-owned `LazyLock<IdLocks>` and both verbs reach it through one private instance helper, `RegistrarVerbs::id_locks`. Neither `RegistrarVerbs` nor `RegistrarVerbsConfig` carries lock state any more, and sharing is not a caller convention.

The helper keeps its `&self` receiver deliberately: routing every instance's lock selection through it is what lets the new cross-instance test drive that path per service. It carries one local `#[allow(clippy::unused_self)]` with a comment stating that reason; the lint's associated-function suggestion is declined there and nowhere wider. A `cfg(test)` accessor, `id_locks_for_test`, delegates to the production helper and is not public.

The absent-only KV compare-and-set is untouched and remains the cross-process ownership primitive. The per-id Tokio mutex is still held across the binding and `OpenBao` awaits; the map's `std::sync::Mutex` is still held across none of them, and weak-entry reclamation is unchanged — an acquisition upgrades or inserts under the map guard, and a release removes only the identical dead `Weak`.

`entry_lock` is removed. Pre-derivation reads only immutable per-instance configuration and calls pure validators, so it had no shared mutable state to guard and the mutex only serialized unrelated requests. The module header and the `pre_derivation` docs now say that the stage is stateless and unsynchronized and that the shared per-id lock is the only in-process serialization boundary; a future entry-stage feature with shared state, such as the rate limiter owned by #788, synchronizes that state itself. Fallible derivation still completes before the per-id lock is acquired.

Closes #875

## Tests

- `two_instances_share_one_per_id_lock` (fast tier): holds a unique id through one service's accessor and proves acquisition through a second service stays pending until release, with the existing yield-and-atomic waiter. Verified to fail its blocking assertion under a mutant helper that returns a per-instance map.
- `tracked_lock_count` and its `assert_untouched` assertion are deleted — a count over a process-wide map is shared mutable state across the test binary. Every pre-derivation refusal still asserts no `OpenBao` request; `derivation_failures_take_no_lock_and_no_openbao_work` is renamed to `derivation_failures_reach_no_openbao_work` to match what it now asserts.
- The three `IdLocks` reclamation tests keep their locally constructed maps and their entry-creation, contention, cleanup and reacquisition assertions.
- `a_concurrent_mint_and_deregister_never_interleave` now drives the mint and the deregister through two separately constructed services, each with its own loaded fixture directory, both `TempDir` guards alive for the whole body; its complete-or-absent final-state assertion is unchanged.
- `two_instances_racing_one_id_produce_exactly_one_claimant` keeps its exactly-one-winner assertions, and its stale comment now records that with shared in-process locking the loser reads the persisted binding and is refused for the host collision, so the verb layer's `KvCreateIfAbsent::AlreadyExists` reread branch is only reachable from another process and is not covered by this suite. The canned-response tests in `src/openbao.rs` still draw the client-level CAS-mismatch distinction and are untouched.

## Test plan

- [x] A private module-owned `LazyLock` supplies the one weak per-id map; neither `RegistrarVerbs` nor `RegistrarVerbsConfig` owns lock state, and both verbs acquire through the single private instance helper.
- [x] The helper keeps its receiver under one local, commented `#[allow(clippy::unused_self)]`; the `cfg(test)` accessor delegates to it and is not public.
- [x] `two_instances_share_one_per_id_lock` uses a unique id and each instance's `cfg(test)` accessor rather than naming the static, and fails its blocking assertion when the helper returns a per-instance map.
- [x] `entry_lock` is gone, no entry-stage mutex or entry-serialization test remains, and the module header and `pre_derivation` docs state the stage is stateless and unsynchronized with the per-id lock as the only in-process boundary.
- [x] `tracked_lock_count` and its `assert_untouched` assertion are deleted; every pre-derivation refusal still asserts no `OpenBao` request.
- [x] `a_released_id_lock_reclaims_its_map_entry`, `a_contended_id_lock_is_shared_and_survives_the_first_release` and `a_reclaimed_id_is_re_acquirable` still run on locally constructed `IdLocks` values.
- [x] Acquisition upgrades or inserts under the map mutex, cleanup removes only the identical dead `Weak` under it, the map mutex crosses no `.await`, and the per-id Tokio mutex is still held through binding and `OpenBao` work.
- [x] The ignored live-`OpenBao` mint/deregister test uses two separately constructed services with two live fixture directories and keeps its complete-or-absent final-state assertion.
- [x] The two-instance claimant test keeps its exactly-one-winner assertions and documents that a cross-process `AlreadyExists` reread is not exercised here; `src/openbao.rs`'s canned-response CAS tests are unchanged.
- [x] `cargo clippy --all-targets -- -D warnings` — clean.
- [x] `scripts/preflight/ci/check.sh` — passes end to end, all nine steps including the `--strict` docs build and `cargo audit` (its only warning is the pre-existing allowed `RUSTSEC-2025-0134` on `rustls-pemfile`). `mkdocs` is not installed system-wide on this host, and without it on PATH the docs step aborts the script at exit 127 before `cargo audit` runs, so it was run with `mkdocs` supplied from a virtualenv. No documentation changed here either way.
- [x] `cargo test --lib` — 571 passed.
- [x] `scripts/impl/run-registrar-verbs-e2e.sh` against a live `OpenBao` — all 15 ignored registrar verb tests pass. That script is what `scripts/preflight/ci/e2e-matrix.sh` invokes for the `registrar-verbs` scenario.

### DIFF
--- a/src/registrar/verbs.rs
+++ b/src/registrar/verbs.rs
@@ -30,12 +30,12 @@
 //! Both verbs run the same three stages, and the order is a correctness
 //! requirement rather than a preference:
 //!
-//! 1. **Pre-derivation**, under a process-wide [`tokio::sync::Mutex`]
-//!    released before any `.await`: validate the two labels, refuse a
-//!    reserved `service_name`, resolve the component's multiplicity from
-//!    the rendered config, and check the instance shape. Mint also checks
-//!    the spec's identity restatement here. Every refusal on this arm
-//!    carries **no** `registration_id`, because none has been computed.
+//! 1. **Pre-derivation**, **stateless and unsynchronized**: validate the
+//!    two labels, refuse a reserved `service_name`, resolve the
+//!    component's multiplicity from the rendered config, and check the
+//!    instance shape. Mint also checks the spec's identity restatement
+//!    here. Every refusal on this arm carries **no** `registration_id`,
+//!    because none has been computed.
 //! 2. **Derivation** of the `registration_id` and the SAN. Fallible even
 //!    after the labels validated — the ≤131-octet bound is enforced on
 //!    the derived string — and a derivation failure takes no per-id lock
@@ -43,6 +43,17 @@
 //! 3. **Per-id work**, under the `registration_id`'s own
 //!    [`tokio::sync::Mutex`], shared by both verbs so a mint and a
 //!    deregister for one identity can never interleave.
+//!
+//! Stage 3's per-id lock is the **only** in-process serialization
+//! boundary either verb has. Its map is owned by this module rather than
+//! by a service instance, so two [`RegistrarVerbs`] in one process
+//! serialize against each other on one derived id instead of each
+//! holding a private lock for it. Stages 1 and 2 read only immutable
+//! per-instance configuration and call pure functions, so nothing there
+//! needs guarding and independent validations run concurrently; a future
+//! entry-stage feature that does introduce shared state — a rate limiter,
+//! say — synchronizes that state itself rather than reinstating a
+//! blanket entry lock over everything before derivation.
 //!
 //! Inside stage 3 the binding is consulted **before** the safe-set:
 //! a different stored host is a collision, a same-host different spec is
@@ -93,7 +104,7 @@ pub(crate) mod wrap_ttl;
 mod tests;
 
 use std::collections::HashMap;
-use std::sync::{Arc, Mutex as StdMutex, PoisonError, Weak};
+use std::sync::{Arc, LazyLock, Mutex as StdMutex, PoisonError, Weak};
 
 use anyhow::Context as _;
 use time::OffsetDateTime;
@@ -137,6 +148,19 @@ pub(crate) const REGISTRAR_TEARDOWN_KV_SUFFIXES: [&str; 5] = [
     SERVICE_SECRET_ID_KV_SUFFIX,
     SERVICE_REISSUE_KV_SUFFIX,
 ];
+
+/// The one per-`registration_id` lock map the whole process shares.
+///
+/// Ownership is deliberately the module's rather than a service
+/// instance's. Two [`RegistrarVerbs`] built over one `OpenBao` namespace
+/// — a re-rendered config, a second endpoint — derive the same
+/// `registration_id` from the same parts, and a per-instance map would
+/// hand each of them a private lock for it, letting one instance's mint
+/// interleave with another's deregister. The absent-only compare-and-set
+/// is the *cross-process* ownership primitive and stays exactly that; it
+/// cannot serialize a mint against a deregister inside one process, and
+/// this map cannot serialize anything outside it.
+static ID_LOCKS: LazyLock<IdLocks> = LazyLock::new(IdLocks::default);
 
 /// How many times a mint re-reads the binding after losing the
 /// compare-and-set race before giving up as unavailable.
@@ -215,11 +239,6 @@ pub(crate) struct RegistrarVerbs {
     token_ttl: String,
     secret_id_ttl: String,
     wrap_ttl_policy: WrapTtlPolicy,
-    /// Serializes the pre-derivation stage of every invocation. Held
-    /// across no `.await`.
-    entry_lock: TokioMutex<()>,
-    /// The per-`registration_id` locks.
-    id_locks: IdLocks,
 }
 
 impl RegistrarVerbs {
@@ -233,8 +252,6 @@ impl RegistrarVerbs {
             token_ttl: config.token_ttl,
             secret_id_ttl: config.secret_id_ttl,
             wrap_ttl_policy: config.wrap_ttl_policy,
-            entry_lock: TokioMutex::new(()),
-            id_locks: IdLocks::default(),
         }
     }
 
@@ -250,17 +267,15 @@ impl RegistrarVerbs {
             )
         };
 
-        // Stage 1. No `.await` inside the guard's scope.
-        let multiplicity = {
-            let _entry = self.entry_lock.lock().await;
-            self.pre_derivation(
+        // Stage 1. Stateless, so it runs unsynchronized.
+        let multiplicity = self
+            .pre_derivation(
                 &request.service_name,
                 &request.host,
                 request.instance,
                 Some(&request.spec),
             )
-        }
-        .map_err(|err| refuse(ProducingArm::PreDerivation, None, err))?;
+            .map_err(|err| refuse(ProducingArm::PreDerivation, None, err))?;
 
         // Still stage 1: the wrap TTL is a purely local property of the
         // request, checked under the construction-supplied policy well
@@ -291,7 +306,7 @@ impl RegistrarVerbs {
             .san_for(request.instance, &request.service_name, &request.host);
 
         // Stage 3.
-        let _id_guard = self.id_locks.acquire(&registration_id).await;
+        let _id_guard = self.id_locks().acquire(&registration_id).await;
         self.mint_locked(request, &request_id, &registration_id, &san, &granted)
             .await
     }
@@ -307,11 +322,9 @@ impl RegistrarVerbs {
             )
         };
 
-        let multiplicity = {
-            let _entry = self.entry_lock.lock().await;
-            self.pre_derivation(&request.service_name, &request.host, request.instance, None)
-        }
-        .map_err(|err| refuse(ProducingArm::PreDerivation, None, err))?;
+        let multiplicity = self
+            .pre_derivation(&request.service_name, &request.host, request.instance, None)
+            .map_err(|err| refuse(ProducingArm::PreDerivation, None, err))?;
 
         let registration_id = derive_registration_id(
             multiplicity,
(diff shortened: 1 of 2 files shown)


AcoPiper:
**[Reviewer Round 1]**

No findings.

The private module-owned `LazyLock<IdLocks>` is reached by both verbs through the same deliberately receiver-bearing `RegistrarVerbs::id_locks` helper, after fallible derivation. The test-only accessor delegates to that production path, and `two_instances_share_one_per_id_lock` holds a unique id through one independently constructed service while proving the second remains blocked; it would fail with the former per-instance maps.

The entry lock and global lock-count assertion are gone, while the local-map reclamation tests preserve the weak-entry lifecycle coverage. The live mint/deregister scenario also now crosses two separately constructed services and retains its complete-or-absent assertion. The PR body correctly closes #875 and includes a checked test plan.

AcoPiper:

PR #884, opened 2026-08-24
Record intent and outcome around both registrar verbs (#777)

### LINKED ISSUE #777: Write intent and outcome audit records around both registrar verbs
# Write intent and outcome audit records around both registrar verbs

## Context

Everything this work builds on is already in `origin/main`, and the work itself is one insertion pass over `src/registrar/verbs.rs`.

**The verb layer.** `RegistrarVerbs::mint` (`src/registrar/verbs.rs:426`) and `RegistrarVerbs::deregister` (`:481`) run three stages in a fixed order. Stages 1 and 2 — label validation, the reserved-name guard, the component lookup, the instance-shape check, the mint's spec-identity restatement, the mint's wrap-TTL grant, then derivation of the `registration_id` and the SAN — are **stateless and unsynchronized**, take no lock at all, touch no `OpenBao`, and every refusal they produce carries `registration_id: None`. Stage 3 does all durable work under the process-wide per-`registration_id` `tokio::sync::Mutex` held in `ID_LOCKS` (`:165`), which is the **only** in-process serialization boundary either verb has and which both verbs share, so a mint and a deregister for one identity can never interleave. Each invocation generates a `RequestId` at entry and carries an opaque `CallerIdentity` through unchanged. Every outcome is already classified: `MintKind::{FirstMint, IdempotentReMint}`, `DeregisterKind::{IdentityRemoved, AlreadyAbsent}`, and the `VerbError` enum for refusals, all in `src/registrar/verbs/outcome.rs`.

**The record store.** `src/registrar/audit.rs` supplies a daemon-owned, root-owned, append-only JSON Lines store. `AuditRecordStore::append` (`:1639`) takes one complete `AuditRecord`, serializes it as one newline-terminated object, appends it, and `sync_data`s it before returning; it holds a store-owned mutex for that operation only and returns a typed `AuditStoreError` for every way it can fail. `AuditRecord::intent` (`:499`) and `AuditRecord::outcome` (`:529`) are the two builders, and `pub(crate) mod bridge` (`:767`) already maps `MintKind`, `DeregisterKind` and every `VerbError` — flattening the whole `RegistrarError` and `WrapTtlRefusal` sets — onto `AuditOutcome`. `RegistrarVerbs` already **holds** the store: it is a fixed field on `RegistrarVerbsConfig` (`:212`) and `InternalVerbsSource` (`:261`), retained on the service (`:352`) and readable through `RegistrarVerbs::audit_store()` (`:420`). No construction, plumbing, schema or configuration work remains. What is deliberately absent is the only thing missing: **no verb arm calls `append`**, and `src/registrar/verbs/tests.rs:208` asserts exactly that.

**What this issue adds.** The record has to be written in two phases around the verb's own work, and an invocation whose record cannot be written has to fail in a way that never leaves `OpenBao` and the caller disagreeing about what exists. The two phases are not negotiable, because the outcome is only known *after* the verb has written the policy, the `AppRole`, the per-service KV and the durable `registration_id → host` binding. A single write-after-the-fact that failed at that point would leave the identity **already live in `OpenBao`** while the caller was told "refused" — and the compensating deregister would then fail the same way, leaving a live, unrecorded, undeletable identity.

The `OpenBao` file audit device stays mandatory and is not replaced. It cannot see a request refused before any `OpenBao` write, and it does not know who asked or for which `(service_name, host, instance)`; this trail does.

### The two writes

Write two records per invocation of either verb.

1. **Intent**, at the very top of `mint` and `deregister`, **before stage 1 and therefore before any `OpenBao` call**. Built with `AuditRecord::intent` from the invocation's `RequestId`, `AuditVerb::Mint` or `AuditVerb::Deregister`, the `CallerIdentity`, and a `RequestedIdentity { service_name, host, instance }` taken straight off the request.
2. The verb, unchanged.
3. **Outcome**, built with `AuditRecord::outcome` from the same `request_id`, the same verb, caller and `RequestedIdentity`, the derived `registration_id` where one exists, and the `AuditOutcome` that `bridge::mint_outcome`, `bridge::deregister_outcome` or `bridge::refusal_outcome` returns for the result.

The intent record carries **no** `registration_id` — the landed `AuditRecord::intent` builder does not take one and writes the key as absent, never as `null` or `""`. That settles placement: there is one intent write per invocation, at entry, identical for both verbs, and it needs no arm analysis at all. The requested parts are exactly `RequestedIdentity`'s three fields. The mint's `spec`, `delivery_mode` and `wrap_ttl` are **not** recorded: the format defines no field for them, and this issue consumes the format rather than extending it.

The verb layer passes the caller identity to the record builder **unchanged and uninterpreted** — whatever `CallerIdentity` holds, which is the socket's peer credentials today and may be a certificate identity later. What the line ends up carrying is then the store's business: `AuditRecord::into_bounded` caps `caller_identity` at `MAX_RECORD_FIELD_BYTES` and records a `truncated.caller_identity` digest when it had to. This issue neither pre-truncates nor works around that.

`bridge` currently carries an `#[allow(dead_code)]` naming this work as its first production caller. Remove it; all three of its functions become live.

### Where the outcome write goes

The outcome record for a result produced **under the per-`registration_id` guard must be written before that guard is released.** That guard is acquired in `mint`/`deregister` immediately before `mint_locked`/`deregister_locked` is called, so the write goes in the enclosing function while the guard is still alive — `mint_locked` and `deregister_locked` need no structural change beyond the disposition threading below. Releasing the guard first would let the next invocation for the same identity claim the binding, change `OpenBao`, and land its outcome line ahead of the line describing the change it just superseded.

Results produced *before* the guard — every stage 1 and stage 2 refusal — write their outcome record where they are produced, under no lock. **Intent lines are ordered only by arrival, and that is correct:** the intent write happens before stage 1, so two invocations for one identity can interleave their intent lines freely. The trail is paired by `request_id`, not by adjacency, and nothing here holds a lock across a whole invocation to make the file globally ordered.

### What a failed write returns

Add exactly two variants to `VerbError` (`src/registrar/verbs/outcome.rs:325`), as `thiserror` variants carrying the underlying `AuditStoreError` as their `#[source]`:

- **`AuditUnwritable { phase: AuditPhase, source: AuditStoreError }`** — a failed **intent** write, and a failed **outcome** write on an invocation that changed no `OpenBao` state.
- **`PostMintUnrecordable { source: AuditStoreError }`** — a failed **outcome** write on an invocation that did, or may have, changed `OpenBao` state.

`phase` is the landed `crate::registrar::audit::AuditPhase` (`src/registrar/audit.rs:214`), reused rather than restated. Do not introduce a second two-variant phase enum beside it: that one is `pub`, `Copy`, `Display`s as the string the format writes, and is documented as having deliberately no catch-all arm so a third phase breaks every downstream `match` at compile time — a private duplicate would opt this error type out of that. Its value is **not** inferable from anything else on the refusal and is fixed by the write that failed:

| Failed write | `phase` |
| --- | --- |
| the entry write, before stage 1 | `AuditPhase::Intent` |
| the outcome write, on an invocation whose disposition is clear | `AuditPhase::Outcome` |

`PostMintUnrecordable` carries no such field, and that is not an oversight: it can arise only from an outcome write, since the disposition cannot be set before the intent write has already succeeded.

Both are returned as an ordinary `VerbRefusal`, so they are individually matchable through `VerbRefusal::error()`. Neither is folded into `VerbError::Unavailable`; every rule below is a caller telling them apart.

**A failed intent write refuses the invocation with nothing created.** This is the point at which refusing is free: no `OpenBao` call has been made. Return `AuditUnwritable { phase: AuditPhase::Intent, .. }`, wrapped in a `VerbRefusal` whose `VerbContext` carries `ProducingArm::PreDerivation` and `registration_id: None` — both literally true, since nothing has been derived — and make no `OpenBao` call at all. The arm and the phase are different things and neither substitutes for the other: `ProducingArm` says how far the *verb* got and has no member that distinguishes an intent write from an outcome write, which is exactly the distinction a later endpoint needs.

**A failed outcome write may arrive after a state change has already happened**, so refusing there would report "refused" for something that exists. Which variant it takes is decided by the disposition rule below and by nothing else.

### Deciding whether the invocation changed `OpenBao` state

**Neither the verb nor the outcome class nor the `ProducingArm` is a usable discriminator.** `origin/main` makes that concrete in three places, each with a test already naming it:

- `claim_and_mint` writes a `creating` binding with the absent-only compare-and-set and only then calls `converge_and_issue`, which can refuse with `VerbError::Unavailable`. The claim is deliberately **retained** through that failure — it is what keeps whatever was created covered by a durable claim — so a *refusal* here follows a durable mutation (`a_failed_convergence_retains_its_creating_binding_and_recovers`, `src/registrar/verbs/tests.rs:1475`).
- `deregister_locked` runs the full material sweep **before** it looks at whether a binding exists, so a `DeregisterKind::AlreadyAbsent` can have deleted a role, a policy and five KV paths (`absent_binding_deregister_sweeps_planted_orphans`, `:1793`).
- A failed teardown refuses only after attempting every resource, so its `TeardownReport` can already show resources `Removed` (`a_failed_teardown_retains_the_binding_and_attempts_every_resource`, `:1701`).

`ProducingArm::Binding` is likewise ambiguous: it covers a read failure, a `RegistrationIdCollision` and a `StoredSpecConflict` — none of which write — as well as a failed claim and a failed binding activation, which do.

So track it explicitly. Carry a **per-invocation mutation disposition** — a monotone boolean created at verb entry, threaded through the locked helpers alongside the values they already take, never cleared — and set it from the *result of each `OpenBao` call that can change state*. It is shared behind `&` across `.await` points in a `Send` future, so an `AtomicBool` is the shape to use; the value is per-invocation, so no ordering stronger than the module already uses is needed:

| Call site | Sets the disposition when |
| --- | --- |
| `claim_binding` → `create_kv_if_absent` (`:1074`) | `Ok(KvCreateIfAbsent::Created)`, or `Err` |
| `write_binding` → `write_kv`, the binding activation in `converge_and_issue` (`:1097`) | `Ok`, or `Err` |
| `provision_service_role`, in `converge_and_issue` (`:791`) | `Ok`, or `Err` |
| `create_secret_id_wrap_only`, in `issue` (`:866`) | `Ok`, or `Err` |
| `teardown_service_material`, in `deregister_locked` (`:930`) | any `TeardownReport` attempt is `ResourceOutcome::Removed` or `ResourceOutcome::Failed` |
| `delete_binding` → `delete_kv` (`:1121`) | `Ok`, or `Err` |

Nothing else sets it. Reads never do: `read_binding`/`try_read_kv`, `read_role_id`, and acquiring an authenticated client all leave it alone. `Ok(KvCreateIfAbsent::AlreadyExists)` leaves it alone too — a lost compare-and-set wrote nothing — which is also why exhausting `MAX_CLAIM_ATTEMPTS` refuses with the disposition still clear.

**Every error sets it, and that direction is the whole point.** A call that failed may have failed after the write reached `OpenBao`, and the two ways of being wrong are not symmetrical: reporting a change as an ordinary refusal leaves live, unrecorded state that nobody is told to clean up, while reporting an unchanged invocation as owed costs at most one idempotent re-drive that finds nothing. A `TeardownReport` entry of `Failed` is indeterminate in exactly the same way and counts the same.

The rule then has one line: **a failed outcome write returns `PostMintUnrecordable` when the disposition is set, and `AuditUnwritable` when it is clear.** The verb, the outcome class and the arm do not enter into it.

Two consequences are worth stating because they are the ones a plausible implementation gets wrong in opposite directions:

- A `StoredSpecConflict` — the identity exists, is bound to *this* host, and was minted with a different spec — is decided from a read and **must not** arm a teardown obligation. A caller told a teardown was owed there would deregister a live identity that this invocation did not create and that is still in use. Same for `RegistrationIdCollision` and for every stage 1 and stage 2 refusal.
- An `AlreadyAbsent` deregister that actually swept something **must** arm it. It is the only trace that the orphan is gone.

There is no mint rule and no deregister rule here. Which verb ran never enters the decision — only whether this invocation touched `OpenBao` in a way that can have changed it.

### The names, and where they diverge from the wire contract

`AuditUnwritable` and `PostMintUnrecordable` are the spellings `docs/reference/registrar-wire-contract.md` §6.5 records for the closed `RegistrarUnavailable` reason set, so adopt them rather than coining new names. **They are internal verb-layer variants here and nothing more.** `origin/main` has no verb-to-wire mapping — `src/registrar/endpoint/handler.rs` hands opaque bytes across and `src/registrar/endpoint/refusal.rs` is transport-level only — and building one is not this issue's work.

Both names are narrower than the conditions above, deliberately; record the divergence, do not resolve it here. That reason set describes `AuditUnwritable` as "intent phase only" and `PostMintUnrecordable` as following "a successful mint", and it is closed, with no member for a completed removal or for a refusal that followed a durable write. Narrowing the internal conditions to fit the names would leave exactly the state changes enumerated above unrecorded and unreported, which is the failure the rule exists to prevent. Carrying `phase` on `AuditUnwritable` is what lets a later endpoint map `AuditPhase::Intent` onto the transcribed reason, which is defined for exactly that case, and decide on its own terms what to tell a caller whose `AuditPhase::Outcome` write failed without changing anything.

**Neither variant is ever recorded**, by construction: the record that would carry one is the record whose write just failed. Enforce that in the type system rather than in prose — change `bridge::refusal_outcome` to return `Option<AuditOutcome>`, returning `None` for exactly these two variants and `Some` for every other. Where a call site meets `None`, write no outcome record and return the refusal unchanged; do not `expect()` or `unreachable!()` on it. `RefusalReason` gains **no** new member, so the store's exhaustive reason table is untouched and no golden fixture is demanded for a value no invocation can produce. The alternative — mapping both onto `RefusalReason::Unavailable` to keep the signature — was rejected: it breaks the documented one-value-per-source-variant property of that enum and puts a durable reason on a condition that can never reach the file.

**A successful mint whose outcome write fails does not return its material.** `MintResult` is `Result<MintOutcome, VerbRefusal>`, so the freshly issued wrapped `secret_id` is dropped and expires unused. That is the right trade: the caller is told the identity exists and a teardown is owed, and re-driving the mint after the store recovers issues fresh material. A deregister whose outcome write fails after a sweep attaches its `TeardownReport` through the existing `VerbRefusal::with_teardown`, so the caller still sees what was deleted.

Log every store failure through `tracing` with its `AuditStoreError` before returning, since the variant a caller eventually sees may be coarser than the store's own error.

### Every `append` error is a write failure for its phase

`AuditRecordStore::append` returns exactly one typed error per way it can fail, and **all of them are write failures for the phase they occurred in**. What the invocation returns is decided by the phase and the disposition rule above, never by which store error it was. This includes `Sync`, which means the line was written but is not durable, and `DirectorySync`, which means the record is written and flushed but the directory entry naming its file may not survive a power failure.

There is nothing to decide about a failed *trim*: the store logs a trim failure and never surfaces it (`trim_logging_failures`), so no `append` a verb makes can be refused by one. Likewise a rotation error can only arise from a rotation the pending record actually required, because the store rotates only when the record does not fit.

**There is no fallback record path to build.** A write that fails is never downgraded to best-effort, buffered, deferred, retried into a second file, or substituted with a log line. Logging the failure through `tracing` is right; treating that log line as the record is not. The two error variants *are* the handling.

### What survives on disk

The invariant is over write **attempts**, not over lines that reach disk. Every invocation attempts the intent write before any `OpenBao` call, and attempts the outcome write on every path where the intent write succeeded and the verb then ran, to success and to refusal alike. Three shapes follow:

1. both writes succeed — a complete intent/outcome pair, which is what a writable store always produces;
2. the intent write succeeded and the **outcome** write was attempted and failed — a durable intent, plus whatever that failed attempt happened to leave behind, while the caller gets `PostMintUnrecordable` or `AuditUnwritable` by the disposition rule. A durable **intent with no outcome** is one form this takes, and where it occurs it is not a gap but exactly the anomaly a trail reader is meant to surface. It is not, however, what a failed outcome write guarantees: some failures leave the outcome line complete and readable, and this issue neither promises nor removes it;
3. the **intent** write failed — the invocation changed nothing in `OpenBao`, so there is no state for the trail to be silent about.

The verb layer performs **no** compensation on the file: no truncation, no repair, no second attempt. Residue after a failed write is the store's business, is deliberately not specified here, and differs by failure mode: a write that fails before any byte reaches the file leaves no line at all; a partial write is truncated back by the store's own recovery; and a `Sync` failure leaves a **complete, newline-terminated line already in the file**, whose survival across a crash is unconfirmed and which the store does not take back — nor may this issue ask it to. What this issue guarantees and tests is the returned variant and the `OpenBao` state, never the physical presence or absence of the record whose write failed. The one file-level property that does hold across every mode is JSONL integrity: no torn line is left behind except where the store's own recovery also failed, which it reports as `PartialAppend`.

### Records only for invocations that reach the verb

State the per-invocation rule as "every invocation that reaches the verb body produces both write attempts". Today that is every invocation. A future admission control that rejects an invocation before it reaches the verb — the token-bucket rate limiting tracked separately, which counts what it suppresses instead of recording it — then lands as an additive change rather than as a regression against this issue's first acceptance criterion. Do not write an unconditional "one pair per request that arrived" invariant that such work would have to weaken, and do not encode any assumption about what a suppressed caller is told.

## Acceptance criteria

- [x] Every invocation of either verb, accepted and refused alike, attempts an intent write before any `OpenBao` call and an outcome write after the verb ran; against a writable store this produces an intent line and a matching outcome line sharing one `request_id` and carrying the timestamp, the verb, the requested `(service_name, host, instance)`, the caller identity, and — on the outcome line, where derivation produced one — the `registration_id`.
- [x] Every intent line omits the `registration_id` key entirely: it is absent from the JSON object, not present as `null` or `""`.
- [x] Refusals produced before the per-id lock is taken — a non-DNS-label `service_name`, a reserved `bootroot-` name, a component with no multiplicity entry, an instance-shape mismatch, a mint spec-identity disagreement, an unusable `wrap_ttl`, and a derived key that exceeds its length bound (`RegistrarError::DerivedKeyInvalid`) — each produce a complete intent/outcome pair, with the outcome line likewise carrying no `registration_id`.
- [x] A `RegistrationIdCollision` still produces a complete pair. It is decided by *reading* the binding, performs no `OpenBao` write, and so is invisible to the `OpenBao` audit device.
- [x] With the store failing the **intent** write, the invocation is refused with `VerbError::AuditUnwritable { phase: AuditPhase::Intent, .. }` and the mock `OpenBao` server received **no** request at all. This is the direct test that the intent write precedes every `OpenBao` write; an implementation that writes one record after the fact and refuses on failure fails it.
- [x] The mutation disposition is set from exactly the six call sites listed in Scope and from nothing else, is never cleared, and is set by **every** error return from each of them. A reviewer can confirm no read path sets it and that `Ok(KvCreateIfAbsent::AlreadyExists)` and an all-`AlreadyAbsent` `TeardownReport` leave it clear.
- [x] **The mint discriminating pair.** A failed outcome write on a mint refused with `StoredSpecConflict` returns `AuditUnwritable { phase: AuditPhase::Outcome, .. }`, and a failed outcome write on a mint refused by a provisioning failure that followed a successful binding claim returns `PostMintUnrecordable`, with the `creating` binding still in `OpenBao`. Both are refusals; an implementation that classifies from success-versus-refusal fails one of them.
- [x] **The deregister discriminating pair.** A failed outcome write on an `AlreadyAbsent` deregister that swept planted orphan material returns `PostMintUnrecordable`, while a failed outcome write on an `AlreadyAbsent` deregister with nothing at all to sweep returns `AuditUnwritable { phase: AuditPhase::Outcome, .. }`. Both are the same outcome class; an implementation that classifies from the outcome class fails one of them.
- [x] A failed outcome write after a `FirstMint`, an `IdempotentReMint`, or a deregister that removed a bound identity returns `PostMintUnrecordable`, and the corresponding `OpenBao` state is present or gone as the verb left it. Re-driving the deregister once the store recovers succeeds and writes its pair.
- [x] A failed outcome write on a deregister refused with `HostMismatch` — decided from a read, before the sweep — returns `AuditUnwritable { phase: AuditPhase::Outcome, .. }`.
- [x] The outcome record for a result produced under the per-`registration_id` guard is written before that guard is released, and no verb-layer lock is added: `ID_LOCKS` remains the only one, the store's own mutex remains innermost, and stages 1 and 2 remain unsynchronized.
- [x] The record carries the verb layer's own refusal variant: a mint refused with `ServiceSpecOutsideSafeSet`, a same-host re-mint refused with `StoredSpecConflict`, and a mint refused for a spec-identity disagreement produce **three different** `outcome.reason` values in the trail.
- [x] The verb layer hands the caller identity to the record builder unchanged and uninterpreted, with no stub value and no waiting on mTLS. An in-bound distinctive value reaches `caller_identity` unchanged; an over-long one is capped by the store's existing `into_bounded` with a `truncated.caller_identity` digest, and this issue neither pre-truncates nor circumvents that.
- [x] `AuditUnwritable` carries both `AuditPhase` values across the suite and never the wrong one: a test matches `AuditPhase::Intent` on an entry-write failure and `AuditPhase::Outcome` on a non-mutating outcome-write failure, and neither assertion passes against a variant that hard-codes one value or omits the field. `PostMintUnrecordable` carries no phase field at all.
- [x] `VerbError::AuditUnwritable` and `VerbError::PostMintUnrecordable` exist by name, are individually matchable through `VerbRefusal::error()`, and can never reach the trail: `bridge::refusal_outcome` returns `Option<AuditOutcome>` and yields `None` for exactly these two, and `RefusalReason` has gained no member.
- [x] No fallback path was added on either failure path: a reviewer can confirm there is no buffer, queue, retry, secondary file, truncation, repair or log-as-record substitute, and that `tracing` carries the store error without standing in for the record.
- [x] The verbs' decision procedure is unchanged: the three stages, the single lock, the request id, the caller-identity parameter and the refusal taxonomy are as before, with only record writes, the disposition threading and the two new error variants inserted.
- [x] `src/registrar/verbs/tests.rs`'s `the_verbs_hold_an_audit_store_and_write_nothing_to_it` and its `assert_no_audit_records` helper are gone, its four other call sites — in `first_mint_creates_exactly_the_derived_role_and_policy` (`:1165`), `same_host_rematch_reuses_the_role_and_returns_fresh_material` (`:1243`), `safe_set_refusal_and_stored_spec_conflict_are_distinct_no_change_outcomes` (`:1399`) and `absent_binding_deregister_sweeps_planted_orphans` (`:1793`) — are inverted to assert the pair each of those invocations must now produce, and the field and module comments in `src/registrar/verbs.rs` and `src/registrar/audit.rs` that say no verb arm appends — including `bridge`'s `#[allow(dead_code)]` — are updated or removed.
- [x] `cargo clippy --all-targets` is warning-free with `-D warnings`, `cargo fmt -- --config group_imports=StdExternalCrate --check` passes, and `cargo test` passes.
- [x] The live-`OpenBao` verb suite passes: `scripts/impl/run-registrar-verbs-e2e.sh` runs green, with all sixteen `#[ignore]`d `registrar::verbs` tests passing now that every one of them writes audit records for the first time.
- [ ] `scripts/preflight/run-all.sh` was run before pushing, and its `scripts/preflight/ci/e2e-matrix.sh` arm — which is where the registrar-verbs scenario lives — is not skipped. The E2E carve-out in `AGENTS.md` does not apply: this change alters registrar verb lifecycle paths that the matrix exercises directly.

## Constraints

- **No invocation that reaches the verb body may *skip* a write.** Skipping and failing are different and only the first is forbidden: an attempted write that fails is handled by the phase and disposition rules and surfaces as an error return. Add no path on which a write is silently omitted, and no best-effort path that lets one fail quietly.
- **Introduce no admission-control or limiter plumbing.** No limiter type, trait, no-op or always-admit implementation, configuration key, call site, or `admits`/`suppressed` abstraction the tests route through. A placeholder built here is a second implementation the work that owns rate limiting would have to reconcile with its own. This issue's code refers to no limiter at all, and its tests assert records over the invocations they drive directly.
- **Do not collapse the two phases into one write**, and do not special-case either verb or any arm into a single record. One record shape and one pairing rule for the whole trail is what keeps a reader from carrying an exception.
- **Do not infer the disposition.** It is not read off the verb, the outcome class, the `ProducingArm`, the success-versus-refusal split, or a re-read of `OpenBao`. It is set from the six enumerated call results and nothing else. Re-reading `OpenBao` to find out what happened is specifically excluded: it is racy under the per-id lock's cross-process limits, and it needs the store to be working to be worth anything.
- **Do not restructure the verbs' decision procedure.** The three stages, their order, the single per-id lock, the request id, the caller-identity parameter and the refusal taxonomy stay as they are. Turning a `?` or a `.map_err(...)` closure into an explicit `match` so a record write or a disposition update can be inserted around it **is** expected and is not restructuring; so is threading one extra per-invocation parameter through the locked helpers. Reordering a check, moving work across the lock boundary, or changing which refusal a path produces is. If a write cannot be placed without one of those, raise it as a finding rather than making the change.
- **Add no verb-layer lock**, and do not hold the per-id guard across anything it does not already cover. The store owns its own innermost mutex and that is the only lock this work relies on.
- **Invent no wire identifier and no `RegistrarUnavailable` reason spelling.** The two variants are internal, and mapping them onto a caller-facing contract belongs at a serialization boundary that does not exist yet.
- **Do not change the record format or the store's behaviour**: not the schema, the field byte caps, the JSON Lines encoding, per-record durability, rotation, retention, the `[registrar]` configuration keys, or the filesystem trust checks. This issue consumes all of them. Changing `bridge::refusal_outcome`'s signature is the one permitted edit inside `src/registrar/audit.rs`, together with the comment and test updates it forces.
- If the fault-injection harness needs extending to fail a *specific* append rather than all of them, keep the extension `#[cfg(test)]`, in the existing `FaultInjection`, and add no production branch. The existing `AppendGate` already makes this possible without any change: it parks each blocking append at its entry and notifies the test, so a fault can be armed between the intent append and the outcome append.
- No `unwrap()` in production code; no `[]` indexing; `thiserror` for the new variants; `tempfile::tempdir()` for every test store; no fixed paths, no network, no `sleep`-based synchronization, and no process-environment mutation in tests.
- No `CHANGELOG.md` entry. The registrar verbs are unreachable from any production code path in the last release and remain so after this change, so nothing here is observable by a user running it.

## Out of scope

- **The record store itself** — the file, the encoding, the `serde` types, the field bounds and their digests, the ownership and symlink checks, the configuration keys, per-record durability, and the rotation and retention bounds. All of it is already in `origin/main` and is consumed unchanged.
- **Reading the trail back.** The intent-without-outcome scan and its settle grace, the malformed-line taxonomy, the retention-shortfall flag and any `bootroot status` audit section belong to the separate anomaly-reporting work (background: #779).
- **Rate limiting** both verbs, coalescing repeated refusals, and reporting suppressed traffic (background: #786, #788, #787). Nothing here anticipates their mechanism beyond stating that this issue's rule covers invocations that reach the verb body.
- **The reserved/quota'd audit store** and its low-water alarm (background: #772–#775).
- Wiring the verbs or the store into `run_daemon`, or into any registrar surface provisioner. Nothing in `src/daemon.rs` constructs `RegistrarVerbs` today and nothing here changes that.
- Any endpoint work: the verb-to-wire mapping, the `RegistrarUnavailable` reason encoding, and authenticating the caller identity.
- Adding a `ProducingArm` variant, a `RefusalReason` variant, or any new refusal to the verbs' taxonomy.
- Changing what either verb does to `OpenBao`. The accepted hazard that an absent-binding deregister still sweeps the full material set stays exactly as it is; this issue only makes sure the sweep is recorded and classified honestly.

## Test plan

Unit tests go in `src/registrar/verbs/tests.rs`, driven against the existing `wiremock` `MockServer` and a store opened over a `tempfile::tempdir()`. `AppendGate` is how a fault is armed between the two appends: install it, drive the verb from a spawned task, await the first entry notification, release it, await the second, arm the fault, then release.

- [x] One complete intent/outcome pair per outcome class: `first_mint`, `idempotent_remint`, `identity_removed`, `idempotent_already_absent`, and a refusal. Each asserts the two lines share one `request_id`, that the intent line has no `registration_id` key, and that the outcome line's `registration_id` is present exactly when derivation produced one.
- [x] Pre-lock refusal coverage: a non-DNS-label `service_name`, a reserved `bootroot-` name, an unconfigured component, an instance-shape mismatch, a mint spec-identity disagreement, a zero `wrap_ttl`, and an over-long derived key each produce a pair with no `registration_id` on either line.
- [x] A `RegistrationIdCollision` produces a complete pair, and the mock server shows the binding read and no write.
- [x] Intent-failure test: with the store's pre-append `append` fault armed from the start, drive a mint and a deregister that would otherwise succeed; assert `VerbError::AuditUnwritable` with `AuditPhase::Intent` and that the existing `assert_untouched(&server)` helper passes. The trail file is then empty — again a property **of this fault mode**, not of failed intent writes in general.
- [x] **Mint discriminating-pair test, one test over the same injected outcome-phase failure.** Drive a `StoredSpecConflict` refusal and a mint whose provisioning fails after the claim landed; assert `AuditUnwritable` with `AuditPhase::Outcome` for the first and `PostMintUnrecordable` for the second, and assert the second left its `creating` binding in the mock `OpenBao`. Asserting them together is what catches an implementation that classifies from success-versus-refusal.
- [x] **Deregister discriminating-pair test, one test over the same injected outcome-phase failure.** Drive an `AlreadyAbsent` deregister whose mock reports material present and deleted, and one whose mock reports every resource already absent; assert `PostMintUnrecordable` for the first and `AuditUnwritable` with `AuditPhase::Outcome` for the second. Same outcome class, opposite variants — this is what catches an implementation that classifies from the outcome class.
- [x] Outcome-failure after a successful mint, with the pre-append `append` fault armed for the outcome write: assert `PostMintUnrecordable` and assert the binding and the role are in the mock `OpenBao`. The trail then holds the intent line and no outcome line for that `request_id` — assert that as a property **of this fault mode**, which fails before any byte reaches the file, and not as a general consequence of a failed outcome write.
- [x] Outcome-failure after a deregister that removed a bound identity: assert `PostMintUnrecordable`, assert the material and the binding are gone, assert the refusal carries a `TeardownReport`, then clear the fault, re-drive the deregister, and assert it succeeds and writes its own pair.
- [x] Outcome-failure on a `HostMismatch` deregister and on a pre-derivation refusal: assert `AuditUnwritable` with `AuditPhase::Outcome`, and assert `PostMintUnrecordable` is not returned. Pairing this with the intent-failure test above is what pins both phase values, since a variant that hard-codes `Intent` satisfies one and a variant that hard-codes `Outcome` satisfies the other.
- [x] Outcome-failure after a failed teardown whose report shows at least one resource `Removed` or `Failed`: assert `PostMintUnrecordable`.
- [x] `sync_data` failure: with the `sync` fault armed, assert an intent-phase failure refuses with `AuditUnwritable` having touched no `OpenBao`, and that an outcome-phase failure after a successful mint returns `PostMintUnrecordable` with the mint's `OpenBao` state present. Assert the returned variant, the `OpenBao` state, and that every line in the trail parses as a complete record. Do **not** assert that the unflushed line is absent: `sync` fails after the complete newline-terminated line has already reached the file, so it may well still be readable there.
- [x] Rotation failure: pre-fill the active file to just under `max_file_bytes` so the next record genuinely requires a rotation, arm the `rotate` fault, and assert the phase's write fails and the disposition rule decides the returned variant.
- [x] Concurrency property test: drive several concurrent mints for one identity against a writable store; assert every invocation produced a correctly paired intent/outcome, that there is exactly one `first_mint` and the rest are `idempotent_remint`, and that the `first_mint` outcome line precedes every `idempotent_remint` outcome line in the file. That ordering is what writing the outcome inside the per-id guard buys. It is a property test rather than a discriminator — an implementation that writes outside the guard races and may still pass — so the in-guard placement is also an explicit code-review criterion above.
- [x] Distinct-reason test: `ServiceSpecOutsideSafeSet`, `StoredSpecConflict` and `SpecIdentityDisagreement` produce three different `outcome.reason` values.
- [x] Caller-identity test: a distinctive value **within** `MAX_RECORD_FIELD_BYTES` reaches `caller_identity` unchanged. Do not assert byte-for-byte survival of an over-long value; the store's bounding owns that and already tests it.
- [x] In `src/registrar/audit/tests.rs`, a test asserting `bridge::refusal_outcome` returns `None` for `AuditUnwritable` and `PostMintUnrecordable` and `Some` for every entry in `every_refusal()`, with that helper's contract comment amended to say the two audit-failure variants are covered here rather than in its table.
- [x] The four live `#[ignore]`d tests that currently call `assert_no_audit_records` are inverted to assert the pair their invocation now produces, and the remaining eleven still pass unchanged. With the concurrency property test above, which is new, that makes sixteen.
- [x] `scripts/impl/run-registrar-verbs-e2e.sh` green. It stands up a single `openbao server -dev` container on a port it allocates and runs the sixteen ignored `registrar::verbs` tests; it is the `registrar-verbs` step of `scripts/preflight/ci/e2e-matrix.sh` and can be run standalone with an `ARTIFACT_DIR` when the whole matrix cannot be driven.
- [ ] `scripts/preflight/run-all.sh` green before pushing.

## Dependencies

Part of #778.

Nothing blocks this. The record store, its `serde` types, its typed write errors, its test-only fault injection, and the `RegistrarVerbs` seam that holds it are all in `origin/main` already (background: #776, landed as PR #879), as are the three stages, the per-`registration_id` lock, the request id, the caller identity and the outcome classification.

## Pointers

- `src/registrar/verbs.rs:426` (`mint`) and `:481` (`deregister`) — where the intent write goes, and where the per-id guard is acquired and the outcome write must happen before it drops
- `src/registrar/verbs.rs:165` (`ID_LOCKS`), `:524` (`pre_derivation`), `:548` (`mint_locked`), `:609` (`claim_and_mint`), `:658` (`mint_against_binding`), `:791` (`converge_and_issue`), `:866` (`issue`), `:930` (`deregister_locked`), `:1074` (`claim_binding`), `:1097` (`write_binding`), `:1121` (`delete_binding`), `:420` (`audit_store`)
- `src/service_material.rs:263` (`TeardownReport`, `ResourceOutcome::{Removed, AlreadyAbsent, Failed}`) and `:311` (`teardown_service_material`)
- `src/registrar/verbs/outcome.rs:325` (`VerbError` — where the two variants go), `:89` (`ProducingArm`), `:424` (`VerbRefusal`, including `with_teardown`)
- `src/registrar/audit.rs:214` (`AuditPhase` — the enum the new error variant reuses), `:499` / `:529` (`AuditRecord::intent` / `::outcome`), `:560` (`into_bounded`), `:767` (`bridge`), `:1639` (`append`), `:918` (`AuditStoreError`), `:1242` (`FaultInjection`), `:1279` (`AppendGate`), `:1440` (`open_for_tests_with_faults`)
- `src/registrar/verbs/tests.rs:208` (the "writes nothing" test to replace), `:289` (`assert_no_audit_records`), `:324` (`assert_untouched`), `:154` / `:165` (the request builders), `:306` (`mock_client`), `:1475` / `:1701` / `:1793` (the three live tests that document the mutation cases above)
- `src/registrar/audit/tests.rs:419` (`every_refusal`)
- `scripts/impl/run-registrar-verbs-e2e.sh`, `scripts/preflight/ci/e2e-matrix.sh`, `scripts/preflight/run-all.sh`
- `docs/reference/registrar-wire-contract.md` §6.5 — the closed `RegistrarUnavailable` reason set the two variant names come from
- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §5.6 and §6

AcoPiper:
## What this does

Both registrar verbs now write two audit records per invocation into the daemon-owned store that has been sitting on `RegistrarVerbs` unused.

The **intent** line goes in at the very top of `mint` and `deregister`, before stage 1 and therefore before any `OpenBao` call. The **outcome** line goes in while the per-`registration_id` guard is still held, so the next invocation for that identity cannot change `OpenBao` and land its outcome line ahead of the line describing the change it superseded. The two are paired by `request_id`; intent lines are ordered only by arrival, which is correct because nothing holds a lock across a whole invocation.

Neither write is best-effort, so each has to say what a failure returns. `VerbError` gains two variants, both ordinary `VerbRefusal`s and both individually matchable:

- `AuditUnwritable { phase: AuditPhase, source: AuditStoreError }` — a failed intent write, and a failed outcome write on an invocation that changed no `OpenBao` state. A failed intent write refuses with nothing created and no `OpenBao` request made at all.
- `PostMintUnrecordable { source: AuditStoreError }` — a failed outcome write on an invocation that did, or may have, changed `OpenBao` state. A successful mint that lands here does not return its material; the wrapped `secret_id` is dropped and expires unused.

No fallback path was added on either: no buffer, queue, retry, secondary file, truncation, repair, or log-as-record substitute. The `tracing` event carries the store's own error and stands in for nothing.

## Why a separate mutation disposition

Nothing already on the refusal says whether the invocation touched `OpenBao`. A refusal can follow a durable claim (`claim_and_mint` retains its `creating` binding through a failed convergence), an `AlreadyAbsent` deregister can have swept a full planted orphan, a failed teardown's report can already show resources `Removed`, and `ProducingArm::Binding` covers reads and writes alike.

So the invocation carries an explicit monotone `AtomicBool`, threaded through the locked helpers, set from the result of exactly the six calls that can change state: `create_kv_if_absent` (only on `Created`), `write_kv`, `provision_service_role`, `create_secret_id_wrap_only`, a teardown attempt that is `Removed` or `Failed`, and `delete_kv`. Reads never set it, and a lost compare-and-set never does — nothing reached `OpenBao`. Every *error* does, because a call that failed may have failed after its write landed: reporting a change as an ordinary refusal leaves live state nobody is told to clean up, while the converse costs one idempotent re-drive that finds nothing.

## `refusal_outcome` returns `Option`

Neither new variant can ever reach the trail, since the record that would carry one is the record whose write just failed. That is now a type error rather than a convention: `bridge::refusal_outcome` returns `Option<AuditOutcome>` and yields `None` for exactly those two. `RefusalReason` gains no member, so the store's exhaustive reason table and its golden fixtures are untouched. `bridge`'s `#[allow(dead_code)]` is gone — all three functions are live.

## Names, and where they diverge from the wire contract

`AuditUnwritable` and `PostMintUnrecordable` are the spellings `docs/reference/registrar-wire-contract.md` §6.5 records for the closed `RegistrarUnavailable` reason set, adopted rather than coined. Both are internal verb-layer variants and nothing more — there is no verb-to-wire mapping in this repository and none is built here. Both names are narrower than the conditions above (that set says "intent phase only" and "after a successful mint"); the divergence is recorded in the variants' rustdoc and deliberately not resolved, because narrowing the conditions to fit the names would leave exactly the state changes above unreported.

## Tests

Fast tier, against `wiremock` and a `tempfile::tempdir()` store:

- a complete pair, with no `registration_id` on either line, for every pre-lock refusal through both verbs, plus the two mint-only ones and a `RegistrationIdCollision` (which is decided from a read and so is invisible to the `OpenBao` audit device)
- **intent-failure**: `AuditUnwritable { phase: Intent }` with `assert_untouched(&server)` passing — the direct test that the intent write precedes every `OpenBao` write
- **the mint discriminating pair**, over one injected outcome-phase failure: a `StoredSpecConflict` gives `AuditUnwritable { phase: Outcome }` and a provisioning failure after a landed claim gives `PostMintUnrecordable`, with its `creating` binding still there. Both are refusals, so an implementation classifying from success-versus-refusal fails one
- **the deregister discriminating pair**, likewise: an `AlreadyAbsent` that swept material gives `PostMintUnrecordable`, one with nothing to sweep gives `AuditUnwritable { phase: Outcome }`. Same outcome class, opposite variants
- the two paths where the outcome write succeeds and the caller is served — a recorded mint returning its material on both arms, and a recorded already-absent deregister — so a regression on the success arm does not hide behind the fault-injecting tests
- outcome-failure after a first mint, after an idempotent re-mint, after an identity removal (which then re-drives cleanly once the store recovers), after a failed teardown, on a `HostMismatch`, on a pre-derivation refusal, and on a claim race lost on every attempt
- `sync` and `rotate` failures, each driven in both phases
- three distinct `outcome.reason` values for `ServiceSpecOutsideSafeSet`, `StoredSpecConflict` and `SpecIdentityDisagreement`
- the caller identity reaching `caller_identity` unchanged
- in `src/registrar/audit/tests.rs`, `refusal_outcome` returning `None` for exactly the two audit-failure variants and `Some` for every entry in `every_refusal()`

The fault for one specific append is armed through the existing `AppendGate`, which parks each blocking append at its entry: the intent append is released before the fault is set, and the outcome append is already parked when it is. No production branch was added.

In the ignored tier, `the_verbs_hold_an_audit_store_and_write_nothing_to_it` and `assert_no_audit_records` are gone and their four call sites are inverted to assert the pairs their invocations now produce. One test is added — concurrent mints for one identity, asserting every invocation is correctly paired, that exactly one is a `first_mint`, and that its outcome line precedes every `idempotent_remint` outcome line, which is what the in-guard placement buys. That takes the ignored `registrar::verbs` suite to sixteen.

No `CHANGELOG.md` entry: the registrar verbs are unreachable from any production code path in the last release and remain so after this change.

Closes #777

## Test plan

- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate` clean
- [x] `cargo clippy --all-targets -- -D warnings` clean (the crate declares no features, so this is the whole configuration)
- [x] `cargo test` green — 745 lib tests plus every integration binary, 0 failed
- [x] `scripts/impl/run-registrar-verbs-e2e.sh` green: all sixteen ignored `registrar::verbs` tests pass, every one of them now writing audit records
- [x] Every invocation of either verb attempts an intent write before any `OpenBao` call and an outcome write after the verb ran; against a writable store the pair shares one `request_id`
- [x] Intent lines omit the `registration_id` key entirely — absent from the JSON object, not `null` and not `""`
- [x] Every pre-lock refusal (non-DNS-label name, reserved `bootroot-` name, unconfigured component, instance-shape mismatch, spec-identity disagreement, unusable `wrap_ttl`, over-long derived key) produces a pair with no `registration_id` on either line
- [x] A `RegistrationIdCollision` produces a complete pair, with the mock showing the binding read and no write
- [x] An intent-write failure returns `AuditUnwritable { phase: AuditPhase::Intent }` and `assert_untouched(&server)` passes
- [x] The disposition is set from exactly the six enumerated call sites, never cleared, set by every error return; `Ok(KvCreateIfAbsent::AlreadyExists)` and an all-`AlreadyAbsent` teardown leave it clear
- [x] Mint discriminating pair: `StoredSpecConflict` → `AuditUnwritable { phase: Outcome }`; provisioning failure after a landed claim → `PostMintUnrecordable`, with the `creating` binding still in `OpenBao`
- [x] Deregister discriminating pair: sweeping `AlreadyAbsent` → `PostMintUnrecordable`; nothing-to-sweep `AlreadyAbsent` → `AuditUnwritable { phase: Outcome }`
- [x] Outcome-write failure after a `FirstMint`, an `IdempotentReMint` and a removal returns `PostMintUnrecordable`, with `OpenBao` state as the verb left it; re-driving the deregister once the store recovers succeeds and writes its pair
- [x] A `HostMismatch` deregister's outcome-write failure returns `AuditUnwritable { phase: Outcome }`
- [x] The outcome record is written before the per-`registration_id` guard is released; `ID_LOCKS` remains the only verb-layer lock and stages 1 and 2 stay unsynchronized
- [x] `ServiceSpecOutsideSafeSet`, `StoredSpecConflict` and a spec-identity disagreement produce three different `outcome.reason` values
- [x] The caller identity reaches `caller_identity` unchanged and uninterpreted
- [x] `AuditUnwritable` carries both `AuditPhase` values across the suite; `PostMintUnrecordable` carries no phase field
- [x] `bridge::refusal_outcome` returns `Option<AuditOutcome>`, `None` for exactly the two audit-failure variants, and `RefusalReason` gained no member
- [x] `scripts/preflight/ci/check.sh` green; the five `validate-*` scripts green; `ci/deploy-no-build-smoke.sh` green; `ci/test-core.sh` green
- [ ] `scripts/preflight/ci/e2e-matrix.sh` end to end — see below; the arms that need passwordless `sudo` are left to CI

## Verification notes

`scripts/preflight/ci/e2e-matrix.sh --skip-hosts` was run and not skipped: local lifecycle, remote lifecycle, the rotation/recovery matrix and reinit recovery all passed, and the `registrar-verbs` step — the one this change exercises directly — is green. It then stopped at `run-stepca-san.sh`, which fails identically on the unmodified base commit: `bootroot infra install` finds host port 8080 still held by the OrbStack helper for the container `compose_stop` just stopped, and moving the port with `HTTP01_ADMIN_HOST_PORT` moves the conflict with it.

The steps after it were run standalone and passed: OpenBao TLS no-delta, two-instance isolation, registrar verbs, and the registrar internal credential. OpenBao TLS re-own and registrar internal credential init need passwordless `sudo`, which this development host does not have, so those two are left to CI.

### Re-verification, 2026-08-25

Re-run independently on the same host. `cargo fmt --check`, `cargo clippy --all-targets -D warnings` and `cargo test` are green (745 lib tests), as are `ci/check.sh`, the five `validate-*` scripts, `ci/deploy-no-build-smoke.sh` and `ci/test-core.sh`. `scripts/impl/run-registrar-verbs-e2e.sh` passes 16/16.

`scripts/preflight/ci/e2e-matrix.sh --skip-hosts` no longer reaches `run-stepca-san.sh`: it now stops earlier, at step 5, `run-rotation-recovery.sh`, which fails its first `scenario-a` verify with `Leaf certificate at certs/daemon-c1.crt does not chain to CA bundle ...; reissue the leaf so it matches the current PKI generation.` That is this development host and not this branch — the same script run from a throwaway worktree at the merge-base `d8d1907`, with binaries built there, fails with the byte-identical message. It also reproduces from a fully cleaned Docker and secrets state, so it is not run-to-run residue.

Steps 1-4 pass in the matrix (local and remote lifecycle, no-hosts arms; the hosts arms are skipped for lack of passwordless `sudo`). Everything after the rotation step was run standalone against a cleaned host and passes: reinit recovery, OpenBao TLS no-delta, two-instance isolation, registrar verbs, and the registrar internal credential. `run-stepca-san.sh`, OpenBao TLS re-own and registrar internal credential init remain blocked here — the first on the OrbStack published-port race described above, the other two on passwordless `sudo`. All three, and the rotation arm, are green in this PR's CI, where all 13 `Docker E2E` jobs pass.

### DIFF
--- a/src/registrar/audit.rs
+++ b/src/registrar/audit.rs
@@ -55,11 +55,11 @@
 //! There is no best-effort mode, no buffering, no fire-and-forget path
 //! and no caller-selectable suppression: [`AuditRecordStore::append`]
 //! either durably wrote one line or returns a typed
-//! [`AuditStoreError`]. Writing records around the two verbs — where
-//! the intent and outcome lines go relative to the `OpenBao` work, and
-//! what a failed write returns to a caller — is not here either. The
-//! verb layer holds the store as a fixed dependency and the sibling
-//! issue that adds those call sites owns that ordering.
+//! [`AuditStoreError`]. Where the intent and outcome lines go relative
+//! to the `OpenBao` work, and what a failed write returns to a caller,
+//! is not here either: `crate::registrar::verbs` owns those call sites
+//! and their ordering, and every failure this module reports reaches it
+//! as one of the two audit-failure refusals.
 
 use std::fs::{DirBuilder, File, OpenOptions, Permissions};
 use std::io::Write as _;
@@ -756,14 +756,6 @@ mod millisecond_rfc3339 {
 /// is one definition of "which reason is this" in the repository and a
 /// second copy in `verbs.rs` would drift from the format it is supposed
 /// to spell.
-// Transitional. The sibling issue that writes intent and outcome
-// records around the two verbs is the first production caller of every
-// function here; until it lands nothing but this module's own tests
-// calls them. Widening them to `pub` to silence the lint would put a
-// crate-private control plane's refusal taxonomy on the library's
-// public surface, which is the opposite of what the visibility rule
-// asks for.
-#[allow(dead_code)]
 pub(crate) mod bridge {
     use super::{AuditOutcome, RefusalReason};
     use crate::registrar::RegistrarError;
@@ -787,7 +779,19 @@ pub(crate) mod bridge {
     }
 
     /// Returns the record outcome for a refusal: its flattened reason
-    /// and the operator-facing detail that goes with it.
+    /// and the operator-facing detail that goes with it, or `None` for a
+    /// refusal that can never reach the trail.
+    ///
+    /// The two audit-failure variants are the whole of that `None`.
+    /// `VerbError::AuditUnwritable` and `VerbError::PostMintUnrecordable`
+    /// are produced *by* a failed record write, so the record that would
+    /// carry one is the record whose write just failed. Saying so in the
+    /// return type rather than in prose is what keeps a call site from
+    /// answering a failed write with a second one: where this returns
+    /// `None`, no outcome record is written and the refusal is returned
+    /// unchanged. [`RefusalReason`] therefore gains no member for either,
+    /// and the reason table stays a total, one-to-one mapping of the
+    /// refusals a durable line can hold.
     ///
     /// The detail is the refusal's *own* message and never a formatted
     /// error chain. `VerbError::Unavailable` is taken from its explicit
@@ -795,8 +799,11 @@ pub(crate) mod bridge {
     /// source — which can carry anything at all, including values from
     /// other systems — cannot reach a durable record even if that
     /// `Display` were later changed to interpolate it.
-    pub(crate) fn refusal_outcome(error: &VerbError) -> AuditOutcome {
+    pub(crate) fn refusal_outcome(error: &VerbError) -> Option<AuditOutcome> {
         let (reason, detail) = match error {
+            VerbError::AuditUnwritable { .. } | VerbError::PostMintUnrecordable { .. } => {
+                return None;
+            }
             VerbError::ReservedServiceName { .. } => {
                 (RefusalReason::ReservedServiceName, Some(error.to_string()))
             }
@@ -816,7 +823,7 @@ pub(crate) mod bridge {
                 (RefusalReason::Unavailable, Some(activity.clone()))
             }
         };
-        AuditOutcome::Refused { reason, detail }
+        Some(AuditOutcome::Refused { reason, detail })
     }
 
     /// Flattens every refusal the config loader and the derivation
--- a/src/registrar/audit/tests.rs
+++ b/src/registrar/audit/tests.rs
@@ -407,13 +407,22 @@ fn appending_preserves_the_caller_supplied_timestamp() {
 // Format: the flattened refusal table
 // ---------------------------------------------------------------------
 
-/// Every variant of the three source enums, paired with the record
-/// reason it must flatten to.
+/// Every variant of the three source enums that a durable line can
+/// hold, paired with the record reason it must flatten to.
 ///
 /// The list is the test's definition of "all of them": adding a variant
 /// upstream without adding it here leaves `refusal_outcome`'s `match`
 /// failing to compile, and adding it here without adding a reason fails
 /// the distinctness assertion below.
+///
+/// The two audit-failure variants — `VerbError::AuditUnwritable` and
+/// `VerbError::PostMintUnrecordable` — are deliberately **not** in this
+/// table and have no [`RefusalReason`]. They are produced by a failed
+/// record write, so the record that would carry one is the record whose
+/// write just failed, and `refusal_outcome` returns `None` for exactly
+/// them. That is covered by
+/// [`the_two_audit_failures_are_the_only_unrecordable_refusals`] rather
+/// than here.
 // One entry per source variant; the exhaustiveness is the point.
 #[allow(clippy::too_many_lines)]
 fn every_refusal() -> Vec<(VerbError, RefusalReason, &'static str)> {
@@ -643,7 +652,7 @@ fn every_source_refusal_maps_to_one_distinct_record_reason() {
 
     let mut seen: Vec<String> = Vec::new();
     for (error, expected_reason, expected_name) in table {
-        let outcome = refusal_outcome(&error);
+        let outcome = refusal_outcome(&error).expect("a recordable refusal produces an outcome");
         let AuditOutcome::Refused { reason, .. } = &outcome else {
             panic!("a refusal must produce the refused class, got {outcome:?}");
         };
@@ -658,6 +667,53 @@ fn every_source_refusal_maps_to_one_distinct_record_reason() {
     assert_eq!(sorted.len(), seen.len(), "every reason must be distinct");
 }
 
+/// The two audit-failure refusals can never reach the trail, and that
+/// is enforced by the return type rather than by prose.
+///
+/// The record that would carry one of them is the record whose write
+/// just failed, so a call site that met a `Some` here would answer a
+/// failed write with a second write of the same kind. Both halves are
+/// asserted together: `None` for exactly these two, and `Some` for every
+/// other refusal the verb layer produces.
+#[test]
+fn the_two_audit_failures_are_the_only_unrecordable_refusals() {
+    let unrecordable = [
+        VerbError::AuditUnwritable {
+            phase: AuditPhase::Intent,
+            source: AuditStoreError::RecordTooLarge {
+                bytes: 1,
+                limit: MIN_AUDIT_MAX_FILE_BYTES,
+            },
+        },
+        VerbError::AuditUnwritable {
+            phase: AuditPhase::Outcome,
+            source: AuditStoreError::Sync {
+                path: PathBuf::from("/var/lib/bootroot/registrar-audit/registrar-audit.jsonl"),
+                source: std::io::Error::from(std::io::ErrorKind::StorageFull),
+            },
+        },
+        VerbError::PostMintUnrecordable {
+            source: AuditStoreError::Append {
+                path: PathBuf::from("/var/lib/bootroot/registrar-audit/registrar-audit.jsonl"),
+                source: std::io::Error::from(std::io::ErrorKind::StorageFull),
+            },
+        },
+    ];
+    for error in &unrecordable {
+        assert!(
+            refusal_outcome(error).is_none(),
+            "{error:?} must produce no record at all"
+        );
+    }
+
+    for (error, _, name) in every_refusal() {
+        assert!(
+            refusal_outcome(&error).is_some(),
+            "{name} is a durable reason and must still be recordable"
+        );
+    }
+}
+
 #[test]
 fn a_refusal_reason_round_trips_through_the_record_types() {
     for (error, _, name) in every_refusal() {
@@ -668,7 +724,7 @@ fn a_refusal_reason_round_trips_through_the_record_types() {
             CALLER.to_string(),
             requested(None),
             None,
-            refusal_outcome(&error),
+            refusal_outcome(&error).expect("a recordable refusal produces an outcome"),
         );
         let rendered = serde_json::to_string(&record).expect("serializes");
         assert!(
@@ -713,7 +769,7 @@ fn an_unavailable_refusal_records_its_activity_and_nothing_else() {
         "reading the durable binding",
         anyhow::anyhow!("connection refused").context("dialling openbao at https://vault.internal"),
     );
-    let AuditOutcome::Refused { reason, detail } = refusal_outcome(&error) else {
+    let Some(AuditOutcome::Refused { reason, detail }) = refusal_outcome(&error) else {
         panic!("an unavailable refusal must produce the refused class");
     };
     assert_eq!(reason, RefusalReason::Unavailable);
@@ -726,7 +782,7 @@ fn an_unavailable_refusal_records_its_activity_and_nothing_else() {
         CALLER.to_string(),
         requested(None),
         None,
-        refusal_outcome(&error),
+        refusal_outcome(&error).expect("an unavailable refusal is recordable"),
     );
     let rendered = serde_json::to_string(&record).expect("serializes");
     for leaked in ["connection refused", "vault.internal", "dialling openbao"] {
--- a/src/registrar/verbs.rs
+++ b/src/registrar/verbs.rs
@@ -44,6 +44,32 @@
 //!    [`tokio::sync::Mutex`], shared by both verbs so a mint and a
 //!    deregister for one identity can never interleave.
 //!
+//! # The audit trail around those stages
+//!
+//! Every invocation that reaches a verb body attempts **two** record
+//! writes into the daemon-owned store: an `intent` line at entry, before
+//! stage 1 and therefore before any `OpenBao` call, and an `outcome`
+//! line once the verb has run. The two are paired by `request_id` rather
+//! than by adjacency — intent lines are ordered only by arrival, which
+//! is correct, because nothing holds a lock across a whole invocation.
+//! The outcome line for a result produced under the per-id guard is
+//! written **before that guard is released**, so the next invocation for
+//! the same identity cannot change `OpenBao` and land its outcome line
+//! ahead of the line describing the change it superseded.
+//!
+//! Neither write is best-effort. A failed intent write refuses the
+//! invocation with nothing created, which is free at that point. A
+//! failed outcome write may arrive after `OpenBao` has already changed,
+//! so what it returns is decided by a per-invocation **mutation
+//! disposition** — see [`MutationDisposition`] — and by nothing else:
+//! not the verb, not the outcome class, not the [`ProducingArm`], and
+//! not a re-read of `OpenBao`.
+//!
+//! The `OpenBao` file audit device stays mandatory and is not replaced.
+//! It cannot see a request refused before any `OpenBao` write, and it
+//! does not know who asked or for which `(service_name, host,
+//! instance)`; this trail does.
+//!
 //! Stage 3's per-id lock is the **only** in-process serialization
 //! boundary either verb has. Its map is owned by this module rather than
 //! by a service instance, so two [`RegistrarVerbs`] in one process
@@ -104,6 +130,7 @@ pub(crate) mod wrap_ttl;
 mod tests;
 
 use std::collections::HashMap;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, LazyLock, Mutex as StdMutex, PoisonError, Weak};
 
 use anyhow::Context as _;
@@ -118,14 +145,16 @@ use self::outcome::{
 };
 use self::wrap_ttl::{GrantedWrapTtl, WrapTtlPolicy};
 use crate::openbao::{KvCreateIfAbsent, OpenBaoClient, SecretIdOptions, WrapInfo};
-use crate::registrar::audit::AuditRecordStore;
+use crate::registrar::audit::{
+    AuditOutcome, AuditPhase, AuditRecord, AuditRecordStore, AuditVerb, RequestedIdentity, bridge,
+};
 use crate::registrar::config::{Multiplicity, RegistrarConfig};
 use crate::registrar::identity::{RequestedSpec, check_instance_shape, derive_registration_id};
 use crate::registrar::internal::{InternalCredential, InternalCredentialError};
 use crate::registrar::{check_spec_identity, is_reserved_service_name, validate_request_labels};
 use crate::service_material::{
-    ProvisionedServiceRole, ServiceRoleTtls, provision_service_role, service_kv_path,
-    service_role_name, teardown_service_material,
+    ProvisionedServiceRole, ResourceOutcome, ServiceRoleTtls, provision_service_role,
+    service_kv_path, service_role_name, teardown_service_material,
 };
 use crate::trust_bootstrap::{
     SERVICE_EAB_KV_SUFFIX, SERVICE_REISSUE_KV_SUFFIX, SERVICE_RESPONDER_HMAC_KV_SUFFIX,
@@ -174,6 +203,106 @@ static ID_LOCKS: LazyLock<IdLocks> = LazyLock::new(IdLocks::default);
 /// a pathological pair of processes from spinning here forever.
 const MAX_CLAIM_ATTEMPTS: usize = 3;
 
+/// Whether one invocation has made an `OpenBao` call that can have
+/// changed durable state.
+///
+/// Set from the *result* of each such call and **never cleared**, so it
+/// is monotone over the invocation. Reads never set it — a binding read,
+/// a `role_id` read, acquiring an authenticated client — and neither
+/// does a lost compare-and-set, which wrote nothing at all; that is why
+/// exhausting [`MAX_CLAIM_ATTEMPTS`] refuses with it still clear.
+///
+/// Every *error* does set it, and that asymmetry is the whole point. A
+/// call that failed may have failed after its write reached `OpenBao`,
+/// and the two ways of being wrong are not symmetrical: reporting a
+/// change as an ordinary refusal leaves live, unrecorded state that
+/// nobody is told to clean up, while reporting an unchanged invocation
+/// as owed costs at most one idempotent re-drive that finds nothing. A
+/// `ResourceOutcome::Failed` teardown attempt is indeterminate in
+/// exactly the same way and counts the same.
+///
+/// It is shared behind `&` across `.await` points in a `Send` future, so
+/// an [`AtomicBool`] is the shape to use; the value is per-invocation and
+/// is read by the same task that set it, so no ordering stronger than
+/// the module already uses is needed.
+#[derive(Debug, Default)]
+struct MutationDisposition(AtomicBool);
+
+impl MutationDisposition {
+    /// Records that a call which can have changed `OpenBao` state has
+    /// completed, however it completed.
+    fn set(&self) {
+        self.0.store(true, Ordering::Relaxed);
+    }
+
+    /// Records the outcome of a state-changing call and hands its result
+    /// straight back, so the call site reads as it did before.
+    fn note<T, E>(&self, result: Result<T, E>) -> Result<T, E> {
+        self.set();
+        result
+    }
+
+    /// Reports whether this invocation is owed a teardown.
+    fn is_set(&self) -> bool {
+        self.0.load(Ordering::Relaxed)
+    }
+}
+
+/// Everything one invocation's two record writes need.
+///
+/// Built at verb entry, before stage 1. The caller identity and the
+/// requested parts are carried into the builders **unchanged and
+/// uninterpreted**: whatever [`CallerIdentity`] holds is what the record
+/// is handed, and what a durable line ends up saying is the store's
+/// business — `AuditRecord::into_bounded` caps the field and records a
+/// digest when it had to. Nothing here pre-truncates or works around
+/// that.
+struct AuditContext {
+    request_id: RequestId,
+    verb: AuditVerb,
+    caller: CallerIdentity,
+    requested: RequestedIdentity,
+    disposition: MutationDisposition,
+}
+
+impl AuditContext {
+    /// Opens the context for one invocation, generating its request id.
+    fn new(verb: AuditVerb, caller: &CallerIdentity, requested: RequestedIdentity) -> Self {
+        Self {
+            request_id: RequestId::generate(),
+            verb,
+            caller: caller.clone(),
+            requested,
+            disposition: MutationDisposition::default(),
+        }
+    }
+
+    /// Builds the record shared by both phases: the timestamp, the
+    /// correlation handle, the verb, the caller and the requested parts.
+    fn intent_record(&self) -> AuditRecord {
+        AuditRecord::intent(
+            OffsetDateTime::now_utc(),
+            self.request_id.as_str().to_string(),
+            self.verb,
+            self.caller.as_str().to_string(),
+            self.requested.clone(),
+        )
+    }
+
+    /// Builds the `outcome` line for a produced result.
+    fn outcome_record(&self, registration_id: Option<&str>, outcome: AuditOutcome) -> AuditRecord {
+        AuditRecord::outcome(
+            OffsetDateTime::now_utc(),
+            self.request_id.as_str().to_string(),
+            self.verb,
+            self.caller.as_str().to_string(),
+            self.requested.clone(),
+            registration_id.map(str::to_string),
+            outcome,
+        )
+    }
+}
+
 /// The fixed dependencies a [`RegistrarVerbs`] is constructed with.
 ///
 /// Every one of these is chosen once, by whatever provisions the
@@ -205,10 +334,9 @@ pub(crate) struct RegistrarVerbsConfig {
     /// fails that surface closed rather than leaving the verbs running
     /// with no trail.
     ///
-    /// No verb arm calls its append API yet. The sibling issue that
-    /// writes intent and outcome records owns the call sites, their
-    /// ordering around the `OpenBao` work, and what a failed write
-    /// returns to a caller; this field is the seam it consumes.
+    /// Both verbs write through it, twice per invocation, and an
+    /// invocation whose record cannot be written is refused rather than
+    /// served silently — see this module's header.
     pub(crate) audit_store: AuditRecordStore,
 }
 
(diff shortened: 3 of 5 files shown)


AcoPiper:
**[Reviewer Round 1]**

No blocking findings.

The integration matches the issue’s fail-closed contract: `mint` and `deregister` attempt intent before local validation or OpenBao access, and successful/locked results record their outcome before releasing the per-registration-id guard. `MutationDisposition` is monotonic and is set at the required write-capable operations, including indeterminate errors and teardown `Removed`/`Failed` attempts, so outcome-write failures distinguish no-change refusals from state that may need recovery.

The refusal bridge also correctly makes the two audit-write errors unrecordable rather than inventing a durable refusal reason. The fast tests exercise the important discriminating paths (read-only conflict vs. claimed mint, empty vs. sweeping deregister), both phase values, and successful recorded paths; the live concurrency test specifically covers the ordering guarantee. PR hygiene is in order: it closes #777, includes a test plan, and the PR discussion contains no out-of-scope issue creation.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Record intent and outcome around both registrar verbs
```

**Body**

```text
The OpenBao file audit device cannot see a request refused before any
OpenBao write, and does not know who asked or for which
(service_name, host, instance). The record store has been in place
for both, unused: no verb arm appended to it.

Both verbs now write two records per invocation. The intent line goes
in at entry, before stage 1 and therefore before any OpenBao call, so
a store that cannot take it refuses while refusing is still free. The
outcome line goes in while the per-registration_id guard is still
held, so the next invocation for that identity cannot change OpenBao
and land its line ahead of the line describing what it superseded.

Neither write is best-effort, which forces a decision about what a
failed one returns. A failed outcome write can arrive after OpenBao
has already changed, and neither the verb, the outcome class nor the
ProducingArm says whether it did: a refusal can follow a durable
claim, an already-absent deregister can have swept a full orphan, and
ProducingArm::Binding covers both reads and writes. So the invocation
carries an explicit mutation disposition instead, set from the result
of each of the six OpenBao calls that can change state and never
cleared. Every error sets it too, because a call that failed may have
failed after its write landed, and reporting a change as an ordinary
refusal leaves live state nobody is told to clean up, where the
converse costs one idempotent re-drive that finds nothing.

AuditUnwritable carries the phase that failed, which nothing else on
the refusal implies; PostMintUnrecordable says a teardown is owed.
Neither can ever reach the trail, since the record that would carry
one is the record whose write just failed — bridge::refusal_outcome
returns Option so that is a type error rather than a convention, and
RefusalReason gains no member.

Most fast-tier tests here inject a write failure, since that is how a
sweep that removed something is told from one that found nothing. The
two paths where the outcome write succeeds and the caller is served —
a mint returning its material on both arms, and an already-absent
deregister — are covered there as well, so a regression that dropped
the wrapped secret_id or recorded the wrong class does not hide behind
the fault-injecting tests until the E2E matrix runs.

Closes #777
```

PR #891, opened 2026-08-24
Add registrar wire protocol and codec (#885)

### LINKED ISSUE #885: Add the registrar endpoint's versioned wire protocol and its codec
# Add the registrar endpoint's versioned wire protocol and its codec

## Context

The registrar endpoint's listener is merged and serves nothing. `src/registrar/endpoint/handler.rs` defines the `RegistrarRequestHandler` trait — one method taking the checked `Operation`, the payload as opaque bytes and the transport-authenticated `CallerIdentity`, returning either response bytes to frame or a zero-data `HandlerRefusal` — and the only implementation is `#[cfg(test)]`. The payload schema was deliberately left out, and this issue is it: the versioned request and response types, the encoding they are serialized in, the codec, the exhaustive mapping from the verb layer's typed refusals onto the caller-facing error identifiers, the endpoint's own reserved health container, and golden fixtures that turn a rename into a test failure instead of a silent cross-repo break.

**The externally owned half of the contract is already checked in and complete.** `docs/reference/registrar-wire-contract.md` carries every externally owned name as tables with one row per name, and states outright that no such name appears only in prose — so a test can recover every name, class and count from the rows. Scope transcribes the closed sets below for review; the file stays the source. Three questions this work would otherwise have had to hedge against are answered there and are not open: the identifier for the pre-derivation non-DNS-label refusal landed upstream as **`ServiceLabelInvalid`**, a unit variant, permanent (§6.3); the retryable identifier's payload member is spelled **`retry_after`** (§6.1); and the identifier set has **no generic, catch-all, `Other` or `Unknown` member** (§7.1), which is what makes an absent identifier the correct wire form for a transient failure.

The reference leaves exactly one item unresolved, and it assigns it here: §10 records that `ca_anchor` is typed `Vec<u8>` upstream with its contents delegated to this repository, that nothing here emits a single `ca_anchor` byte string today, and that the container packing this repository's two anchor values into those bytes is chosen by "the endpoint transport/codec issue". This issue chooses it below and closes that row.

This issue writes no daemon wiring. Registering a production handler, provisioning the verb layer inside the daemon and lifting the endpoint's not-yet-usable state belong to the sibling issue that consumes what this one builds.

### The encoding is bootroot's own, and this issue picks it

The reference supplies names, not bytes, and could not supply bytes: this socket is not review-protocol's wire, whose framing and encoding belong to the transport between the co-located registrar and the control plane. The encoding here is on the same footing as the listener's framing and operation identifiers.

- **`serde`-derived JSON, encoded with `serde_json`, one object per frame.** `serde_json` is already a dependency (`Cargo.toml:25`) and is what the audit record store writes its own records with. Its field names are literally the mirrored names rather than a positional layout, which is what makes a rename visible at the byte level — and with the payload names owned elsewhere, a format in which a rename is invisible would defeat the fixtures' entire purpose.
- **Wire enums that carry data use the internally-tagged `serde` representation `ReloadStrategy` already uses (`src/state.rs:22`); fieldless enums are plain strings.** Concretely, **the `error` object is the identifier enum in that representation with `id` as the tag key** — the same idiom, whose tag key on `ReloadStrategy` is `type`. A payload-free identifier therefore serializes as `{"id": "ServiceHostMismatch"}` and nothing more; a payload-carrying one adds its own member beside the tag.
- **An externally owned enum variant rides the wire in the reference's own spelling, exactly, with no `rename_all` applied to it** — `"ServiceHostMismatch"`, `"NotProvisioned"`, `"LocalFile"`. Bootroot-owned enums (`outcome`, `class`) take the `snake_case` spellings given below. The two vocabularies sit side by side inside one object and that is deliberate: recasing an external variant to match this repository's own convention is exactly the silent cross-repo break the fixtures exist to catch.

### The externally owned half, transcribed here for review

Every externally owned name this protocol carries is already checked in at `docs/reference/registrar-wire-contract.md`, and that file stays the **sole source** the codec, the fixtures and the conformance test take names from. The transcription exists so a reviewer can catch a misreading without opening the file. **Where this transcription and the file disagree, the file wins, and the difference is a finding raised against this issue — never a codec change made to match this text.** The constraint against retyping an externally owned name into a test or a fixture is unchanged; part 2 of the fixture work below is what proves it was obeyed.

**The closed sets and their counts.** §11 of the reference carries the counts as a table of their own, so the conformance test recovers them from the file rather than hard-coding them.

| Set | Count | Members |
| --- | --- | --- |
| `Register` fields | 7 | `service_name`, `delivery_mode`, `host`, `instance`, `spec`, `wrap_ttl`, `idempotency_key` |
| `Deregister` fields | 4 | `service_name`, `host`, `instance`, `idempotency_key` |
| `ServiceSpec` fields | 4 | `component`, `service_name`, `reload`, `cert_group` |
| `BootstrapMaterial` fields | 4 | `role_id`, `wrapped_secret_id`, `ca_anchor`, `expires_at` |
| `DeliveryMode` variants | 2 | `LocalFile`, `RemoteBootstrap` |
| Typed enroll errors | 7 | the identifier table below |
| `RegistrarUnavailable` reasons | 6 | the reason table below |
| Retryable typed errors | 1 | `RegistrarBusy` |

**The seven identifiers, with the payload each carries and the class the reference assigns it** (§6.1). The payload column is what decides an `error` object's members beyond `id`: five identifiers carry none, so their whole `error` object is `{"id": "…"}`.

| Identifier | Payload member | Class |
| --- | --- | --- |
| `ServiceSpecConflict` | none | permanent |
| `ServiceNameCollision` | none | permanent |
| `ServiceInstanceMismatch` | none | permanent |
| `ServiceHostMismatch` | none | permanent |
| `RegistrarUnavailable` | `reason` | permanent |
| `RegistrarBusy` | `retry_after` | retryable |
| `ServiceLabelInvalid` | none | permanent |

**The six `RegistrarUnavailable` reasons** (§6.5), which is a closed set. The response type must be able to **decode** all six; this codec ever *emits* only the three marked.

| Reason | Emitted by this codec | Why |
| --- | --- | --- |
| `NotProvisioned` | yes | the ten config-loading `RegistrarError` variants map to it |
| `PostMintUnrecordable` | not yet | assigned below; no verb variant produces it today |
| `AuditUnwritable` | not yet | assigned below; no verb variant produces it today |
| `CredentialInvalid` | no | the registrar's own client certificate, observed registrar-side |
| `EndpointUnreachable` | no | this daemon being down; nothing it can answer with |
| `RegistrarUnreachable` | no | manager-synthesized, never returned by a registrar |

A reason this codec never emits is still a variant of the decodable enum. The response types round-trip in both directions, so a reader has to decode a `RegistrarUnavailable` carrying any of the six; an enum holding only the emitted three is incomplete and the conformance test fails it. The same holds for the identifier set: all seven decode, whatever this daemon emits.

### The bootroot-owned wire names, spelled here

Nothing downstream can supply these, and pinning them afterwards with generated fixtures would be circular. Spell them exactly as written, in `snake_case` matching the rest of this repository's `serde` types.

**The protocol version.**

- The field is **`protocol_version`**, a JSON integer, **required on every message in both directions**. It takes no `serde` default, so a message without it does not decode.
- Its value is **`1`**, held in a `PROTOCOL_VERSION` constant at the top of the protocol module rather than written inline at each construction site.
- The daemon accepts exactly the versions it implements — that single value — and every response carries the version it spoke.
- **The check belongs to the type, not to either direction's decoder.** `protocol_version` is a `ProtocolVersion` newtype whose hand-written `Deserialize` accepts `PROTOCOL_VERSION` and rejects every other value, and every request and response type carries that one type. A derived decoder over a plain integer would take any number the peer sent and leave the rule to a per-direction check — and the forgotten direction is the one that silently misreads.
- **An absent or unimplemented `protocol_version` fails to decode in both directions — the rule is symmetric.** On a request that reaches the caller as the listener's existing pre-verb refusal: zero response bytes, a typed log line, a clean close, never version-`1` semantics and never a wire refusal, because no verb ran and no request id exists to put on one. On a response it surfaces as the codec's decode `Err`, because the reader is a caller with no refusal to answer with — and what it would otherwise misread is minted material or a refusal it acts on.
- **Unknown-field tolerance does not reach the version.** Tolerance covers a *field* this daemon does not know; an unimplemented version is a *meaning* this daemon does not have, and serving it anyway is the silent misinterpretation the version exists to prevent.

**The response envelope.** Every response carries `protocol_version`, `request_id` (the verb layer's per-invocation handle, a string) and `registrar_health` (the reserved container, an object of named members, serialized as `{}` until the three member-owning issues fill it). Beyond those three:

| Response shape | Its own members |
| --- | --- |
| Mint success | `registration_id`, `outcome`, `material` |
| Deregister success | `registration_id`, `outcome` |
| Refusal | `class`, optional `error`, optional `registration_id` |

- **`outcome`** is a string-valued enum: **`first_mint`** and **`idempotent_remint`** on the mint response, **`removed`** and **`already_absent`** on the deregister response. These are the wire form of `MintKind` and `DeregisterKind` (`src/registrar/verbs/outcome.rs:189`, `:270`), and are bootroot's — the mirrored `Done` acknowledgement carries no fields at all, so every member of the deregister success shape is bootroot-owned.
- **`material`** nests the mirrored `BootstrapMaterial` payload under one member, carrying the reference's four field spellings verbatim and unflattened. The nesting puts every externally owned name in a response beneath a single member, which is what makes the conformance test's checked/not-checked split mechanical instead of a hand-maintained exclusion list.
- **`class`** is a string-valued enum with exactly **`retryable`** and **`permanent`**, and is **never absent** — it is what a caller reads when there is no identifier to read.
- **The `class` member is bootroot-owned; the classification it carries is not.** The reference has no such wire field: it records each identifier's class as a column on its identifier table, so *which* class an identifier has is transcribed and conformance-checkable, while the member carrying it on this socket is this endpoint's own. Its two values reuse the reference's own words — `permanent` and `retryable` (§6.1, §7) — so no reader maps one vocabulary onto another. Do not spell the retryable value `transient`.
- **`error`** is optional. When present it is an object whose **`id`** member carries the mirrored identifier spelling and whose remaining members are that identifier's own mirrored payload — `retry_after` for `RegistrarBusy`, `reason` for `RegistrarUnavailable`. When absent it is **omitted entirely**, never serialized as `null`; the same holds for `registration_id`.
- **Requests carry `protocol_version` and, beyond it, only the reference's own field names.** No bootroot-owned member is added to a request shape.

**The JSON value type of every member, so two implementations cannot differ on one.** These are the canonical v1 representations and there is no second accepted form for any of them.

| Member | JSON value |
| --- | --- |
| `service_name`, `host`, `idempotency_key`, `role_id`, `wrapped_secret_id` | String. `idempotency_key` is an opaque correlation value. `wrapped_secret_id` is the `OpenBao` response-wrapping token copied verbatim — not the raw `secret_id`, and not re-encoded. |
| `instance` | Unsigned integer in the `u32` range when present, **omitted** when absent. Never `null`, never a quoted number. |
| `delivery_mode` | String, the external variant spelling exactly: `"LocalFile"` or `"RemoteBootstrap"`. |
| `spec` | Required object carrying exactly `component`, `service_name`, `reload` and optional `cert_group`. The first three are strings; `cert_group` is a string when present and **omitted** when absent. No privilege member exists. |
| `wrap_ttl`, `retry_after` | Non-negative whole-second integer. |
| `expires_at` | RFC 3339 **UTC** string, produced with the `time` crate's `Rfc3339` well-known format, which is what this repository already writes elsewhere (`src/hooks.rs:99`, `src/bin/bootroot-remote/bootstrap.rs:320`). It carries the granted absolute deadline, not the requested `wrap_ttl`. |
| `ca_anchor` | Standard **padded** RFC 4648 base64 string, encoded with the existing `base64` dependency's `engine::general_purpose::STANDARD` (`Cargo.toml:8`) — the engine this repository already uses for opaque binary elsewhere (`src/acme/flow.rs:6`, `src/daemon.rs:1215`). Not `URL_SAFE_NO_PAD`, which the repository also uses but reserves for JWS, and never `serde_json`'s default numeric array. Its decoded byte framing is settled just below. |
| `protocol_version` | Integer. |
| `request_id`, `registration_id` | String. |
| `outcome`, `class` | String, from the bootroot-owned spellings above. |
| `error` | Object, or **omitted**. |
| `registrar_health` | Object. |

**Every optional member's canonical serialized form is omission, never `null`** — `instance`, `spec.cert_group`, `error` and `registration_id` alike. All strings use ordinary JSON escaping. A fixture pins each literal encoding, so a base64 engine swap or a `null` regression fails as a byte diff.

**A payload that does not decode is never answered on the wire.** The rule stated above for the version generalises to the whole payload: any request the codec cannot decode into its typed shape — an absent required member, a wrong JSON type, a `wrap_ttl` that is negative or fractional or past `u64`, an `instance` outside `u32`, a `delivery_mode` outside the two spellings — is a typed `Err` the codec returns, never a wire refusal: no verb ran, so no request id exists to put one on. The sibling issue's handler turns that `Err` into the already-merged `HandlerRefusal` (`src/registrar/endpoint/handler.rs`), whose documented purpose is exactly "a payload a production codec cannot decode", and the listener answers it with zero response bytes and a clean close. This codec constructs none itself and writes no response for one.

**One consequence, stated here rather than discovered mid-implementation.** With `wrap_ttl` typed as a non-negative whole-second integer, `WrapTtlRefusal::Negative` and `WrapTtlRefusal::NotWholeSeconds` (`src/registrar/verbs/wrap_ttl.rs:23`) are not reachable *through this socket* — such a request fails to decode instead — while `Zero` and `ExceedsOpenBaoRange` stay reachable. All four still get their own arm and round-trip test: completeness is measured against the enum, so a variant that becomes reachable later finds an arm already written.

**`ca_anchor`'s byte framing, settled here to the byte.** The decoded `ca_anchor` bytes are the compact UTF-8 JSON serialization of exactly this object, with the members in this order:

```json
{"trusted_ca_sha256":["<64-character-lowercase-hex-sha256>",...],"ca_bundle_pem":"<PEM bundle>"}
```

- `trusted_ca_sha256` holds one lowercase hexadecimal SHA-256 fingerprint per certificate, computed over that certificate's **DER** bytes and ordered exactly like the certificates in `ca_bundle_pem`.
- `ca_bundle_pem` holds the complete public CA bundle as UTF-8 PEM text, with the certificates in that same order, **LF** line endings, and exactly one trailing LF.
- The compact JSON carries no BOM, no insignificant whitespace and no trailing newline; the PEM newlines are JSON-escaped by `serde_json`.
- **Both members are required**, the fingerprint array and the certificate bundle are non-empty, and each fingerprint must match the certificate at the same position.
- Those JSON bytes are then encoded as a standard padded RFC 4648 base64 JSON string, per the table above.
- **No length prefix, no delimiter outside the JSON structure, no private key and no additional member is included.**

This is not a new format: it is the shape `parse_trust_payload` (`src/kv_payload.rs:48`) already reads, under those two key names (`src/trust_bootstrap.rs:12`/`:13`), so the two values cannot drift apart in transit and no second anchor format is grown.

**The conversion between bytes and that parser is named here, not left to either side of the seam.** `parse_trust_payload` takes `&serde_json::Value`, **not** bytes, so "hand the decoded bytes to the existing parser" is one step short. The full path is: base64-decode the member to bytes, `serde_json::from_slice::<serde_json::Value>` those bytes, then `parse_trust_payload(&value)`. This issue provides that whole path as a **single codec-owned pair of functions** — `TrustPayload` to the base64 member and back — so the handler sourcing the anchor, this codec and any future reader share one implementation rather than each writing its own base64, `Value` and compact-serialization steps. A failure at any stage is an ordinary decode error under the rule above, never a wire refusal.

**Acceptance by `parse_trust_payload` does not by itself prove this framing, and the tests must not treat it as if it did.** Its `validate_bundle_consistency` matches each fingerprint against a `HashSet` of the bundle's digests, lowercased at comparison time, so it accepts uppercase hex and any ordering — while the framing above requires lowercase hex, one per certificate, in the bundle's own order. Those two rules are this issue's own assertions; the parser is the consistency check underneath them, not a substitute.

Record all of the above in the local half of the reference and mark §10's open row resolved.

**One worked example per response shape** (`…` elides a value, never a member):

```json
{"protocol_version": 1, "request_id": "…", "registration_id": "…", "outcome": "first_mint",
 "material": {"role_id": "…", "wrapped_secret_id": "…", "ca_anchor": "…base64…", "expires_at": "…"},
 "registrar_health": {}}

{"protocol_version": 1, "request_id": "…", "registration_id": "…", "outcome": "already_absent",
 "registrar_health": {}}

{"protocol_version": 1, "request_id": "…", "registration_id": "…", "class": "permanent",
 "error": {"id": "ServiceHostMismatch"}, "registrar_health": {}}

{"protocol_version": 1, "request_id": "…", "class": "retryable", "registrar_health": {}}
```

The fourth is the **unclassified refusal form**: no `error`, `class` always `retryable`. It is the *identifier* that is unconditionally absent in that form; `registration_id` follows the ordinary rule below.

**Member order is canonical on the way out and irrelevant on the way in.** JSON objects are unordered, so member order is not something a conforming reader may depend on — but the fixtures are compared as bytes and a regeneration check has to be stable, so the order the codec *emits* is fixed here rather than left to whichever declaration order the first implementation picked. The rule is one line: **`serde` struct declaration order is the emitted order, and the declaration order is the reference's own table order for externally owned members, with the bootroot-owned envelope leading and `registrar_health` last.** That yields exactly these:

| Object | Emitted member order |
| --- | --- |
| `Register` request | `protocol_version`, `service_name`, `delivery_mode`, `host`, `instance`, `spec`, `wrap_ttl`, `idempotency_key` |
| `Deregister` request | `protocol_version`, `service_name`, `host`, `instance`, `idempotency_key` |
| `spec` | `component`, `service_name`, `reload`, `cert_group` |
| Mint success | `protocol_version`, `request_id`, `registration_id`, `outcome`, `material`, `registrar_health` |
| `material` | `role_id`, `wrapped_secret_id`, `ca_anchor`, `expires_at` |
| Deregister success | `protocol_version`, `request_id`, `registration_id`, `outcome`, `registrar_health` |
| Refusal | `protocol_version`, `request_id`, `registration_id`, `class`, `error`, `registrar_health` |
| `error` | `id`, then that identifier's own payload member (`reason` or `retry_after`) |
| decoded `ca_anchor` JSON | `trusted_ca_sha256`, `ca_bundle_pem` |

An omitted optional member is simply absent; the remaining members keep their relative order. **Decoding ignores order entirely**: a message whose members arrive in any order decodes, and re-encoding it yields the canonical order above. So what the round-trip test catches is a *codec* that emits a different order, not a *caller* that sent one — which is the only half of it that is contractual.

### The message shapes

**Requests.** Take the field set from the reference and nothing else — the two rows the order table above lists, and no other member. `instance` is `Option<u32>` on both. `spec` is the four-field `component` / `service_name` / `reload` / `cert_group` shape, with no privilege field.

**`spec.reload` and `spec.cert_group` cross this layer as opaque strings, and this issue does not interpret them.** The reference types both as newtypes over `String` (§2) and assigns their *contents* to the provisioning tool (§4.4). **How either string is spelled is recorded nowhere yet**, and it is not this repository's to choose; settling it is a prerequisite of the sibling issue that converts them into `ReloadSpec { kind, target }` and `cert_group: Option<u32>` (`src/registrar/config.rs:175`, `:214`), and that sibling is blocked on it. Keeping the interpretation out of this issue keeps that block off the codec: the field *names* are pinned by the reference and fixture-testable today, so the two strings ride through opaquely and no grammar is needed here. Do not narrow `spec` to the safe-set pair and do not collapse it to an opaque blob: all four names ride the wire, and the two identity-restatement checks read `component` and `service_name`.

**`delivery_mode` is carried on the request**, spelled `"LocalFile"` or `"RemoteBootstrap"` as the reference spells the variants (§4.5), which is **not** the kebab-case `local-file` / `remote-bootstrap` this repository serializes its own `DeliveryMode` with (`src/state.rs:318`; reference §8.2 records both sides). Do not reach for that local type's `serde` impl, and take the value from the wire rather than from local configuration.

**No request carries the caller identity, the request id, a composed name, a `registration_id`, the domain, a `token_policies` value or a policy body.** The caller identity is the transport-authenticated `CallerIdentity` the listener supplies. The rule is about the *source*, not today's implementation of it: no request shape may gain a field for it in any protocol version, and no test here may assert that peer credentials specifically are what arrives.

**Strictness is asymmetric, and it is a rule about direction.** *Serialization* emits exactly the fields listed here and nothing else; *deserialization* ignores fields it does not recognise, so a newer caller does not break an older daemon. The authority-bearing names above are the one carve-out: they are never read, in any version, so tolerance cannot become the channel through which a caller supplies its own identity.

**Responses decode as well as encode.** A caller on this socket has to read a refusal back off the wire and act on its identifier and class, so the response types round-trip in both directions and the refusal arm is expressible by a reader rather than only by this codec. Encode-only types would force an in-repository caller to re-derive the shape by hand.

**The mint success** carries the derived `registration_id`, the outcome class, the health container, and `material`: `role_id` and `wrapped_secret_id` from `MintOutcome` (`src/registrar/verbs/outcome.rs:201`), `expires_at` from `MintOutcome::expires_at`, and `ca_anchor`. Serialize what the verb returned; this layer does not unwrap, re-wrap, log or persist it — `MintOutcome::into_wrapped_secret_id` is consuming for exactly that reason. **`expires_at` is the granted absolute deadline, not the requested `wrap_ttl` echoed back**: `MintOutcome::expires_at` is the wrap token's creation time plus the TTL `OpenBao` actually reported, so both the registrar's clamp and `OpenBao`'s are already in it.

**`MintOutcome` carries no CA anchor**, so `ca_anchor`'s value has no source in the verb layer's result. Sourcing it is the handler issue's; this issue defines the field, its framing and its fixture.

**The deregister success** carries the derived `registration_id`, the outcome class and the health container. Every member is bootroot-owned: the external contract returns one bare acknowledgement for both outcomes, and collapsing them is the co-located registrar's translation step, not this codec's. They stay distinguishable here because the audit outcome classes and the operator reading them differ, and because a codec that fuses them leaves the registrar nothing to collapse from.

**The refusal** carries the request id, the class, the optional identifier, the optional `registration_id` and the health container.

**`registration_id` is read off the verb's envelope, not re-derived per variant.** `VerbContext::registration_id()` (`src/registrar/verbs/outcome.rs:144`) returns `None` for every outcome produced before derivation completed and `Some` for every one produced after, and both verbs already construct it that way: the derivation-arm refusal is built with `None` (`src/registrar/verbs.rs:469`, `:501`) because the composed candidate failed the rule that makes a value a `registration_id`. So the codec serializes the member when the envelope has one and omits it otherwise, and the failing candidate never reaches the response: it belongs in the audit record and the daemon's log, escaped and length-bounded there like every other caller-supplied byte string.

### The error mapping

**Enumerate the arms off the enums the compiler sees.** `VerbError` (`src/registrar/verbs/outcome.rs:325`) has seven variants, one of which wraps the whole of `RegistrarError` (`src/registrar/error.rs`, seventeen variants) and one of which wraps `WrapTtlRefusal` (`src/registrar/verbs/wrap_ttl.rs:23`, four variants). The `match` is exhaustive over all three, with **no `_ =>` arm**, so an unmapped or newly added variant is a compile error here rather than a wrong answer a caller acts on. The table below is the complete assignment as of `origin/main`; where it and the enum disagree, the enum wins and the difference is a finding to raise.

| Internal variant | Wire identifier | Class |
| --- | --- | --- |
| `ReservedServiceName` | none — unclassified form | `retryable` |
| `Registrar(InvalidServiceName)` | `ServiceLabelInvalid` | `permanent` |
| `Registrar(InvalidHost)` | `ServiceLabelInvalid` | `permanent` |
| `Registrar(ComponentNotConfigured)` | `ServiceInstanceMismatch` | `permanent` |
| `Registrar(ServiceInstanceMismatch)` | `ServiceInstanceMismatch` | `permanent` |
| `Registrar(SpecIdentityDisagreement)` | `ServiceSpecConflict` | `permanent` |
| `Registrar(ServiceSpecOutsideSafeSet)` | `ServiceSpecConflict` | `permanent` |
| `Registrar(DerivedKeyInvalid)` | none — unclassified form | `retryable` |
| `Registrar(ConfigUnreadable)`, `FingerprintLineMalformed`, `FingerprintMismatch`, `ConfigMalformed`, `UnsupportedSchemaVersion`, `UnknownMultiplicity`, `UnknownReloadKind`, `InvalidReloadTarget`, `InvalidDomain`, `InvalidComponentKey` | `RegistrarUnavailable { reason: NotProvisioned }` | `permanent` |
| `InvalidWrapTtl(_)` — all four | none — unclassified form | `retryable` |
| `RegistrationIdCollision` | `ServiceNameCollision` | `permanent` |
| `StoredSpecConflict` | `ServiceSpecConflict` | `permanent` |
| `HostMismatch` | `ServiceHostMismatch` | `permanent` |
| `Unavailable` | none — unclassified form | `retryable` |

**The two collapses onto a named identifier, and why they are collapses rather than a missing identifier.**

- **`ServiceSpecConflict` carries three members** — the safe-set refusal, the stored-spec disagreement, and the spec-identity disagreement. The last is *one* internal variant covering both `spec.component` and `spec.service_name` (`SpecIdentityField` records which, for the audit record's sake), so it is one member and not two. The collapse happens at this boundary and nowhere earlier: the internal types and the audit outcome classes keep the three apart because they mean different things to an operator.
- **`ServiceInstanceMismatch` carries two** — the instance-presence mismatch and the component absent from the multiplicity table. That second meaning is intentional and is recorded as such in the reference (§6.4), fixed by three merged design documents, and the limiter's rationale is written around `ServiceInstanceMismatch` being the cheapest refusal. Do not re-map it and do not widen `ServiceLabelInvalid` to take it.

**The unclassified refusal form, and its four members.** The identifier set has no generic member, by design (reference §7.1), so a failure that is not one of the seven is exactly what the caller reads as generic and classifies as retryable. That is the wire form for the four variants above, and the `error` member is optional for them and for nothing else. `class` is `retryable` beside the absent identifier — not because a retry will help in every case, but because a caller receiving no identifier classifies the failure as retryable by its own rule, so a `permanent` class there would assert something the caller cannot act on.

- **`Unavailable`** must not serialize as `RegistrarUnavailable`. Every reason in that closed set means *permanent until an operator acts*, and the caller's contract is to discharge an owed teardown on a permanent refusal but retry on a retryable one — so mapping a briefly unreachable `OpenBao` onto it converts a retryable failure into a silently dropped obligation. The cause stays in the daemon's log; a caller's only decision from a transient failure is to retry.
- **`Registrar(DerivedKeyInvalid)`** is **reachable**, and the mapping is written on that basis. `instance` is an `Option<u32>` and derivation formats it with `{instance:03}` (`src/registrar/identity.rs:242`), so a ten-digit value composes past the 131-octet bound while `service_name` and `host` stay inside their 63-octet label limits. Do not write this arm as unreachable, do not `expect()` it away, and do give it a round-trip test. Accept the consequence: a caller retries a request whose parts can only recompose the same unusable key until an operator renames a host or a component, while the audit record carries the internal reason and the limiter bounds the cost.
- **`ReservedServiceName`** refuses a wire `service_name` inside bootroot's reserved `bootroot-` namespace — a *valid* DNS label refused for what it names. No identifier in the closed set describes it: `ServiceLabelInvalid` refuses a name that is not a single DNS label (§6.2) and this one is, and the rest are about derivation, binding or spec state. It also must not be distinguishable on the wire: a distinct identifier would confirm to a prober which names are reserved, and that guard is what holds the non-self-propagation guarantee. The internal variant stays distinct in the audit record.
- **`InvalidWrapTtl`** refuses a zero, negative, sub-second or out-of-`OpenBao`-range `wrap_ttl`. Nothing in the identifier set is about a requested lifetime, and the requested value is the co-located registrar's own choice rather than an end user's, so an unusable one is a peer-side invariant violation on the same terms as the derived-key failure. All four `WrapTtlRefusal` variants take this form; each gets its own arm, not a wildcard.

**Do not extend the identifier set or the reason set to avoid any of the above.** Extending either is a cross-repo change made in review-protocol and transcribed afterwards, never made here.

**Three refusals the verb layer will eventually raise do not exist as variants yet, and their wire assignment is settled here anyway** — so that whatever adds each variant adds a `match` arm rather than making a wire decision at whatever point the compiler happens to stop. Record all three beside the mapping and in the local half of the reference.

- **An outcome audit record that could not be written after a successful mint** → `RegistrarUnavailable` under the reason `PostMintUnrecordable`; permanent. It is the one identifier-and-reason pair in the whole contract that leaves the caller's teardown owed, which is why it may not be folded in with anything else.
- **An intent audit record that could not be written, so the verb refused before creating anything** → `RegistrarUnavailable` under the reason `AuditUnwritable`; permanent. The internal variant that will carry this is expected to be spelled `AuditUnwritable` too, and **the names coinciding licenses no shortcut**: match the internal variant explicitly like every other arm and do not route it by name equality. It is a *different* refusal from the one above and must not be merged with it, because only one of the two leaves a teardown owed.
- **A refusal by the registrar's own rate limit** → `RegistrarBusy`, the one retryable identifier, with `retry_after` as whole seconds. This issue defines its wire form and its retryable class and covers it with a fixture; whatever later produces the internal variant supplies the arm.

**A transport refusal is not a wire error and appears nowhere in this mapping.** An operation identifier the daemon does not recognize is refused by the listener without writing a response, and the same will hold for the authorization refusal mTLS adds. The identifier set names neither, and neither may be invented nor overloaded to carry one.

### The reserved health container

Reserve a health field, **spelled `registrar_health`**, shaped as a container of named members rather than as any one payload's shape, and populate none of it. **Three member names are reserved with it and left empty: `certificates` for certificate lifetimes and renewal state, `limiter` for limiter pressure, and `audit_capacity` for audit-store capacity signals.** They are reserved in the type's doc comment and in the local half of the reference, and **not** in the fixtures, which pin the container's spelling, its placement and its current `{}` encoding and cannot pin a member nothing emits. Reserving the names now makes each of those three a purely additive change later: adding a member to a container is additive, growing a payload a second shape is not. Nothing about *what* each member reports, or how often it is refreshed, is settled here.

These four names are bootroot-owned and there is no fallback-to-the-reference arm for them — the transcribed half does not spell them and must not be asked to. In particular this is **not** the ecosystem's `audit_health` `AgentInfo` tail (reference §9): that is a different field on a different protocol, populated by the co-located registrar out of what it reads from this container, and nothing in this repository ever writes it.

**The container rides all three response shapes — mint success, deregister success and refusal — and that placement is fixed here.** The success-only shape is the tempting one and it is wrong, because each reserved member reports a condition whose arrival makes success responses stop. The audit store is fail-closed, so as it fills, invocations are refused and a success-only container would stop carrying the low-water alarm in the exact state the alarm exists to announce. A throttled caller receives refusals by definition. A lapsed certificate stops every enrollment in the deployment.

**Serialize it from a snapshot the encoder is handed; never compute it per request.** With no member populated yet, "does not compute health" cannot be tested as the absence of code — there is no code to be absent. So the rule is a *seam*, and this issue builds it:

- The protocol module defines a `RegistrarHealth` value type — `Debug`, `Clone`, `Default` — whose serialized form is today `{}`. `RegistrarHealth::default()` is its only constructor, and it reads nothing.
- **Every response encoder takes a `&RegistrarHealth` as an explicit parameter**, and that parameter is its only source for the member. Not a closure, not a trait object, not a channel, not a handle to the audit store, the limiter or the certificate state — nothing the encoder could invoke. Computing health inside the encoder is not expressible, which is what makes the property hold for the three member-owning issues rather than only for today's empty container.
- The protocol module correspondingly imports nothing from the audit store, the limiter or the certificate state, and a reviewer can confirm that from its `use` list.
- Encoding is a pure function of its arguments: the same outcome or refusal and the same snapshot encode to byte-identical output every time, and encoding never mutates the snapshot.

Producing the snapshot, deciding how often it is refreshed and threading it in from the daemon are the sibling issue's and the three member-owning issues'; this issue defines the parameter, the type and the empty encoding. Keep the value bounded and fixed-shape: it must never grow with request volume, with the number of registrations, or with the size of the audit store. A transport refusal carries nothing, because it writes no response bytes at all.

### The fixtures and the conformance test

"Generated from the reference" means the externally owned names in the fixtures are derived from and checked against the reference rather than retyped into a test. Three parts, all of them this issue's work:

1. **A checked-in test-only fixture generator** that constructs each message value and serializes it with the production codec into a committed file — one per message shape, including at least one permanent and one retryable refusal, and one of the unclassified form. Regenerating is a command an implementer runs; the output is committed, so a codec change shows up as a byte diff in review.
2. **A name-conformance test that reads `docs/reference/registrar-wire-contract.md`** and asserts in both directions that every externally owned field name, identifier and reason in the codec and the fixtures appears in the reference's tables, and that every one the tables list appears in the codec. The file's path is stable and is what the test reads it by; its rows are the contract and its prose is commentary.
3. **A decode-re-encode round trip per fixture**, asserting that re-encoding a decoded fixture reproduces the fixture **byte for byte** — so a dropped member, an added member, a changed encoding or a codec that emits the members in an order other than the canonical one above all fail. A separate case feeds each fixture back with its members shuffled and asserts it still decodes to the same value, so the byte comparison is not read as making decoding order-sensitive.

Bootroot-owned names — the framing, the operation identifiers, `protocol_version`, the envelope members, the health container, and the deregister success shape's members — are **not** subject to part 2. They are pinned by the fixtures — except the health container's three reserved member names, which no fixture can carry because nothing emits them yet, and which the local half of the reference and the type's doc comment pin instead.

### The local half of the reference

Add a clearly separated **bootroot-owned local half** beside the transcribed tables, recording every decision Scope settles and nothing else: the format and its enum representation, the version rule, the envelope members with their value types and canonical order, the `ca_anchor` framing and conversion (with §10's row marked resolved), the undecodable-payload rule, the full internal-variant-to-identifier mapping including all three many-to-one forms and the three not-yet-existing variants, and the health container's reserved member names, snapshot parameter, placement and member ownership. It may **not** add, rename, re-spell, reclassify or re-transcribe anything in the transcribed half, nor settle an external name that half left unresolved; a mapping that cannot be written because the transcribed half lacks a name is a finding raised against the reference, never filled in here. Keep the two halves visibly separated so a reader can tell at a glance which side of the mirror rule a line sits on.

Where this adds operator-facing text to the mirrored documentation pages, mirror it in `docs/en/` and `docs/ko/` alike, with no `mkdocs.yml` nav change.

## Acceptance criteria

- [x] The mapping is **complete over `VerbError`, `RegistrarError` and `WrapTtlRefusal` as the compiler sees them**, with no `_ =>` arm anywhere in it, and every variant's wire form — the identifier it maps to, or the deliberate absence of one — pinned with its class. Completeness is measured against the enums, not the table in Scope.
- [x] Exactly **two** many-to-one collapses onto a named identifier exist and no others: `ServiceSpecConflict` with three members and `ServiceInstanceMismatch` with two, every member of each reaching the intended shared identifier and class. The collapse is **total on the wire**: the `class` and `error` members are byte-identical across a collapse's members, with no discriminator, payload or message revealing which internal variant produced the response. The envelope's `registration_id` follows the derivation-stage rule alone and is never a variant discriminator.
- [x] A **third** many-to-one exists and is not an identifier collapse: the unclassified refusal form, carrying no `error` member and `class` of `retryable`, whose members are exactly `Unavailable`, `Registrar(DerivedKeyInvalid)`, `ReservedServiceName` and all four `InvalidWrapTtl` variants, and no other variant takes that form.
- [x] `Unavailable` serializes with **no typed identifier at all** — not `RegistrarUnavailable`, not any other, and not a placeholder, sentinel or empty string — with the retryable class intact: no `error` member is emitted and the response reads as retryable.
- [x] `Registrar(DerivedKeyInvalid)` is exercised through a request whose parts each validate but whose composed key exceeds 131 octets, and the response carries the request id, no `registration_id`, no `error` member and `class` of `retryable`, with the failing candidate absent from the response bytes.
- [x] `ReservedServiceName` and each of the four `InvalidWrapTtl` variants have their own arm and their own round-trip test, and none of them emits an identifier from the closed set.
- [x] The ten config-loading `RegistrarError` variants map to `RegistrarUnavailable { reason: NotProvisioned }` with the permanent class, each with its own arm.
- [x] The three not-yet-existing variants — `PostMintUnrecordable`, `AuditUnwritable` and `RegistrarBusy` — have their assignment recorded beside the mapping and in the local half of the reference, and `RegistrarBusy`'s wire form, its `retry_after` whole-second payload and its retryable class are defined and fixture-covered even though no verb variant produces it yet.
- [x] `protocol_version` behaves exactly as Scope states, covered by one test so the version rule and the unknown-field tolerance cannot be read as one rule: required in both directions, equal to the `PROTOCOL_VERSION` constant, absent or unimplemented failing to decode on a request and a response alike, and an unknown field at the implemented version still decoding. The rejection comes from the `ProtocolVersion` type's own `Deserialize`, so neither direction can be given a derived decoder that accepts an arbitrary integer.
- [x] The bootroot-owned wire names are exactly those spelled in Scope — `protocol_version`, `request_id`, `registration_id`, `outcome` with its four variant spellings, `material`, `class` with its two, the optional `error` with its `id` member, and `registrar_health` — with each member's JSON value type and the canonical emitted order as stated there, all pinned by the golden fixtures so a reviewer can confirm none was renamed, added or dropped.
- [x] `ca_anchor` is a standard **padded** base64 string whose decoded bytes are exactly the compact UTF-8 JSON object Scope spells out — both members required and non-empty in the stated order, fingerprints lowercase hex over DER and positionally matching the bundle, LF-only PEM with one trailing LF, and nothing else in the object. One codec-owned pair of functions converts between `TrustPayload` and the base64 member — base64, `serde_json::from_slice` to a `serde_json::Value`, then `parse_trust_payload(&value)` (`src/kv_payload.rs:48`) — and no second conversion exists in the codec. That parser accepts a mint fixture's converted value; the lowercase-hex and positional-order rules and the rest of the framing are asserted separately at the byte level, because the parser checks neither. §10 of the reference no longer lists an unresolved item.
- [x] Every member carries the JSON value type tabulated in Scope, including the ones two implementations could differ on: `instance` unsigned and unquoted in the `u32` range, `delivery_mode` as `"LocalFile"` / `"RemoteBootstrap"` rather than this repository's `local-file` / `remote-bootstrap`, `wrap_ttl` and `retry_after` as whole-second integers, and `expires_at` as an RFC 3339 UTC string.
- [x] An absent optional member is **omitted** rather than serialized as `null`, for `error`, `registration_id`, `instance` and `spec.cert_group` alike — a missing key, never a null.
- [x] A payload the codec cannot decode — an absent required member, a wrong JSON type, an out-of-range `instance` or `wrap_ttl`, an unknown `delivery_mode` — yields a typed decode `Err` and **no response bytes and no wire refusal**: the codec neither constructs a `HandlerRefusal` nor emits a response.
- [x] All seven identifiers and all six `RegistrarUnavailable` reasons **decode**, whether or not this daemon emits them. The conformance test measures the decodable sets against the reference's tables and its §11 counts, not against the subset the mapping emits.
- [x] Serialization emits members in the canonical order tabulated in Scope, and decoding is order-insensitive: each fixture re-encodes byte for byte, and the same fixture with its members shuffled decodes to the same value.
- [x] `registration_id` is serialized from `VerbContext::registration_id()` and from nowhere else, so it is present exactly when the envelope carries one: a pre-derivation and a during-derivation refusal omit it, a post-derivation one carries it.
- [x] **Serialization** emits exactly the field set listed in Scope and nothing else; **deserialization** accepts and serves a message carrying an unknown field. A test covers each direction.
- [x] No request shape has a field for a caller identity, the request id, a composed name, a `registration_id`, a domain, a `token_policies` value or a policy body; a request supplying those names anyway decodes with them never read, and the identity carried through is the one supplied to the codec rather than a peer-credential value, so the mTLS swap does not rewrite it.
- [x] `spec` carries all four fields, with `reload` and `cert_group` as opaque strings that this codec does not interpret; a test asserts a spec round-trips byte-identically without either being parsed.
- [x] The response types decode as well as encode, and a refusal is expressible by a caller reading the wire, not only by this codec.
- [x] Both success outcome classes survive serialization: first mint versus idempotent re-mint, and identity removed versus idempotent already-absent, with no collapse of the deregister pair.
- [x] The mint success carries the granted absolute deadline as an RFC 3339 instant taken from `MintOutcome::expires_at`, distinct from the requested `wrap_ttl`.
- [x] The codec matches the reference exactly on every externally owned name, enforced by a **name-conformance test that reads `docs/reference/registrar-wire-contract.md`** and fails in both directions, so no externally owned name reaches the codec, a fixture or a test from this issue's text rather than from the reference.
- [x] Bootroot-owned names are pinned by the fixtures and are **not** asserted against the transcribed half; a reviewer can confirm the conformance test does not demand the reference carry them.
- [x] The golden fixtures are emitted by a checked-in test-only generator running the production codec, committed as encoded files, and cover every message shape including at least one permanent refusal, one retryable refusal and one of the unclassified form; a regeneration check matches the committed fixtures against what the generator emits from the current codec, **byte for byte**.
- [x] `registrar_health` exists under that spelling, documented as a local endpoint name distinct from the ecosystem's `audit_health` tail, is a container of named members, and is present with an identical shape on **all three** response shapes — mint success, deregister success and refusal, permanent and retryable alike.
- [x] The container is serialized from a snapshot the encoder is **handed**: a `RegistrarHealth` value type exists with `Default` as its only constructor, every response encoder takes it as an explicit `&RegistrarHealth` parameter and has no other source for the member, and the protocol module imports nothing from the audit store, the limiter or the certificate state. All four response cases encoded from one snapshot carry identical `registrar_health` bytes, and encoding is byte-reproducible for the same arguments — so the encoder signatures show per-request computation is not expressible, not merely absent.
- [x] The container is documented as the single place later signals are added, with its three member names assigned to the three issues that populate them and the placement rule stated so no member owner narrows it.
- [x] The encoding, the protocol version, the message shapes, the completed mapping, the resolved `ca_anchor` framing and the health container's names, placement and member assignment are documented in a clearly separated bootroot-owned local half of the reference; the transcribed half is byte-unchanged, with no external name, class or reason added, renamed, reclassified or re-transcribed. Operator-facing text is mirrored in `docs/en/` and `docs/ko/`, with no `mkdocs.yml` nav change.
- [x] `cargo clippy --all-targets` is warning-free under `-D warnings` and `cargo fmt -- --check --config group_imports=StdExternalCrate` passes.

## Constraints

- Keep the protocol boring and versioned enough that adding a field later does not break a running caller. **Adding a field is the compatible change; changing what one means is not** — that is what the version is for, and the daemon serves only versions it implements rather than guessing at one it does not.
- Do not rename, add or drop a bootroot-owned wire member at implementation time, and do not leave one to the codec's default. Every one is spelled in Scope precisely so two implementations cannot produce incompatible JSON while both satisfying the prose; a member that needs to change is a change to this issue's text, not a local choice.
- Map the verb layer's error types with an **exhaustive `match`**, never a catch-all `_ =>` arm. A catch-all is exactly what turns an unmapped refusal into a silently wrong identifier — including the two pre-derivation refusals, which map to *different* identifiers and which a catch-all would fuse.
- Treat the transcribed half of `docs/reference/registrar-wire-contract.md` as the sole source of every externally owned name, and as read-only from this issue. Do not invent, rename, merge, re-spell or reclassify an identifier, a reason or a field, and do not guess a spelling to unblock a test — a guessed name is a cross-repo break that surfaces at deploy time. Writing to the reference is permitted only in the bootroot-owned local half.
- Do not extend the identifier set or add a `RegistrarUnavailable` reason. In particular, do not create one for the spec-identity disagreement, the derived-key format failure, the reserved-name refusal, an unusable `wrap_ttl`, or a transport-level refusal — each has a settled assignment in Scope.
- Do not read `idempotency_key` for control flow or derive idempotence from it. The durable binding and the stored-spec comparison are what make a re-mint idempotent; a key-keyed shortcut would let a caller replay or split an identity's history by choosing keys.
- Do not take the caller identity from the wire, in any protocol version, under any field name.
- Do not interpret `spec.reload` or `spec.cert_group`. They cross this layer as opaque strings.
- Do not touch the socket, the activation contract, the peer-credential check, the enablement gate, the framing, the transport's bounds or the unit pair. A frame reaching this codec is already length-bounded, so add no second size check; an over-long frame never reaches decoding at all.
- Do not register a production handler, construct `RegistrarVerbs`, or change daemon startup. Those belong to the sibling issue.
- Do not populate the health container. All three of its members land elsewhere.
- Do not treat a bootroot-owned name as externally owned, and in particular do not rename `registrar_health` toward the ecosystem's `audit_health` tail or reshape it to match.
- Do not retype an externally owned name into a test or a fixture by hand. Names reach the codec and the fixtures from the reference, and the conformance test is what proves it.
- No `unwrap()` in production code; no `[]` indexing; `thiserror` for error types; `pub(crate)` visibility unless a compiler error forces wider.

## Out of scope

- Everything the listener already merged: the socket and its activation, the root-owned directory rule, peer-credential authentication, the off-by-default gate, the unit pair, the operation enumeration and its dispatcher, and the transport's frame ceiling, timeouts and connection cap.
- The production `RegistrarRequestHandler`, provisioning `RegistrarVerbs` inside the daemon, lifting the no-handler startup refusal, and withdrawing the documentation's "unsupported" statement — the sibling issue's, which depends on this one.
- Converting the wire `spec` into the verb layer's `RequestedSpec`, and settling the `reload` and `cert_group` string spellings that conversion needs. The spellings are externally owned and unrecorded; neither this issue nor its sibling may choose one.
- Sourcing the `ca_anchor` value. This issue defines the field, its framing, its conversion functions and its fixture; the handler supplies the anchor to convert.
- Settling, negotiating or transcribing any externally owned spelling, and any edit to the reference's transcribed half.
- mTLS termination, client-certificate verification, and scoping the endpoint to a caller identity.
- Populating the reserved health container's three members.
- The verb layer's refusal taxonomy, outcome classification, arms or locks; this issue maps what they produce and changes none of it.

## Test plan

Every case here runs in-process under plain, unprivileged `cargo test`. Build refusals and outcomes directly from the verb layer's public constructors (`VerbRefusal::new`, `MintOutcome::new`, `DeregisterOutcome::new`) rather than by driving a verb.

- [x] Round-trip test per variant of `VerbError`, `RegistrarError` and `WrapTtlRefusal`, asserting the exact identifier or its deliberate absence, plus the preserved class. The covered set is read off the enums, so a variant added later has no arm and fails to compile.
- [x] Collapse tests driving all three members of `ServiceSpecConflict` and both of `ServiceInstanceMismatch` as separate cases, asserting each reaches the shared identifier and class with `class` and `error` byte-identical across a collapse's members. Distinguishability survives only in the internal variant and the audit record, neither of which this issue changes.
- [x] Unclassified-form test enumerating all seven members (`Unavailable`, `DerivedKeyInvalid`, `ReservedServiceName`, four `InvalidWrapTtl`), asserting no `error` member, `class` of `retryable`, and that no other variant takes the form.
- [x] `DerivedKeyInvalid` test built from a ten-digit `instance` with 63-octet `service_name` and `host`, asserting the request id, the absent `registration_id`, the absent `error`, the retryable class, and the candidate's absence from the response bytes.
- [x] `registration_id` presence test over a pre-derivation envelope, a derivation-arm envelope and a post-derivation envelope.
- [x] Version tests, each run against both a request type and a response type so neither direction is covered only by inference: a message with no `protocol_version` fails to decode; a message at an unimplemented version fails to decode; a message at the implemented version carrying an unknown field decodes; every response the codec emits carries `PROTOCOL_VERSION`.
- [x] Value-type tests: `wrap_ttl` and `retry_after` as whole-second integers, `expires_at` as an RFC 3339 UTC string, `ca_anchor` as a padded base64 string rather than a numeric array, `instance` as an unquoted `u32`, `delivery_mode` as `"LocalFile"` / `"RemoteBootstrap"`.
- [x] Decode-failure tests: an absent required member, a wrong JSON type, a negative and a fractional `wrap_ttl`, an `instance` past `u32`, and an unknown `delivery_mode` each produce a typed decode `Err` with no response bytes emitted.
- [x] Decodable-set tests: a response carrying each of the seven identifiers and each of the six `RegistrarUnavailable` reasons decodes, including the reasons this codec never emits.
- [x] Member-order tests: re-encoding each fixture reproduces it byte for byte in the canonical order, and the same fixture with its members shuffled decodes to the same value.
- [x] `ca_anchor` framing tests: run a mint fixture's member through the codec's own base64 → `serde_json::Value` → `parse_trust_payload` path, with negative cases for a fingerprint that matches no certificate and for decoded bytes that are not valid JSON; assert the decoded bytes are compact JSON in the stated member order, LF-only PEM, no trailing newline, lowercase hex and positional fingerprint order — the last two because the parser checks neither; and round-trip a `TrustPayload` to the member and back.
- [x] Optional-member omission test decoding a fixture of each refusal form and asserting absent members are missing keys, not nulls.
- [x] Direction tests: serialization emits exactly the listed field set; deserialization accepts an unknown field. Paired so the two rules cannot be read as contradicting each other.
- [x] Authority carve-out test sending each of the forbidden names explicitly and asserting none is read and that the identity carried through is the one supplied to the codec.
- [x] `spec` opacity test asserting all four fields round-trip and that `reload` and `cert_group` are neither parsed nor validated here.
- [x] Response-decode test asserting a caller can read each response shape back off the wire, including a refusal with its identifier and class intact.
- [x] Outcome-class tests asserting first-mint versus idempotent-remint and removed versus already-absent survive the round trip without collapsing.
- [x] Granted-deadline test asserting the mint response carries `MintOutcome::expires_at` rather than the requested `wrap_ttl`, driven through an outcome whose deadline is shorter than the request.
- [x] `idempotency_key` test decoding two requests that differ only in that member: both decode, each preserves its own key verbatim and re-encodes to the byte string it came from, and the two decoded values are equal in every other member and differ in exactly that one — it is a required, opaque member, so they *must* differ, and what the test proves is that the codec carries it through unread. A reviewer confirms the rest by inspection: the member appears in the struct, the serializer and the deserializer, and in no branch, comparison or lookup.
- [x] Golden fixtures for every message shape, emitted by the checked-in generator and committed, decoded and re-encoded by the tests. The refusal shape gets its own — at least one permanent, one retryable and one unclassified — carrying the empty health container, so the three member-owning issues extend a pinned refusal encoding rather than introducing one.
- [x] Regeneration check asserting the committed fixtures match what the generator emits from the current codec.
- [x] Name-conformance test reading the reference's tables and failing in both directions, with the bootroot-owned names excluded by construction.
- [x] Health-container tests: present with an identical shape on all four response cases; all four encoded from one `RegistrarHealth` value carry identical `registrar_health` bytes; encoding the same arguments twice is byte-identical.
- [x] `./scripts/check-docs.sh` passes for the documentation changes.

## Dependencies

**Nothing blocks this issue: every input it needs is already in `origin/main`.** The listener is there — the socket, the activation contract, the framing, the `Operation` enum, the transport's bounds, the peer-credential identity, the single pre-verb refusal path, and the `RegistrarRequestHandler` trait whose payload half this issue defines. So is the verb layer: `VerbError`, `RegistrarError`, `WrapTtlRefusal`, `MintOutcome`, `DeregisterOutcome`, `VerbContext` and `RequestId` all live under `src/registrar/`, and this issue maps them without changing any of them. So is the wire-contract reference, complete and with its one open row assigned here.

Part of the registrar endpoint group under the registrar endpoint umbrella.

## Pointers

- `src/registrar/endpoint/handler.rs` — the `RegistrarRequestHandler` trait, `HandlerRefusal`, and the seam this payload rides
- `src/registrar/endpoint/frame.rs:42` (`Operation`) — the two operations a payload is decoded for
- `src/registrar/verbs/outcome.rs` — `VerbError:325`, `VerbContext:144` (`registration_id`), `MintKind:189`, `DeregisterKind:270`, `MintOutcome:201`, `RequestId:34`, `CallerIdentity:70`
- `src/registrar/error.rs` — `RegistrarError`'s seventeen variants and `SpecIdentityField`
- `src/registrar/verbs/wrap_ttl.rs:23` (`WrapTtlRefusal`) — the four variants behind `VerbError::InvalidWrapTtl`
- `src/registrar/identity.rs:231` (`derive_registration_id`) and `src/input_validation.rs` (`validate_registration_id`) — where `DerivedKeyInvalid` comes from and why it is reachable
- `docs/reference/registrar-wire-contract.md` — the sole source of every externally owned name; §6.1 the identifier set, §6.5 the reason set, §7.1 the absent generic member, §8.1 the anchor encodings, §10 the open row this issue closes
- `src/kv_payload.rs:48` (`parse_trust_payload`, which takes `&serde_json::Value`) and `src/trust_bootstrap.rs:12`/`:13` — the existing anchor shape the `ca_anchor` framing reuses
- `src/state.rs:22` (`ReloadStrategy`) — the tagged-enum `serde` idiom; `src/state.rs:318` (`DeliveryMode`) — this repository's own delivery-mode spellings, which are not the wire's
- `src/registrar/audit.rs`, `Cargo.toml:8` (`base64`), `:25` (`serde_json`)
- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §4, §5.1–§5.2 and §6

sehkone:
## Summary

- Add the registrar endpoint versioned JSON request/response protocol, codec, exhaustive refusal mapping, and snapshot-supplied health container.
- Define canonical TrustPayload base64 framing and commit golden fixtures plus reference-name conformance tests.
- Resolve the local wire-contract item and update English and Korean endpoint documentation.

Closes #885

## Test plan

- [x] Run `cargo fmt -- --check --config group_imports=StdExternalCrate`.
- [x] Run `cargo clippy --all-targets -- -D warnings` and `cargo doc --no-deps --document-private-items` with `RUSTDOCFLAGS=-D warnings`.
- [x] Run `./scripts/check-docs.sh`.
- [x] Compile the Linux endpoint and its unit-test target with `cargo check --target aarch64-unknown-linux-gnu` and `cargo test --target aarch64-unknown-linux-gnu --lib --no-run`.


sehkone:
**[Reviewer Round 1]**

## Review findings

- P1: [protocol.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:184) accepts a full register payload when its frame operation is `Deregister`. `DeregisterRequest` shares all required fields and silently ignores `delivery_mode`, `spec`, and `wrap_ttl`, so the production handler could tear down an identity when given a valid mint body under the wrong operation. Retain generic unknown-field tolerance, but reject members that are known to belong to the other v1 request shape.

- P2: The required shuffled-fixture decode coverage is missing. [golden_fixtures_decode_and_reencode_byte_for_byte](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:1729) only decodes canonical generated fixtures; it never shuffles their object members. This leaves the explicit order-insensitive decoding acceptance criterion untested.

sehkone:

sehkone:
**[Author Round 1]**

Both review items were fixed and pushed in commit `46d0e817`.

- Fixed P1: deregistration frames now reject all v1 mint-only members (`delivery_mode`, `spec`, `wrap_ttl`) while retaining tolerance for genuinely unknown fields.
- Fixed P2: every golden fixture is now recursively member-shuffled, decoded, and re-encoded to its canonical bytes.

Validation passed:

- Linux container protocol tests: 20 passed.
- Cross-target test compilation and strict Clippy passed.
- Preflight quality/docs/core stages passed; the E2E matrix stopped later because `172.17.0.1` is not assigned to a local interface on this machine.

PR description remains accurate and still includes `Closes #885`: [PR #891](https://github.com/aicers/bootroot/pull/891).

sehkone:
**[Reviewer Round 2]**

- P2: Commit `46d0e817` has a malformed body: it contains literal `\n\nCloses #885` characters rather than real line breaks. As a result, the issue reference is not a proper commit-body trailer and violates the repository’s commit-message requirements. Amend the commit with a real multiline body. The PR body itself correctly includes `Closes #885` and a test plan.

Round 1 items are resolved:

- The deregister decoder now rejects each known mint-only member while retaining unknown-field tolerance.
- The golden-fixture test now recursively shuffles object members and verifies canonical re-encoding.

sehkone:

sehkone:
**[Author Round 2]**

- Fixed P2: amended the prior fix commit so its body has real, 72-column-wrapped lines and a separate Closes #885 trailer. The history-only amendment is now commit bc968d71 and has been force-with-lease pushed.

The PR description remains accurate and retains Closes #885. A new CI run is queued.

sehkone:
**[Author Round 2]**

Review feedback addressed and pushed.

- Fixed — Round 2 P2: amended the prior commit with real multiline, 72-column-wrapped body text and a separate `Closes #885` trailer. New commit: `bc968d71`.
- Already resolved — Round 1 P1/P2 remain fixed as reviewer confirmed: cross-operation request members are rejected, and shuffled fixtures verify order-insensitive decoding.

The PR description was rechecked and remains accurate with `Closes #885`; I left it unchanged. I also posted the round response on [PR #891](https://github.com/aicers/bootroot/pull/891).

The force-with-lease push succeeded and the new CI run is queued. The prior full CI suite was green.

sehkone:
**[Reviewer Round 3]**

- P2: The required `ca_anchor` framing coverage is incomplete. [protocol.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:1630) tests uppercase fingerprints and missing members, but not a valid-base64 value whose bytes are not JSON, nor a lowercase 64-hex fingerprint that does not match the PEM bundle. It also only asserts successful decoding at [line 1521](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:1521), rather than round-tripping the decoded `TrustPayload` and comparing it to the original. Add those cases required by the issue’s framing test plan.

Round 2’s prior item is resolved: `bc968d71` has a real multiline, wrapped commit body with a separate `Closes #885` trailer. The earlier cross-operation rejection and shuffled-fixture coverage remain correctly present.

sehkone:

sehkone:
**[Author Round 3]**

- Fixed P2: added `ca_anchor` rejection coverage for valid base64 whose decoded bytes are not JSON and for a syntactically valid lowercase 64-hex fingerprint that does not match the PEM bundle. The successful path now decodes the anchor and compares the returned `TrustPayload` with the original.

Validation passed: Linux container protocol tests (23 passed), cross-target test compilation, formatting, strict Clippy, Rustdoc, and the quality/docs/core preflight stages. The E2E matrix completed the no-hosts lifecycle, then stopped at the hosts lifecycle because this machine lacks non-interactive `sudo -n`; the script requires it for that mode.

The PR description remains accurate and retains `Closes #885`.

sehkone:
**[Author Round 4]**

Both P2 findings are fixed in commit `2a5fdcd4`.

- Fixed — classification conformance: the test now parses every identifier and its `Class` column from the mirrored reference, checks the codec accepts that pairing, rejects its opposite, and compares the complete identifier-to-class mapping.
- Fixed — external-name sourcing: invalid wire cases are now produced by serializing typed valid values and mutating only the value under test. The redundant literal-spelling assertion was removed because golden fixtures and the reference conformance suite already pin the encoded names.

Validation passed:

- `cargo fmt -- --check --config group_imports=StdExternalCrate`
- `cargo clippy --target aarch64-unknown-linux-gnu --all-targets -- -D warnings`
- `cargo test --target aarch64-unknown-linux-gnu --lib --no-run`
- Linux-container protocol suite: 19 passed, 1 intentional fixture-print test ignored

The PR description was rechecked and remains accurate with `Closes #885`.

sehkone:
**[Author Round 4]**

Fixed both Round 4 findings and pushed `2a5fdcd4`.

- Fixed: reference-driven identifier classification conformance.
- Fixed: tests now derive valid external wire values from typed serialization and mutate only the invalid aspect.

Validation passed: formatting, Linux cross-target Clippy/test compilation, and Linux protocol suite (19 passed).

PR description remains accurate; Round 4 response was posted. CI is running.

sehkone:
**[Reviewer Round 5]**

Round 4 classification conformance is resolved: the test now reads each identifier’s `Class` from the reference and checks both the accepted and opposite class.

Still unresolved:

- P2: externally owned request field names are still manually retyped in test JSON, contrary to the issue’s explicit constraint. For example, [protocol.rs:905](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:905)–[920](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:920) hard-code `service_name`, `host`, `instance`, and `idempotency_key`; [line 1461](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:1461) does the same for the nullable-option case. Build a valid typed request (or use a generated fixture), serialize it, then remove/mutate the member under test. This preserves the reference-backed conformance path and prevents these tests from silently retaining stale external spellings.

sehkone:

sehkone:
**[Author Round 5]**

Fixed — P2: request decode-failure tests no longer hand-write externally owned request-field names in JSON. They now serialize typed DeregisterRequest values and remove or replace the serialized member under test; the unknown-member case likewise starts from the typed serialized request. This retains the reference-backed conformance path while preserving the missing-version, unsupported-version, invalid idempotency-key, out-of-range instance, and null optional-member coverage.

Pushed commit 0e48ffae (Derive malformed request payloads).

Validation passed:

- scripts/preflight/ci/check.sh
- cargo test --target aarch64-unknown-linux-gnu --lib --no-run
- Linux container protocol suite: 22 passed, 1 intentional fixture-printer ignored

I also attempted scripts/preflight/run-all.sh --skip-hosts; its quality and static stages passed, then its Docker deployment smoke stopped on Docker Desktop reporting no such object for a container object. The test change does not affect that deployment path. The PR description remains accurate and retains Closes #885.

sehkone:
**[Author Round 5]**

Fixed — P2: request tests now serialize typed requests and mutate them, eliminating hard-coded external wire-field names.

Committed and pushed `0e48ffae` (`Derive malformed request payloads`) to PR #891. The PR description remains accurate.

Validated:

- `scripts/preflight/ci/check.sh`
- Cross-target test compilation
- Linux protocol tests: 22 passed, 1 ignored

The full preflight stopped in Docker deployment smoke due to Docker Desktop reporting a missing container object; CI is running.

sehkone:
**[Reviewer Round 6]**

[Reviewer Round 6] — NOT_APPROVED

Round 5’s P2 is resolved: malformed request tests now derive payloads from serialized typed requests.

New findings:

- P2: The local reference does not record the required full internal-refusal mapping. It only summarizes categories in [registrar-wire-contract.md](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/docs/reference/registrar-wire-contract.md:472), omitting the individual `RegistrarError` and `WrapTtlRefusal` variants and their explicit arms. The issue requires the complete mapping, including every collapse and future assignment, in the local half of the reference.

- P2: The required reachable `DerivedKeyInvalid` test is absent. The tests only construct that error manually ([protocol.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:1208)); they never derive it from individually valid request parts that compose beyond 131 octets, nor verify the encoded refusal omits both `registration_id` and `error`. This leaves the specifically called-out reachability path untested.

- P2: `RegistrarHealth` is publicly constructible throughout the crate as a unit struct ([protocol.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-885-review/src/registrar/endpoint/protocol.rs:261)), contrary to the contract that `RegistrarHealth::default()` is its only constructor. Make its representation non-constructible outside this module while preserving decoding and `Default`.

The PR body has appropriate `Closes #885` linkage and a test-plan checklist; I found no indication that an implementation-side GitHub issue was created.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add registrar wire protocol and codec
```

**Body**

```text
Define the registrar endpoint's versioned JSON wire protocol with
canonical trust-payload framing, exhaustive refusal conversion, golden
fixtures, and reference conformance checks.

Closes #885
```

PR #893, opened 2026-08-24
Harden audit-status failure reporting (#892)

### LINKED ISSUE #892: Harden audit-status failure reporting
# Harden audit-status failure reporting

## Context

`bootroot status` reports registrar-audit health while preserving the rest of an operator’s status output when audit settings cannot be loaded or the store cannot be scanned. An explicitly supplied `--agent-config` path that is absent at the metadata gate is the sole intended command error; every other audit-related failure belongs in the audit section’s failed-scan state.

The optional file source used by `Settings::from_file` deliberately treats an absent or unreadable source as defaults. It therefore cannot safely load a supplied path: a directory, an unreadable regular file, a regular file with an unsupported extension, or a path that vanishes after inspection could otherwise become a scan of the default store. Explicit status configuration needs a strict, file-only loading path, while the existing metadata gate remains responsible for the distinct missing-path command error and for recognizing a known non-regular-file condition.

Two failures still need hardening. The settings `spawn_blocking` join is propagated with `?`, so a task panic — or cancellation during runtime shutdown before a blocking task begins — can abort the command instead of rendering a failed scan. This differs from the existing scanner `TaskJoin` mapping, which is already a failed scan. Also, the metadata gate turns an explicit non-regular file into a hand-written English `String`; that implementation-owned reason is embedded in the localized failed-scan warning instead of being translated as a known status condition. Other error values in the warning remain diagnostic text and are not translated by this issue.

## Scope

- Add a `Settings` constructor for one explicitly supplied file that has no environment overlay and makes its `File` source `required(true)`. Keep `Settings::from_file` unchanged for callers that intentionally accept a missing optional source. Use the strict constructor only for `bootroot status` after an explicit path has passed the metadata gate.
- Keep the explicit-path metadata gate, with these outcomes: no `--agent-config` keeps the optional default-file behavior; an explicit `NotFound` path at the metadata gate remains the command error; an explicit non-regular path becomes a typed failed-audit condition; and any other metadata error remains a failed audit status with its underlying error. A regular explicit file is loaded through the strict constructor, so read, unsupported-format, parse, deserialization, and check-then-use failures become failed audit statuses rather than default settings.
- Represent the explicit non-regular-file condition without a pre-rendered English `String`, and render its reason through a new `Messages` accessor and English/Korean catalogue entries inside the existing localized failed-scan warning.
- Convert a settings-loading task join failure into `AuditStatus::Failed`, using the selected settings path and the join error text, so normal status collection and rendering continue. Extract the join-result mapping into a small testable helper rather than adding a test-only input to `StatusArgs`.
- Preserve the current scan mapping: `StoreAbsent` is not configured and every other scanner error, including `TaskJoin`, is a failed scan.
- Continue to validate only `validate_registrar_settings`; do not call `Settings::validate()`.

## Acceptance criteria

- [x] An explicit missing `--agent-config` path returns the existing command error before status output only when it is absent at the metadata gate, and no `--agent-config` continues to allow a missing default `agent.toml`.
- [x] An explicit directory, unreadable regular file, regular file with an unsupported extension, path that fails metadata inspection, or path that fails after inspection never scans the default audit-store settings; it renders the audit section as failed and names the supplied path.
- [x] The directory/non-regular-path warning states the condition in the selected locale, including a Korean explanation in Korean output, while the warning still identifies the supplied path.
- [x] A settings task join failure returns `AuditStatus::Failed`, identifies the selected settings path, retains the join error text, and does not abort normal status collection.
- [x] Parse failures, invalid registrar settings, absent stores, and existing scanner failures retain their present audit-state mappings.

## Constraints

- Do not change the audit record format, scanner semantics, configuration keys, or the status command’s pre-existing infrastructure/OpenBao exit conditions.
- Do not change `Settings::from_file`'s optional-source contract. The new strict constructor reads exactly one supplied file, has no environment overlay, and is used only by this status-loading path.
- An explicit path absent at the metadata gate remains the command error; an explicit path that passes the gate and then fails to load becomes a failed-scan warning, even if the later error is `NotFound`.
- Keep blocking settings I/O off Tokio runtime workers.
- Do not add `Settings::validate()` or duplicate registrar validation.

## Out of scope

- Translating arbitrary I/O, configuration-parser, validation, or task error text carried in failed-scan warnings; this issue localizes only the known explicit non-regular-file condition.
- Changing the audit scan’s path trust checks, file selection, pairing, retention derivation, or lookback window.
- Adding audit-status fields, changing clean or anomalous audit output, or expanding the registrar endpoint health relay.
- Changing the reserved audit-store layout, reserve, capacity reporting, or unrelated status sections.
- Editing `docs/en/operations.md` or `docs/ko/operations.md`: they document the scan's successful observations, not exact configuration-load failure wording, and this issue preserves the documented failed-scan versus missing-config behavior.

## Test plan

- [x] Test the optional and strict `Settings` file-loading contracts independently: a missing optional source yields defaults, while a missing strict source fails rather than yielding defaults. Test an existing regular explicit file with an unsupported extension through the strict constructor and through `load_audit_status`, verifying both fail rather than load defaults; this existing file is the deterministic post-metadata-gate source failure.
- [x] Cover every explicit metadata-gate outcome separately: no explicit path with no default file, missing explicit path, directory/non-regular path, an underlying metadata error where deterministic, a valid parseable file, and malformed settings. Verify that a non-regular path cannot yield `AuditStatus::Absent` from default settings.
- [x] Test that an unreadable regular explicit file maps to a failed audit state rather than default settings by removing its read permissions; skip this permission-based test when the effective UID is 0 because root can read the file despite its mode.
- [x] Drive a deliberately panicking `spawn_blocking` task to obtain a real `JoinError`, pass it to the extracted join-result helper, and verify the selected path and join text produce `AuditStatus::Failed`.
- [x] Render the known non-regular-file failure in English and Korean, asserting each warning contains the path and its locale's explanation; retain the existing rendering coverage for diagnostic error text.
- [x] Confirm invalid registrar settings, absent stores, and representative scanner failures keep their existing audit states.
- [x] Run `cargo fmt -- --check --config group_imports=StdExternalCrate` and `cargo clippy --all-targets -- -D warnings`. Run both `cargo test --lib` for the `Settings` constructor tests and `cargo test --bin bootroot` for the `commands::*` status tests; neither target alone covers this issue. Run `scripts/preflight/run-all.sh`; the E2E matrix may be skipped because this change touches neither Docker lifecycle nor E2E scripts or paths. Do not run `./scripts/check-docs.sh`, since no documentation changes are in scope.

## Dependencies

Part of #780.

## Pointers

- `src/commands/status.rs:258` — `load_audit_status`, which maps settings-task join failures to a failed audit status.
- `src/commands/status.rs:247` and `:316` — `AuditSettingsLoad` and `load_audit_settings`, including the explicit-path metadata gate.
- `src/commands/status.rs:434` — `audit_section_lines`, which renders failed audit states.
- `src/config.rs:484` — `Settings::from_file` and its `file_builder` optional file source; add the strict explicit-file constructor beside it without changing this method's behavior.
- `src/i18n.rs:319` — the `Strings` field declaration for the existing failed-scan warning; declare the new known-condition string beside it.
- `src/i18n/en.rs`, `src/i18n/ko.rs`, and `src/i18n/status.rs` — English and Korean catalogue values and the `Messages` accessor for the known non-regular-file condition.

sehkone:
## Summary

- Load an explicitly supplied agent configuration file strictly after its metadata gate, preventing fallback to default audit settings.
- Preserve status output by mapping settings task joins and configuration failures to failed audit scans.
- Localize the known non-regular configuration path condition in English and Korean.

Closes #892

## Test plan

- [x] Verify optional versus strict settings loading, including unsupported explicit files.
- [x] Verify explicit configuration metadata outcomes, malformed settings, registrar validation, absent stores, and scanner failures.
- [x] Verify settings task join failures and localized English/Korean non-regular-path warnings.
- [x] Run `cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib`, and `cargo test --bin bootroot`.
- [x] Run the completed CI-equivalent checks, core coverage, and non-hosts E2E lifecycle from `scripts/preflight/run-all.sh`.
- [ ] Complete the hosts-mode E2E lifecycle locally (blocked because it requires non-interactive `sudo -n` on this workstation).

### DIFF
--- a/src/commands/status.rs
+++ b/src/commands/status.rs
@@ -257,17 +257,14 @@ enum AuditSettingsLoad {
 
 async fn load_audit_status(args: &StatusArgs, messages: &Messages) -> Result<AuditStatus> {
     let settings_path = audit_settings_path(args.agent_config.as_ref());
-    let settings = match map_audit_settings_join_result(
+    let settings = map_audit_settings_join_result(
         settings_path,
         tokio::task::spawn_blocking({
             let agent_config = args.agent_config.clone();
             move || load_audit_settings(agent_config)
         })
         .await,
-    ) {
-        Ok(settings) => settings,
-        Err(status) => return Ok(status),
-    };
+    );
 
     let settings = match settings {
         AuditSettingsLoad::Ready(settings) => settings,
@@ -306,8 +303,8 @@ fn audit_settings_path(agent_config: Option<&PathBuf>) -> PathBuf {
 fn map_audit_settings_join_result(
     settings_path: PathBuf,
     result: std::result::Result<AuditSettingsLoad, tokio::task::JoinError>,
-) -> std::result::Result<AuditSettingsLoad, AuditStatus> {
-    result.map_err(|error| AuditStatus::Failed {
+) -> AuditSettingsLoad {
+    result.unwrap_or_else(|error| AuditSettingsLoad::Failed {
         path: settings_path,
         reason: AuditFailureReason::Diagnostic(error.to_string()),
     })
@@ -772,8 +769,7 @@ mod tests {
     async fn unreadable_regular_agent_config_is_a_failed_scan() {
         use std::os::unix::fs::PermissionsExt;
 
-        // SAFETY: geteuid only reads the process identity and has no preconditions.
-        if unsafe { libc::geteuid() } == 0 {
+        if bootroot::fs_util::current_process_euid() == 0 {
             return;
         }
 
@@ -903,12 +899,12 @@ mod tests {
             .to_string();
         let path = PathBuf::from("selected-agent.toml");
 
-        let Err(AuditStatus::Failed {
+        let AuditSettingsLoad::Failed {
             path: failed_path,
             reason,
-        }) = map_audit_settings_join_result(path.clone(), join_result)
+        } = map_audit_settings_join_result(path.clone(), join_result)
         else {
-            panic!("the join failure must return failed audit status");
+            panic!("the join failure must return failed audit settings");
         };
         assert_eq!(failed_path, path);
         let AuditFailureReason::Diagnostic(reason) = reason else {
--- a/src/commands/status.rs
+++ b/src/commands/status.rs
@@ -229,8 +229,13 @@ enum AuditStatus {
     Scan(AuditScan),
     Failed {
         path: std::path::PathBuf,
-        error: String,
+        reason: AuditFailureReason,
     },
+}
+
+enum AuditFailureReason {
+    ExplicitConfigNotRegularFile,
+    Diagnostic(String),
 }
 
 struct AuditScanSettings {
@@ -241,22 +246,33 @@ struct AuditScanSettings {
 
 enum AuditSettingsLoad {
     Ready(AuditScanSettings),
-    Failed { path: PathBuf, error: String },
-    MissingExplicitConfig { path: PathBuf },
+    Failed {
+        path: PathBuf,
+        reason: AuditFailureReason,
+    },
+    MissingExplicitConfig {
+        path: PathBuf,
+    },
 }
 
 async fn load_audit_status(args: &StatusArgs, messages: &Messages) -> Result<AuditStatus> {
-    let settings = tokio::task::spawn_blocking({
-        let agent_config = args.agent_config.clone();
-        move || load_audit_settings(agent_config)
-    })
-    .await
-    .context("loading registrar audit settings off runtime")?;
+    let settings_path = audit_settings_path(args.agent_config.as_ref());
+    let settings = match map_audit_settings_join_result(
+        settings_path,
+        tokio::task::spawn_blocking({
+            let agent_config = args.agent_config.clone();
+            move || load_audit_settings(agent_config)
+        })
+        .await,
+    ) {
+        Ok(settings) => settings,
+        Err(status) => return Ok(status),
+    };
 
     let settings = match settings {
         AuditSettingsLoad::Ready(settings) => settings,
-        AuditSettingsLoad::Failed { path, error } => {
-            return Ok(AuditStatus::Failed { path, error });
+        AuditSettingsLoad::Failed { path, reason } => {
+            return Ok(AuditStatus::Failed { path, reason });
         }
         AuditSettingsLoad::MissingExplicitConfig { path } => {
             anyhow::bail!(messages.status_error_agent_config_missing(&path.display().to_string()));
@@ -276,9 +292,25 @@ async fn load_audit_status(args: &StatusArgs, messages: &Messages) -> Result<Aud
         Err(AuditScanError::StoreAbsent { .. }) => Ok(AuditStatus::Absent),
         Err(error) => Ok(AuditStatus::Failed {
             path: settings.directory,
-            error: error.to_string(),
+            reason: AuditFailureReason::Diagnostic(error.to_string()),
         }),
     }
+}
+
+fn audit_settings_path(agent_config: Option<&PathBuf>) -> PathBuf {
+    agent_config
+        .cloned()
+        .unwrap_or_else(|| PathBuf::from("agent.toml"))
+}
+
+fn map_audit_settings_join_result(
+    settings_path: PathBuf,
+    result: std::result::Result<AuditSettingsLoad, tokio::task::JoinError>,
+) -> std::result::Result<AuditSettingsLoad, AuditStatus> {
+    result.map_err(|error| AuditStatus::Failed {
+        path: settings_path,
+        reason: AuditFailureReason::Diagnostic(error.to_string()),
+    })
 }
 
 fn load_audit_settings(agent_config: Option<PathBuf>) -> AuditSettingsLoad {
@@ -288,7 +320,7 @@ fn load_audit_settings(agent_config: Option<PathBuf>) -> AuditSettingsLoad {
             Ok(_) => {
                 return AuditSettingsLoad::Failed {
                     path: path.clone(),
-                    error: "agent configuration path is not a regular file".to_string(),
+                    reason: AuditFailureReason::ExplicitConfigNotRegularFile,
                 };
             }
             Err(error) if error.kind() == std::io::ErrorKind::NotFound => {
@@ -297,27 +329,29 @@ fn load_audit_settings(agent_config: Option<PathBuf>) -> AuditSettingsLoad {
             Err(error) => {
                 return AuditSettingsLoad::Failed {
                     path: path.clone(),
-                    error: error.to_string(),
+                    reason: AuditFailureReason::Diagnostic(error.to_string()),
                 };
             }
         }
     }
-    let settings_path = agent_config
-        .clone()
-        .unwrap_or_else(|| PathBuf::from("agent.toml"));
-    let settings = match Settings::from_file(agent_config) {
+    let settings_path = audit_settings_path(agent_config.as_ref());
+    let settings_result = match agent_config {
+        Some(path) => Settings::from_required_file(path),
+        None => Settings::from_file(None),
+    };
+    let settings = match settings_result {
         Ok(settings) => settings,
         Err(error) => {
             return AuditSettingsLoad::Failed {
                 path: settings_path,
-                error: error.to_string(),
+                reason: AuditFailureReason::Diagnostic(error.to_string()),
             };
         }
     };
     if let Err(error) = validate_registrar_settings(&settings.registrar) {
         return AuditSettingsLoad::Failed {
             path: settings_path,
-            error: error.to_string(),
+            reason: AuditFailureReason::Diagnostic(error.to_string()),
         };
     }
     AuditSettingsLoad::Ready(AuditScanSettings {
@@ -423,8 +457,17 @@ fn audit_section_lines(messages: &Messages, audit_status: &AuditStatus) -> Vec<S
                 );
             }
         }
-        AuditStatus::Failed { path, error } => lines
-            .push(messages.status_warning_audit_scan_failed(&path.display().to_string(), error)),
+        AuditStatus::Failed { path, reason } => {
+            let reason = match reason {
+                AuditFailureReason::ExplicitConfigNotRegularFile => {
+                    messages.status_warning_audit_config_not_regular_file()
+                }
+                AuditFailureReason::Diagnostic(error) => error,
+            };
+            lines.push(
+                messages.status_warning_audit_scan_failed(&path.display().to_string(), reason),
+            );
+        }
     }
     lines
 }
@@ -664,7 +707,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn unreadable_agent_config_path_is_a_failed_scan() {
+    async fn supplied_directory_agent_config_is_a_failed_scan() {
         let dir = tempfile::tempdir().expect("create test directory");
         let config = dir.path().join("agent-config-directory");
         std::fs::create_dir(&config).expect("create config directory");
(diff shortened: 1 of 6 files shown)


sehkone:
**[Reviewer Round 1]**

I read the issue, the diff, and the surrounding code in `src/commands/status.rs`, `src/config.rs`, and the i18n catalogues. The approach is the right one and I believe it satisfies every acceptance criterion. Two things I'd like changed (one is a repo-rule violation), plus a couple of notes.

## What works

The core of the issue is the check-then-use hole, and `Settings::from_required_file` closes it properly. Because the source is `required(true)` with no extension fallback that can silently succeed, every post-gate failure mode the issue enumerates lands in `AuditSettingsLoad::Failed` rather than in default settings: an unsupported extension, a path with no extension at all (the `config` crate's exact-match branch rejects an unknown extension instead of falling through), a read that fails on permissions, a parse error, and the file vanishing between `std::fs::metadata` and the load. `src/commands/status.rs:316-345` keeps the metadata gate's three outcomes distinct exactly as specified, and a FIFO or device node still short-circuits at the gate rather than blocking forever inside the `config` reader.

Splitting `AuditFailureReason` into `ExplicitConfigNotRegularFile` and `Diagnostic(String)` is the right shape — it moves the one condition the code owns into the catalogue and leaves foreign error text untranslated, which is precisely the scope line the issue draws. The new accessor at `src/i18n/status.rs:280` returns `&'static str` like every other placeholder-free string in that file, and the Korean wording lines up with the neighbouring `status_error_agent_config_missing`.

Test coverage is genuinely good and most of it would have failed before the change: `supplied_unsupported_agent_config_is_a_failed_scan` and `unreadable_regular_agent_config_is_a_failed_scan` both used to reach default settings and could report `Absent`, and `supplied_directory_agent_config_is_a_failed_scan` now asserts the typed variant rather than a non-empty string. The join test drives a real panicking `spawn_blocking` and asserts the join text is carried verbatim, rather than fabricating a `JoinError`.

Not touching `CHANGELOG.md` is correct: the audit-status feature is still under `[Unreleased]`, so no user of the last release can observe this hardening.

## Please change

**1. `src/commands/status.rs:775-776` introduces a new `unsafe` block for something the crate already exposes safely.**

```rust
// SAFETY: geteuid only reads the process identity and has no preconditions.
if unsafe { libc::geteuid() } == 0 {
```

`AGENTS.md` says not to introduce new `unsafe`, and here it is avoidable: `bootroot::fs_util::current_process_euid()` (`src/fs_util.rs:1061`) is a safe `pub` wrapper over the same call, already imported by test modules in this same binary crate (`src/commands/init/steps/registrar_internal.rs:1318`, `src/commands/rotate/registrar_internal.rs:810`). `src/registrar/audit/scan.rs:1853` uses the identical root-skip idiom through it. Please use it and drop the `unsafe` block and the SAFETY comment.

Related note, not a request: under `euid == 0` this test returns green without asserting anything, so in any job that runs as root the permission path is never exercised. The issue asks for the skip, so I'm not asking you to change it — just flagging that its coverage depends on the test job's user.

**2. `map_audit_settings_join_result` returns `Result<AuditSettingsLoad, AuditStatus>`, which inverts what `Err` means and costs the caller a second match.**

At `src/commands/status.rs:306-314` the `Err` arm carries a perfectly successful outcome, and the caller has to unwind that at `:259-269`:

```rust
Ok(settings) => settings,
Err(status) => return Ok(status),
```

followed immediately by a second `match settings` that handles `AuditSettingsLoad::Failed` the same way. `AuditSettingsLoad::Failed` already means "this is a failed audit status" — having the helper return `AuditSettingsLoad` instead would fold the two matches into one and make the helper's contract self-evident:

```rust
fn map_audit_settings_join_result(
    settings_path: PathBuf,
    result: std::result::Result<AuditSettingsLoad, tokio::task::JoinError>,
) -> AuditSettingsLoad {
    result.unwrap_or_else(|error| AuditSettingsLoad::Failed {
        path: settings_path,
        reason: AuditFailureReason::Diagnostic(error.to_string()),
    })
}
```

The test asserts on `AuditSettingsLoad::Failed { path, reason }` instead and stays just as sharp, and the helper is still the small testable unit the issue asked for. This is a readability call rather than a defect, so I won't block on it, but I'd take the simpler shape.

## Smaller points

- `src/config.rs:504` — "Builds the defaults-plus-file layers **both** constructors share" is now stale. There are three file-loading constructors and `from_required_file` deliberately bypasses this one. Worth rewording to name `new` and `from_file` explicitly, so the next reader doesn't take the new constructor's duplication of `apply_defaults` for an oversight.
- `src/config.rs:488-490` — the doc says "refuses an absent or unreadable file". It also refuses a file whose format it cannot infer, which is the case `load_audit_settings` most relies on. Mentioning that would make the contrast with `from_file` complete.
- `supplied_agent_config_metadata_error_is_a_failed_scan` (`:751`) depends on `ENOTDIR` from a path under a regular file. That is dependable on Linux and macOS, so it is fine here, but unlike its neighbour it carries no `#[cfg(unix)]`. Consistency nit only.
- Rendering a config-load failure through `status_warning_audit_scan_failed` produces "registrar audit scan failed for /etc/bootroot/agent.toml", naming a config file as though it were the store. That is pre-existing and the issue explicitly asks for the existing warning, so no change here — noting it only in case it is worth a follow-up.

## Hygiene

`Closes #892` with a full `## Test plan` is right, and the one unchecked box is a local environment limitation that is explained rather than a dropped requirement, so no `## Not addressed` section is needed. No stray debug output, no drive-by changes, no dead code.

Nothing here is a correctness defect — item 1 is a repo-rule fix and the rest are polish.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed** — Replaced the new test-only `unsafe { libc::geteuid() }` call with the existing safe `bootroot::fs_util::current_process_euid()` wrapper. This removes the avoidable unsafe block while preserving the required root-only skip.

2. **Fixed** — Changed `map_audit_settings_join_result` to return `AuditSettingsLoad` directly. A join failure now produces `AuditSettingsLoad::Failed`, so `load_audit_status` has one failed-settings mapping rather than an inverted `Result` and second caller match.

3. **Fixed** — Updated the strict-loader documentation to include unsupported formats and clarified that `file_builder` is shared only by `Settings::new` and `Settings::from_file`.

4. **Pushed back** — The metadata-error test needs no `#[cfg(unix)]`: unlike the adjacent permission-mode test, it uses no Unix-only API. The repository targets Unix systems, where the regular-file path component deterministically produces the asserted metadata failure. Adding a cfg would reduce applicable coverage without improving portability.

5. **Pushed back** — I left the existing failed-scan warning wording unchanged. The issue explicitly requires rendering this path through the existing warning and identifies only the known non-regular-path reason for localization; changing the store-oriented wording would be an out-of-scope user-visible change.

6. **Pushed back** — The effective-UID skip remains intentional and matches the issue test plan. In root-run environments permission bits cannot reliably make the file unreadable, so the test must skip rather than assert a false premise.

Validation: `cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo test --lib`, `cargo test --bin bootroot`, and `cargo clippy --all-targets -- -D warnings` pass. `scripts/preflight/run-all.sh` completed its CI-equivalent checks, core coverage, and non-hosts E2E lifecycle; hosts-mode remains unavailable locally because this workstation lacks non-interactive `sudo -n`.

sehkone:
**[Reviewer Round 2]**

I re-read the diff against `origin/main`, checked each Round 1 item against the updated code, and type-checked the binary locally (`cargo check --bin bootroot`, clean). All three requested changes landed, and all three pushbacks are sound. Nothing from Round 1 remains open, and I found no correctness defect this round.

## Round 1 items

**1. Avoidable `unsafe` — fixed, verified.** `src/commands/status.rs:772` now reads `bootroot::fs_util::current_process_euid() == 0`, and there is no `libc::` reference or `SAFETY` comment left anywhere in the file (`grep -n 'geteuid\|libc::' src/commands/status.rs` is empty). This matches the idiom already used at `src/registrar/audit/scan.rs`.

**2. Inverted `Result` from the join helper — fixed, verified.** `map_audit_settings_join_result` (`src/commands/status.rs:303-313`) now returns `AuditSettingsLoad`, and the caller at `:258-277` collapsed to a single `match settings`. The test at `:892` destructures `AuditSettingsLoad::Failed { path, reason }` and still asserts the join text verbatim, so the helper's coverage did not weaken.

**3. Stale doc comments — fixed, verified.** `src/config.rs:488-490` now says "refuses an absent, unreadable, or unsupported-format file", which names the case `load_audit_settings` actually leans on, and `:516-517` names `Settings::new` and `Settings::from_file` explicitly. That is exactly the signal I wanted: the next reader will see `from_required_file`'s separate `apply_defaults` call as deliberate rather than an oversight.

**4. `#[cfg(unix)]` on the metadata-error test — accepted, closing.** Your reasoning holds. The test uses no Unix-only API, so a `cfg` would gate it on nothing, and `ENOTDIR` from a path component that is a regular file is deterministic on the platforms this crate targets. Not re-raising.

**5. Store-oriented warning wording — accepted, closing.** Agreed. The issue names the existing warning as the render target and scopes localization to the one known reason; rewording it would be an out-of-scope user-visible change. This was flagged as a possible follow-up, not a request.

**6. Root skip in the permission test — accepted, closing.** Agreed, and the issue's test plan asks for the skip by name. My note was about where coverage actually runs, not about the code.

## This round's read of the change

I re-checked the parts that matter most and they hold up:

- `run_status` calls `load_audit_status` first (`src/commands/status.rs:29`), so `MissingExplicitConfig` still bails before any status output, and `AuditStatus::Failed` still reaches `print_status_summary` with only the pre-existing infra/OpenBao conditions setting a non-zero exit (`:126-132`). The constraint about exit conditions is preserved.
- The env-overlay question I did not ask in Round 1 turns out to be a non-issue: `Settings::from_file` has no `BOOTROOT_*` overlay either (only `Settings::new` does, `src/config.rs:454`), so both branches of `load_audit_settings` resolve the audit directory identically with respect to the environment. No asymmetry between `status --agent-config x.toml` and bare `status`.
- `audit_settings_path`'s `"agent.toml"` default agrees with `file_builder`'s (`src/config.rs:513`), so the path reported for a join failure under no `--agent-config` names the file that would actually have been read.
- The two tests that carry the real weight would both have failed before this change: with `required(false)`, the `config` crate swallows a resolve error and returns an empty map, so an existing `.unsupported` file and an unreadable `0o000` file each used to yield defaults and could report `Absent`. They now assert `Failed`.

## Two cosmetic points, neither blocking

**a. The new catalogue key is named as a warning but is a fragment.** `status_warning_audit_config_not_regular_file` (`src/i18n.rs:320`, `src/i18n/en.rs:308`, `src/i18n/ko.rs:308`) holds `"agent configuration path is not a regular file"` — no `WARNING:` prefix, because it is substituted as `{reason}` into `status_warning_audit_scan_failed`. It is the only `status_warning_*` value in either catalogue that is not a complete warning line; the surrounding fragments live under `status_audit_*` (`status_audit_retention_shortfall`, `status_audit_yes`). `status_audit_config_not_regular_file` would put it with its own kind and stop a future reader from rendering it standalone. Purely a rename across the four declaration sites plus the accessor.

**b. `audit_settings_path(agent_config: Option<&PathBuf>)`** (`src/commands/status.rs:297`) — `Option<&Path>` is the form `AGENTS.md`'s prefer-references rule points at, and both call sites reach it with `.as_deref()` instead of `.as_ref()`. Clippy's `ptr_arg` does not look inside `Option`, so nothing flags it. Trivial.

## Hygiene

Unchanged and correct: `Closes #892`, a full `## Test plan`, and the single unchecked box is a local `sudo -n` limitation that the issue's own test plan already permits skipping for this change — not a dropped requirement, so no `## Not addressed` section is warranted. No GitHub issue was created in this run. `CHANGELOG.md` is rightly untouched: the audit-status feature is still under `[Unreleased]` (`CHANGELOG.md:7-14`), so no user of 0.3.0 can observe this hardening. No stray debug output, no drive-by changes, no dead code.

Approving. Points (a) and (b) are yours to take or leave.

sehkone:

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Harden audit status failures
```

**Body**

```text
Keep explicit agent configuration failures within the audit status section
so operators retain the rest of the health report. Load supplied files
strictly and localize known non-regular path failures.

Closes #892
```

Issue #780, opened 2026-08-04
Record and surface every registrar verb invocation


sehkone:
# Record and surface every registrar verb invocation

## Summary

The restricted registrar surface bounds *what* a compromised caller can mint, but not *how often* or *for whom*. Repeated mints, and a teardown driven against a live instance, are therefore **detected, not prevented** — and that detection only means anything if it lives on the bootroot side, where the registrar cannot reach it. The registrar holds no OpenBao credential at all, and the verb-level record is not an OpenBao write: it is an append to a root-owned file made by the daemon process itself. What makes it one the registrar can neither forge nor erase is therefore **process and filesystem ownership** — the daemon's own identity and a root-owned store the caller cannot reach — not a credential, since no credential is involved in appending to a local file. "Under the bootroot-internal credential" describes the *OpenBao* writes a mint performs, which are recorded separately by the device below; the two artifacts carry the same property by different mechanisms, and the children state each one where it applies.

This is a **grouping issue and carries no work of its own.** It exists because its three children share one on-disk contract and one security argument, and reviewing any of them against a restatement of that contract rather than against the contract itself is how they drift apart.

The split is the write / read / hold seam. The first child — itself a grouping issue — defines the store and puts records into it: the JSON Lines record schema, the two-phase intent-before-any-OpenBao-write / outcome-after ordering, the bounds on attacker-influenced fields, rotation, the startup ownership and symlink checks, and the verb return semantics that make *unrecordable implies refused* actually hold. The second child consumes that contract without redefining any of it, and turns the store into a signal — the intent-without-outcome scan, the malformed-line count, the retention shortfall, and the `bootroot status` section that reports them. The third child — also a grouping issue — gives both this store and the mandatory OpenBao audit device a bounded, measured home on the bootroot host, and relays the reader's signals off it onto the endpoint. The first two are strictly ordered: the reader depends on the writer. The third depends on both.

## Children

- [x] Registrar audit records: the store and its two-phase verb integration — Detection *(grouping issue)*
- [x] Scan the audit store for anomalies and report them on `bootroot status` — Detection
- [x] Reserved audit store: layout, enforced reserve, and capacity reporting — Detection *(grouping issue)*
- [x] Harden audit-status failure reporting — Detection *(verify follow-up)*
- [x] Prevent audit config path substitution — Detection *(verify follow-up)*
- [x] Complete the audit scanner contract and coverage — Detection *(verify follow-up)*

## Shared background

[...]

PR #895, opened 2026-08-25
Prevent audit config path substitution (#894)

### LINKED ISSUE #894: Prevent audit config path substitution
# Prevent audit config path substitution

## Context

`bootroot status --agent-config` must retain normal status output when a supplied audit configuration cannot be loaded. Issue #892 added a required `config::File` source after an explicit-path metadata gate, but `config` 0.15.25 resolves a pathname that is no longer a file by looking for supported filename extensions. An extension-less supplied path can therefore disappear after the gate and cause a same-directory `*.toml` sibling to be loaded. That violates the requirement to read exactly the supplied file and can scan default audit settings when the sibling has no registrar audit overrides.

The audit scanner described by #779 is already consumed by the status loader; this follow-up changes only explicit configuration loading, not scanner semantics.

## Scope

- Change the strict explicit-file loading path used by `bootroot status` so it reads configuration bytes from exactly the supplied pathname and never invokes filename or extension discovery.
- Preserve defaults plus the supplied file and preserve the absence of a `BOOTROOT_*` environment overlay for this path. Derive the format only from the supplied file's supported extension; an unsupported or extension-less supplied file remains a load failure.
- Preserve the explicit metadata gate: a supplied `NotFound` at that gate remains the command error, a supplied non-regular path remains the typed localized failed-audit condition, and every other metadata error remains a diagnostic failed-audit status.
- After a supplied regular file has passed the gate, any subsequent read, format, parse, deserialization, or disappearance failure must render the audit section as failed with the supplied pathname. It must not load a sibling configuration or scan its registrar audit store.
- Keep the no-`--agent-config` path on `Settings::from_file(None)`, including its optional missing-default-file behavior.

## Acceptance criteria

- [x] An extension-less explicit file that passes inspection, then disappears while a same-directory `agent.toml` exists, produces `AuditStatus::Failed` naming the extension-less supplied path; it does not become `Absent` or scan the sibling's configured store.
- [x] The strict loader reads only the supplied file and rejects an unsupported or extension-less supplied filename rather than probing candidates with additional extensions.
- [x] Existing explicit-path outcomes remain unchanged: missing at the metadata gate is the command error; directory/non-regular paths use the localized known-condition reason; unreadable, malformed, unsupported, and invalid registrar settings are failed audit statuses naming the supplied path.
- [x] No explicit configuration retains optional default `agent.toml` behavior, and scanner mappings remain `StoreAbsent` to not-configured and all other scanner errors to failed scan.

## Constraints

- Keep blocking configuration I/O off Tokio runtime workers.
- Do not change audit record format, scanner trust or selection semantics, configuration keys, the normal status output, or OpenBao exit conditions.
- Do not call `Settings::validate()` or duplicate registrar validation; continue to use only `validate_registrar_settings`.
- Do not weaken the metadata gate or convert a missing explicit path at that gate into a failed audit status.
- Do not introduce an environment overlay to the strict explicit-file path.

## Out of scope

- Changing the optional-source contract of `Settings::from_file`.
- Translating arbitrary configuration, I/O, validation, or task diagnostics.
- Changing audit scanning, audit-store layout, retention handling, or the registrar health relay.
- Documentation changes; documented `--agent-config` examples already use supported `.toml` paths and this work changes failure containment, not operator workflow.

## Test plan

- [x] Add a deterministic test for the post-metadata-gate sequence: prepare an extension-less explicit file and a sibling `agent.toml`, remove the supplied file before the strict-load step, and assert the selected supplied path produces a failed audit status without loading the sibling. Extract a small production helper if needed to test the load-after-gate step without a test-only `StatusArgs` input.
- [x] Test strict loading directly for a missing path, unsupported extension, and extension-less filename, verifying no extension candidate is accepted.
- [x] Retain or extend status tests for missing-at-gate, non-regular, unreadable, malformed, invalid registrar, no-explicit-config, absent-store, and scanner-error mappings.
- [x] Run `cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib`, `cargo test --bin bootroot`, and `scripts/preflight/run-all.sh`. The E2E matrix may be skipped because this change does not affect Docker lifecycle or E2E scripts.

## Dependencies

Part of #892. Part of #780.

## Pointers

- `src/commands/status.rs:313` — explicit metadata gate and load result mapping.
- `src/commands/status.rs:335` — branch selecting the strict explicit-file loader.
- `src/config.rs:497` — current required pathname-based `config::File` constructor to replace or refactor for exact-file semantics.
- `src/config.rs:484` — optional `Settings::from_file` contract that must remain unchanged.
- `src/commands/status.rs:707` and `:729` — existing explicit-path status tests to extend.
- `src/config.rs:618` and `src/config.rs:628` — direct optional-versus-required loader contract tests.

sehkone:
## Summary

- Read an explicit audit configuration directly from its supplied pathname, deriving only the supported TOML format.
- Keep all post-metadata-gate configuration failures in the failed audit status, preventing sibling configuration discovery.
- Cover extension-less, unsupported, missing, and post-gate-disappearance paths.

Closes #894. Part of #892. Part of #780.

## Test plan

- [x] Confirm a vanished extension-less explicit path cannot load a sibling `agent.toml`.
- [x] Confirm strict loading rejects missing, unsupported, and extension-less supplied paths.
- [x] Verify explicit-path audit failure mappings and the optional default configuration path.
- [x] Run `cargo fmt -- --check --config group_imports=StdExternalCrate`.
- [x] Run `cargo clippy --all-targets -- -D warnings`.
- [x] Run `cargo test --lib` and `cargo test --bin bootroot`.
- [x] Run the CI-equivalent checks and core lifecycle portions of `scripts/preflight/run-all.sh`.
- [x] Verify E2E matrix hosts-mode coverage in green branch CI.

### DIFF
--- a/src/commands/status.rs
+++ b/src/commands/status.rs
@@ -265,7 +265,13 @@ async fn load_audit_status(args: &StatusArgs, messages: &Messages) -> Result<Aud
         })
         .await,
     );
+    load_audit_status_from_settings_load(settings, messages).await
+}
 
+async fn load_audit_status_from_settings_load(
+    settings: AuditSettingsLoad,
+    messages: &Messages,
+) -> Result<AuditStatus> {
     let settings = match settings {
         AuditSettingsLoad::Ready(settings) => settings,
         AuditSettingsLoad::Failed { path, reason } => {
@@ -331,11 +337,30 @@ fn load_audit_settings(agent_config: Option<PathBuf>) -> AuditSettingsLoad {
             }
         }
     }
-    let settings_path = audit_settings_path(agent_config.as_ref());
-    let settings_result = match agent_config {
-        Some(path) => Settings::from_required_file(path),
-        None => Settings::from_file(None),
-    };
+
+    match agent_config {
+        Some(path) => load_explicit_audit_settings_after_metadata_gate(path),
+        None => {
+            load_audit_settings_from_result(audit_settings_path(None), Settings::from_file(None))
+        }
+    }
+}
+
+/// Loads one explicit configuration after its caller has confirmed it is a
+/// regular file.
+///
+/// Keeping this step separate from the metadata gate ensures a file that
+/// disappears in between is reported as an audit failure instead of falling
+/// back to a discovered sibling configuration file.
+fn load_explicit_audit_settings_after_metadata_gate(path: PathBuf) -> AuditSettingsLoad {
+    let settings_result = Settings::from_required_file(&path);
+    load_audit_settings_from_result(path, settings_result)
+}
+
+fn load_audit_settings_from_result(
+    settings_path: PathBuf,
+    settings_result: Result<Settings, config::ConfigError>,
+) -> AuditSettingsLoad {
     let settings = match settings_result {
         Ok(settings) => settings,
         Err(error) => {
@@ -686,6 +711,45 @@ mod tests {
         .await
         .expect("a missing default store does not abort status");
         assert!(matches!(status, AuditStatus::Absent));
+    }
+
+    #[tokio::test]
+    async fn vanished_extensionless_agent_config_does_not_load_a_toml_sibling() {
+        let dir = tempfile::tempdir().expect("create test directory");
+        let supplied = dir.path().join("agent");
+        let audit_dir = dir.path().join("sibling-registrar-audit");
+        let sibling = dir.path().join("agent.toml");
+        std::fs::create_dir(&audit_dir).expect("create sibling audit directory");
+        std::fs::write(&supplied, "[registrar]\n").expect("write supplied config");
+        std::fs::write(
+            &sibling,
+            format!(
+                "[registrar]\naudit_store_dir = \"{}\"\naudit_record_dir = \"{}\"\n",
+                dir.path().display(),
+                audit_dir.display()
+            ),
+        )
+        .expect("write sibling config");
+
+        assert!(
+            std::fs::metadata(&supplied)
+                .expect("inspect supplied config")
+                .is_file(),
+            "the explicit metadata gate must pass before the file disappears"
+        );
+        std::fs::remove_file(&supplied).expect("remove supplied config after metadata gate");
+
+        let status = load_audit_status_from_settings_load(
+            load_explicit_audit_settings_after_metadata_gate(supplied.clone()),
+            &test_messages(),
+        )
+        .await
+        .expect("a strict-load failure must not abort status");
+        let AuditStatus::Failed { path, reason } = status else {
+            panic!("the vanished supplied config must not load its TOML sibling");
+        };
+        assert_eq!(path, supplied);
+        assert!(matches!(reason, AuditFailureReason::Diagnostic(_)));
     }
 
     #[tokio::test]
--- a/src/commands/status.rs
+++ b/src/commands/status.rs
@@ -349,9 +349,9 @@ fn load_audit_settings(agent_config: Option<PathBuf>) -> AuditSettingsLoad {
 /// Loads one explicit configuration after its caller has confirmed it is a
 /// regular file.
 ///
-/// Keeping this step separate from the metadata gate ensures a file that
-/// disappears in between is reported as an audit failure instead of falling
-/// back to a discovered sibling configuration file.
+/// This seam lets tests drive the load-after-gate step without constructing
+/// [`StatusArgs`]. [`Settings::from_required_file`] provides the exact-path
+/// guarantee that prevents fallback to a discovered sibling configuration.
 fn load_explicit_audit_settings_after_metadata_gate(path: PathBuf) -> AuditSettingsLoad {
     let settings_result = Settings::from_required_file(&path);
     load_audit_settings_from_result(path, settings_result)
@@ -747,6 +747,33 @@ mod tests {
         .expect("a strict-load failure must not abort status");
         let AuditStatus::Failed { path, reason } = status else {
             panic!("the vanished supplied config must not load its TOML sibling");
+        };
+        assert_eq!(path, supplied);
+        assert!(matches!(reason, AuditFailureReason::Diagnostic(_)));
+    }
+
+    #[tokio::test]
+    async fn vanished_toml_agent_config_is_a_failed_scan_after_metadata_gate() {
+        let dir = tempfile::tempdir().expect("create test directory");
+        let supplied = dir.path().join("agent.toml");
+        std::fs::write(&supplied, "[registrar]\n").expect("write supplied config");
+
+        assert!(
+            std::fs::metadata(&supplied)
+                .expect("inspect supplied config")
+                .is_file(),
+            "the explicit metadata gate must pass before the file disappears"
+        );
+        std::fs::remove_file(&supplied).expect("remove supplied config after metadata gate");
+
+        let status = load_audit_status_from_settings_load(
+            load_explicit_audit_settings_after_metadata_gate(supplied.clone()),
+            &test_messages(),
+        )
+        .await
+        .expect("a strict-load failure must not abort status");
+        let AuditStatus::Failed { path, reason } = status else {
+            panic!("a supplied config that vanishes after the gate must fail the audit status");
         };
         assert_eq!(path, supplied);
         assert!(matches!(reason, AuditFailureReason::Diagnostic(_)));
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 
 use anyhow::Result;
 use config::builder::DefaultState;
-use config::{Config, ConfigBuilder, ConfigError, Environment, File, FileFormat};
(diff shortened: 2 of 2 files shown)


sehkone:
**[Reviewer Round 1]**

I've read the issue, the PR (#895), the diff, the surrounding code, and the `config` 0.15.25 / `toml_parser` sources the change reimplements.

## Verdict

This is the right fix, and it's built carefully. Replacing `File::from(path).required(true)` with an explicit extension check + `fs::read` + `File::from_str` is the correct shape: it removes the discovery step that `FileSourceFile::find_file` performs (`config-0.15.25/src/file/source/file.rs:53-85` — for an extension-less path it goes straight to `set_extension("toml")` and finds the sibling), while keeping every other layer identical. The reimplementation is faithful to what `config`'s `resolve()` actually does, down to the BOM skip and the lossy decode, so the blast radius really is confined to discovery. The metadata gate, the `MissingExplicitConfig` → command-error mapping, the `ExplicitConfigNotRegularFile` typed reason, the no-env-overlay property, and `Settings::from_file(None)` for the no-flag path are all untouched. `Cargo.toml:10` enables only the `toml` feature, so restricting to `.toml` loses nothing, and the case-sensitive `extension == "toml"` matches how `config` compares extensions exactly.

The headline test is genuine: under the old code, the sibling in `vanished_extensionless_agent_config_does_not_load_a_toml_sibling` deserializes and passes `validate_registrar_settings` (absolute store dir, record dir nested inside it), so the status would have come back `Scan`/`Absent` rather than `Failed` — it fails before the change. All eight pre-existing explicit-path status tests are retained and still land on the branches they were written for (`unreadable_regular_agent_config_is_a_failed_scan` and `malformed_agent_config_is_a_failed_scan` both use `.toml` names, so they still exercise `fs::read` and the parser rather than short-circuiting on the format gate). No CHANGELOG entry is correct: the feature itself is still under `[Unreleased] ### Added` (CHANGELOG.md:11-16), so no released user can observe this. Linkage (`Closes #894. Part of #892. Part of #780.`) and the test-plan checklist are in order, and there's no sign of an author-created issue.

Points below are all minor.

## 1. The doc comment on the new seam claims a guarantee it doesn't provide

`src/commands/status.rs:350-354`:

> *Keeping this step separate from the metadata gate ensures a file that disappears in between is reported as an audit failure instead of falling back to a discovered sibling configuration file.*

That's not what makes it safe. Splitting `load_explicit_audit_settings_after_metadata_gate` out of `load_audit_settings` ensures nothing — the two run back to back in the same blocking task, and inlining the one-liner would change no behaviour at all. The containment comes entirely from `Settings::from_required_file` reading the exact path. The function exists as a test seam, which the issue explicitly sanctioned.

This matters because it points a future maintainer at the wrong invariant: someone could relax `from_required_file` back toward `File::from(path)` and still believe the boundary was holding the line. Say what's true — this is a seam so the load-after-gate step can be driven without a `StatusArgs`, and the exact-path guarantee lives in `from_required_file`.

## 2. `required_file_format` hardcodes `"toml"` with nothing tying it to the enabled `config` features

`src/config.rs:562-573`. `Settings::from_file`/`Settings::new` still go through `file_builder`'s `File::from(path)`, which accepts every format `config` was compiled with. Today they agree because `Cargo.toml:10` is `default-features = false, features = ["toml"]`. Add another feature and they diverge silently: `from_file` starts accepting `agent.json` while `from_required_file` rejects it as "not of a supported file format", with nothing in either place cross-referencing the other.

`FileFormat::all()`/`extensions()` are `pub(crate)`, but `FileStoredFormat::file_extensions()` is public (`config-0.15.25/src/file/mod.rs:23`), so `FileFormat::Toml.file_extensions().contains(&ext)` would stay tied to the crate. Failing that, a comment naming the Cargo feature this depends on.

Separately, the summary sentence — "Returns the only configuration format compiled into this application" — describes a constant, not this function. It validates an extension and errors on anything else; the doc should say that.

## 3. `strip_utf8_bom` has no observable effect here, and no test

`src/config.rs:576-584`. It mirrors `config`'s `resolve()` exactly, so as fidelity it's defensible. But the only format this build supports is TOML, and `toml_parser`'s lexer already skips a leading `EF BB BF` itself (`toml_parser-1.1.2/src/lexer/mod.rs:35`). A BOM'd `agent.toml` parses identically with or without the helper — so this is new production code with no reachable effect and no test, whose stated rationale ("accepted by `config::File` sources") is about a layer this path deliberately no longer uses. Either drop it, or keep it with a one-line test and a comment saying it's insurance against a future non-TOML format rather than something the current path needs.

## 4. The lossy decode reads as an oversight but is a deliberate fidelity choice

`src/config.rs:501`. `String::from_utf8_lossy` means an `agent.toml` containing invalid UTF-8 gets U+FFFD substituted instead of being rejected — a byte-mangled `audit_record_dir` would load as a path with replacement characters rather than failing. That is precisely what `config` did before (`resolve()` calls `from_utf8_lossy` too), so it's not a regression, and preserving it is right for a change scoped to discovery only. But nothing in the code says so, and it's exactly the kind of line someone later "fixes" into `String::from_utf8` — quietly changing behaviour — or mistakes for deliberate hardening. One line of comment settles it.

## 5. The two new tests are slightly weaker than their names

- `required_file_loading_never_discovers_a_toml_sibling` (`src/config.rs:662-670`) asserts only `is_err()`. It does fail before the change (the sibling would resolve and return `Ok`), so it's real coverage — but it can't distinguish "rejected the extension" from "read the wrong file and then failed on it". The sibling already carries a distinctive `email`; asserting on that, or on the error text, would pin the intent.
- `vanished_extensionless_agent_config_does_not_load_a_toml_sibling` (`src/commands/status.rs:716-753`) verifies the acceptance criterion, but the disappearance is inert: `required_file_format` short-circuits on the missing extension before `fs::read` is ever reached, so deleting the `remove_file` line wouldn't change the outcome. The `.toml`-named-and-vanished case is the one that actually exercises the `fs::read` → `ConfigError::Foreign` branch after the gate. That branch isn't unreachable — `unreadable_regular_agent_config_is_a_failed_scan` hits it via `EACCES` — so this is a naming/precision nit rather than a coverage hole, but a second case would make the test match its title.

## 6. Residual surface, correctly out of scope

`bootroot-agent` still loads its config through `Settings::new` (`src/bin/bootroot-agent.rs:128`) → `file_builder` → `File::from(path).required(false)`, so `bootroot-agent -c /etc/bootroot/agent` still silently resolves to `/etc/bootroot/agent.toml`, and a missing file still yields defaults. The issue puts `Settings::from_file`'s optional contract explicitly out of scope, so leaving it alone is the right call for this PR — noting it only so the remaining surface is on the record, not as something to fix or file here.

None of the above blocks the change. Items 1 and 2 are the ones I'd want addressed before merge, since both are comments/structure that will mislead the next reader; 3–5 are cleanup.

sehkone:

sehkone:
[Author Round 1]

1. Fixed — Corrected the helper documentation: it is a test seam, while `Settings::from_required_file` owns the exact-path guarantee.

2. Fixed — Tied the TOML extension check to `FileFormat::Toml.file_extensions()` and documented that `Cargo.toml` enables only config's `toml` feature. The comment requires explicit strict-loader support whenever another config format is enabled, since config's complete enabled-format list is not public.

3. Fixed — Kept BOM removal for source-semantics compatibility with future formats, documented that rationale, and added a focused leading-BOM test.

4. Fixed — Documented that lossy UTF-8 decoding intentionally preserves `config::File` behavior rather than introducing a new failure mode.

5. Fixed — The extension-less sibling test now verifies the unsupported-format error, and a new `.toml` post-metadata-gate disappearance test exercises the actual read-failure mapping.

6. Pushed back — `bootroot-agent` retains its optional `Settings::new` behavior. Changing that path is explicitly out of scope for #894 and would alter the optional-source contract the issue requires us to preserve.

Validation: `cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib`, `cargo test --bin bootroot`, and `scripts/preflight/run-all.sh` all passed locally. Pushed as `11dd688b`.

sehkone:
**[Author Round 1]**

Addressed and pushed review round 1 in `11dd688b`.

1. Fixed — clarified the test seam versus exact-path invariant.
2. Fixed — tied TOML extension metadata to `config` and documented feature coupling.
3. Fixed — documented retained BOM behavior and added coverage.
4. Fixed — documented intentional lossy UTF-8 compatibility.
5. Fixed — strengthened format-boundary coverage and added vanished-`.toml` post-gate coverage.
6. Pushed back — `bootroot-agent` optional config behavior is explicitly out of scope.

Posted the full response on [PR #895](https://github.com/aicers/bootroot/pull/895#issuecomment-5404369687).

Local validation passed: formatting, clippy, library/binary tests, and full preflight. New GitHub CI is running for `11dd688b`.

Issue #790, opened 2026-08-04
Runtime registrar surface and its non-self-propagation guarantee


sehkone:
# Runtime registrar surface and its non-self-propagation guarantee

## Summary

Today the only way to mint a `bootroot-service-*` identity is to hold raw OpenBao write on `sys/policies/acl/*` and `auth/approle/role/*`. The runtime registrar (the bootroot-co-located roxyd) needs to mint and tear down service identities long after the one-shot install CLI is gone, but handing it those raw paths lets it author an arbitrary policy body under a permitted name, or bind a new AppRole to the registrar's own policy — turning one leaked credential into fleet-wide identity minting.

This effort closes that by construction. bootroot keeps the broad role/policy-write authority in a bootroot-internal credential held by its own daemon, and exposes to the registrar only two narrow verbs — mint and deregister — that take an identity's *parts* (`service_name`, `host`, `instance`) and never a role definition or a policy body. The registrar ends up holding no OpenBao credential at all: it authenticates to a root-owned unix-domain socket with an identity-scoped client certificate, and every invocation of either verb (accepted and refused alike) leaves a bootroot-side audit record the registrar cannot forge or erase.

Independently of the confinement work, this effort also splits `registration_id` out of `service_name`. Today one string is both the SAN's service label and the sole key of every per-service namespace bootroot owns, which permits exactly one registration of a component deployment-wide. Splitting the key from the label is what admits per-host and multi-instance registrations, and it is the piece that would be expensive to retrofit later.

Design source: `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` (accepted). Read §2 for the verified current state, §3–§4 for the threat model and the decided mechanism, §5 for what bootroot must provide, and §6 for the acceptance criteria. Children below carry their own self-contained specs; the RFC is background, not a substitute for reading the issue.

## Children

- [x] #756 Split the namespace key from the SAN label with a required `registration_id` — Foundation
- [x] #757 Read the bootler-rendered registrar config and derive registration identities — Verb building blocks
- [x] #758 Implement the restricted registrar mint and deregister verbs — Verb building blocks
- [x] #759 Settle and transcribe the registrar endpoint's caller-facing wire contract — Verb building blocks
- [x] #760 Issue a distinguishable registrar client identity and pin the endpoint's server identity — Surface
- [x] #763 Registrar endpoint: the socket and its wire protocol — Surface *(grouping issue)*
[...]

PR #904, opened 2026-08-25
Terminate mTLS on the registrar endpoint and scope it to the registrar identity (#764)

### LINKED ISSUE #764: Terminate mTLS on the registrar endpoint and scope it to the registrar identity
# Terminate mTLS on the registrar endpoint and scope it to the registrar identity

## Context

The registrar endpoint's serving path is already in `origin/main`. `src/registrar/endpoint/` carries the systemd activation contract, the socket ownership policy, the bootroot-owned frame envelope, one pre-verb refusal path and the accept loop with its bounded connection fleet, all with tests. A connection is authenticated today by exactly one check — `serve::authorize_peer` (`src/registrar/endpoint/serve.rs:88`) requires the peer's uid to equal the daemon's effective uid — and the identity that reaches the verbs is `serve::caller_identity` (`:78`), which renders `unix-peer:uid=<uid>`. That authenticates *root on this host*, not *the registrar*.

The design requires the caller to be authenticated as a specific certificate identity, and requires that identity to be the only one accepted at this endpoint, so that the authority the registrar wields is exactly two calls. This issue adds that layer onto the existing serving path and changes nothing else about it.

**The identity primitives are already in the library and are consumed here, never re-derived.** All are public items of `src/registrar.rs` and `src/registrar/endpoint_pin.rs`:

- `recognize_registrar_client(end_entity_der, domain)` and `recognize_registrar_client_name(name, domain)` (`src/registrar.rs:328`, `:346`) apply the client name rule: exactly one SAN, of type DNS, spelled `<instance>.bootroot-registrar.<host>.<domain>` with the domain matched as a multi-label suffix on a label boundary, every comparison ASCII-case-insensitive. No CN is consulted, no chain is built, no validity window is checked and no authorization decision is made. They return `RegistrarClientIdentity { instance, host, domain }` or a typed `RegistrarIdentityError`.
- `single_dns_san(der)` (`:237`) extracts that one name, and `registrar_endpoint_identity(instance, host, domain)` (`:180`) composes the endpoint's own server name `<instance>.bootroot-registrar-endpoint.<host>.<domain>` from `REGISTRAR_ENDPOINT_LABEL` (`:116`).
- `endpoint_pin::RegistrarEndpointVerifier` (`:314`), reached through `endpoint_server_verifier` (`:264`) and `build_endpoint_client_config` (`:290`), is the caller's side of the pin: it accepts only a server whose single DNS SAN equals the expected endpoint name **and** whose presented chain builds to a pinned, CA-capable, time-valid trust anchor, with a distinct `EndpointVerifyRejection` for each way that can fail. A caller that also authenticates itself builds its `ClientConfig` from that verifier and finishes with `rustls::ConfigBuilder::with_client_auth_cert`, which is exactly what this issue's tests do.

Both reserved labels sit under the `bootroot-` prefix that `service add` refuses, so no operator-authored identity can collide with either.

**The refusal shape already exists and needs no change to work over TLS.** `refusal::refuse` (`src/registrar/endpoint/refusal.rs:294`) is generic over `S: AsyncWrite + Unpin`. It logs a typed reason under a daemon-generated connection diagnostic id — `ConnectionId::next()`, assigned at accept before any byte is read — writes **zero** response bytes, and calls `close()`, which is `stream.shutdown()`; on a `tokio_rustls::server::TlsStream` that emits `close_notify`. `RefusalCause` was deliberately left at two variants so a non-transport cause could share the path. This issue adds one caller to that path — the identity refusal in Scope item 4 — and builds no second one.

**The endpoint's operation set is already closed, and this issue leaves it exactly as it is.** `frame::Operation` (`src/registrar/endpoint/frame.rs:41`) has two variants and no more; `Operation::from_name` (`:55`) returns `None` for any other syntactically valid name, which `serve::read_header` refuses as `TransportRefusalReason::UnrecognizedOperation` with the identifier logged as received; and `serve::dispatch` (`:456`) matches the enum exhaustively with no `_ =>` arm, so a third operation cannot reach a handler without that match being made to account for it. The half of the design's requirement that says *these two calls and no more* is therefore already enforced by the serving path. What is missing, and what this issue supplies, is the other half: that only the registrar identity may make them.

**Server-side TLS does not exist in this repository.** There is no `ClientCertVerifier` anywhere in the tree, and no `rustls::ServerConfig` in production library code. The one production `ServerConfig` is the http01 responder's (`src/bin/bootroot-http01-responder/tls.rs:149`), built `.with_no_client_auth()`. The `with_no_client_auth()` calls in `src/tls.rs` are on `ClientConfig`s and mean the opposite thing — that bootroot's outbound clients present no certificate — so they are not what this issue departs from.

**The endpoint is not usable in a shipped build yet, and this issue does not change that.** `endpoint::production_handler()` (`src/registrar/endpoint.rs:172`) returns `None`, so `endpoint::activate` refuses startup when the endpoint is enabled, and `docs/en/configuration.md` and `docs/en/operations.md` tell operators so. Registering that handler, provisioning the verb layer in the daemon and withdrawing the unsupported statement are work in progress elsewhere and are not dependencies of this issue; see Out of scope. The consequence here is an ordering one, stated in Scope item 1: the material refusals below are wired in *after* the missing-handler refusal, so on today's build `endpoint::activate` returns the missing-handler diagnostic before it ever looks at certificate material. That is the right priority — an endpoint that cannot serve is not made more serviceable by a certificate — but it means the material refusals are not reachable through `activate` in a shipped build, and this issue does not pretend otherwise. They are asserted on the builder that produces the configuration, which is a `pub(crate)` function over its inputs; the resulting acceptor is then asserted on the real serving path through `adopt`, the seam the merged tests already drive.

### 1. The endpoint's server certificate: load it, present it, expose the swap seam

Add two keys to the existing `[registrar_endpoint]` table of `agent.toml`, deserialized onto `RegistrarEndpointSettings` (`src/config.rs:212`) as `Option<PathBuf>`: **`server_cert_path`** and **`server_key_path`**. The spellings follow the tree's conventions — a `*_path` suffix for a `PathBuf` setting (`role_id_path`, `secret_id_path`, `ca_bundle_path`) and the `cert`/`key` pairing of `Paths` (`src/config.rs:111`). Both are optional, so a configuration with no `[registrar_endpoint]` table parses exactly as it does today. The struct loses its `Copy` derive; `validate_registrar_endpoint_settings` (`src/config/validation.rs:199`) takes it by reference. Add both keys, commented out, beside `enabled` in `agent.toml.example`.

Add `src/registrar/endpoint/tls.rs` holding:

- A loader that reads the PEM certificate and key from those two paths and verifies the key matches the leaf. Mirror `load_certified_key` (`src/bin/bootroot-http01-responder/tls.rs:73`), which proves the match by signing and verifying a test message; that function lives in a binary crate and is not importable from the library.
- The material checks below.
- `EndpointCertResolver`, a `pub(crate)` implementation of `rustls::server::ResolvesServerCert` over an `RwLock<Arc<CertifiedKey>>` with a `swap(CertifiedKey)` method — the shape of `ReloadableCertResolver` (`src/bin/bootroot-http01-responder/tls.rs:31`, `swap` at `:49`), mirrored rather than reused for the same reason. A poisoned lock must not take the endpoint down: recover the guard rather than `expect`ing on it.

**The typed resolver handle has to survive, or the seam it exists for is unreachable.** `ConfigBuilder::with_cert_resolver` takes an `Arc<dyn ResolvesServerCert>` and `ServerConfig` keeps only that trait object; the trait does not extend `Any`, so there is no way back to `EndpointCertResolver` from a built configuration, and code holding only the `ServerConfig` or the `TlsAcceptor` cannot call `swap` at all. So the builder returns the `Arc<EndpointCertResolver>` alongside the `Arc<ServerConfig>`, `ActivatedEndpoint` retains it, and it is reachable through a `pub(crate)` accessor beside the existing `listener()`, `socket_path()`, `daemon_uid()` and `handler()` (`src/registrar/endpoint.rs:135`). No wider API is added: the daemon already reaches the `Arc<ActivatedEndpoint>` through `RegistrarEndpoint::activated()`, so the later renewal work needs nothing from this issue but that accessor. Nothing in this issue calls `swap` outside a test, and the test drives it through that owner rather than through a resolver value of its own — which is what makes the accessor exercised rather than speculative.

Call `crate::tls::install_crypto_provider()` (`src/tls.rs:226`) before building anything, and complete the `ServerConfig` with `.with_cert_resolver(...)` exactly as the responder does at `:151`.

**The PEM at `server_cert_path` is the leaf followed by its issuer chain, not a bare leaf, and the loader proves it.** A caller verifies this endpoint with `RegistrarEndpointVerifier`, whose anchor selection scans only the certificates the server actually presents — the end-entity and the intermediates — and keeps those whose DER SHA-256 is pinned and that are CA-capable and inside their validity window. A leaf-only file loads cleanly, matches its key and carries the right SAN, and then **every correctly pinned caller rejects the handshake** with `AnchorMismatch`. Because the material is supplied out of band until issuance lands, that failure would otherwise reach an operator as a client-side error with no local symptom to read.

So the loader self-checks its own material through the very rule a caller applies, rather than through a second one written here. It reads the leaf's single DNS SAN, requires it to be an endpoint identity in the configured domain (the recognizer Scope item 4 adds), then builds `RegistrarEndpointVerifier::new(pins, san)` — `pins` being the deployment's configured anchors, `trust.trusted_ca_sha256`, which is the same value a caller's pin file is written from — and calls its `verify(leaf, intermediates, now)` on the loaded chain. Any `EndpointVerifyRejection` is a startup refusal reported with its typed variant, so a missing chain, a chain that does not build to a configured anchor, an expired or not-yet-valid anchor and a non-CA anchor are each diagnosed as themselves. This adds no verification logic to the repository: it runs merged public code against the daemon's own material.

With the endpoint enabled, each of these is a **startup refusal**: an absent `server_cert_path` or `server_key_path` setting, and material at a configured path that is missing, unreadable, key-mismatched, SAN-mismatched, or rejected by the self-check above. Split that set as finely as the diagnostic needs: a file that is not PEM at all, a certificate setting pointing at a file holding no certificate, a key setting pointing at a file holding no private key, and a key of a type this build cannot sign with are each their own variant, because the two paths transposed is the likeliest way to get this wrong and both halves of it parse as PEM. The diagnostic always names the setting at fault (`[registrar_endpoint] server_cert_path` or `server_key_path`) and, when that setting has a value, the configured path as well; an absent setting has no path to report, so its diagnostic names the missing setting and states that it is required when the endpoint is enabled. There is no fallback: do not generate a self-signed certificate, do not present the host's ordinary service leaf, and do not add an issuance path.

**Factor the build so every refusal above is testable, and be precise about where each half is asserted.** Put the whole of it — read both files, check the key against the leaf, recognize the SAN, run the self-check, assemble the client verifier from item 2, and build the `ServerConfig` — behind one `pub(crate)` function of `src/registrar/endpoint/tls.rs` that takes its inputs as values (the two configured paths, the trust bundle path and its pinned fingerprints, and the configured domain) and returns the `Arc<ServerConfig>` and the `Arc<EndpointCertResolver>`, or the typed refusal. It touches no descriptor, no socket and no process environment, so **every startup-refusal case in the acceptance criteria is a direct test of this function** under `tempfile::tempdir()`.

`endpoint::activate` calls it after the unprivileged-daemon warning and the missing-handler check and before the activation contract is consumed, so an operator on a build with no handler still gets the handler diagnostic first and a build that has one fails before it touches the inherited descriptor. That ordering is a specification of where the call goes, not something this issue can demonstrate: while `production_handler()` returns `None`, `activate` cannot reach the material step at all, so no test here asserts a material failure through it. The ordering becomes observable when a production handler is registered.

`adopt` takes the already-built configuration and resolver, and `ActivatedEndpoint` stores the acceptor, the resolver and the configured `domain` — the last being what the identity check in item 4 is run against. `adopt` therefore carries the *serving* half: the merged harness hands it a descriptor and a configuration and drives real handshakes through the production accept loop. It cannot produce a material failure, and this issue does not ask it to.

`RegistrarEndpoint::activate` (`src/registrar.rs:438`) accordingly takes `&Settings` in place of the bare `enabled` flag; its one call site is `src/bin/bootroot-agent.rs:65`.

No issuance here: no ACME call, no `rcgen` self-signing in production code, no scheduler. Until certificate issuance for this leaf lands, the material at those paths is supplied out of band — by a fixture in tests, by the operator in a manual deployment. Say so in the documentation, together with the chain requirement.

### 2. Require and verify a client certificate

Build the `ServerConfig` with `rustls::server::WebPkiClientVerifier::builder(roots).build()` and `.with_client_cert_verifier(verifier)`. The builder's `.allow_unauthenticated()` arm makes the client certificate optional and **must not** be used: a connection presenting no client certificate, or one whose chain does not verify, fails the handshake before any request byte is read.

The roots are the deployment's own CA material the daemon already holds: the certificates in `trust.ca_bundle_path`, restricted to those whose DER SHA-256 appears in `trust.trusted_ca_sha256` (`src/config.rs:374`). Use `crate::tls::parse_pem_to_cert_list` (`src/tls.rs:261`) and `crate::tls::sha256_hex` (`:459`); `certs_to_root_store` (`:308`) is private today and becomes `pub(crate)`. Restricting to the pinned subset rather than trusting the whole bundle mirrors what the outbound path already does (`build_pinned_root_verifier`, `src/tls.rs:339`) and stops a bundle file that gained a certificate from silently widening who may connect. With the endpoint enabled, an absent `trust.ca_bundle_path`, an unreadable or unparseable bundle, an empty `trust.trusted_ca_sha256`, or a bundle in which no certificate matches a pin is a startup refusal naming the setting at fault, an empty pin list naming the pin list rather than the bundle — an empty root store cannot back a `WebPkiClientVerifier`, and the outbound path's direct-pin fallback has no meaning for client authentication.

### 3. Wrap the accepted connection

`serve::handle_connection` (`src/registrar/endpoint/serve.rs:241`) is the only function in the serving path that is concrete over `UnixStream`; `serve_request`, `read_header`, `dispatch`, `write_response` and `refusal::refuse` are all already generic over `S`. So the wrap is one change at one place: hand the authenticated stream to a `tokio_rustls::TlsAcceptor` built from the configuration above and pass the resulting `TlsStream` down. Nothing below that call site changes.

The per-connection order is deliberate and is part of this issue's contract: accept → connection id → capacity permit → spawn → peer-credential check → TLS handshake → registrar-identity check → read the frame. The peer check stays **first** and stays a gate of its own — certificate authentication and socket-level authentication answer different questions, the peer check costs one syscall and no handshake, and a connection failing it is refused even when its client certificate is the registrar's.

Add `HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(5)` beside the endpoint's other deadlines (`src/registrar/endpoint.rs:90`), cumulative from acceptance (`accepted_at + HANDSHAKE_TIMEOUT`), so an unauthenticated peer cannot hold one of the sixteen connection slots open indefinitely. `HEADER_IDLE_TIMEOUT` then restarts at handshake completion instead of at acceptance — a slow handshake must not consume the header budget. Update that constant's doc comment and pass the post-handshake instant to `serve_request` as its `accepted_at`.

A failed handshake reaches the caller only as a TLS alert, so the daemon's log is the only diagnosis and it is a requirement rather than a nicety: log it under the connection id with a typed label recovered from the error. `tokio_rustls` surfaces a `rustls::Error` inside the `io::Error`, so downcast it and distinguish `NoCertificatesPresented` (no client certificate) from `InvalidCertificate(_)` (the chain did not verify) from everything else. Nothing is written and the stream is dropped — there is no application stream to close cleanly.

### 4. Scope the endpoint to the registrar identity

After a successful handshake, take the peer's end-entity certificate from the session's `peer_certificates()` and apply the client name rule against the `domain` the endpoint captured at activation from `Settings::domain` (`src/config.rs:47`). Take it in its two steps — `single_dns_san` and then `recognize_registrar_client_name`, which is exactly what `recognize_registrar_client` composes — rather than the composed call, because the identity rendered below needs the SAN string itself and the composed form does not hand it back. A certificate that verified against the deployment CA but is not the registrar client identity — an ordinary service leaf issued for the registrar's own host, for instance — is refused through `refusal::refuse`: zero bytes, clean close, and a typed log carrying the `RegistrarIdentityError` so an operator can tell "not the registrar" from "the chain did not verify" from "no certificate at all".

**Deciding the identity here, after the handshake, rather than inside a custom `ClientCertVerifier`, is settled, and both halves of the reason matter.** A verifier lives in the `ServerConfig` and is shared by every connection, so it cannot log under the connection diagnostic id that correlates the rest of that connection's lines; and a rejection from inside it is a TLS alert, which is a second refusal shape beside the silent clean close every other pre-verb refusal already uses. Nothing is read from the caller before this check, so "refused before any request is read" still holds. Chain verification stays entirely inside `WebPkiClientVerifier`, and this issue writes no `ClientCertVerifier` at all.

Replace `serve::caller_identity(peer_uid)` with the certificate identity, rendered at that one site and passed to the handler unchanged: **`registrar-client:<san>`**, where `<san>` is the ASCII-lowercased single DNS SAN of the presented leaf. The prefix mirrors the existing `unix-peer:` convention. Spelling it in this issue rather than leaving it to a later one is deliberate: this value is the caller of record the verb layer receives and the one that audit records will carry. Delete the uid-derived renderer; the peer's uid, gid and pid stay logged connection diagnostics. Update the "What authenticates a caller" section of `src/registrar/endpoint.rs`'s module documentation.

The endpoint's own leaf needs the mirror-image rule for item 1's SAN check, and the daemon cannot compose the expected name because it holds no configured instance or host label — only the domain. So generalize the merged rule rather than duplicating it: factor `recognize_registrar_client_name` (`src/registrar.rs:346`) onto a private helper parameterized by the expected reserved label and the error a mismatch produces, then add `recognize_registrar_endpoint_name` and `recognize_registrar_endpoint` over `REGISTRAR_ENDPOINT_LABEL`, returning the parsed instance, host and domain as a `RegistrarEndpointIdentity` — a type of its own rather than a reused `RegistrarClientIdentity`, because the two names authenticate opposite ends of one connection and a value of one is never a value of the other — with a label mismatch reported as a new `RegistrarIdentityError::NotRegistrarEndpoint`. `recognize_registrar_client`'s behaviour and signature do not change.

### 5. Reload, and documentation

`[registrar_endpoint] enabled` is fixed for the process lifetime and a reload that changes it is rejected outright (`check_registrar_endpoint_reload`, `src/config.rs:231`). The two new paths are fixed for the same reason — the material is loaded once, above the `SIGHUP` loop, so a changed path would be read by nothing. Extend that check to compare the whole table and reject a reload that changes any of the three, with a diagnostic naming the key that changed. The *contents* at those paths remain re-readable through the resolver's `swap` seam, which is what certificate renewal drives; only the paths themselves are fixed.

Document, in `docs/en/configuration.md` (the "Registrar endpoint (Linux only)" section) and `docs/en/operations.md` (its counterpart), with `docs/ko/` mirrored page for page: the accepted client identity and that it is the only one; the two new settings by name, the requirement that `trust.ca_bundle_path` and `trust.trusted_ca_sha256` be configured when the endpoint is enabled, and **the requirement that `server_cert_path` hold the leaf followed by the issuer chain up to a configured trust anchor, with the symptom a bare leaf produces**; that the server material is supplied out of band until issuance lands; the handshake rejection behaviour and the server-side logging that is the caller's only diagnosis; the startup refusals; and that the two paths are not reloadable. Do **not** touch the "unsupported until a production request handler lands" statements on either page — withdrawing those belongs to the work that registers the handler.

## Acceptance criteria

- [x] The endpoint requires a client certificate: a connection presenting none fails the handshake, and the daemon logs it under the connection id as no-client-certificate. The verifier is not built with `.allow_unauthenticated()`.
- [x] A client certificate that does not verify against the pinned subset of the configured CA bundle fails the handshake, logged as a distinct chain-verification reason.
- [x] A certificate that verifies but is not the registrar client identity — including an ordinary service leaf issued for the registrar's own host — is refused after the handshake and before any request byte is read, through the existing refusal path, with a typed log naming the `RegistrarIdentityError`.
- [x] The registrar client identity is accepted and both operations are served over TLS, with responses unchanged from the plaintext serving path's.
- [x] The verbs receive `registrar-client:<san>` as their caller-identity parameter, not the peer-credential value, and the uid-derived renderer is gone.
- [x] The peer-credential check still runs and still refuses a mismatched peer, even when its client certificate is the registrar's.
- [x] The full pre-verb refusal shape still holds once the stream is TLS: on the unrecognized-operation path the caller reads **zero application bytes**, then sees an orderly `close_notify`-terminated end of stream — not an `UnexpectedEof`, reset or broken pipe — and no error identifier is emitted.
- [x] The endpoint presents the chain loaded from `server_cert_path`, and a caller built from `endpoint_pin::endpoint_server_verifier` against a pin file naming the deployment anchor connects; a caller whose pin set does not match refuses without sending a request.
- [x] Material whose PEM carries the leaf alone, or whose chain does not build to a certificate pinned in `trust.trusted_ca_sha256`, is refused by that builder with the `EndpointVerifyRejection` that names the failure — so the case that would otherwise present only as every caller failing is caught at the daemon.
- [x] With the endpoint enabled, each of an absent `server_cert_path` or `server_key_path` setting, a missing, unreadable, key-mismatched or SAN-mismatched material file, and an absent, unreadable or unpinned `trust.ca_bundle_path` makes the configuration builder refuse, so the endpoint cannot start. Every diagnostic names the setting at fault; the ones with a configured value name the path too, and each absent-setting case says the setting is required when the endpoint is enabled. No self-signed or fallback certificate is generated in any of them.
- [x] The build is a `pub(crate)` function over its inputs that touches no descriptor, socket or process environment, so each refusal above is asserted on it directly, and `endpoint::activate` calls it at the specified point.
- [x] `[registrar_endpoint]` parses `server_cert_path` and `server_key_path` into `Settings`; a configuration with no `[registrar_endpoint]` table still parses as it does today; an unknown key in the table is still an error; and `agent.toml.example` carries both keys commented out beside `enabled`.
- [x] A `SIGHUP` reload that changes any key in `[registrar_endpoint]` is rejected with a diagnostic naming the changed key, and the running daemon is left as it is.
- [x] The `Arc<EndpointCertResolver>` survives configuration assembly and is reachable from `ActivatedEndpoint` through a `pub(crate)` accessor. Swapping its `CertifiedKey` **through that accessor, on a running endpoint,** changes the certificate presented on the next handshake with no restart — so the seam certificate renewal drives is reachable from what the daemon holds, not only from a resolver a test built itself.
- [x] A connection that never completes a handshake is dropped at `HANDSHAKE_TIMEOUT` and logged, ending its task and releasing its capacity permit; a connection that completes one gets the full `HEADER_IDLE_TIMEOUT` budget measured from that completion.
- [x] `frame::Operation`, `Operation::from_name` and `serve::dispatch`'s exhaustive match are unchanged, and the merged tests over them still pass.
- [x] The English and Korean documentation pages carry the items listed in Scope item 5, `./scripts/check-docs.sh` passes, `cargo fmt -- --check --config group_imports=StdExternalCrate` passes, and `cargo clippy --all-targets -- -D warnings` is clean.

## Constraints

- Do not write a `ClientCertVerifier`, and do not disable or weaken verification to make a test pass. Chain verification is `WebPkiClientVerifier` over the pinned subset of the configured bundle; the identity rule is applied on top of it, after the handshake, in bootroot's own code.
- Reuse the existing helpers in `src/tls.rs` and the merged identity and pinning primitives rather than adding a second TLS stack, a second name rule or a second server-material check. Add no dependency: `rustls`, `tokio-rustls`, `rustls-pemfile`, `ring`, `x509-parser` and `rcgen` are all already in `Cargo.toml`.
- Do not add certificate issuance or renewal — no ACME call, no `rcgen` self-signing in production code, no scheduler. This issue loads material and exposes a swap point.
- Do not change the socket, its ownership, the activation contract, the frame envelope, the payload codec or the enablement gate; they must keep behaving as they do.
- **Do not add a second gate on the operation set.** `Operation`, `Operation::from_name` and the exhaustive `match` in `serve::dispatch` already close it, and a further permitted-set function or table here would duplicate that rule in a place a later change could leave inconsistent with it. If a third operation is ever added, the compile error at that match is where it gets classified.
- Do not add a second refusal path. Every post-handshake refusal goes through `refusal::refuse`.
- Do not extend, rename or overload the caller-facing error identifier set to express an authentication or authorization refusal, and in particular do not reuse `RegistrarUnavailable` for it — that would stop it being uniquely diagnostic of a broken registrar. A refused connection is silent and cleanly closed, and needs no identifier.
- Do not hold a `std::sync` guard across an `.await`; the resolver's lock is taken and released inside synchronous `resolve` and `swap`.
- No `unwrap()` in production code; no `[]` indexing on slices or maps; `thiserror` for typed errors; `pub(crate)` unless a compiler error forces wider; constants at the top of the file.

## Out of scope

- The endpoint's operation set and its dispatch. Which operations exist, how an unrecognized one is refused and how a third would be forced into an explicit decision are all settled in the serving path already; this issue neither widens nor re-implements them.
- Registering a production request handler, provisioning the verb layer inside the daemon, and withdrawing the "enabling the endpoint is unsupported" statements from the documentation.
- The payload schema, its codec, the mapping of verb outcomes onto caller-facing responses, and the response health container.
- Issuing or renewing either registrar leaf, and the configuration settings for the *client* pair's paths.
- Writing the pin file, which the provisioning tool does at install.
- The audit record contents and the rate limiter.
- Adding a reusable in-repo client for the endpoint. The tests here build their `ClientConfig` inline from `endpoint_pin::endpoint_server_verifier` plus `with_client_auth_cert`, and add no helper.

## Test plan

The tier is settled: these cases run in ordinary unprivileged `cargo test`. The merged endpoint tests bind their own listener under `tempfile::tempdir()`, `chmod` it, and hand the descriptor through the production `adopt` seam with a substituted handler, so the whole accept path — and now the whole TLS path — is reachable without root and without systemd. Extend that harness to carry a `ServerConfig` rather than writing a second one. A case needing a *successful* mint against real OpenBao still belongs to a Docker arm, and none is added here.

- [x] Handshake tests with a locally generated CA and leaves under `tempfile::tempdir()`, following the `rcgen` pattern the pin tests already use: no client certificate; a leaf from a foreign CA; a leaf from a CA in the bundle whose fingerprint is not pinned. Each asserts the connection fails and the logged typed reason.
- [x] Identity tests: an ordinary `001.roxyd.<host>.<domain>` leaf, one issued for the registrar's own host, a leaf with two SANs, and a leaf whose SAN ends in a domain that is only a string suffix of the configured one — each refused with its `RegistrarIdentityError` logged; the registrar client leaf accepted.
- [x] Caller-identity test asserting `registrar-client:<san>` reaches the handler, and asserting no `unix-peer:` value can.
- [x] Peer-credential regression test: a connection failing the peer check is refused even with a valid registrar client certificate.
- [x] Refusal-shape regression test over TLS, driven through the unrecognized-operation case: the caller reads zero application bytes, its read ends cleanly rather than as `UnexpectedEof` or a reset, and the typed refusal log carries the connection id and the operation identifier as received.
- [x] Timeout tests with `tokio::time` paused: a peer that opens a connection and sends no `ClientHello` is dropped at `HANDSHAKE_TIMEOUT`, with the typed log asserted; a peer that completes a handshake and then sends nothing is refused as `NoHeader` a full `HEADER_IDLE_TIMEOUT` after the handshake rather than after acceptance.
- [x] Server-material self-check tests, run directly on the configuration builder: leaf-only material, material whose chain builds only to an unpinned CA, and material under an expired anchor are each refused with the matching `EndpointVerifyRejection`; leaf-plus-chain material builds a configuration and a resolver.
- [x] Pinning round-trip: a client built from `endpoint_pin::endpoint_server_verifier` against a pin file written by the test, presenting the registrar client leaf via `with_client_auth_cert`, completes a mint and a deregister; a client whose pin set does not match refuses without sending a request. This is the end-to-end proof that the loaded chain is one a pinned caller accepts.
- [x] Resolver swap test, driven through the owner: take the resolver from the `ActivatedEndpoint` the harness produced, `swap` a second `CertifiedKey` into it while the endpoint is accepting, and assert the next handshake presents the new leaf with no restart. Reaching the resolver only through that accessor is the point of the test — a test that swapped a resolver it had built itself would pass while renewal remained unable to reach one.
- [x] Startup-refusal tests for each remaining case in the acceptance criteria, all on the configuration builder, asserting the diagnostic names the setting at fault and, where there is one, the configured path — and that no certificate is generated. None of them goes through `endpoint::activate`, which returns the missing-handler diagnostic before reaching this step in the current build.
- [x] Config tests: the three-key `[registrar_endpoint]` table deserializes; an absent table parses unchanged; an unknown key is still refused; a reload changing any key is rejected and names it.
- [x] Every test uses `tempfile::tempdir()` for socket roots, certificate material and pin files, binds no fixed path, reaches no network, and mutates no process environment.

## Dependencies

Part of #790. Nothing blocks this issue: the socket, the refusal path, the identity recognizer and the pinning verifier are all in `origin/main` and described above, and the work is implementable and testable against test-generated certificate material.

Certificate issuance and renewal for the two registrar leaves depend on *this* issue rather than the other way round — renewal asserts that a freshly issued leaf is still accepted here, and drives the `swap` seam Scope item 1 exposes. That is why item 1 stops at loading and swapping.

## Pointers

- `src/registrar/endpoint/serve.rs` — `caller_identity:78`, `authorize_peer:88`, `handle_connection:241` (the one concrete-stream function, where the handshake goes), `serve_request:300`, `read_header:344` (the merged unrecognized-operation refusal), `dispatch:456` (the exhaustive match that already closes the operation set; unchanged here)
- `src/registrar/endpoint/refusal.rs` — `TransportRefusalReason:101`, `Refusal:207`, `refuse:294`, `close:315`
- `src/registrar/endpoint.rs` — the deadline constants at `:90`, `ActivatedEndpoint:128` and its accessors at `:135` (where the resolver accessor joins them), `production_handler:172`, `activate:229` and its ordered startup steps, `adopt:262`
- `src/registrar/endpoint/frame.rs` — `Operation:41`, `from_name:55`
- `src/registrar.rs` — `REGISTRAR_ENDPOINT_LABEL:116`, `registrar_endpoint_identity:180`, `single_dns_san:237`, `RegistrarClientIdentity:270`, `RegistrarIdentityError:281`, `recognize_registrar_client:328`, `recognize_registrar_client_name:346`, `RegistrarEndpoint::activate:438`, `RegistrarEndpoint::activated:472` (how the daemon reaches the endpoint, and so the resolver)
- `src/registrar/endpoint_pin.rs` — `EndpointVerifyRejection:117`, `load_anchor_pins:236`, `endpoint_server_verifier:264`, `build_endpoint_client_config:290`, `RegistrarEndpointVerifier:314` with `new:339` and `verify:370`, `pinned_anchors:402`, `anchor_usability:473`, and the `rcgen` CA/leaf helpers in its test module
- `src/tls.rs` — `install_crypto_provider:226`, `parse_pem_to_cert_list:261`, `certs_to_root_store:308` (becomes `pub(crate)`), `build_pinned_verifier:327`, `build_pinned_root_verifier:339` (the pinned-subset precedent), `ca_bundle_fingerprints:451`, `sha256_hex:459`
- `src/bin/bootroot-http01-responder/tls.rs` — `ReloadableCertResolver:31`, `swap:49`, `resolve:59`, `load_certified_key:73`, `verify_cert_key_match:101`, `build_tls_config:141` and `.with_cert_resolver` at `:151`
- `src/config.rs` — `domain:47`, `Paths:111`, `RegistrarEndpointSettings:212`, `check_registrar_endpoint_reload:231`, `TrustSettings:374`; `src/config/validation.rs:199` and `:320`; `agent.toml.example` `[registrar_endpoint]` at `:52`
- `src/registrar/endpoint/tests.rs` — `Harness:1493`, `RunningEndpoint:1534`, `round_trip:1557`, `capture_logs:708`, the substitutable handlers at `:720`–`:866`, and `the_only_operations_are_mint_and_deregister:551`
- `src/bin/bootroot-agent.rs:65` — the one `RegistrarEndpoint::activate` call site
- `docs/en/configuration.md` and `docs/en/operations.md` (Registrar endpoint sections), mirrored in `docs/ko/`; `docs/reference/registrar-client-identity.md`; `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §4, §5.3, §6

AcoPiper:
## Summary

The registrar endpoint authenticated a connection by peer uid alone, which says "root on this host" rather than "the registrar". Every accepted connection is now wrapped in mutual TLS, and the endpoint admits exactly one certificate identity.

- **Server material.** `[registrar_endpoint]` gains `server_cert_path` and `server_key_path`. A new `src/registrar/endpoint/tls.rs` loads the PEM pair, proves the key matches the leaf by signing and verifying, recognizes the leaf's SAN as an endpoint identity, and then runs the *caller's own* verifier (`endpoint_pin::RegistrarEndpointVerifier`) over its own chain. A bare leaf would otherwise load cleanly and be rejected by every correctly pinned caller with no symptom on the daemon, so that case — along with a chain that builds only to an unpinned or expired anchor — is a startup refusal naming the `EndpointVerifyRejection`. The whole build is one `pub(crate)` function over its inputs, touching no descriptor, socket or process environment, so every refusal is asserted on it directly. There is no issuance, no self-signing and no fallback.
- **Client verification.** `WebPkiClientVerifier` over the *pinned subset* of `trust.ca_bundle_path` — the certificates whose DER SHA-256 appears in `trust.trusted_ca_sha256`, mirroring what the outbound path already does. `.allow_unauthenticated()` is not used, so a caller presenting no certificate fails the handshake. `certs_to_root_store` becomes `pub(crate)`; no new TLS stack and no `ClientCertVerifier` is written.
- **Identity.** After the handshake, the peer's leaf is taken through `single_dns_san` and `recognize_registrar_client_name` against the configured domain. Anything that verified but is not the registrar client name — an ordinary service leaf issued for the registrar's own host, for instance — is refused through the existing `refusal::refuse` path: zero bytes, clean close, and a typed log carrying the `RegistrarIdentityError`. The verbs now receive `registrar-client:<san>`; the uid-derived renderer is gone. `recognize_registrar_client_name` is factored onto a private helper parameterized by the reserved label, and `recognize_registrar_endpoint`/`recognize_registrar_endpoint_name` join it over `REGISTRAR_ENDPOINT_LABEL`, so the two ends of one connection cannot drift apart.
- **Ordering and budgets.** accept → connection id → permit → spawn → peer check → handshake → identity → first byte. The peer check stays first and stays a gate of its own. `HANDSHAKE_TIMEOUT` is cumulative from acceptance, and `HEADER_IDLE_TIMEOUT` now restarts at handshake completion. A failed handshake is logged under the connection id with `NoCertificatesPresented`, `InvalidCertificate(_)` and everything else distinguished.
- **Renewal seam.** The `Arc<EndpointCertResolver>` survives configuration assembly — there is no way back to it from a built `ServerConfig` — and is reachable from `ActivatedEndpoint` through a `pub(crate)` accessor. The swap test drives it through that owner rather than a resolver of its own.
- **Reload and documentation.** `check_registrar_endpoint_reload` now compares the whole table and names the key that changed. The English and Korean configuration and operations pages carry the accepted identity, the two settings, the chain requirement with the symptom a bare leaf produces, the out-of-band material, the handshake logging, the startup refusals and the non-reloadable paths. The "unsupported until a production request handler lands" statements are untouched. `docs/reference/registrar-client-identity.md` gains the endpoint name rule beside the client one — the same rule under the other reserved label, applied by the daemon to its own leaf rather than to a peer.

`frame::Operation`, `Operation::from_name` and `dispatch`'s exhaustive match are unchanged — the operation set was already closed, and no second gate on it was added.

Closes #764

Part of #790

## Test plan

- [x] Handshake tests with an `rcgen` CA and leaves under `tempfile::tempdir()`: no client certificate, a leaf from a foreign CA, a leaf from a bundle CA that is not pinned — each fails the handshake with its typed reason logged.
- [x] Identity tests: an ordinary service leaf, one issued for the registrar's own host, a two-SAN leaf, and a leaf whose SAN ends in a string-only suffix of the domain — each refused with its `RegistrarIdentityError`; the registrar client leaf accepted.
- [x] Caller-identity test asserting `registrar-client:<san>` reaches the handler and no `unix-peer:` value can.
- [x] Peer-credential regression: a connection failing the peer check is refused even with a valid registrar client certificate.
- [x] Refusal shape over TLS on the unrecognized-operation path: zero application bytes, a clean `close_notify`-terminated end of stream rather than `UnexpectedEof`, no error identifier, and the connection id plus the operation identifier as received in the log.
- [x] Timeout tests with `tokio::time` paused: a peer sending no `ClientHello` is dropped at `HANDSHAKE_TIMEOUT`; a peer that completes a handshake gets the full `HEADER_IDLE_TIMEOUT` from that completion.
- [x] Server-material self-check on the builder: leaf-only material, a chain building only to an unpinned CA, and material under an expired anchor are each refused with the matching `EndpointVerifyRejection`; conforming material builds a configuration and a resolver.
- [x] Pinning round-trip: a client built from `endpoint_pin::endpoint_server_verifier` plus `with_client_auth_cert` completes a mint and a deregister; a client whose pin set does not match refuses without sending a request.
- [x] Resolver swap driven through the `ActivatedEndpoint` accessor changes the presented chain on the next handshake with no restart.
- [x] Startup refusals for every remaining case, on the builder, each naming the setting at fault and — where there is one — the configured path, with nothing generated. Each way the material can be wrong is diagnosed as itself: a certificate or key path that is missing or unreadable, the two paths transposed, a PEM block that does not decode, a file carrying no PEM at all, and a key this build cannot sign with — and the key diagnostics quote none of the material they failed on.
- [x] Config tests: the three-key table deserializes, an absent table parses unchanged, an unknown key is still refused, and a reload changing any key is rejected by name.
- [x] Every test uses `tempfile::tempdir()`, binds no fixed path, reaches no network and mutates no process environment.
- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate`, `cargo clippy --all-targets -- -D warnings` and `./scripts/check-docs.sh` all clean. The endpoint is Linux-gated, so clippy and `cargo test --lib registrar::` were re-run in a `rust:bookworm` container: clippy clean, 406 passed with only the three known root-in-container artifacts failing (they fail identically on `main`), including all 83 `registrar::endpoint::tests`.


AcoPiper:
**[Reviewer Round 1]**

## Findings

- P2 — Startup-refusal coverage is incomplete despite the PR’s claim that every case is directly tested. The table exercises an unreadable certificate path at [tests.rs:3151](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-764-review/src/registrar/endpoint/tests.rs:3151), but for `server_key_path` it tests only a missing file and a mismatched key ([tests.rs:3165](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-764-review/src/registrar/endpoint/tests.rs:3165)). There is also no direct cert/key invalid-PEM test—the only malformed-PEM test targets the CA bundle ([tests.rs:3289](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-764-review/src/registrar/endpoint/tests.rs:3289)). Add builder tests for those paths, asserting `SERVER_KEY_SETTING`/`SERVER_CERT_SETTING` and their configured paths. This is specifically required by the issue’s “every startup-refusal case” criterion and protects the diagnostics operators rely on.

The implementation otherwise follows the intended design closely: pinned-root client verification, post-handshake registrar-name scoping, preserved peer-UID gate, clean TLS refusal path, and retained resolver seam are all present. PR linkage and test-plan structure are also appropriate.

AcoPiper:

AcoPiper:
**[Author Round 1]**

Pushed as `354e992`; PR #904 description updated.

## Review item dispositions

**P2 — Startup-refusal coverage is incomplete (unreadable `server_key_path`, no direct invalid-PEM test for cert/key) — Fixed, and extended.**

The finding is correct. The builder splits its diagnostics finely (the issue requires "a file that is not PEM at all, a certificate setting pointing at a file holding no certificate, a key setting pointing at a file holding no private key, and a key of a type this build cannot sign with are each their own variant"), but only some of those variants were asserted. Checking every variant of `EndpointTlsError`, three had no direct test at all: `Unreadable` for the key setting, `Unparsable` for either material file, and `UnsupportedKey`.

- Added `"an unreadable key file"` to the `every_startup_refusal_names_the_setting_at_fault_and_generates_nothing` table (`src/registrar/endpoint/tests.rs:3172`), mirroring the certificate case — key path at a directory, asserting `SERVER_KEY_SETTING` and the configured path.
- Added `malformed_server_material_is_diagnosed_by_what_is_wrong_with_it` (`src/registrar/endpoint/tests.rs:3358`), covering four cases on the builder, each asserting both the variant and that the diagnostic names the setting and the configured path:
  - a certificate PEM block that does not base64-decode → `Unparsable` / `SERVER_CERT_SETTING`
  - a certificate path at a file with no PEM block at all → `NoCertificate` / `SERVER_CERT_SETTING`
  - a key PEM block that does not decode → `Unparsable` / `SERVER_KEY_SETTING`
  - a well-formed `PRIVATE KEY` block holding something `any_supported_type` rejects → `UnsupportedKey` / `SERVER_KEY_SETTING`

I went one case beyond what the review asked for: `UnsupportedKey` was the last untested variant, and the PR claims every startup refusal is directly asserted, so leaving it out would have kept that claim false.

The key cases also assert the rendered diagnostic contains none of the file's body — `load_certified_key` carries a comment saying the key parse error deliberately quotes no bytes, and nothing was holding it to that.

No production code changed; no doc change was needed, since `docs/en/configuration.md:375` already documents this refusal set and the behaviour is unchanged.

## Verification

The endpoint is Linux-gated, so everything ran in a `rust:bookworm` container:

- `cargo clippy --all-targets -- -D warnings` — clean (first pass flagged `items_after_statements` on a test-local `const`; moved to the top of the function per the repo's constant rule, re-run clean).
- `cargo test --lib registrar::endpoint::tests` — 83 passed, 0 failed.
- `cargo test --lib registrar::` — 406 passed, with only the three known root-in-container artifacts failing (`reports_an_unreadable_active_file_as_a_scan_failure`, `reports_an_unlistable_store_directory_as_a_scan_failure`, `a_capture_that_cannot_complete_leaves_no_partial_snapshot` — all "unreadable file" tests that cannot fail as root, and that fail identically on `main`).
- `cargo fmt -- --check --config group_imports=StdExternalCrate` — clean, run on the host.

AcoPiper:
**[Reviewer Round 2]**

## [Reviewer Round 2]

No findings. Approved.

The prior P2 is resolved: the updated tests cover unreadable `server_key_path` plus malformed certificate/key PEM, no-certificate material, and unsupported keys, while asserting the affected setting and path. See [tests.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-764-review/src/registrar/endpoint/tests.rs:3177) and [tests.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-764-review/src/registrar/endpoint/tests.rs:3318).

The current diff continues to meet the issue’s mTLS, pinned client verification, post-handshake registrar-name scoping, resolver seam, reload, and documentation requirements.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Terminate mTLS on the registrar endpoint
```

**Body**

```text
The endpoint authenticated a connection by peer uid alone, which says
"root on this host" rather than "the registrar". The authority it
wields is exactly two calls, so who may make them has to be an
identity and not a uid.

Every accepted connection is now wrapped in mutual TLS. The endpoint
presents the configured leaf and its chain, requires a client
certificate that verifies against the pinned subset of the
deployment's own CA bundle, and then admits exactly one name: the
registrar client identity. A leaf that verified against the
deployment CA but carries any other name is refused through the
existing pre-verb path — zero bytes, a clean close and a typed log —
rather than from inside a verifier, so the refusal shape stays one
shape and every line correlates under the connection id. The peer
check stays first and stays a gate of its own: it costs one syscall
and answers a different question.

The material self-check is what turns a whole class of failure into a
startup refusal instead of every caller failing. A caller selects its
trust anchors from the certificates the server actually presents, so
a file holding the leaf alone loads cleanly, matches its key, carries
the right name, and is then rejected by every correctly pinned caller
with no symptom on the daemon at all. The loader therefore runs the
caller's own verifier over the daemon's own chain rather than a
second rule written here, and the diagnostics are split as finely as
an operator needs — the two paths transposed is the likeliest way to
get this wrong, and both halves of it parse as PEM.

The certificate resolver is retained on the activated endpoint
because a built ServerConfig keeps only a trait object and the trait
does not extend Any: renewal reaches the resolver there or nowhere.
Nothing here issues, renews or self-signs anything.

Closes #764

Part of #790
```

PR #905, opened 2026-08-25
Provision the reserved audit store and bind the OpenBao audit device into it (#888)

### LINKED ISSUE #888: Provision the reserved audit store and bind the OpenBao audit device into it
# Provision the reserved audit store and bind the OpenBao audit device into it

## Context

The registrar's detection story rests on two artifacts on the bootroot host: the verb-level records the daemon writes for every `mint` and `deregister`, and the OpenBao file audit device that records the mint's OpenBao writes. Both are fail-closed inputs to a live security argument, and today they are written in unrelated places with no shared budget.

Two facts drive the design. The OpenBao file audit device is **mandatory** — `openbao/openbao.hcl` declares it at `file_path = "/openbao/audit/audit.log"`, `bootroot init` refuses to proceed without it (`verify_audit_file`, `src/openbao.rs:664`), and OpenBao fails requests it cannot write — so a filesystem that fills stops OpenBao serving, the renewal path for every certificate in the deployment, on the one host that must not be restarted. And the two writers sit on **different sides of the container boundary**: OpenBao is a Compose service writing into the `openbao-audit` named volume (`docker-compose.yml:12`, `docker-compose.deploy.yml:29`), while the daemon that will write the verb records is a host process. A budget covering only the volume leaves the records on the host root filesystem, where filling them still stops OpenBao when the two share a backing device.

The configuration that names the store already exists. What does not exist is the directory itself, or any reason for OpenBao to write into it. This issue creates that layout on the host during a root-run `bootroot init` and renders the Compose override that binds the OpenBao audit device into it.

**Be precise about which writer actually starts writing here.** The OpenBao device moves for real: after this change its bytes land in `<audit_store_dir>/openbao` on the host. The daemon's verb records do not, because in this build there is no writer to move — `AuditRecordStore` is `#[allow(dead_code)] // Transitional` and nothing outside its own tests constructs one. What this issue delivers for that half is the directory at `<audit_store_dir>/records`, root-owned and `0700`, which the record store's trust checks accept and which the configuration already resolves `audit_record_dir` to. Wiring the daemon to open a store is separate work, and nothing here — code comments, error text or documentation — may suggest records are being written today.

This issue enforces nothing and measures nothing: the budget stays a number, and separate issues in this group make it a kernel-enforced ceiling and report on the store's capacity.

### What the configuration already provides

The `[registrar]` table in the daemon's configuration file carries the store's description, and this issue consumes it without changing it:

- `audit_store_dir` — an absolute path, default `/var/lib/bootroot/audit-store`, holding `records/` for the daemon's verb records and `openbao/` for the OpenBao file audit device. Validation rejects a relative value.
- `audit_store_reserve_bytes`, `audit_store_low_water_bytes` and `audit_store_enforcement` — the budget, the alarm threshold and the operator's enforcement mode. This issue reads none of them; they are named so it is clear the table is complete and needs no fifth key.
- `audit_record_dir` — resolves to `<audit_store_dir>/records` when omitted, and is a configuration error when set to a path outside `audit_store_dir`. The daemon's record directory is inside the store by construction, so this issue's only job for it is to create the directory the daemon will find.
- `validate_registrar_settings` (`src/config/validation.rs`) is `pub` and re-exported from `src/config.rs`, so a binary-crate caller reaches it as `bootroot::config::validate_registrar_settings` — the way `src/commands/` already reaches `validate_cert_duration_vs_default_renew_before`. The install side calls that same function, so one struct and one validation block govern the keys on both sides.

That configuration work is the subject of the sibling issue this one depends on, and it lands first.

### What the repository already provides

Everything here is in `main` today, stated so the work below can assume it without reading anything else.

- **The record store exists** as `AuditRecordStore` (`src/registrar/audit.rs`). `open` creates missing ancestors at `0755`, audits the store directory's **immediate parent**, creates the store directory at `0700` and audits that too (`prepare`, `:1797`); an audited path is refused when it is a symlink, not owned by uid 0, or group- or world-writable (`check_directory`, `:2318`). It is `#[allow(dead_code)] // Transitional` — nothing outside its own tests constructs one, and there are no records on disk to migrate.
- **`[registrar_endpoint] enabled` exists** (`src/config.rs:137`), defaults to `false` including when the table is absent, and is rejected off Linux. **Setting it to `true` makes the daemon refuse to start** in this build — `activate` bails because no request handler is registered (`src/registrar/endpoint.rs:237`) — and both manuals say enabling it is unsupported until the registrar protocol work lands.
- **The host's registrar role is recorded separately, in `state.json`**, as `registrar_endpoint: { enabled, host, domain }` (`src/state.rs:123`). No command writes that entry — storing it belongs to the registrar endpoint work, and `src/state.rs:114` says so — but `bootroot init` **consumes** it through `registrar_endpoint_intent` (`src/commands/init/steps/registrar_internal.rs:92`) and already gates on it: an endpoint-enabled host is what forces TLS on `:8200` (`src/commands/init/steps/orchestrator.rs:1133`). `scripts/impl/run-registrar-internal-init-e2e.sh:324` seeds it and drives a full endpoint-enabled `init` in the default CI E2E matrix.
- **An endpoint-enabled `bootroot init` already runs as root**, because it publishes the bootroot-internal credential under uid 0 and refuses otherwise — which is why that E2E invokes that one command through `sudo -n`. Every other command in the scenario, `infra install` and `infra up` included, runs unprivileged.
- **Rendered Compose overrides have an established shape**, and this issue reuses the two parts that fit: a renderer writing under `<compose dir>/secrets/openbao/` with `!override` rather than `!reset` (`write_openbao_exposed_override`, `src/commands/guardrails.rs:1152`), and a bring-up resolver that returns a path only when the deployment should carry it, hard-errors when it should and the file is gone, and is collected in `run_infra_up` (`src/commands/infra.rs:118`) and passed to Compose with `-f` (`resolve_openbao_exposed_override`, `:1426`; error text `error_openbao_override_file_missing`). The port override's third part — an operator intent in `state.json` (`save_openbao_bind_intent`, `:1609`) — is deliberately **not** reused: this override carries no operator choice, being a function of one configuration key and one recorded predicate.
- **The `openbao` service's entrypoint runs as root.** Neither compose file sets `user:` on that service, and both run `chown openbao:openbao /openbao/audit && chmod 700 /openbao/audit` before delegating to `docker-entrypoint.sh` (`docker-compose.yml:17`, `docker-compose.deploy.yml:34`).
- **Command-side error text goes through the `Messages` i18n catalog** (`src/i18n/en.rs`, `src/i18n/ko.rs`), unlike configuration validation, which bails with inline strings.
- **`bootroot init` recreates the OpenBao container** through `openbao_recreate_invocation` (`src/commands/init/steps/openbao_transition.rs:176`), which takes a single optional override path today. Its doc comment notes that an endpoint-enabled loopback host transitions with no exposed-port override at all, so on exactly the path this issue cares about the audit override is the only one.
- **No install-side command reads `agent.toml`.** There is no `Settings::new` anywhere under `src/commands/infra.rs`, `src/commands/guardrails.rs` or `src/commands/init/`.

### The store directory, and the contract it must satisfy

On an endpoint-enabled host, create the **host filesystem directory** named by `audit_store_dir`, with `records/` and `openbao/` beneath it.

**The store directory contract, stated once here and referred to by name below:** `audit_store_dir` is a directory and not a symlink, owned by uid 0, at mode **exactly `0700`**, reachable through ancestors that are all non-symlink directories carrying `o+x`. `records/` is held to the same ownership and mode. **`openbao/` is outside this contract**: this issue requires, compares and asserts nothing about its owner or its mode, and the only thing checked about it anywhere here is that it is a directory and not a symlink. That silence is deliberate and is not a prohibition — finding it owned by the container's uid, or by whatever another component set, is equally fine, and nothing here will contradict, repair or reject it.

Exactly `0700`, rather than the record store's looser "not group- or world-writable", is load-bearing: the install-side division of labour rests on an unprivileged process being unable to traverse the store, which is what lets `bootroot infra up` check the store directory and nothing beneath it, and a `0755` store repeals that while still passing every record-store trust check. The ancestors' `o+x` keeps the store directory `lstat`able without privilege and grants no listing without `o+r`. Both are checks over what is at the path when they run, never over how it got there: a plain directory and the root of a filesystem mounted there are the same thing to `lstat`.

### Moving the OpenBao device: a rendered Compose override

Render a **new** override file — its own name, alongside and independent of the exposed-port override, under `<compose dir>/secrets/openbao/` — replacing the `openbao` service's `volumes:` list with `!override` so `/openbao/audit` is a bind mount of `<audit_store_dir>/openbao` instead of the `openbao-audit` named volume. The exposed-port override rewrites `ports:` and this one rewrites `volumes:`, so the two compose without interacting and both may be passed with `-f` in one invocation. The container path is unchanged, so `openbao/openbao.hcl` needs no edit, the audit device's `file_path` is untouched and `verify_audit_file` keeps passing.

- `!override` replaces the **whole** list, so the override must re-declare the service's other mounts — `openbao-data:/openbao/file` and `./openbao:/openbao/config:ro` — verbatim. Dropping one silently unmounts OpenBao's storage or its configuration. Use `!override` and not `!reset`: `!reset` discards the new value and removes the attribute entirely, which is why the port override uses `!override`.
- **Keep the `openbao-audit` volume declared** in both compose files. It stops being mounted where the override applies; it does not stop existing. `OPENBAO_NAMED_VOLUMES` (`src/commands/clean.rs:34`), the reinit rendering assertion (`src/commands/reinit.rs:2213`), the reinit plan messages (`src/i18n/en.rs:541`, `src/i18n/ko.rs:541`) and `scripts/impl/run-reinit-recovery.sh:704` name it, and its contents are an existing deployment's audit history.
- **`openbao/` ends up owned by the container's user without this issue doing anything.** The service's entrypoint runs as root, and against a bind mount its `chown`/`chmod` of `/openbao/audit` operate on the host directory. So the install path added here needs no host-side chown, discovers no numeric uid and edits no entrypoint. That is a statement about what this issue has to do, not a rule about what may be done to that directory elsewhere.
- **`audit_store_dir` staying `0700` and root-owned does not block the container.** A bind mount's source is resolved by the Docker daemon, which runs as root; the container reaches the directory through its own mount point and never traverses the host parent. Do not loosen the parent's mode to fix a permission problem that is not there.

### The store's location: one authoritative value

**`[registrar] audit_store_dir` in the daemon's configuration file is the only definition of where the store is.** Nothing records a second copy of it — not `state.json`, not a flag, not an environment variable — because both writers must resolve into one directory, and two values that can drift are two directories waiting to happen.

`bootroot init` gains `--agent-config <path>`, naming **the operator's `bootroot-agent` configuration file**. `InitArgs` (`src/cli/args.rs:1099`) carries no such option today; the new one follows the spelling and meaning of `ServiceAddArgs::agent_config` (`:1404`) and `VerifyArgs::agent_config` (`:1661`), and names the operator's file — *not* the bootroot-internal `agent.toml` that `init` itself generates under the internal credential directory. This issue reads the operator's file and neither reads nor writes the generated one.

- `init` reads **exactly two tables** from that file and nothing else: `[registrar]`, for the store's location, and `[registrar_endpoint]`, for the enablement value the lifecycle section below cross-checks. **Structural deserialization and semantic validation part company here.** Both tables are deserialized under exactly the daemon's structural rules — same fields, same defaults, same `deny_unknown_fields` — so a file the daemon would refuse to parse is refused here too. But `validate_registrar_settings` runs over `[registrar]` only, and **no semantic rule at all** is applied to the endpoint's enablement, whose only such rule is a platform check belonging to a process about to listen. `init` does not construct a whole `Settings` either: that requires a non-empty `[[profiles]]`, and an installer has no business demanding a complete daemon configuration to learn one path and one boolean.
- **Read it through the `config` crate, already a dependency, and add no new one.** Follow `parse_toml` (`src/registrar/config.rs:512`): a `config::Config` over a single `config::File` in `config::FileFormat::Toml`, deserialized into **one partial-file struct carrying both tables and nothing else** —
  - `#[serde(default)] registrar: RegistrarSettings`, so an absent `[registrar]` table yields `RegistrarSettings::default()` and the documented defaults apply;
  - `#[serde(default)] registrar_endpoint: RegistrarEndpointSettings` — **the daemon's own type, reused rather than cloned**. It is `pub` (`src/config.rs:137`), carries `#[serde(deny_unknown_fields)]` (`:136`) and exactly one field, `#[serde(default)] pub enabled: bool`, so reusing it inherits the daemon's structural rules by construction; a hand-written copy would be silently wrong the day a second key joins that table. An absent `[registrar_endpoint]` table, or one without the key, reads as `false`.
  - **Reusing the type does not import the platform rule, and cannot.** The off-Linux rejection is not in that type's `Deserialize` impl; it lives in `validate_registrar_endpoint_settings` (`src/config/validation.rs:159`), a **private `fn`** not among the items re-exported from `src/config.rs:14`, so the binary crate has no way to call it. `deny_unknown_fields` comes along and the semantic rule provably does not — the split guaranteed by the module boundary rather than by an instruction someone has to remember.
  - The practical consequence to test for: `[registrar_endpoint] enable = true` — the missing `d` — is a **hard error** naming the path and the underlying deserialization failure. It must never read as `enabled: false` and then be cross-checked as a disagreement, or silently agree with a recorded `false`, leaving an operator who thought they had enabled the endpoint with a run that provisioned nothing and said nothing.

  Every other table the operator's file carries — `[[profiles]]`, `[acme]` and the rest — is ignored rather than fought with. Do **not** reuse `Settings::file_builder` (`src/config.rs:414`): it layers `apply_defaults` and marks the file `required(false)`, so a nonexistent path would silently deserialize into defaults instead of failing.
- **The flag is required on any run that would provision or unwind**, and omitting it there is a hard error naming the flag, with nothing created, rendered or deleted. Either condition alone makes it required: the recorded predicate is enabled, or **a rendered audit override is present on disk** — the mark of a host that was provisioned, which keeps the deletion of an override from being the one operation that never reads the daemon's configuration.
- The flag carries two readings: `audit_store_dir`, from `[registrar]`, and the endpoint's enablement, from `[registrar_endpoint]`, cross-checked against `state.json` as the lifecycle section below sets out. Falling back to the documented defaults would be the two-location failure this section exists to prevent: an operator whose `audit_store_dir` is non-default would get a store provisioned and mounted at the default path while the daemon resolved the configured one, with no later check able to catch it. The path must name the configuration the daemon actually runs with, so provisioning comes **after** that file exists — an ordering requirement the documentation states.
- **Where neither condition holds — predicate not enabled, no rendered override — the flag is optional and the run is a no-op.** That is the ordinary non-endpoint host, with nothing to provision and nothing to unwind. It is still never ignored when supplied there: `init` parses the file and cross-checks its enablement like any other run, so a file enabling the endpoint against a recorded `false` is refused rather than quietly disregarded, and its `audit_store_dir` is read and validated even though no store is provisioned.
- A supplied path that does not exist, cannot be read, does not parse as TOML, whose `[registrar]` table does not **deserialize** into `RegistrarSettings`, or whose `[registrar]` table fails validation is a **hard error naming the path and the underlying error**. Deserialization is a third distinct failure, separate from parsing and from validation: the table is `deny_unknown_fields` and `audit_store_enforcement` is an enum, so perfectly good TOML still fails here on an unrecognised key or enforcement value, before any validation rule runs (`src/config.rs:1362` proves that rejection fires through this crate). Surface whichever error occurred rather than flattening the three into one message. A file with no `[registrar]` table is not an error: the documented defaults apply, because that is the daemon's own configuration stating them.

`bootroot infra up` needs no such flag and reads no configuration: the rendered override names its own bind source, so the override file is the record of what `init` resolved.

### The install-side lifecycle: which command does what, and what decides it

**One enablement answer, checked against both places it is recorded.** `bootroot init` gates on `state.json`'s `registrar_endpoint.enabled`, read through the existing `registrar_endpoint_intent` — the predicate `init` already uses to force TLS on `:8200` and to publish the internal credential, so gating the store on it keeps one `init` run from treating a host as an endpoint host for one purpose and not for another.

That recorded state is not the daemon's own answer, though. The daemon reads `[registrar_endpoint] enabled` from its configuration file, the two are written by different hands, and nothing today makes them agree. So `init` **also reads `[registrar_endpoint] enabled` from the `--agent-config` file when it is given one, and refuses to proceed when the two disagree** — a hard error naming both sources and both values, with no directory created, no override rendered and no existing override deleted on that run. Only a run that gets past that check has an enablement answer at all, and that answer is the recorded predicate.

- The dangerous direction is a recorded `enabled` over a configuration that disables the endpoint: it would bind the deployment's mandatory audit device into a store for a daemon whose own configuration says the endpoint does not exist, leaving the audit path somewhere that configuration does not account for. Failing closed on it is the point of the check.
- The other direction — a configuration enabling the endpoint on a host whose `state.json` records `false` or nothing — is caught by the same check whenever `--agent-config` is supplied, which is why the flag is cross-checked rather than ignored there. It catches the operator who enabled the endpoint in the daemon's configuration and never recorded the state, whose run would otherwise provision nothing, publish no credential and force no TLS, silently.
- **Be honest about the check's reach.** It fires only on the file `init` was handed, so where the flag is legitimately omitted — no enabled predicate and no rendered override — a mismatch on disk goes undetected. That is not a hole this issue opens: `init` on such a host forces no TLS, publishes no internal credential and provisions no store, exactly as today.
- **The check is a refusal, never a resolution.** `init` must not pick a winner between the two sources, must not rewrite `state.json` from the configuration, and must not rewrite the configuration from `state.json`. Which one is wrong is the operator's to know.
- Read that boolean as a plain field of the file being parsed. Do **not** run the daemon's own `[registrar_endpoint]` validation over it: the rule rejecting an enabled endpoint off Linux belongs to a daemon that is about to listen, and an installer is not one.

**`bootroot infra install` is not touched, and provisioning is deliberately not placed there.** It runs unprivileged, so it cannot create the uid-0 `0700` directory the record store's trust checks require without a privilege escalation it has never needed, and on a fresh host it runs before the daemon's configuration file exists, which would make the ordering requirement install-blocking rather than a documented precondition. `bootroot init` already runs as root on an endpoint-enabled host and already gates on the recorded predicate, so provisioning lives there. The residual this accepts, stated rather than hidden: OpenBao comes up on the `openbao-audit` named volume during install and moves to the store at the first `init` that follows, and what it wrote before that stays in the volume, as the documentation says.

**`bootroot init` — provisions, on an endpoint-enabled host, as root.** This is the one command in this flow that runs as root, which is why every privileged step lives in it. With the enablement answer `true` it resolves `audit_store_dir` as above and then runs **both preflight refusals first** — the stale-override check and the ancestor-chain check, in that order — and only afterwards creates, renders or recreates anything. The order is normative, not incidental: an acceptance criterion requires a stale override to fail with neither the store nor the mount touched, and a check that ran after `mkdir`, after the render, or after the Compose recreate could not deliver that.

- **Refuses a stale override, before anything else.** A rendered override naming a different store directory than the configuration now resolves is a hard error naming both paths, raised before any directory is created, before the override is rewritten and before any Compose invocation, so the run leaves the store and the running mount as it found them. `init` does not silently re-point the mount: moving the store leaves the old audit history behind. The documented move is to stop the stack, relocate the directory, delete the rendered override and re-run `init`.
- **Checks the ancestor chain, still before anything is created.** Every existing component of the path above the store must be a non-symlink directory — checked with `lstat` — carrying the world-execute bit. Missing components are created at `0755`, the same `ANCESTOR_DIR_MODE` the record store uses. The default `/var/lib/bootroot/audit-store` satisfies the rule wherever its ancestors are the conventional `0755` or are created by this step; a host on which `/var/lib/bootroot` already exists at a tighter mode is refused like any other. An ancestor without `o+x` is a **hard error naming that directory**, raised before any directory is created and before any override is rendered, so a rejected path leaves no half-provisioned host behind.
- **Creates `audit_store_dir` and `records/`** if missing, root-owned and `0700`. Where one already exists it is **checked, not repaired**: any departure from the store directory contract above is a hard error naming the path and reporting what was found. **The message states the remedy as well as the fault** — the owner and mode required, and that `chown 0:0` followed by `chmod 0700` fixes it. Widening a store directory silently is how a store stops being one, and narrowing somebody's deliberate change is no better, so the error is the only place an operator is told which two commands to run.
  - Every mode `init` accepts and does not repair has to be a mode the later unprivileged `infra up` accepts too. A `0750` store passes a group-writable test, passes the record store, then fails every subsequent `infra up`; `init` succeeding into a deployment that cannot be brought up is what the exact-mode rule prevents. `records/` is held to the same mode because `init` created it at `0700`, so any other mode means something changed it and `init` does not know what.
  - These checks look only at the store directory itself, so a `lost+found` inside a mounted store is not something they see or object to.
- **Creates `openbao/`** root-owned and `0700` **only when it does not exist**, and otherwise leaves it exactly as it is, with one requirement on an existing one: it must be a **directory and not a symlink**, checked with `lstat`. Anything else there is a hard error naming it, because that path is about to become a bind-mount source — Docker would otherwise resolve a symlink to a directory nobody chose, or fail the container start with a message about a mount rather than about the store. Beyond that one check, **`init` requires, compares and asserts no owner and no mode**: after the first container start the entrypoint owns that directory, so `init` must not re-assert root ownership, chown or chmod it, or compare its owner or mode against anything — reusing the `audit_store_dir` check here would fail every re-run on a deployment that has ever started.
- **Renders the override** with `<audit_store_dir>/openbao` as the bind source and passes it to Compose, so the `up -d --force-recreate openbao` at `src/commands/init/steps/openbao_transition.rs:195` comes up on the bind mount. The invocation builder takes a single optional override path today and must be widened to carry both it and the audit override. The command-line order is pinned so it is testable and cannot drift: base compose file, then the exposed-port override when there is one, then the audit override. The E2E's loopback host has no exposed-port override, so the case where **both** are present is guarded by the unit test below instead. `run_infra_up` collects the same pair the same way.

With the enablement answer `false` — the recorded predicate disabled and the `--agent-config` file's `[registrar_endpoint] enabled` also `false` — `init` deletes any rendered audit override, creates nothing, touches no store directory and selects nothing. Because a rendered override makes the flag mandatory, that deletion is never reached without an agreeing configuration in hand; the switch-off block below sets out what each of the other cases does instead.

**`bootroot infra up` — selects, and checks only what an unprivileged process can.** With the recorded predicate enabled it requires the rendered override to be present and fails with an error naming `bootroot init` when it is not, mirroring the existing missing-override error. It then reads the bind source out of that override — the way `validate_openbao_override_binding` already reads a rendered mapping back — and `lstat`s **the store directory that bind source sits in, and nothing else**: the *parent* of the bind source, which is `audit_store_dir` by construction, held to the store directory contract.

The parent and not the bind source itself, stated so no implementer reaches for the more obvious path. Resolving `<audit_store_dir>/openbao` needs search permission on a root-owned `0700` directory, so an unprivileged `lstat` of the bind source returns `EACCES` on every correctly provisioned host — failing exactly where it should pass — while `lstat`ing `audit_store_dir` needs search permission only on the directories **above** it, which the ancestor rule guarantees.

**The expected owner is a parameter, not a hard-coded `0`.** Production passes uid 0; a test passes its own uid — the shape `check_directory(path, expected_uid)` already uses in the record store. Compare against a literal and the check becomes untestable outside a privileged runner.

`infra up` deliberately does **not** look inside the store: establishing what is inside it is `bootroot init`'s job and re-checking it at open time is the record store's. It creates nothing and writes no state, so with the predicate off it selects nothing and a rendered file left on disk stays inert until an `init` removes it.

**Switching an endpoint host off is a coordinated change, and this issue defines it end to end.** Both sources have to reach `false`, because the cross-check refuses on disagreement and a refused run changes nothing:

- Flipping only `state.json` to `false` stops the bind mount immediately — `infra up` reads no configuration and gates on the recorded predicate alone, so it selects no audit override from that moment, and OpenBao returns to the `openbao-audit` named volume on its next recreate. The rendered file stays on disk, inert.
- The rendered file is deleted by the next `init` whose two sources **agree on `false`**, and while that override is on disk `--agent-config` is mandatory, so the agreement is always verified rather than assumed. An `init` still handed the enabling configuration is refused on disagreement; an `init` given no configuration fails naming the flag. Both leave the rendered file where it is, and both are the operator's cue to finish the change in the daemon's configuration — the only run that deletes the file.
- **The store directories and everything in them are left exactly as they are** throughout, whichever order the two flips happen in. The documentation states the two-step procedure, says where those files remain, and says removing them is manual.

### Documentation

Update **both** `docs/en/` and `docs/ko/`, on the existing pages, with no `mkdocs.yml` nav change:

- The audit-logging sections (`docs/en/operations.md:95`, `docs/ko/operations.md:87`), which currently state unconditionally that the log is persisted on the `openbao-audit` Docker volume: say where it lives on a provisioned endpoint host, **that this requires `state.json`'s `registrar_endpoint.enabled` and the daemon configuration's `[registrar_endpoint] enabled` to agree — a mismatch makes `init` refuse and leaves the audit log where it was** — that records written before the store was provisioned remain in that volume and are read as before, and what switching the host off does and does not move.
- The install-side flow: that `[registrar] audit_store_dir` is the single definition of the store's location, that `bootroot init --agent-config` is how the install side reads it and is **required** on an endpoint-enabled run — so the daemon's configuration must exist first — that `bootroot init` creates the store and must run as root on such a host, that every directory above the store must be world-traversable and why, and the procedure for moving a provisioned store: stop the stack, relocate the directory, delete the rendered override, re-run `init`.
- Switching an endpoint host off: set `[registrar_endpoint] enabled = false` in the daemon's configuration **and** `registrar_endpoint.enabled` to `false` in `state.json`; until both agree `init` refuses, while a bring-up stops applying the override as soon as the recorded predicate is `false`. That final run still needs `--agent-config`, since a rendered override keeps the flag mandatory. The override is deleted by the first `init` that sees the two agree, and the store's contents are never touched.
- **The audit-record sections of all four pages say that bootroot does not create the store, and this change makes that false.** `docs/en/configuration.md`, `docs/ko/configuration.md`, `docs/en/operations.md` and `docs/ko/operations.md` each open their audit-record section with a status paragraph written as two separable sentences: one saying bootroot does not create the store, one saying it writes no verb records. The first is superseded here and is replaced on all four pages, in each page's own language, by what is now true: that a root-run `bootroot init` creates `audit_store_dir` with `records/` and `openbao/` on an endpoint-enabled host, that it must run as root, and that a host whose registrar endpoint is not enabled still has no store. The second sentence stays: creating the directory writes no record, nothing in production constructs a record store, and the trail stays empty until the separate issue that opens one at daemon startup. Revise the four paragraphs together or none of them — an English page saying the store is created beside a Korean page saying it is not is worse than leaving both alone.
- That this change establishes a shared location and that nothing enforces the configured budget yet, so nobody reads a configured reserve as a guarantee it is not yet backed by.

## Acceptance criteria

- [x] With `state.json` recording `registrar_endpoint.enabled = true` and `--agent-config` naming a file whose `[registrar_endpoint] enabled` is `true` and whose `[registrar] audit_store_dir` names the store, `bootroot init` creates `audit_store_dir`, `records/` and `openbao/` root-owned and `0700` when they are missing, renders the audit override and passes it to Compose. `audit_store_dir` and `records/` are still root-owned and `0700` after the whole run, so no recursive ownership change reached them.
- [x] A record store opens successfully over the layout `bootroot init` produces: a test drives the directory-creation function `init` uses, then opens a store at `<audit_store_dir>/records`, exercising the same parent-and-directory trust checks production uses.
- [x] A re-run of `bootroot init` over an already-started deployment succeeds with `openbao/` owned by the container's user: `init` requires of an existing `openbao/` only that it is a non-symlink directory, leaves its owner and mode untouched, and asserts no expected owner or mode for it. A test passes it a directory owned by a non-root uid and asserts the run succeeds; a symlink or a regular file there is a hard error naming it.
- [x] A re-run over an `audit_store_dir` or `records/` that is not a directory, is a symlink, is not owned by uid 0, or is at any mode other than exactly `0700` — `0750` and `0755` included — fails naming that path and what was found, and repairs nothing. Every mode `init` accepts is a mode the unprivileged `infra up` check also accepts.
- [x] Both the `init` and the `infra up` store-directory errors state the **remedy** as well as the fault — the owner and mode required, and that `chown 0:0` then `chmod 0700` on the named path fixes it — asserted for both.
- [x] `bootroot init --agent-config <path>` reads exactly two tables from that file — `[registrar]` for `audit_store_dir` and `[registrar_endpoint]` for the enablement boolean — through one partial-file struct, without requiring a `[[profiles]]` block, applying `validate_registrar_settings` to the first and no semantic rule to the second. Both deserialize under the daemon's structural rules: the endpoint half reuses `RegistrarEndpointSettings` rather than a local copy, so `deny_unknown_fields` holds on both.
- [x] An unknown key in **either** table is a hard error naming the path and the underlying deserialization failure — `[registrar] audit_store_dirr` and `[registrar_endpoint] enable` alike — never silently read as a default. A path that is missing, unreadable, not valid TOML, whose `[registrar]` table carries an `audit_store_enforcement` outside the two variants, or whose `[registrar]` table fails validation is likewise a hard error naming the path and the underlying error; a file with no `[registrar]` table yields the documented defaults.
- [x] No semantic rule is applied to the endpoint boolean: a file whose `[registrar_endpoint] enabled` is `true` loads successfully on a target where the daemon's own validation would refuse to start, and nothing in the install path calls `validate_registrar_endpoint_settings` — which it could not, that function being private to the library crate and unexported.
- [x] Omitting `--agent-config` is a hard error naming the flag, with nothing created, rendered or deleted, on both runs that require it: one whose recorded predicate is enabled, and one whose predicate is `false` but which finds a rendered audit override on disk. With the predicate not enabled and no rendered override, omitting it is accepted and the run is a no-op — but supplying it there still cross-checks the file, so a file enabling the endpoint against a recorded `false` is refused. A test covers every combination of recorded value, flag presence and rendered-override presence.
- [x] `bootroot init` refuses to proceed when `state.json`'s `registrar_endpoint.enabled` and the `--agent-config` file's `[registrar_endpoint] enabled` disagree, with an error naming both sources and both values. Nothing is created, rendered or deleted on that run, and neither source is repaired from the other. A test covers both directions, and asserts the installer does not apply the daemon's off-Linux rejection to that boolean.
- [x] Every existing directory above `audit_store_dir` is verified to be a non-symlink directory carrying `o+x`, and missing ones are created at `0755`, before anything is created or rendered. An ancestor without `o+x` is a hard error naming that directory, leaving no directory created and no override rendered.
- [x] Nothing records a second copy of the store's location: `state.json` gains no audit-store key, no new flag records a path, and `bootroot infra install` behaves exactly as it does today.
- [x] A rendered override naming a different store directory than the configuration now resolves is a hard error from `init` naming both paths, raised before any directory is created, before the override is rewritten and before any Compose invocation, so neither the mount nor the store contents change. A test asserts the ordering: over a layout that does not yet exist, the run leaves the store directory uncreated, the rendered file byte-identical and no Compose command issued.
- [x] With no `registrar_endpoint` entry in `state.json`, and with one whose `enabled` is `false`, `init` creates no store directory and renders no override, and no compose invocation carries an audit override.
- [x] **Switching off:** from a host provisioned with both sources `true`, flipping `state.json` to `false` makes every selection path stop applying the audit override immediately. An `init` still handed the enabling configuration is refused on disagreement, and an `init` given no `--agent-config` fails naming the flag since the rendered override makes it mandatory; both leave the rendered file on disk. Only an `init` supplied a configuration whose `[registrar_endpoint] enabled` is `false` deletes it. The store directories and their contents are untouched throughout.
- [x] With the predicate enabled, `infra up` selects the rendered override and `lstat`s exactly one path — the **parent** of the bind source named in it, which is `audit_store_dir` — holding it to the store directory contract. It never stats the bind source itself, never descends into the store, creates nothing and writes no state. A missing override, or a store directory failing that check, is an error naming `bootroot init`. The expected owner is a parameter rather than a literal `0`, and a test drives the check **unprivileged**, under its own uid, against a `0700` store reached through world-traversable ancestors.
- [x] A deployment carrying **both** an OpenBao non-loopback bind intent and an enabled registrar predicate reaches Compose with the base compose file and both overrides, in that pinned order, and the merged result keeps both the intended host port publication and the audit bind mount. Asserted for the `init` recreate and for `infra up`.
- [x] The rendered override binds `<audit_store_dir>/openbao` at `/openbao/audit` under `volumes: !override` and re-declares `openbao-data:/openbao/file` and `./openbao:/openbao/config:ro` verbatim; `openbao/openbao.hcl` is unmodified; and `openbao-audit` is still declared in both compose files.
- [x] Every existing path in the tree that drives an endpoint-enabled `bootroot init` — the one E2E script that seeds the predicate, and any unit test reaching that branch — supplies `--agent-config` and still passes.
- [x] The endpoint-enabled Docker E2E asserts that the container's `/openbao/audit` is backed by `<audit_store_dir>/openbao` on the host, that `verify_audit_file` succeeds at init, and that `assert_openbao_audit_log` passes unchanged — that helper reads the container path through `docker exec`, so a correct bind mount is transparent to it.
- [x] The default lifecycle E2E, which records no registrar endpoint, is unchanged: no override rendered, no store directory created, and the audit log still found on the `openbao-audit` volume.
- [x] No page still says bootroot does not create the audit store. Each of `docs/en/configuration.md`, `docs/ko/configuration.md`, `docs/en/operations.md` and `docs/ko/operations.md` says, in its own language and in the same opening paragraph of its audit-record section that carried the superseded claim, that a root-run `bootroot init` creates `audit_store_dir` with `records/` and `openbao/` on a host whose registrar endpoint is enabled, and that a host without one still has no store. The neighbouring statement that no verb record is written yet survives on all four pages, and nothing else in those sections — the record format, the rotation and retention bounds, the file names, the ownership and permission rules, the provisioning arithmetic and the reading recipes — is restructured or changed.
- [x] The audit log's location, the requirement that the two enablement values agree and what a mismatch does, the install-side flow, the required flag, the ancestor requirement, the store-relocation procedure, the two-step switch-off and the statement that the budget is not yet enforced are documented in both `docs/en/` and `docs/ko/` with no `mkdocs.yml` nav change. `./scripts/check-docs.sh` passes.
- [x] `cargo clippy --all-targets` is warning-free and `cargo fmt -- --config group_imports=StdExternalCrate --check` passes.

## Constraints

- `audit_store_dir` and `records/` must satisfy the store directory contract stated in Scope, whether this run created them or found them, and a run that finds otherwise fails naming the path, what it found and the two commands that fix it. Nothing here repairs a mode or an owner. `openbao/` is outside that contract: the code added here must not chown it, chmod it, or assert an owner or mode for it, and must accept whatever it finds. Never chown `audit_store_dir` recursively and never loosen its mode.
- Do not add an exemption, a mount-point special case or a "skip the check when `audit_store_enforcement` is `filesystem`" branch. The checks are about what is at the path when they run, never about what put it there, and an exemption keyed on the enforcement mode would disable them on the deployments that reserve the most.
- This issue creates the store's directories from `bootroot init` and nowhere else, because `init` is the only command in this flow that runs as root on an endpoint-enabled host. It does not modify `bootroot infra install` at all, and `bootroot infra up` creates nothing, chowns nothing and repairs nothing — in particular it never falls back to creating a store directory at a lesser ownership when it finds none.
- No command may require read access below `audit_store_dir` unless it runs as root — the directory is root-owned `0700`, so a check on `records/` or `openbao/` from `infra install` or `infra up` is unimplementable. Conversely `audit_store_dir` itself must be reachable without privilege, because `infra up` is required to check it; do not answer an unreachable store with a caught `EACCES` treated as "not provisioned".
- `--agent-config` is mandatory on an `init` whose recorded predicate is enabled **and** on one that finds a rendered audit override on disk, so that both provisioning and unwinding are cross-checked against the daemon's own configuration. Do not delete a rendered override on the strength of `state.json` alone. Do not add a fallback, a search path, a conventional default location, or an inference from `state.json` — each reintroduces a second answer to where the store is. There is exactly one definition of the store's location, `[registrar] audit_store_dir`; do not record it in `state.json` and do not add a flag carrying a path of its own.
- Gate on `state.json`'s `registrar_endpoint.enabled` and nothing else, and cross-check it against `[registrar_endpoint] enabled` in the `--agent-config` file, failing closed when they disagree. Do not add a third enablement key, do not gate on `audit_store_dir` being recorded or present on disk, and do not infer enablement from the socket or from installed units. The cross-check refuses; it never resolves or reconciles either source. Do not write that `state.json` entry, and do not add a flag that sets it.
- Do not change install behaviour, the rendered Compose files, `state.json`'s contents or the documented audit-log location for a deployment whose `state.json` records no enabled registrar endpoint.
- Switching an endpoint host off leaves the store directories and their contents alone: it changes what is mounted, never what is stored, and never deletes an audit artifact.
- Do not edit `openbao/openbao.hcl`'s audit stanza, and do not change the audit device's container path. The backing store moves; the device's configuration does not.
- Do not edit either compose file's `openbao` entrypoint — its chown and chmod of `/openbao/audit` are what give the bind-mounted directory its owner — do not remove the `openbao-audit` volume declaration, and do not delete or migrate its contents. Do not modify the exposed-port override or its renderer: the audit override is a separate file with a separate name, rendered and resolved alongside it. Every new command-side error carries its text in both `src/i18n/en.rs` and `src/i18n/ko.rs`.
- Do not redefine the record encoding, the rotation bounds, the retention policy or the record store's trust checks. Those are settled and in `main`; this issue creates a layout they accept.
- Do not add, rename, retype or re-default any `[registrar]` key, and do not duplicate its validation. This issue reads the table through the re-exported validator and changes nothing about it.
- Do not claim, in code comments, error text or documentation, that the configured reserve is enforced, or that the daemon's verb records are being written to the store. This change moves the OpenBao device onto the store and creates the directory the records will one day land in; enforcing a budget over it, and wiring a record writer, are both separate.
- Revise only the creating half of the four audit-record status paragraphs. Do not delete or soften the sentence saying no verb record is written yet, do not reword the surrounding claims that read as the contract a future writer honours, and do not add a section to any of the four pages: where the store's contents come from, and the point at which a verb record is first written, are documented by the issue that wires the writer.
- Do not **change** `[registrar_endpoint]` or the daemon's use of it: its shape, default, validation, `Settings::registrar_endpoint`, `check_registrar_endpoint_reload`, the socket units and `activate` are untouched. This issue only **reads** that table from the operator's file, reusing `RegistrarEndpointSettings` for deserialization. Adding a key to it, relaxing its `deny_unknown_fields`, or making its platform rule reachable from the binary crate are all out.
- Add no new dependency. The partial-file read uses the `config` crate already in `Cargo.toml`, by the idiom already in the tree; `toml` is not a dependency of this package and must not become one.
- Add **no `CHANGELOG.md` entry**. The `[registrar]` table, the record store and the registrar endpoint all landed after the most recent release tag and have never shipped, so nothing here is observable to a user of the last release.
- No `unwrap()` in production code; no `[]` indexing. Tests use `tempfile::tempdir()` and never a fixed path, and never mutate the process environment.

## Out of scope

- **Defining the `[registrar]` configuration keys** — adding `audit_store_dir`, `audit_store_reserve_bytes`, `audit_store_low_water_bytes` or `audit_store_enforcement`, deriving `audit_record_dir` from the store, or validating any of them. Those land in the sibling this one depends on, and are read here.
- **Enforcing the reserve** — a loopback-backed filesystem, a mount unit and its ordering ahead of both writers, a daemon-side mount-point check, or any behaviour that makes `audit_store_enforcement` mean something. An open sibling covers this; this issue mounts nothing.
- **Measuring the store and reporting it** — a `statvfs` probe, a subtree walk, headroom arithmetic, a capacity alarm, or any health member carrying one. An open sibling covers this.
- **Bounding the OpenBao audit device's own growth** — rotating, pruning or capping it.
- **Wiring the record store into the daemon**: opening it from the `[registrar]` table at startup, writing a verb record, or failing the registrar surface closed on an unusable store. Nothing constructs a store in production today, and that stays true here.
- Making the registrar endpoint serve, registering a request handler, or anything else that would let `[registrar_endpoint] enabled = true` start a daemon. Likewise, writing the `state.json` predicate this issue reads.
- The **rate limiter** in every part: the token buckets, the `rate_limit_*` keys, the coalesced counted records, `RegistrarBusy { retry_after_seconds }` and any limiter health member.
- Migrating audit records already in the `openbao-audit` volume, moving them into the store, or deleting store contents when the host is switched off. Relocating an already-provisioned store is likewise a documented manual procedure, not an automated one.

## Test plan

Unit and integration tests, all over `tempfile::tempdir()` and none mutating the process environment:

- [x] **Predicate**: a recorded `true` with an agreeing configuration creates the layout at the expected owner and mode and renders the override; a recorded `false` and an absent entry each create nothing and render nothing.
- [x] **Store over the created layout**: build the layout through the directory-creation function `init` uses, then open a record store at `<audit_store_dir>/records` and assert it succeeds.
- [x] **Idempotency**: a second run over an existing layout succeeds and changes nothing; an existing `openbao/` at a different owner and mode is left untouched, while a symlink and a regular file there each fail naming it; an `audit_store_dir` or `records/` that is a symlink, foreign-owned, at `0750`, or at `0755` each fail naming the path and what was found, with nothing repaired.
- [x] **Config resolution**: a `[registrar]`-only file resolves `audit_store_dir` with no `[[profiles]]` present; a file with no `[registrar]` table yields the defaults; a missing path, an unreadable path, malformed TOML, an unknown key in `[registrar]`, an `audit_store_enforcement` outside the two variants, and a `[registrar]` table failing validation each fail naming the path and the underlying error.
- [x] **Required flag**: with the predicate enabled and no `--agent-config`, the run fails naming the flag with nothing created or rendered; with the predicate `false`, a rendered audit override on disk and no flag, the run fails naming the flag and the rendered file is still there afterwards; with the predicate not enabled, no rendered override and no flag, the run succeeds as a no-op.
- [x] **Enablement agreement**: recorded `true` over a configuration whose `[registrar_endpoint] enabled` is `false` or whose table is absent, and recorded `false` over a configuration enabling it, each fail naming both sources and both values, with nothing created, rendered or deleted and neither source rewritten; agreement in either direction proceeds as that value dictates.
- [x] **Partial-file load, structural half**: `[registrar_endpoint] enable = true` fails the load naming the path and the deserialization error rather than reading as `false`; an unknown key under `[registrar]` fails the same way; an absent `[registrar_endpoint]` table and a present-but-empty one both yield `enabled: false`.
- [x] **Partial-file load, semantic half**: `[registrar_endpoint] enabled = true` loads on a target where the daemon's own platform validation would reject it. Where the suite runs only on Linux, assert the weaker property — that the load path calls no endpoint validator.
- [x] **Ancestor traversal**: a store under a `0700` ancestor is refused naming that ancestor with nothing created or rendered; `0755` ancestors are accepted; a symlinked ancestor is refused; a missing ancestor is created at `0755`. One case then confirms the `infra up` check succeeds under the test process's own uid against the accepted layout.
- [x] **Stale override**: render an override for one store directory, resolve a different one from the configuration, and assert `init` fails naming both paths. Prove the ordering, not only the message: point the configuration at a store directory that does not exist yet and assert the run creates no directory, leaves the rendered file byte-identical and issues no Compose invocation.
- [x] **Switching off**: provision with both sources `true`, flip `state.json` to `false`, and assert a bring-up selects no audit override while the rendered file is still on disk; that an `init` still passed the enabling configuration is refused and leaves that file in place; that an `init` given no `--agent-config` fails naming the flag and also leaves it in place; that an `init` supplied a configuration whose `[registrar_endpoint] enabled` is `false` deletes it; and that both store subdirectories and their contents are unchanged throughout.
- [x] **`infra up` resolution**, driven under the test process's own uid by passing that uid as the expected owner: a present override over a store directory at the expected owner and `0700` resolves; an absent override, a symlinked store directory, one at `0750`, and one owned by a different uid each return an error naming `bootroot init`, and no case creates a directory or writes state. One case puts an unreadable `openbao/` under an otherwise valid store and asserts the check still succeeds.
- [x] **Override rendering**: the rendered file binds `<audit_store_dir>/openbao` at `/openbao/audit` under `volumes: !override`, re-declares the other two mounts verbatim, leaves `openbao/openbao.hcl` byte-identical, and leaves `openbao-audit` declared in both compose files.
- [x] **Combined overrides**: with a recorded non-loopback OpenBao bind intent **and** an enabled predicate, assert the recreate invocation carries the base compose file, the exposed-port override and the audit override in that order, that the merged service keeps both the publication and the bind mount, and that the same pair reaches `infra up`'s collected overrides. Cover port override only, audit override only, and neither.
- [x] **Remedy text**: both store-directory errors name the required owner and mode and the `chown 0:0` / `chmod 0700` that fixes them.

Docker E2E, extending `scripts/impl/run-registrar-internal-init-e2e.sh` — the only script that seeds the predicate and drives an endpoint-enabled `init`:

- [x] Its phases already run in this issue's order (install, seed the predicate, `bootroot init` under `sudo -n`, then unprivileged `infra up`), so nothing is reordered and `install_infra` is untouched. In the seed phase, beside the `state.json` predicate, write the operator configuration file this run passes to `--agent-config`: `[registrar] audit_store_dir` naming the store base below, and `[registrar_endpoint] enabled = true` so the two sources agree. Nothing in this script starts a `bootroot-agent` daemon, so an enabled `[registrar_endpoint]` in a file only the installer reads starts nothing and refuses nothing.
- [x] Assert after `init`, through `sudo -n` since the store is root-owned `0700`: the store exists with `records/` and `openbao/` beneath it; `openbao/` is no longer root-owned once the container has started while `audit_store_dir` and `records/` still are; OpenBao's `/openbao/audit` is backed by it; `verify_audit_file` passed; and `assert_openbao_audit_log` passes unchanged. Then assert the closing unprivileged `infra up` succeeds and the audit log is still found — the proof that the unprivileged select path works against a store it cannot descend into.
- [x] **The store's base is a physical, world-traversable temporary directory — not the run root, not `$TMPDIR`, not the checkout.** Every obvious location fails the ancestor rule: `RUN_ROOT` is `mktemp -d` and so `0700`, and loosening it is not the answer since it holds the run's secrets tree; `$TMPDIR` is a `0700` per-user directory on macOS; logical `/tmp` is a symlink to `private/tmp` there; and anywhere under the checkout inherits a developer home that is often `0750`. So resolve `/tmp` **physically** with `cd -P /tmp && pwd -P`, create a directory there from an explicit `mktemp -d` template under that resolved base so `$TMPDIR` is never consulted, and `chmod 0755` it. That directory is the store's **parent**; `audit_store_dir` is a path beneath it that `bootroot init` creates. Assert every component of that physical path is a non-symlink directory carrying `o+x` before `init`, so a drifting harness fails with its own message rather than through `init`'s ancestor error. Teardown removes the base alongside `RUN_ROOT` through the `sudo -n` the script already requires.
- [x] Regression: the default local lifecycle E2E, which records no registrar endpoint, still passes with no override rendered, no store directory created and the audit log still on the `openbao-audit` volume.
- [x] **Documentation read-through**: the audit-record sections of all four pages are read end to end in their own language — none still says bootroot does not create the store, each says a root-run `bootroot init` creates it on an endpoint-enabled host and that a host without an enabled endpoint has none, each still says no verb record is written yet, and the four agree. A search over `docs/` finds no surviving sentence in either language claiming the store is not created.
- [x] `scripts/preflight/ci/e2e-matrix.sh` before pushing, and `./scripts/check-docs.sh` for the manuals.

## Dependencies

Depends on the sibling issue in this group that adds the four `audit_store_*` keys to the `[registrar]` table, derives `audit_record_dir` from `audit_store_dir`, makes `validate_registrar_settings` `pub` and re-exported so a binary-crate caller can reach it, and writes the audit-record status paragraph this issue's documentation work revises. This issue reads that configuration and creates the directory it describes; it defines no key of its own. Everything else it builds on is in `main`: the record store and its trust checks, the `state.json` registrar-endpoint predicate and the `init` path that consumes it, the rendered-override lifecycle, and the OpenBao recreate invocation.

The two checks this issue implements — `init` re-checking `audit_store_dir` and `records/` on every run, `infra up` checking `audit_store_dir` unprivileged on every bring-up — apply to whatever presents that path, a directory this issue created or the root of a filesystem mounted over it. That is a consequence of the checks, not an instruction to anyone, and this issue waits on nothing for it: at merge time there is no mount anywhere, and every criterion above is evaluated against a plain directory this issue's own `bootroot init` created.

Part of #772.

## Pointers

- `src/config.rs:88` (`RegistrarSettings`) and `src/config/validation.rs:120` (`validate_registrar_settings`), reached as `bootroot::config::validate_registrar_settings` through the re-export at `src/config.rs:14`; `src/commands/ca.rs:24` is an existing binary-crate call through that door; `src/config/validation.rs:98` is the non-empty-`[[profiles]]` rule that makes a whole-`Settings` load the wrong tool here
- `src/config.rs:136`-`:142` (`RegistrarEndpointSettings`) — the type the partial-file struct reuses; `src/config/validation.rs:159` (`validate_registrar_endpoint_settings`) — the private, unexported platform rule the install side neither calls nor can call
- `src/registrar/config.rs:512` (`parse_toml`) — the single-file `config::Config` idiom to follow; `src/config.rs:414` (`Settings::file_builder`) — the one **not** to reuse; `src/config.rs:1362` — the test proving an unknown `[registrar]` key is refused through this crate
- `src/registrar/audit.rs:1797` (`prepare`) and `:2318` (`check_directory`) — the trust checks the created layout must satisfy, with `ANCESTOR_DIR_MODE`/`STORE_DIR_MODE` at `:111` and `:114`
- `src/state.rs:123` (`StateFile::registrar_endpoint`) — the predicate; `src/commands/init/steps/registrar_internal.rs:92` (`registrar_endpoint_intent`) is its reader and `src/commands/init/steps/orchestrator.rs:1133` an existing gate on it
- `src/commands/infra.rs:118` (`run_infra_up`) — where resolved overrides are collected and passed to Compose; `:1426` (`resolve_openbao_exposed_override`) — the resolver to model the new one on, with its missing-file error; `validate_openbao_override_binding`, in `src/commands/guardrails.rs`, called from `src/commands/infra.rs:1446` — how a mapping is read back out of a rendered override
- `src/commands/guardrails.rs:1152` (`write_openbao_exposed_override`) — the renderer to model the new one on
- `src/commands/init/steps/openbao_transition.rs:176` (`openbao_recreate_invocation`) — the single-optional-override argument that must carry two, and `:195` — the `up -d --force-recreate openbao`
- `src/cli/args.rs:1099` (`InitArgs`, spanning to `ReinitArgs` at `:1297`) — no `agent_config` field today; the spelling to follow is `ServiceAddArgs::agent_config` at `:1404` and `VerifyArgs::agent_config` at `:1661`, the latter read by `src/commands/verify.rs:329`
- `src/i18n/en.rs` and `src/i18n/ko.rs` — the catalogs every new command-side error text goes in
- `docker-compose.yml:10`-`17` and `docker-compose.deploy.yml:27`-`34` — the `openbao` service's `volumes:` list and root-run entrypoint; the `openbao-audit` declarations are at `docker-compose.yml:146` and `docker-compose.deploy.yml:152`
- `src/commands/clean.rs:34` (`OPENBAO_NAMED_VOLUMES`), `src/commands/reinit.rs:2213`, `src/i18n/en.rs:541`, `src/i18n/ko.rs:541`, `scripts/impl/run-reinit-recovery.sh:704` — the places naming `openbao-audit` that must keep working
- `src/openbao.rs:664` (`verify_audit_file`) and `openbao/openbao.hcl` — the mandatory device and the stanza whose `file_path` stays `/openbao/audit/audit.log`
- `scripts/impl/run-registrar-internal-init-e2e.sh` — `install_infra` at `:277`, predicate seeding at `:324`, the root `init` at `:348`, the closing unprivileged `infra up` at `:610`; `scripts/impl/lib/audit-log.sh:21` (`assert_openbao_audit_log`), called from `run-local-lifecycle.sh:1423` and `run-remote-lifecycle.sh:1126`
- `docs/en/operations.md:95` / `docs/ko/operations.md:87` — the audit-logging sections to update
- `docs/en/configuration.md`, `docs/ko/configuration.md`, `docs/en/operations.md`, `docs/ko/operations.md` — the four audit-record sections, each opening with the two-sentence status paragraph whose creating half this issue retires; the paragraph sits in the same place on every page so the edit is mechanical
- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §5.6 and §6

AcoPiper:
## Summary

`bootroot init` now provisions the reserved audit store on a host whose `state.json` records an enabled registrar endpoint, and renders a Compose override that binds the OpenBao file audit device into it. Only the OpenBao half actually starts writing there: nothing in this build constructs a record store, so `<audit_store_dir>/records` is created and stays empty, and no code, error text or documentation added here claims otherwise.

**The store's location has exactly one definition.** `init` gains `--agent-config <path>`, naming the operator's `bootroot-agent` configuration file, and reads exactly two tables from it — `[registrar]` for `audit_store_dir` and `[registrar_endpoint]` for the enablement boolean — through one partial-file struct built from the daemon's own types, so `deny_unknown_fields` holds on both without requiring a `[[profiles]]` block. `validate_registrar_settings` runs over the first; no semantic rule at all is applied to the second, and the install path calls no endpoint validator. Nothing records a second copy of the path: no `state.json` key, no flag, no environment variable.

**The two enablement answers are cross-checked, never reconciled.** `init` gates on `state.json`'s `registrar_endpoint.enabled` through the existing `registrar_endpoint_intent`, and refuses when the `--agent-config` file's `[registrar_endpoint] enabled` disagrees, naming both sources and both values, with nothing created, rendered or deleted. The flag is mandatory wherever a run would provision or unwind — an enabled predicate, or a rendered audit override on disk — so neither happens on the strength of `state.json` alone.

**One store directory contract, checked and never repaired.** `audit_store_dir` and `records/` must be non-symlink directories, owned by uid 0, at mode exactly `0700`, reached through non-symlink ancestors carrying `o+x`; missing ancestors are created at `0755`. `init` re-checks them on every run and the unprivileged `infra up` checks `audit_store_dir` — the *parent* of the bind source named in the rendered override, never the bind source itself — on every bring-up, with the expected owner a parameter rather than a literal `0`. Both errors state the remedy: `chown 0:0` then `chmod 0700` on the named path. A run that would provision and is not running as that expected uid is refused before anything is created: the directories are created owned by whoever creates them, so an unprivileged `init` would otherwise leave behind exactly the store the next root run refuses as foreign-owned and does not repair. `reinit` inherits that refusal through the shared preflight. Exact `0700` is what lets the unprivileged check exist at all. `openbao/` is deliberately outside the contract — the container entrypoint owns it from the first start, so `init` requires only that it is a non-symlink directory and asserts no owner and no mode.

**Both preflight refusals run before anything is created.** A rendered override naming a different store than the configuration now resolves, and an ancestor without `o+x`, are hard errors raised before any `mkdir`, any render and any Compose invocation, so a refused run leaves the store and the running mount as it found them.

**Compose.** A new override file, alongside and independent of the exposed-port override, replaces the `openbao` service's `volumes:` list with `!override` so `/openbao/audit` binds `<audit_store_dir>/openbao`, re-declaring `openbao-data:/openbao/file` and `./openbao:/openbao/config:ro` verbatim. `openbao/openbao.hcl` is untouched, both compose files still declare `openbao-audit`, and the recreate invocation and `run_infra_up` both carry base file, then exposed-port override, then audit override in that pinned order. A store path Compose cannot carry is refused at the render, naming both characters: Compose splits a bind mount's value on `:` to find its source, target and mode, so a colon in `audit_store_dir` moves that boundary rather than travelling inside it whatever the value is quoted with, and a control character has no spelling Compose reads back. Nothing downstream would have caught it — the reader strips the same container-path suffix it wrote, so the stale-override check sees no disagreement and the failure surfaces as a mount error two steps later.

Beyond the issue's literal scope, two consequences of it are handled where they fall: `bootroot reinit` gains the same `--agent-config`, threads it into the init pass and raises those refusals in its pre-wipe preflight, since every one of them was otherwise reachable only after the wipe — the partial-init trap reinit exists to recover from — plus one `init` itself never raises: reinit brings the stack back up before it re-runs `init`, and that bring-up gates on the recorded predicate alone, so it requires a rendered override the init pass has not written yet on a host that has never provisioned one; and the renderer publishes into the secrets tree through that tree's existing owner-aware writers, because a root-run `init` writing `secrets/openbao/` with `create_dir_all` left the sidecars' directory root-owned and the override unreadable to the unprivileged bring-up that has to read it back.

The endpoint-enabled E2E arm this issue's assertions live in also caught an `init` defect one layer down, fixed here because the audit assertions are what fail on it. The two infra OpenBao Agent sidecars are written before the OpenBao TLS transition and started after it, but that gate read `bind_intent` — true only for a non-loopback bind — so an endpoint-enabled loopback host generated them for the plaintext listener and started them before the recreate that takes it away. The containers came up, `init` succeeded, and the deferred phase-2 apply found them `Running` with no config change to make; both agents then spoke HTTP to a TLS port for the rest of the deployment and renewed nothing. The gate is now the same binding the TLS transition itself reads, and the harness asserts what the sidecars are generated for directly, so a regression fails with its own reason rather than as a store-backed audit log with no `auth/approle/login` in it. The same harness now captures step-ca's logs by container name rather than service name, having collected nothing but docker's `No such container` for exactly the run that needed them.

Documentation is updated on the existing pages in both `docs/en/` and `docs/ko/`, with no `mkdocs.yml` nav change. No `CHANGELOG.md` entry: the `[registrar]` table, the record store and the registrar endpoint all landed after the last release tag and have never shipped.

Closes #888

Part of #772

## Test plan

- [x] `cargo fmt -- --config group_imports=StdExternalCrate --check` and `cargo clippy --all-targets -- -D warnings` are clean
- [x] `cargo test --no-fail-fast` passes with no failures
- [x] `./scripts/check-docs.sh` passes
- [x] **Predicate**: a recorded `true` with an agreeing configuration creates the layout at the expected owner and mode and renders the override; a recorded `false` and an absent entry each create nothing and render nothing
- [x] **Store over the created layout**: the layout is built through the directory-creation function `init` uses, then a record store opens at `<audit_store_dir>/records`
- [x] **Idempotency**: a second run changes nothing; an existing `openbao/` at a different owner and mode is left untouched, while a symlink and a regular file there each fail naming it; a symlinked, foreign-owned, `0750` or `0755` `audit_store_dir` or `records/` each fail naming the path and what was found, repairing nothing
- [x] **Config resolution**: a `[registrar]`-only file resolves with no `[[profiles]]`; a file with no `[registrar]` table yields the defaults; a missing path, an unreadable path, malformed TOML, an unknown key, an out-of-range `audit_store_enforcement` and a validation failure each name the path and the underlying error
- [x] **Required flag**: every combination of recorded value, flag presence and rendered-override presence, including the no-op case and the cross-check that still fires where the flag is optional
- [x] **Enablement agreement**: both directions of disagreement are refused naming both sources and both values, with nothing created, rendered or deleted and neither source rewritten
- [x] **Partial-file load, structural half**: `[registrar_endpoint] enable = true` fails the load rather than reading as `false`; an absent and a present-but-empty table both yield `enabled: false`
- [x] **Partial-file load, semantic half**: an enabled endpoint loads with no semantic rule applied, and a source scan asserts the install path names no endpoint validator
- [x] **Unprivileged provisioning**: a run whose expected uid is not the one it runs as is refused naming both, with no store created and no override rendered; the reinit preflight raises the same refusal before the wipe; an unwinding run is not held to it
- [x] **Ancestor traversal**: a `0700` ancestor is refused naming it with nothing created or rendered; `0755` ancestors are accepted; a symlinked ancestor is refused; a missing one is created at `0755`, and the `infra up` check then succeeds under the test process's own uid
- [x] **Stale override**: fails naming both paths, and the ordering is proved — no directory created, the rendered file byte-identical, no Compose invocation issued
- [x] **Switching off**: a flipped `state.json` stops the mount immediately with the rendered file inert on disk; an `init` still handed the enabling configuration is refused; one given no flag fails naming it; only one given a disabling configuration deletes the file; the store's contents are unchanged throughout
- [x] **`infra up` resolution**, driven unprivileged under the test process's own uid: a conforming store resolves; an absent override, a symlinked store, a `0750` store and a foreign-owned store each name `bootroot init`, creating no directory and writing no state; an unreadable `openbao/` under an otherwise valid store still passes
- [x] **Override rendering**: binds `<audit_store_dir>/openbao` at `/openbao/audit` under `volumes: !override`, re-declares the other two mounts verbatim, leaves `openbao/openbao.hcl` byte-identical, and leaves `openbao-audit` declared in both compose files
- [x] **Unrenderable store path**: an `audit_store_dir` carrying a colon, and one carrying a control character, are each refused at the render naming both characters, with the message reachable in both catalogs
- [x] **Combined overrides**: base file, exposed-port override and audit override reach both the `init` recreate and `infra up` in the pinned order, with port-only, audit-only and neither also covered
- [x] **Remedy text**: both store-directory errors name the required owner and mode and the `chown 0:0` / `chmod 0700` that fixes them
- [x] **Reinit preflight**: every refusal above is raised before the wipe, and a run that gets past it has created, rendered and deleted nothing; `--agent-config` reaches the init pass rather than being a no-op on the one command that re-runs it; and an endpoint-enabled host carrying no rendered override yet is refused there too
- [x] **Documentation read-through**: the audit-record sections of all four pages read end to end in their own language; none still says bootroot does not create the store, each says a root-run `bootroot init` creates it on an endpoint-enabled host and that a host without one has none, each still says no verb record is written yet
- [x] **Infra agents over TLS**: the rendered agent override and both `agent.hcl` files name this install's `https://<instance>-openbao:8200` and the CA bundle that verifies the listener's leaf, and the shared `assert_openbao_audit_log` finds the sidecars' AppRole login and KV read in the store-backed device. Part of the same E2E arm, so verified in CI
- [x] `scripts/impl/run-registrar-internal-init-e2e.sh` — the endpoint-enabled Docker E2E: the store exists with `records/` and `openbao/`, `openbao/` is no longer root-owned once the container has started while the other two still are, an unprivileged `bootroot init` before the root one is refused leaving no store and no override, the container's `/openbao/audit` is backed by the store, `verify_audit_file` passed, `assert_openbao_audit_log` passes unchanged, and the closing unprivileged `infra up` succeeds. Needs passwordless `sudo`, which this development host does not have — verified in the default CI E2E matrix
- [x] `scripts/preflight/ci/e2e-matrix.sh` — the default lifecycle E2E is unchanged: no override rendered, no store directory created, the audit log still on the `openbao-audit` volume. Verified in CI


AcoPiper:
**[Reviewer Round 1]**

Found one blocking issue.

- [P1] Enforce root before creating the audit store — [orchestrator.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-888-review/src/commands/init/steps/orchestrator.rs:338) calls `apply_audit_store` with `expected_uid = 0` before the existing endpoint-only work that eventually requires root. But `create_layout` creates `audit_store_dir` before checking its owner ([audit_store.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-888-review/src/registrar/audit_store.rs:339)). Thus an unprivileged endpoint-enabled `init` creates a user-owned `0700` store, then fails its own uid-0 check. A later correct `sudo bootroot init` refuses that directory as foreign-owned and deliberately will not repair it, leaving the host unable to provision without manual intervention. Check that the caller is root before `apply_audit_store` can mutate anything, and add a regression test that an unprivileged attempt leaves no store or override behind.

I found no PR-hygiene issue: the body correctly uses `Closes #888`, the permitted `Part of #772`, and includes a substantive test plan.

AcoPiper:

AcoPiper:
**[Author Round 1]**

I've pushed the fix; CI is running now. Here's the disposition of the round-1 review:

## [P1] Enforce root before creating the audit store — **Fixed**

The finding is correct. `create_layout` creates before it checks, so an endpoint-enabled `init` run without privilege built `audit_store_dir` and `records/` under the caller's own uid and only then failed its own owner check — leaving exactly the directory the later `sudo bootroot init` refuses as foreign-owned and, by contract, does not repair.

**What changed:**

- `src/commands/audit_store.rs:494` — the refusal lives in `plan_audit_store`, beside every other refusal that must precede a write, and fires only on the branch that would provision. It compares `fs_util::current_process_euid()` against `inputs.expected_uid` rather than a literal `0`, keeping the rule identical to the one the layout checks apply — so a test running under its own uid provisions exactly as production does under `0`, and production's `expected_uid = 0` means "must be root".
- Because it is in the plan rather than in `apply_audit_store`, `reinit` inherits it through `preflight_audit_store` and raises it **before the wipe**, which is where the other audit-store refusals already moved.
- An **unwinding** run is deliberately not held to it: switching a host off deletes a rendered override and creates no store. That is asserted rather than left implicit.
- New message `error_audit_store_not_privileged` in `en`/`ko`, naming the store path, the required uid and the uid the run is actually under, and stating that nothing was created, rendered or deleted.
- `create_layout`'s rustdoc now records that creating precedes checking and that the caller must already be running as `expected_uid`.

**Tests** (all three requested behaviours, plus the boundary):

- `an_unprivileged_provisioning_run_creates_nothing` — refused naming both uids, no store directory, no rendered override.
- `the_preflight_raises_the_unprivileged_refusal_before_the_wipe` — same refusal out of the reinit path, with a rendered override present so the missing-override refusal cannot be what fires.
- `an_unwinding_run_is_not_held_to_the_provisioning_uid` — the deletion path still works.
- `scripts/impl/run-registrar-internal-init-e2e.sh` gains a `refuse-unprivileged` phase that runs the real binary unprivileged against `expected_uid = 0` before the root `init`, asserting the refusal message and that no store and no override exist afterwards. That arm needs passwordless `sudo`, which this host does not have, so it verifies in CI.

**Docs:** `docs/en/operations.md`, `docs/ko/operations.md`, `docs/en/cli.md`, `docs/ko/cli.md` — the "must run as root" statements now say the run is refused up front and why. No `CHANGELOG.md` entry, unchanged reasoning: none of this has shipped.

**Local verification:** `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --no-fail-fast` (0 failures), `scripts/preflight/ci/check.sh` (clean; the one `cargo audit` warning is the pre-existing allowed `RUSTSEC-2025-0134`), and all three E2E validators pass. The PR body is updated with the new refusal and two test-plan items.

AcoPiper:
**[Reviewer Round 2]**

**[Reviewer Round 2]**

No remaining findings.

The Round 1 P1 is resolved: `plan_audit_store` now rejects a provisioning run whose effective UID differs from the required UID before `create_layout` can create anything ([audit_store.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-888-review/src/commands/audit_store.rs:506)). The regression coverage verifies no store or override is left behind, and the same preflight reaches `reinit` before its destructive work ([audit_store.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-888-review/src/commands/audit_store.rs:1481)).

The updated implementation covers the issue’s lifecycle, store contract, override composition, configuration cross-check, unprivileged `infra up` behavior, and documentation requirements. PR linkage and test-plan hygiene are also correct.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Provision the audit store and bind OpenBao into it
```

**Body**

```text
The registrar's detection story rests on two artifacts written in
unrelated places with no shared budget: the daemon's verb records and
OpenBao's mandatory file audit device.  The configuration naming a
shared store already existed; the directory did not, and nothing gave
OpenBao a reason to write into it.

`bootroot init` now creates that layout on a host whose `state.json`
records an enabled registrar endpoint, and renders a Compose override
binding the file audit device into it.  Only the OpenBao half starts
writing: nothing in this build constructs a record store, so `records/`
is created and stays empty, and no code, error text or documentation
added here claims otherwise.

`[registrar] audit_store_dir` stays the single definition of where the
store is.  `init` gains `--agent-config`, naming the operator's
`bootroot-agent` file, and reads exactly two tables from it through one
partial-file struct built from the daemon's own types, so
`deny_unknown_fields` holds on both without demanding a `[[profiles]]`
block.  Nothing records a second copy of that path — two values that
can drift are two directories waiting to happen.

The two enablement answers are cross-checked and never reconciled.
`init` gates on the recorded predicate and refuses when the
configuration disagrees, naming both sources and both values, because
which one is wrong is the operator's to know.  The flag is mandatory
wherever a run would provision or unwind, so neither happens on the
strength of `state.json` alone.

`audit_store_dir` and `records/` are checked and never repaired:
non-symlink directories, uid 0, mode exactly `0700`, reached through
non-symlink ancestors carrying `o+x`.  Exact `0700` is what lets the
unprivileged `infra up` check exist at all — it can `lstat` the store
without being able to descend into it.  A run that would provision and
is not the expected uid is refused before anything is created, since
the layout is created owned by whoever creates it and the next root run
would refuse it as foreign-owned rather than repair it.  `openbao/` is
deliberately outside the contract: the container entrypoint owns it
from the first start.

Two consequences are handled where they fall.  `reinit` raises those
refusals before its wipe rather than after it, plus one `init` never
raises, because the bring-up between the wipe and the init pass needs
an override that pass has not written yet.  And the renderer publishes
through the secrets tree's owner-aware writers, because a root-run
`init` writing `secrets/openbao/` with `create_dir_all` left the
override unreadable to the unprivileged bring-up that reads it back.

The endpoint-enabled E2E arm also caught an `init` defect one layer
down.  The two infra OpenBao Agent sidecars are written before the TLS
transition and started after it, but that gate read `bind_intent` —
true only for a non-loopback bind — so an endpoint-enabled loopback
host generated them for the plaintext listener and started them before
the recreate that takes it away.  Both agents then spoke HTTP to a TLS
port for the rest of the deployment and renewed nothing.  The gate is
now the same binding the TLS transition itself reads.

No `CHANGELOG.md` entry: the `[registrar]` table, the record store and
the registrar endpoint all landed after the last release tag and have
never shipped.

Closes #888

Part of #772
```

AcoPiper:
Rebased onto `62eb2cc` ("Serve the registrar verbs from the daemon"), which is the current `main`. The branch is now a single commit, `0c4cf7c`, on top of it.

Four documentation files conflicted; `src/commands/init/constants.rs` and `src/registrar.rs` merged without intervention, and no source or script hunk changed as a result of the rebase (the branch's diff over `src/` and `scripts/` is byte-identical to what it was before).

**`docs/{en,ko}/configuration.md` — "Registrar audit records".** `main` replaced the base paragraph with one describing the two groups of keys in `[registrar]` and their `SIGHUP` behaviour; this branch had replaced the same paragraph with one describing what `bootroot init` provisions. Both were kept, `main`'s first. The trailing sentence "bootroot does not write registrar verb records in this build" was dropped: `main` now serves the verbs and writes the records, so it had become false.

**`docs/{en,ko}/operations.md`, after "Audit logging".** Two different sections were added at the same anchor — `main`'s `#### Registrar verb rate limiting` and this branch's `### The shared audit store`. Both kept, in that order, which preserves the heading hierarchy: the `####` stays a subsection of "Audit logging" and the `###` opens a new sibling section under "OpenBao".

**`docs/{en,ko}/operations.md` — "Audit records" under the registrar endpoint.** `main` removed the "Once a writer is wired in" hedge from this paragraph, since there is now a writer. `main`'s unhedged wording was taken and this branch's provisioning paragraph placed ahead of it, again without the stale "does not write records" sentence.

The resolution leaves the branch's documentation changes purely additive over `main`; the only deletions are line rewraps. Cross-references to `#the-shared-audit-store` / `#공용-감사-저장소` from `cli.md`, `configuration.md` and `operations.md` all still resolve.

Verified locally before pushing: `cargo build --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check` with `group_imports=StdExternalCrate`, and the full `cargo test` suite (0 failures across every target, run twice). Because the change touches `docs/`, `markdownlint-cli2` and `./scripts/check-docs.sh` (`mkdocs build --strict`) were run as well; both pass.

PR #906, opened 2026-08-25
Serve the registrar verbs through the endpoint's production handler (#886)

AcoPiper:
## Summary

Makes the registrar endpoint usable. Until now `production_handler()` returned `None` and `activate` bailed whenever `[registrar_endpoint] enabled = true`, because nothing in the crate constructed a `RegistrarVerbs`. This provisions the verb layer inside the daemon, implements `RegistrarRequestHandler` on top of the merged codec, registers it, and withdraws the "unsupported" statement from the operator documentation.

**The handler comes off the activated endpoint.** `ActivatedEndpoint` no longer holds one and `adopt` no longer takes one; `serve::run` takes it as a second argument and `spawn_registrar_endpoint` supplies it, so "never accept a connection without a handler behind it" is a signature rather than a runtime invariant — no slot, no lock, no interior mutability. `endpoint::activate(enabled)` keeps its signature and its once-only call site above the reload loop. A `SIGHUP` is a coordinated stop followed by a fresh invocation, so the handler is rebuilt from the reloaded settings with no window in which one is swapped under a live accept loop.

**The verb layer's ten dependencies now have somewhere to come from.** Eight new `[registrar]` keys carry the provisioning config's path, the wrap-TTL ceiling, the two role TTLs and the per-issuance `secret_id` options; three more — the secrets directory, the recorded `OpenBao` URL and the KV mount — are read out of the `state.json` that the new `state_file` key names, through a three-member `Deserialize`-only projection in the daemon composition layer. Nothing under `src/registrar/` learns that file exists, `[openbao]` is not read for any of it, and the active root fingerprint stays derived rather than configured. Range rules are enforced unconditionally in `validate_registrar_settings`; the cross-table rule that `state_file` is required when the endpoint is enabled lives in `validate_settings`, where both tables are visible. Everything is resolved with `?` before the invocation spawns anything, because a socket-activated endpoint with nobody accepting turns every caller into a hang.

**Build nothing when the endpoint is disabled** — no provisioning config, no audit store, no credential, no state file.

**`ca_anchor` is this issue's to settle, and is settled here.** The handler reads `bootroot/ca` with its own `InternalCredential`, per mint and never cached, before the verb runs. The emitted fingerprint array is derived from the bundle in bundle order rather than copied from the stored array, the derived set is compared with the stored one case-insensitively (matching `parse_trust_payload`), the bundle is normalized to LF with exactly one trailing LF, and the validated pair is handed to the codec's own `Serialize` type so the fixed member order is pinned by declaration rather than by a `serde_json::Value`. The `bootroot/ca` path is now one `pub const` in the library that the binary's `PATH_CA_TRUST` refers to.

Part of #886

## Not addressed

The wire `spec` conversion, and everything that needs a mint to reach the verb layer. The issue declares itself blocked on an external prerequisite and forbids working around it; that prerequisite has not landed.

- **`spec.reload` and `spec.cert_group` have no recorded grammar.** All three prerequisites in Scope fail as of this branch: `docs/reference/registrar-wire-contract.md` §4.4 still records the two only as externally owned opaque strings with no spelling and no provenance, §10 carries no row for either, and the provisioning tool's `ReloadHook::DockerSighup` still has no counterpart among the four `ReloadKind` values the rendered config admits. Settling this belongs to the owner of the provisioning-config contract, and the issue is explicit that a spelling chosen here would pass this repository's tests by construction and break every enrollment at deploy time. So `requested_spec` is one named seam that returns `SpecConversionError::GrammarNotSettled` for every input, and its refusal takes the existing pre-verb `HandlerRefusal` path unchanged: zero response bytes, the transport's fixed `handler-rejected-payload` line, a clean close. Filling it in from the transcribed rows is mechanical and is the whole of what remains. The handler's own line for it is at `warn` rather than `debug`, because on a correctly provisioned deployment every mint stops there and an operator at the default log level would otherwise see only the transport's fixed line.
- **The `spec` tests enumerated from those rows.** The round-trip cases over every form the grammar admits and every form it rejects, the two undecodable-payload cases over the socket for a rejected `spec.reload` and a rejected `spec.cert_group`, and the grammar-independent property that each component in the shared config fixture's rendered spec decodes to a `RequestedSpec` equal to its `RegistrationSpec`. The issue requires all of these to be enumerated from the reference's rows and never from its own text, so there is nothing to enumerate them from yet.
- **Mint-path coverage that needs a successful mint.** A socket-carried mint reaching the verbs, `ca_anchor` decoded from an end-to-end mint response, a rewritten `bootroot/ca` observed on the next mint, and an anchor-read failure driven over the socket. The `ca_anchor` behaviour itself is implemented and covered — the read, the bundle-ordered derivation, the set comparison, the CRLF normalization, the fixed member order and the unreadable-path diagnostic are all asserted at the handler seam and at the codec framing over `wiremock`, rather than from a mint response. Deregister is covered end to end over the real socket, and the mint path is covered over the socket up to its refusal.

## Test plan

- [x] Signature-level tests hold: `ActivatedEndpoint` carries no handler, the accept task cannot be spawned without one, and nothing but the operation, the payload and the caller crosses the trait.
- [x] An enabled endpoint with its dependencies in place starts and serves, and the missing-handler refusal is gone.
- [x] A disabled endpoint loads no provisioning config, opens no audit store, builds no credential and opens no state file.
- [x] Each invocation-level dependency fails startup by name before anything is spawned: absent provisioning config, digest-mismatched provisioning config, unusable audit store, absent internal credential, stale internal credential.
- [x] A changed `[registrar]` value takes effect on the next invocation, a rewritten `state.json` does too — so the projection is read per invocation rather than cached — and `check_registrar_endpoint_reload` still rejects a change to `enabled`.
- [x] The seven defaulted `[registrar]` keys apply their defaults when absent and when the whole table is absent, read their explicit values at the documented spelling, reject an unknown key, and reject one case per range rule with the key named — including a `max_wrap_ttl` that is not a whole number of seconds and a bare integer for `secret_id_ttl`.
- [x] `state_file` is rejected when relative, required when the endpoint is enabled with the key and the enablement named, absent-and-accepted when it is disabled, and carried through when absolute.
- [x] The state file's three members are read, extra members are ignored, the projection has no `Serialize` impl, and the file is byte-unchanged after startup; an unreadable, malformed, empty-member or plaintext-URL file fails the invocation with the path and member named.
- [x] Secrets-directory resolution: absent member resolves to `secrets` beside the state file, a relative one resolves against the state file's directory rather than the process working directory, an absolute one is used as recorded, and a nonexistent resolved directory fails at startup with its path named.
- [x] No path under `src/registrar/` names the state file or the `state_file` key, and `RegistrarVerbs::internal` has exactly one production call site.
- [x] A `[registrar]` range violation is rejected on a host whose endpoint is disabled.
- [x] `SecretIdOptions` reaches the factory with `metadata: None`, `num_uses` defaulting to `0`, `ttl` absent unless set, and `token_bound_cidrs` from its key; no key can set `metadata`.
- [x] A socket-carried deregister returns a framed codec response, and reaches the real verbs with the transport-supplied `CallerIdentity` unchanged — asserted on what the verbs recorded, not on a double. The mint arm stops at the unsettled `spec` grammar before the verbs, so the deregister arm is the one that carries this.
- [x] Two socket-carried requests differing only in `idempotency_key` produce identical verb inputs and identical responses.
- [x] A typed verb refusal is framed as a wire refusal with its identifier and class, produced by the verbs themselves, with the health container riding it.
- [x] An undecodable payload is zero response bytes, a clean close and no wire identifier, with the transport's fixed line unchanged and the handler's own line separate and carrying no connection id.
- [x] `ca_anchor`: the read returns what the codec frames and `parse_trust_payload` accepts it; entry *n* is the digest of certificate *n* even when the stored array is in another order; an anchor whose stored and derived sets differ is refused; a CRLF bundle with no trailing newline is emitted as LF with exactly one; the framed bytes carry `trusted_ca_sha256` before `ca_bundle_pem`; an unreadable anchor names its mount and KV path; a rewritten object is reflected on the next read with no restart.
- [x] The library's restated defaults are pinned against the binary crate's originals, and the three `state.json` member names are pinned against the CLI's writer.
- [x] No page in `docs/en/` or `docs/ko/` still says enabling the endpoint is unsupported, and each states the supported configuration; `./scripts/check-docs.sh` passes.
- [x] `CHANGELOG.md` gains one `### Added` entry with no issue or pull request reference.
- [x] `cargo clippy --all-targets -- -D warnings` and `cargo fmt -- --check --config group_imports=StdExternalCrate` pass, on Linux as well as on the development host.
- [x] `scripts/preflight/ci/e2e-matrix.sh` passes in CI. The daemon lifecycle and the configuration surface both change here, so it is not skippable.


AcoPiper:
**[Reviewer Round 1]**

Review finding:

- [P3] Update the now-stale codec module header in [src/registrar/endpoint/protocol.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-886-review/src/registrar/endpoint/protocol.rs:8). It still says the production handler is a sibling issue and that these types are unused outside focused tests, but this PR adds and uses `ProductionHandler`. The accompanying crate-wide `#![allow(dead_code)]` is therefore justified by an obsolete explanation.

Otherwise, the PR’s `Part of #886` / `## Not addressed` handling is appropriate for the explicitly unsettled mint grammar, and the implemented deregistration and handler-wiring path is well covered.

AcoPiper:

AcoPiper:
**[Author Round 1]**

Pushed as `07383e1`. PR [#906](https://github.com/aicers/bootroot/pull/906) now has a new CI run.

## Dispositions

### [P3] Stale codec module header in `src/registrar/endpoint/protocol.rs` — **Fixed**

The finding is correct: the comment said the production handler was a sibling issue and that these types were unused outside focused tests, and this PR is what made both false.

Before rewriting it I checked whether the attribute could just be deleted, by removing it and running `cargo clippy --all-targets` on both the dev host and Linux. It cannot. On Linux, seven items are still reported, and they are a coherent set — the **caller's** half of the codec: `encode_request`, `decode_mint_response`, `decode_deregister_response`, `decode_refusal_response`, `decode_ca_anchor`, its `DecodedWireTrustPayload` helper, and `WrappedSecretId::as_str`. This daemon is only the server end of the wire: it decodes a request and encodes a response. The enrolling agent that decodes those responses is separate work, so the attribute stays — but the header now names exactly which items it covers and why they are kept beside the half that is live, rather than deleted and rewritten against the same reference later.

One correction to the finding's wording, which doesn't change the fix: `#![allow(dead_code)]` inside a non-root module file is module-scoped, not crate-wide. Nothing outside `protocol.rs` was ever affected.

### Same defect one file over: `src/registrar/verbs.rs` — **Fixed (beyond the finding)**

`src/registrar/verbs.rs:123` carried the mirror-image stale justification — "the endpoint that drives these verbs is separate work… no production code path reaches them yet" — falsified by this same PR. Leaving it would have been the reviewed defect, unrepaired, in the file the finding's reasoning applies to most directly. It also cannot be dropped, for a different reason that is now stated: the socket-activated endpoint that constructs `RegistrarVerbs` is Linux-only, so on any other host the whole layer is dead and takes the parts of `crate::registrar::audit` only these verbs call down with it (that cascade is why the codec's own dead items are invisible on macOS and only surface on Linux). On Linux what remains is the injectable constructor and the accessors only tests read.

I kept both as module-level allows rather than narrowing to `cfg_attr(not(target_os = "linux"), …)` — that form, used at `src/registrar/audit.rs:1180`, would be wrong here, since ten items in the verb layer and seven in the codec are dead on Linux too.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --document-private-items` all pass on the dev host and in a Linux container matching CI's toolchain.

No documentation, CHANGELOG, or PR-description change: this is comment-only and changes nothing a user or operator can observe. I re-read the PR body against the branch and it is still accurate, so I left it as is.

AcoPiper:
**[Reviewer Round 2]**

Approved — no follow-up findings.

The Round 1 item is resolved: [protocol.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-886-review/src/registrar/endpoint/protocol.rs:8) now accurately explains the module-scoped dead-code allowance and the still-unreached caller-side codec paths. The related update in [verbs.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-886-review/src/registrar/verbs.rs:123) is also sound.

The PR’s partial-delivery framing remains appropriate: the unsettled mint `spec` grammar is explicitly retained under `## Not addressed`, while the implemented production wiring and deregistration path match the issue’s available scope.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Serve the registrar verbs from the daemon
```

**Body**

```text
The registrar endpoint's listener was merged and refused to start:
nothing in the crate constructed a RegistrarVerbs, so
production_handler() returned None and activate bailed whenever the
endpoint was enabled. This provisions the verb layer inside the daemon
and registers a handler over the merged codec, so an enabled endpoint
starts and serves, and withdraws the "unsupported" statement from the
operator documentation.

The handler comes off the activated endpoint rather than living on it.
ActivatedEndpoint no longer holds one and adopt no longer takes one;
serve::run takes it as a second argument, so "never accept a connection
without a handler behind it" is a signature rather than a runtime
invariant, needing no slot, no lock and no interior mutability. A SIGHUP
is a coordinated stop followed by a fresh invocation, so the handler is
rebuilt from the reloaded settings with no window in which one is
swapped under a live accept loop.

The factory's ten dependencies now have somewhere to come from. Eight
new [registrar] keys carry the provisioning config's path, the wrap-TTL
ceiling, the two role TTLs, the per-issuance secret_id options and the
state file's location; three more — the secrets directory, the recorded
OpenBao URL and the KV mount — are read out of that state.json through a
Deserialize-only projection in the daemon composition layer, so a URL or
mount a later init rewrites reaches the daemon with no operator edit and
no second spelling can disagree with the deployment's own record.
Everything resolves with ? before anything is spawned, because a
socket-activated endpoint with nobody accepting turns every caller into
a hang; a disabled endpoint builds none of it.

The CA anchor is read per mint and never cached, before the verb runs so
a failed read refuses before anything is minted. Its fingerprint array
is derived from the bundle in bundle order rather than copied from the
stored array, and the validated pair goes to the codec's own Serialize
type so the fixed member order is pinned by declaration.

The wire spec conversion is left as one named seam that refuses every
input. Its two strings are spelled under a grammar no repository
records, and choosing one here is not this repository's to do: it would
pass these tests by construction and break every enrollment at deploy
time. Mint therefore stops at that refusal until the owner of the
provisioning-config contract settles it and it is transcribed, which is
what keeps this a partial delivery.

Part of #886
```

AcoPiper:
Rebased onto `d018620` ("Route opened audit entries through the classifier"), which brought in mTLS termination on the registrar endpoint, the two-bucket verb rate limiter, and the audit-entry classifier.

Conflicts resolved in seven files:

- **`src/registrar/endpoint.rs`** — main added the TLS fields (`acceptor`, `resolver`, `domain`) to `ActivatedEndpoint` and to `adopt`, while this branch removed the handler from both. Kept the TLS material, dropped only the handler field, the `handler()` accessor, the `production_handler()` gate in `activate` and the handler parameter of `adopt`. The `activate` doc steps were renumbered accordingly: an enabled endpoint now goes handler-free from the unprivileged-daemon warning to the certificate material to the activation contract.
- **`src/registrar/endpoint/serve.rs`** — kept main's mTLS connection path (handshake, `recognize_caller`, the typed handshake-failure labels) and threaded the handler through as an argument, so `serve_request` takes the parameter rather than `endpoint.handler()`.
- **`src/registrar/endpoint/tests.rs`** — the harness moved the handler from `Harness::bind(handler)` to `RunningEndpoint::start(&endpoint, handler)` across every call site, `adopt_for_test` lost its handler parameter, this branch's new production-handler tests were converted from the plain `round_trip(&socket_path, …)` to the harness's TLS `round_trip`, and the caller-identity expectations moved from `unix-peer:uid=…` to `registrar-client:<san>`.
- **`src/config.rs` / `src/config/validation.rs`** — both sides added `[registrar]` keys, so both sets are kept: main's four `rate_limit_*` keys and this branch's eight verb-layer keys. The rate-limit range checks moved into a `validate_rate_limit_sizing` helper to keep `validate_registrar_settings` under the `too_many_lines` limit. `validate_registrar_state_file_requirement` now takes `&RegistrarEndpointSettings`, since that struct stopped being `Copy` when main added the two certificate paths.
- **`docs/en/configuration.md` / `docs/ko/configuration.md`** — took this branch's "two groups of keys" paragraph; main's "that is the whole table" sentence no longer holds now that the verb-layer settings are a second block.

Beyond the conflicted hunks, three things had to follow from main's limiter: the daemon's `InternalVerbsSource` now builds a `VerbRateLimiter` from the four `rate_limit_*` keys over the shipped counting sink, and the two `RegistrarVerbsConfig` test harnesses construct one at the default sizing over the no-op sink. Three source-scraping assertions were updated to match — the factory's field list (now eleven dependencies), the `ActivatedEndpoint` field list (now including the TLS material), and `an_enabled_endpoint_no_longer_refuses_for_a_missing_handler`, which now asserts the refusal is the certificate material's rather than the activation contract's, since the handler check that used to precede it is gone.

Verified on Linux (the registrar endpoint is `#[cfg(target_os = "linux")]`, so a macOS run compiles none of it): `cargo clippy --all-targets -- -D warnings` clean, `cargo doc` clean under `RUSTDOCFLAGS=-D warnings`, `cargo fmt --check` clean, markdownlint clean, and the test suite green apart from tests that require an unprivileged uid, a resolvable gid or a `docker` binary — all environment artifacts of the container, each of which passes in a run where its precondition holds. The full suite is also green on macOS.

PR #907, opened 2026-08-25
Bound both registrar verbs with two token buckets (#786)

### LINKED ISSUE #786: Bound both registrar verbs with two token buckets
# Bound both registrar verbs with two token buckets

## Context

Every registrar verb invocation is now recorded, refusals included. `RegistrarVerbs::mint` and `RegistrarVerbs::deregister` in `src/registrar/verbs.rs` write an **intent** record when the invocation arrives and an **outcome** record when it finishes, into the daemon-owned `AuditRecordStore`. That closed a detection gap and opened a resource one: nothing limits either verb, so a compromised caller — the actor this whole mechanism exists for — can flood refused invocations and grow the record store without bound. The cheapest refusals are the ones settled before any OpenBao work, so the flood path is also the free path.

The consequence is not merely a large file. When the filesystem fills, every subsequent mint is refused, so fleet-wide enrollment stops. Worse, the OpenBao file audit device this design builds on is **mandatory** — `openbao/openbao.hcl` declares an `audit { type = "file" ... }` stanza and `bootroot init` refuses to proceed without it (`src/openbao.rs::verify_audit_file`) — and OpenBao fails requests when it cannot write its audit device. So a full disk can stop OpenBao serving, which is the renewal path for every certificate in the deployment, on the one host that must not be restarted. "Mints are rare in steady state" is true of legitimate mints and irrelevant to the attacker.

This issue builds the **mechanism** that bounds the rate: the buckets, where they are charged, and what each check point tells its caller. Two words are used precisely throughout and are not synonyms. An invocation is **limited** when this issue suppressed its audit records, which happens at *either* check point, and **throttled** when it is limited at the **admission** check point and therefore receives the retryable throttle outcome instead of an attempted verb. Every throttled invocation is limited; the reverse does not hold.

### What is already in the tree, and what this issue may therefore assume

- **The design is settled and the accepted RFC states it.** `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §5.6 specifies buckets keyed per-`(client identity, verb, bucket)` over the two-value enum `predecision_refusal` / `admission`, defines *limited* against *throttled* in those words, and states that the **admission** check point is the only one that throttles, while an invocation limited on the pre-decision path is **not** throttled and keeps the refusal its own input earned. §6 and §7 restate both. There is no open design question here and no RFC text for this issue to change; implement what that section says.
- **Both verbs write exactly one intent record per invocation, at verb entry, before any arm analysis.** `mint` and `deregister` each build an `AuditContext`, call `write_intent`, and only then run their checks — a single write for the whole invocation rather than one per arm. `ProducingArm` (`src/registrar/verbs/outcome.rs`) is therefore a classification stamped on a refusal or an outcome, not a second entry point with a second record. Any statement that each arm writes its own intent describes code that does not exist; the ordering rules below are written against the flow that does.
- **The checks that decide the pre-decision refusals are pure.** `RegistrarVerbs::pre_derivation` is a synchronous method that runs label validation, the reserved-namespace guard, the requested-spec identity check, the component multiplicity lookup and the instance-shape check. It performs no I/O, takes no lock and reaches nothing outside the process. For `mint`, `WrapTtlPolicy::grant` is the same kind of check. This is what makes the two-value bucket key knowable at the charge point.
- **The refusals the pure checks produce are not all permanent, and this issue does not make them so.** `map_refusal` in `src/registrar/endpoint/protocol.rs` classifies most of them permanent with a named identifier — an invalid `service_name` or `host` as `ServiceLabelInvalid`, an unconfigured component or a wrong instance shape as `ServiceInstanceMismatch`, a spec-identity disagreement as `ServiceSpecConflict` — but `VerbError::ReservedServiceName` and every `VerbError::InvalidWrapTtl` variant map to `RefusalClass::Retryable` with **no** identifier. Both are produced by the pure checks this issue charges the pre-decision bucket for. So the property this issue preserves is not permanence; it is that limiting changes the caller's answer in no way at all.
- **The caller identity is already stable across connections.** `src/registrar/endpoint/serve.rs::caller_identity` is the one site it is rendered at, and it renders `unix-peer:uid=<uid>` from the kernel-reported peer uid alone. The peer's pid is deliberately absent because it is reused and races the credential it was read alongside, and the gid is absent because it is not the subject. The value carries no caller input, so one caller spends one bucket rather than getting a fresh full budget per connection. The endpoint additionally admits exactly one uid (`authorize_peer` is an equality against the daemon's own uid), so in this build there is effectively one identity.
- **The throttle's caller-facing wire form already exists and is fixed.** `EnrollError::RegistrarBusy { retry_after: u64 }` in `src/registrar/endpoint/protocol.rs` is classified `RefusalClass::Retryable` by `validate_refusal_class`, is the family's only retryable identifier, and is pinned by the golden fixture `src/registrar/endpoint/fixtures/refusal-busy.json`. `docs/reference/registrar-wire-contract.md` §6.1 spells the payload `retry_after`, §7 records "retryable identifiers — exactly one: `RegistrarBusy`", and its local half already assigns that form to the registrar's future rate-limit refusal. What does not exist is a verb-layer variant that produces it: `map_refusal` is exhaustive and catch-all-free, and its own comment records that the change introducing that variant adds the arm.
- **The verbs have no production construction site yet, and this issue does not add one.** `RegistrarVerbs` is built only from `RegistrarVerbsConfig`, and today only tests build it: the endpoint's `RequestHandler` trait exists and `serve::run` is spawned from the daemon, but no production handler holds a `RegistrarVerbs` and calls `mint` or `deregister`. The limiter is therefore injected the way the audit store already is — one more fixed dependency on `RegistrarVerbsConfig`, chosen by whoever constructs the verbs and reachable from no request. Whatever wires the verbs into a production handler picks the limiter up for free; building that wiring is not this issue's.
- **The `[registrar]` configuration table exists and rejects keys it does not know.** `RawRegistrarSettings` in `src/config.rs` carries `#[serde(deny_unknown_fields)]` so that a misspelled key is a load error rather than a setting that silently does nothing, and a test pins that. Validation lives in `src/config/validation.rs::validate_registrar_settings` and names the offending key as `registrar.<key>`.

### Where the limiter runs, against the control flow that exists

The bucket an invocation is charged against is decided by whether the verbs' pure checks refuse it, and the charge must precede the intent write. Both verbs therefore run in this order:

1. **Classify.** Run the verbs' existing pure checks first. For `deregister` that is `pre_derivation` alone; for `mint` it is `pre_derivation` followed by the wrap-TTL grant, in the order they already run in. They perform no I/O and take no lock, so running them ahead of the intent write costs nothing and changes no answer they produce.
2. **Charge.** With those checks refusing, charge the invocation's `predecision_refusal` bucket. With them passing, charge its `admission` bucket.
3. **Proceed.** With a token available, the invocation continues into the flow that exists today — the intent write, then the remaining stages, then the outcome record — **reusing** the values step 1 produced rather than recomputing them. No check is evaluated twice.

With the charged bucket empty:

- **`predecision_refusal`** — write **neither** record, emit one limited-invocation event, and return the refusal step 1 already produced. The caller's answer is preserved exactly: the same `VerbError`, mapping to the same wire identifier — or the same absence of one — and the same `RefusalClass` as the identical unlimited invocation, whatever those turn out to be. Only the records are suppressed. This is the free-flood path, and it is the only place where records are suppressed and the caller is still given its own answer.
- **`admission`** — the invocation is **not attempted at all**. Return the typed throttle outcome below, take no per-id lock, make no OpenBao call, write **neither** record, and emit one limited-invocation event.

Two consequences of the landed flow, stated so they are not decided twice:

- A refusal raised **after** the pure checks pass — a `derive_registration_id` failure, a binding or safe-set refusal, an OpenBao failure — is an `admission` invocation and spends an `admission` token. The bucket key is fixed at the charge point and is never revised by what the invocation turns out to be.
- **The wrap-TTL grant belongs in the pre-decision set even though the RFC's arm description does not name it.** It is a purely local check on the request, it reaches no OpenBao and writes nothing, so a flood of out-of-range `wrap_ttl` values is as free to the attacker as a flood of malformed labels. Charging it against `admission` would let that free traffic starve legitimate mints, which is the one thing the two buckets exist to prevent.
- **`predecision_refusal` is the enumerated set of checks in step 1 and nothing else.** It is not "every refusal reached without I/O": `derive_registration_id` is synchronous and touches nothing outside the process, and its failures are `admission` refusals all the same, because they are settled after the charge. The membership rule is the list — `pre_derivation`, plus `WrapTtlPolicy::grant` for `mint` — and a check added to the verbs later joins that set only by being added to step 1, deliberately and with its own reason.
- An intent write that fails is classified `ProducingArm::PreDerivation` but is **not** a pre-decision refusal for the limiter's purposes: it happens after the charge, on an invocation whose pure checks passed, so it spends an `admission` token like any other admitted invocation.

### The buckets

- **Bucket key: `(client identity, verb, bucket)`**, where `bucket` is an enum with exactly two values, defined here and exported for the sibling issue that turns limited invocations into counted records to reuse:
  - `predecision_refusal` — the invocation was refused by the pure checks of step 1. Every such invocation is a refusal by construction, it performs no OpenBao work, and it is the cheapest thing a caller can make the daemon do.
  - `admission` — the invocation passed those checks and may therefore reach OpenBao, whatever it turns out to be.
- Two verbs × two buckets gives **four buckets per client identity**.
- **The two-value enum is the maximal refinement knowable where the limiter is charged, and that is why the key has this shape.** A pre-decision refusal *is* knowable there, because the checks that produce it are pure and can run before the charge. The outcome of an admitted invocation is not: a spec conflict, a name collision and a host mismatch are all settled by OpenBao work that follows the intent write. Keying a bucket on the outcome class would therefore require charging **after** the durable write the limiter exists to prevent, or predicting the outcome. Do not widen the key toward outcome classes.
- **The bucket key never contains the verbs' fine-grained refusal reason**, even where step 1 knows it. A key that varied with the reason would let a caller multiply its budget simply by varying its refusal.
- **The client identity is the opaque `CallerIdentity` the verbs already carry** into every invocation. Key the buckets on it as-is. Do not parse it, normalise it or strip fields out of it. It is the peer-uid rendering today and becomes the client certificate's identity when mTLS lands, with no change here and no dependency on that work.
- **Bound the map with an explicit cap, counted in bucket entries.** The limiter holds one entry per `(client identity, verb, bucket)` key, and the cap is a file-top constant on the **number of entries**, not on the number of identities: `256`, which is the reference deployment's `64` identities × the four buckets each identity can create. Entries are what consume the memory, and counting them keeps the cap in the same unit as every removal rule below — a cap counted in identities cannot be honoured by removing a bucket at all, since an identity holding one drained bucket and three full ones is not removable and freeing its three full ones frees no identity slot. With one registrar identity the cap is never reached in normal operation; it exists so that a future identity source which varies per connection becomes a bounded degradation rather than memory exhaustion on the bootroot host.
- **What a charge does, stated exhaustively, because this path is the whole defence against a caller-varying identity.** A charge whose key is already in the map charges that entry. A charge whose key is absent:
  1. **Below the cap** — insert a bucket holding a full burst, and charge it.
  2. **At the cap** — first **reclaim every entry currently refilled to its full burst**. A full entry is indistinguishable from an absent one, since both yield a full bucket on next use, so this frees slots without handing anybody a token. It never touches a drained entry, and it needs no victim-selection policy: the rule is *every full entry or none*, which is deterministic and directly testable. If it frees a slot, insert and charge as above.
  3. **Still at the cap** — every remaining entry is drained, so insert nothing and charge the invocation against the **overflow bucket** for its `(verb, bucket)`.
- **The four overflow buckets.** One per `(verb, bucket)` pair, created with the limiter, sized from the same configured burst and refill interval as the map entries for that `bucket`, never reclaimed, and **not counted against the cap**. They exist to make the miss path *total*: whatever the map holds, there is always exactly one bucket to charge, so no invocation is ever admitted unlimited for want of a slot. Charging "some existing entry for the same `(verb, bucket)`" would not be total — a saturated map need not contain that pair at all — and it would spend one identity's budget on another's traffic by an unspecified choice of victim.
- **The residual this leaves, stated rather than implied.** While the map is saturated with drained entries, every identity without an entry of its own shares the four overflow buckets, so in that degraded state one caller's flood can throttle another caller's traffic. That is inherent to any bounded map — memory cannot be bounded while a private bucket is guaranteed to an unbounded number of identities — and what matters is that it is bounded rather than unlimited.
- **The limiter is a token bucket**, not a fixed or sliding window. A fixed window lets a caller spend two full budgets back to back across a boundary; a token bucket states the legitimate burst and the sustained rate as two independent numbers, which is exactly the distinction the sizing below turns on. Tokens accrue at one token per refill interval, are capped at the burst, and are computed from a **monotonic** clock — a wall-clock step must not be able to mint tokens.
- **A bucket starts full, is held in memory only, and is created lazily.** A bucket comes into existence on its key's first use holding a full burst, so a freshly started endpoint can absorb the full documented bring-up wave immediately — which is what the sizing formula promises an operator, and an empty-start bucket would refuse the first wave on every restart. Nothing about the limiter is persisted: durable limiter state would have to live on the very store this mechanism is protecting. The cost is that restarting the daemon resets every bucket, which is acceptable because restarting it requires root on the bootroot host, and root compromise of that host is outside what this design defends against.
- **An entry is removed only while it holds a full burst — never below it, anywhere, for any reason.** That is the rule the reclaim step obeys, and it governs any other place the map is pruned. Removing a full entry is indistinguishable from recreating it; removing a drained one silently hands that caller a fresh budget, which is a limiter bypass reachable by waiting. Reclaiming full entries at other times — an idle sweep, say — is permitted and changes no observable behaviour.

### The throttle outcome

- Produce it as a **distinct variant of `VerbError`**, carrying an unsigned whole-second retry-after duration. It must be distinguishable in that surface from every permanent unavailability variant, whose reasons all mean *until an operator acts*.
- **The throttle is returned only from the admission check point, and the invariant that keeps it there is preservation rather than permanence.** Limiting suppresses the **record**, never the **answer**: a limited invocation returns exactly the refusal, identifier and `RefusalClass` it would have returned unlimited, and `RegistrarBusy` is never substituted for an answer the daemon has already determined. That holds for the permanent pre-decision refusals and equally for the retryable-unclassified ones, and it is what makes limiting invisible to a caller except in the record store. The admission check point is different in kind: there the daemon has determined nothing, so a throttle adds an answer rather than replacing one.
- **Why substituting `RegistrarBusy` on the pre-decision path would be wrong, not merely different.** `RegistrarBusy` is the enroll family's only retryable **identifier**, and its contract is *this clears on its own after `retry_after`*. None of the pure checks' refusals clear on their own — a reserved `service_name`, a malformed label, an unconfigured component and an out-of-range `wrap_ttl` all fail identically on the next attempt — so the ones already classified permanent would be contradicted in the type, and the ones classified retryable-unclassified would gain a false promise about when to retry. Either way a broken caller retries a request that can never succeed, on the path this design already names as the cheapest to flood, and the limiter feeds the traffic it was added to damp.
- **The payload is a duration, not a deadline**: an unsigned whole-second count, never an absolute timestamp, so it does not depend on the caller's clock agreeing with the daemon's. Its value is the time until the charged `admission` bucket holds one token, derived from that bucket's elapsed accrual against the admission refill interval, rounded up to whole seconds, with a floor of `1` so a value of `0` can never invite an immediate retry. It is deterministic; document that a caller with several outstanding requests should jitter its own retries.
- **The variant and its `map_refusal` arm land together, here.** `map_refusal` in `src/registrar/endpoint/protocol.rs` is exhaustive with no wildcard, so the tree stops compiling the moment the variant exists with no arm for it, and the mapping's own comment already assigns the arm to the change that introduces the variant: an arm added ahead of the variant would be unreachable, and a variant added ahead of the arm would not build. Add both in this issue. The arm carries no design content — it is one line of routing that maps the variant onto the **already-defined** `EnrollError::RegistrarBusy { retry_after }`, carrying the payload the limiter produced, with the `RefusalClass::Retryable` that `validate_refusal_class` already requires for that identifier. It introduces no second wire identifier, no second refusal class, no competing fixture and no change to `docs/reference/registrar-wire-contract.md`: the identifier, its payload spelling, its retryable class and its golden fixture `src/registrar/endpoint/fixtures/refusal-busy.json` are all in the tree already, and this issue only routes to them.
- **Suppressing the record is not suppressing the evidence, and this issue's entire share of that is emitting the event.** A limited pre-decision invocation still emits its one limited-invocation event through the sink, exactly as step 1 requires, and that emission is the whole obligation here. Coalescing those events into a counted record, and amending the record reader's anomaly definition to account for them, belong to the sibling record issue and are listed in Out of scope; do not implement a coalescing counter here. The event may never be skipped along with the records: skip it and the chain that counts a flood has nothing to count, making a flood least visible exactly while it is worst.

**What the two buckets do and do not guarantee — stated plainly, because the boundary matters.** A flood on the `predecision_refusal` path cannot consume `admission` budget, so the path an attacker can drive for free, with no valid input and no OpenBao cost, **cannot starve legitimate mints**. That is the isolation this design buys, and it is the flood the Context describes. A caller that can produce well-formed, derivable requests can still consume `admission` budget with invocations that end in refusal, because at admission nothing distinguishes them from a real mint. That residual is bounded by **attacker cost** rather than by the limiter — every such attempt spends OpenBao work as well as an admission token — and the documentation must say so rather than imply a guarantee the design does not provide.

### How the limiter is constructed

The limiter is a field on `RegistrarVerbs`, supplied through `RegistrarVerbsConfig` beside the audit store and the wrap-TTL policy, and built from the four configuration values below plus a sink. It is a fixed dependency like every other field there: the verbs receive it and can neither choose its sizing nor decline to have one, and nothing in a request reaches it. Add no production wiring path for the verbs themselves — there is none today, and the change that adds one takes the limiter along with everything else on that struct.

### The limited-invocation event

Both check points emit one event per limited invocation, carrying the client identity, the verb and the bucket that had no token. Publish it through a **sink trait** the limiter is constructed with, plus a default in-process implementation that keeps a `u64` count **per bucket** — one for `predecision_refusal`, one for `admission`, both since daemon start — and nothing else. Two counters rather than one sum, because the sibling record issue publishes them separately and the two mean opposite things to an operator: a `predecision_refusal` count rising says someone is flooding malformed input while those callers still got their real answers, and an `admission` count rising says legitimate traffic is being held back and a bring-up may be stalling. A single total would have to be split again downstream, and it cannot be.

**The limiter holds exactly one sink, so say what happens when the sibling installs its own: it wraps this one rather than replacing it.** The sibling's sink is constructed around the default counting sink and delegates every event to it before coalescing, so the two per-bucket counters stay the **single** implementation in the tree — which matters because that sibling's health member publishes exactly those counters, and a sink that displaced them would leave the member reading a counter nobody increments. Keep the trait's shape composable enough for that: an implementation takes another implementation and forwards to it. This issue builds the trait, the default counting sink, and nothing that prevents the wrapping.

### Configuration and sizing

- Size the limiter from the **legitimate** fan-out, not from "mints are rare in steady state". There is effectively one registrar identity, so "per client identity" is close to global: throttling an attacker throttles the control plane in the same bucket, and an ordinary bring-up — an onboarding wave plus several modules per host — is a burst.
- Add four keys to the `[registrar]` table in `src/config.rs`, beside the audit-record and audit-store keys already there. **Every one is an unsigned integer; the configuration surface carries no floating-point value.** A refill *rate* would have to be fractional to express "one token per second or slower", which drags in negative zero, NaN, infinity and precision questions that all have to be validated away; a refill *interval* expresses the same rates as plain integers and has none of them.
  - `rate_limit_admission_burst` — `u32`, default `512`.
  - `rate_limit_admission_refill_interval_ms` — `u32` milliseconds per token, default `500` (two tokens per second).
  - `rate_limit_predecision_refusal_burst` — `u32`, default `32`.
  - `rate_limit_predecision_refusal_refill_interval_ms` — `u32` milliseconds per token, default `1000` (one token per second).
- The four keys join `RawRegistrarSettings` and `RegistrarSettings` the way the existing keys do, with `serde` defaults and a matching `Default` impl, and they **leave every existing key's value, default and derivation untouched**. Keep `deny_unknown_fields` on that table as it is: rejecting a misspelled key is deliberate there, and this issue does not relax it.
- **Validation rejects zero for all four keys** in `validate_registrar_settings`, naming the offending key as `registrar.<key>` in the same shape as the existing failures, with the reason documented: a zero burst throttles the first legitimate mint, and a zero refill interval is an unbounded token supply that disables the limiter silently. A negative or non-integer TOML value must surface as a load error naming the key rather than reaching the operator raw or panicking — and getting that takes a hand-written `Deserialize` on the value, because declaring the field `u32` does **not** produce it. The `config` crate's own conversion is permissive and silent: it turns `0.5` into `1`, `1.9` into `2`, `"512"` into `512` and even `true` into `1`, reporting nothing. Only a negative and a value past `u32::MAX` fail on their own. For a limiter's sizing that is a silently wrong bound rather than a typo the operator is told about — `rate_limit_admission_burst = 0.5` becomes a burst of `1`, which throttles the first legitimate mint of every bring-up, and nothing in the running daemon says why. So each of the four keys reads its value's *own* type first and refuses everything that is not a non-negative integer in range, which is what turns each of those cases into a load error naming the offending key. The same applies to any later unsigned key added to this table.
- **The sizing formula, to be stated in the shipped documentation and not only in code:** `rate_limit_admission_burst >= wave_hosts × modules_per_host`, where `wave_hosts` is the largest number of hosts an operator brings up at once and `modules_per_host` the largest number of components on one host. The shipped default assumes the reference deployment **64 hosts × 8 modules = 512 mints in one wave**, which is where `512` comes from. A wave larger than the burst still completes rather than being refused, taking `(mints − burst) × rate_limit_admission_refill_interval_ms / 1000` extra seconds. Legitimate pre-decision refusals are operator typos arriving one at a time, hence the much smaller refusal budget.
- The documentation must show **the deployment's own numbers** run through that formula, not only the shipped defaults, so an operator whose fleet is larger than the reference can see that they must raise the burst before their first bring-up rather than after it.
- Document the two buckets and what their isolation does and does not guarantee, the sizing arithmetic with the deployment's own numbers, and the four configuration keys with their defaults, in **both** `docs/en/operations.md` (under "Audit logging") and `docs/ko/operations.md` (under "감사 로깅"), with no `mkdocs.yml` nav change. The four keys additionally join the `[registrar]` table on `docs/en/configuration.md` and `docs/ko/configuration.md`, which is where every other key of that table is already listed; keep that table **one** block on those pages, since TOML refuses a table declared twice and a reader pastes what the page shows.

## Acceptance criteria

- [x] A single client identity flooding invocations that the verbs' pure checks refuse is limited: after the refusal burst is spent, further such invocations write neither an intent nor an outcome record, and each emits exactly one limited-invocation event.
- [x] That flood does **not** starve legitimate mints: a test drives the `predecision_refusal` bucket to empty and asserts a concurrent admitted mint is still performed and still recorded.
- [x] Limiting preserves the caller's answer exactly: for each pre-decision refusal the verbs can produce — including `ReservedServiceName` and an out-of-range `wrap_ttl`, which are retryable with no identifier today — a limited invocation returns the same `VerbError`, the same wire identifier (or the same absence of one) and the same `RefusalClass` as the identical unlimited invocation, and never `RegistrarBusy`.
- [x] Both check points sit before the intent write: a test asserts that a limited invocation on either bucket appends nothing to the audit record store, and that a limited `admission` invocation performs no OpenBao call.
- [x] The pure checks are evaluated once per invocation, not once for the limiter and again for the verb.
- [x] Draining the `admission` bucket makes the next invocation return the typed retryable throttle variant, distinguishable in `VerbError` from every permanent unavailability variant.
- [x] The tree builds with `map_refusal` still exhaustive and wildcard-free: the new `VerbError` variant gets its own arm in this issue, mapping it to the existing `EnrollError::RegistrarBusy { retry_after }` with `RefusalClass::Retryable`, and no new wire identifier, refusal class, fixture or protocol-reference change is introduced.
- [x] A refusal raised after the pure checks pass spends an `admission` token, and the bucket charged is never revised by the invocation's eventual outcome.
- [x] The retry payload is an unsigned whole-second duration of at least 1, never an absolute timestamp, and its value tracks the charged admission bucket's accrual against the configured refill interval.
- [x] The limiter is a token bucket driven by a monotonic clock: a bucket drained at a boundary does not immediately refill to a second full burst, and a wall-clock jump adds no tokens.
- [x] A bucket starts full and holds no persisted state: a freshly created limiter admits a full burst before limiting, and a limiter recreated from scratch starts full again. No entry is ever removed from the map while it is below its full burst.
- [x] Buckets are keyed on a value stable across connections: two sequential connections from one caller spend the **same** bucket rather than getting a fresh full budget, so the limiter cannot be reset by reconnecting.
- [x] The map is capped by a file-top constant counted in **bucket entries** (`256`, the reference `64` identities × four buckets): driving far more distinct client identities than the cap allows leaves the entry count at or below it, reclaims only entries at full burst, and never removes a drained entry.
- [x] The miss path is total: with the map at the cap and every entry drained, an invocation for an absent key is charged against the overflow bucket for its `(verb, bucket)` — including when the map holds no entry carrying that pair — and is limited on that bucket's tokens rather than admitted unlimited. The four overflow buckets are sized from the same keys as the map entries and are not counted against the cap.
- [x] The bucket enum has exactly two values and is exported for the sibling record issue; no bucket key, anywhere, carries the verbs' fine-grained refusal reason.
- [x] The four `rate_limit_*` keys are unsigned integers that load with the documented defaults; validation rejects a zero value for each with an error naming `registrar.<key>`, and a negative, fractional, string, boolean or past-`u32::MAX` value produces a load error naming the key rather than a silent coercion, a raw deserialization failure or a panic. A `[registrar]` table that sets only the pre-existing keys still loads with every one of their values unchanged.
- [x] The limiter is constructed with a limited-invocation sink; the shipped default sink counts limited invocations since daemon start as a `u64` **per bucket**, readable in process, and a test floods each bucket separately and asserts only that bucket's counter moves. The trait admits a wrapping implementation — one that takes another sink and forwards to it — so the sibling record issue can add coalescing without displacing these counters.
- [x] The documented sizing includes the `wave_hosts × modules_per_host` arithmetic and the deployment's own numbers, and a realistic bring-up burst (an onboarding wave plus several modules per host) passes without being limited.
- [x] The buckets, their isolation boundary, the sizing and the four keys are documented in **both** `docs/en/operations.md` and `docs/ko/operations.md`, on the existing pages, with no `mkdocs.yml` nav change, and the four keys also appear in the single `[registrar]` block on `docs/en/configuration.md` and `docs/ko/configuration.md`. A test loads that documented block verbatim and asserts it equals the shipped defaults, so `deny_unknown_fields` fails a rename the pages do not follow.
- [x] `cargo clippy --all-targets` is warning-free with `-D warnings`, and `cargo fmt -- --config group_imports=StdExternalCrate --check` passes.
- [x] `./scripts/check-docs.sh` passes. Both operations manuals change here, and that script is what CI runs for `docs/`: it verifies the vendored theme, builds with `mkdocs build --strict`, and asserts the built pages link the theme stylesheets.
- [ ] `scripts/preflight/run-all.sh` passes locally before the branch is pushed, including `scripts/preflight/ci/e2e-matrix.sh`. This issue changes daemon construction, the `[registrar]` configuration table and the verbs' entry path, all of which the E2E suite exercises, so the preflight is not skippable here.

## Constraints

- The limiter must not be sized by intuition about steady state. Legitimate bring-up bursts must pass; the sizing rationale must be written down where an operator can find it.
- Both check points must sit **before** the invocation's intent write. A limiter that runs after it cannot suppress the record it exists to suppress, and it would make the intent record the amplifier for the exhaustion the limiter is preventing.
- Moving the pure checks ahead of the intent write is the only reordering this issue makes. Do not restructure the verbs' arms, their per-id locks, their stage boundaries or their refusal taxonomy, do not move any OpenBao work, and do not change which refusal a given input produces, which `ProducingArm` it is stamped with, or which identifier and `RefusalClass` `map_refusal` gives it. Reclassifying an existing refusal — making the pre-decision set uniformly permanent, say — is a wire-contract change this issue does not make.
- Do not key a bucket on the verbs' fine-grained refusal reason, and do not interpret, re-derive or normalise `CallerIdentity` — it is opaque, and a bucket key that assumes a certificate identity breaks the pre-mTLS caller.
- The throttle must never be substituted for an outcome the daemon has already determined, and must not be expressed by overloading an existing variant with a boolean.
- Do not weaken the "unrecordable intent implies refused" rule the audit-record work establishes. Suppression here is a deliberate skip decided before the write is attempted; it is not a failed write, and it never makes a performed mint unrecorded.
- Keep the configuration surface free of floating-point values; express rates as integer millisecond intervals.
- Do not hold a bucket lock across an OpenBao call or a record write. The limiter's decision is taken and released at the check point; everything after it runs under the locks the verbs already define.

## Out of scope

- **Changing any RFC text.** The accepted RFC already specifies the bucket key and the admission-only throttle this issue implements.
- **The counted record and everything about it**: the coalescing window and its configuration key, the record variant, its encoding and fixture, the reader classification amendments, and the field bound on the identity it carries. This issue emits events through a sink; the sibling record issue turns them into records.
- **Everything on the wire boundary beyond the throttle variant and its `map_refusal` arm**: extending the golden fixtures beyond the existing `refusal-busy.json`, the protocol reference's local-half schema for the health container, and the `registrar_health.limiter` member that publishes the two counters. All the sibling record issue's.
- **Wiring the verbs into a production request handler.** No such wiring exists today; the limiter rides on `RegistrarVerbsConfig` and arrives with whatever change adds it.
- A per-`(component, host)` **ceiling** on how many instances a deployment may legitimately run — a different object from a rate limit, and deliberately deferred.
- Isolating legitimate mints from a flood of refusals raised after the pure checks pass. Those consume admission budget by construction, because nothing distinguishes them from a real mint before the intent write. This is the design's stated residual and it carries its own bound: an attacker driving it pays OpenBao work per attempt, so the path is limited by attacker cost rather than by this mechanism.
- Changing the caller identity's source or spelling, the endpoint's socket policy, or anything about mTLS.
- Defining the record encoding, the rotation bounds, the anomaly scan or `audit_record_dir` itself.
- The **reserved audit store** in every part: its configuration keys, layout, Compose override, capacity probe, headroom arithmetic and low-water alarm.

## Test plan

- [x] Pre-decision flood test: after the burst, neither record phase is written, one event is emitted per invocation, and the caller still receives the same refusal it would have received unlimited.
- [x] Answer-preservation test across the whole pre-decision refusal set, driven pairwise limited against unlimited and covering at least one permanent-with-identifier case, `ReservedServiceName`, and an out-of-range `wrap_ttl`: the `VerbError`, the mapped identifier and the `RefusalClass` are equal in both runs.
- [x] Bucket-isolation test: drain the `predecision_refusal` bucket and assert an admitted mint is still performed and recorded.
- [x] Ordering tests: a limited invocation on either bucket appends nothing to the record store, and a limited `admission` invocation is asserted through a seam that fails if OpenBao is reached.
- [x] Admission-throttle test asserting the typed retryable variant with a retry value of at least 1, no OpenBao call, no record, and one event carrying the `admission` bucket; plus a codec test that the variant encodes as `RegistrarBusy { retry_after }` with the retryable class.
- [x] Bucket-selection test: an invocation refused only after the pure checks pass — a derivation or binding refusal — spends an `admission` token and not a `predecision_refusal` one.
- [x] Token-bucket behaviour test: monotonic-clock refill, no double budget across a boundary, no tokens from a wall-clock jump, a fresh bucket starting full, and no removal of an entry below its burst. Drive time with `tokio::time` pause/advance rather than sleeping.
- [x] Bucket-key stability test: two sequential connections from one caller spend the same bucket.
- [x] Bounded-map tests: driving far more distinct identities than the cap keeps the entry count at or below it; a map at the cap holding a full entry reclaims it and admits the new key; a map at the cap with every entry drained charges the overflow bucket for the invocation's `(verb, bucket)`, covering the case where no map entry carries that pair; and no path removes an entry below its full burst.
- [x] Config tests: the four `rate_limit_*` keys load with their defaults; a zero value for each, and a negative, a fractional, a string and a past-`u32::MAX` value for one, each fail with an error naming `registrar.<key>`; a table setting only the pre-existing keys loads with their values unchanged.
- [x] Sink tests: the default sink's two per-bucket counts each track a flood driven at their own bucket and neither moves for the other's; the limiter functions with a sink that does nothing; and the limiter driven through a trivial forwarding wrapper around the default sink still advances both counters, which is the composability the sibling record issue's coalescing sink depends on.
- [x] Legitimate-burst test asserting a realistic bring-up wave is not limited.
- [x] Documentation build: `./scripts/check-docs.sh` succeeds with the new English and Korean sections in place, and `mkdocs.yml` is unchanged.
- [ ] Preflight: `scripts/preflight/run-all.sh` succeeds, with `scripts/preflight/ci/e2e-matrix.sh` covering the daemon and configuration paths this issue touches.
- [x] Tests use `tempfile::tempdir()` and never a fixed path, and reach no network.

## Dependencies

Nothing blocks this issue. The design decision it rests on is settled and the accepted RFC states it; the verbs, their two-phase audit records, the opaque caller identity, and the throttle's wire identifier with its retryable class and golden fixture are all in the tree. Part of the registrar verb rate-limiting umbrella.

The sibling issue that records limited invocations and reports the throttle consumes this issue's bucket enum, its sink trait and its typed throttle variant, so this issue lands first and must leave those three seams usable without being revisited.

The `map_refusal` arm for the throttle variant belongs to this issue and not to that sibling, because the variant and its arm cannot land apart — `map_refusal` is exhaustive and wildcard-free, and the sibling depends on this issue for the variant it would route, so no order exists in which they land separately and the tree builds. That is also what the mapping's in-tree comment says. Everything else on that boundary is the sibling's and is untouched here — the fixture work beyond the existing `refusal-busy.json`, the protocol reference's local-half health schema, and the `registrar_health.limiter` member.

## Pointers

- `src/registrar/verbs.rs` — `RegistrarVerbs::mint` and `::deregister` (the entry sequence the limiter is inserted into), `write_intent`, `pre_derivation`, `record_refusal`, `record_mint`, and `RegistrarVerbsConfig` (how the audit store is injected, and the seam the limiter follows)
- `src/registrar/verbs/outcome.rs` — `VerbError` (the variant to add), `CallerIdentity`, `ProducingArm`
- `src/registrar/verbs/wrap_ttl.rs` — `WrapTtlPolicy::grant`, the other pure check on the mint path
- `src/registrar/endpoint/protocol.rs` — `map_refusal` (exhaustive, catch-all-free), `EnrollError::RegistrarBusy`, `validate_refusal_class`
- `src/registrar/endpoint/serve.rs::caller_identity` — the one site the caller identity is rendered at
- `src/config.rs` — `RegistrarSettings` and `RawRegistrarSettings`, where the four keys join the `[registrar]` table, and where the keys' strict `Deserialize` lives
- `src/config/defaults.rs` — where the four keys' `serde` default functions join the existing ones
- `src/config/validation.rs::validate_registrar_settings` — where the zero-value rejections go
- `src/registrar/audit.rs` — `AuditVerb`, which the bucket key needs `Hash` on, and `bridge::refusal_outcome`, whose `None` set the throttle variant joins: it is produced before the intent write, on an invocation whose records are suppressed by construction, so it never reaches an outcome write and gets no reason-table row
- `src/daemon.rs::spawn_registrar_endpoint` and `src/registrar/endpoint/serve.rs::run` — how the endpoint reaches the daemon today
- `src/openbao.rs::verify_audit_file` — the mandatory audit device whose exhaustion this limiter exists to bound
- `docs/en/operations.md` ("Audit logging") and `docs/ko/operations.md` ("감사 로깅") — the sections the limiter documentation sits beside
- `docs/reference/registrar-wire-contract.md` §6.1 and §7 — the identifier spelling and the retryable/permanent split
- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §5.6 — the bucket key, the limited/throttled vocabulary and the admission-only throttle

AcoPiper:
## Summary

Both registrar verbs are now rate limited by token buckets keyed on `(client identity, verb, bucket)`, where `bucket` is a two-value enum — `predecision_refusal` and `admission`. Every mint and deregister writes an intent record at entry and an outcome record when it finishes, refusals included, so a compromised caller could grow the record store for free by flooding the refusals the verbs' pure checks settle before any OpenBao work. A full filesystem stops every mint, and because the OpenBao file audit device is mandatory and OpenBao fails requests it cannot audit, it can stop OpenBao serving on the one host that must not be restarted.

Both verbs now run their pure checks first and exactly once — `pre_derivation`, plus `WrapTtlPolicy::grant` for `mint` — charge the bucket that result selects, and only then write the intent line. Moving those checks ahead of that write is the only reordering: they are pure, so running them early costs nothing and changes no answer, and what they produced is carried forward rather than recomputed.

A limited invocation writes nothing per invocation anywhere, the daemon's journal included: no `tracing` line stands in for the suppressed records, because one line per flooded request is the unbounded write the buckets exist to remove, on the cheapest path a caller has. What a flood leaves behind is the sink's per-bucket counts, held in memory.

Limiting suppresses the record, never the answer. On the pre-decision path the caller keeps exactly the refusal its own input earned — the same `VerbError`, the same wire identifier or the same absence of one, the same `RefusalClass` — and `RegistrarBusy` is never substituted there. At admission the daemon has determined nothing, so the new `VerbError::Throttled { retry_after }` adds an answer rather than replacing one; its `map_refusal` arm lands with it, because that match is exhaustive and wildcard-free, and routes to the `EnrollError::RegistrarBusy` identifier, retryable class and golden fixture already in the tree. No new identifier, class, fixture or protocol-reference change.

The map is capped at `256` bucket entries — the reference `64` identities × four buckets each. At the cap a charge first reclaims every entry refilled to its full burst, which frees slots without handing anybody a token, since a full entry and an absent one both yield a full bucket on next use. When that frees nothing, four overflow buckets — one per `(verb, bucket)`, sized from the same keys, not counted against the cap — keep the miss path total. No entry is ever removed below its burst.

Four `rate_limit_*` keys join the `[registrar]` table with a hand-written `Deserialize` that reads each value's own type first, so a negative, fractional, string, boolean or past-`u32::MAX` value is a load error naming `registrar.<key>` rather than the `config` crate's silent coercion; validation rejects zero for each. Sizing follows `wave_hosts × modules_per_host`, documented with the reference deployment's own numbers in both operations manuals, alongside what the isolation does and does not guarantee; the four keys also join the single `[registrar]` block on both configuration pages.

The limiter is injected on `RegistrarVerbsConfig` the way the audit store already is, so whatever wires the verbs into a production handler picks it up for free.

Closes #786

## Test plan

- [x] Pre-decision flood: after the burst, neither record phase is written, one event is emitted per invocation, and the caller still receives the same refusal it would have received unlimited (`a_pre_decision_flood_is_limited_without_writing_a_record`)
- [x] Answer preservation across the whole pre-decision refusal set, driven pairwise limited against unlimited and covering a permanent-with-identifier case, `ReservedServiceName` and an out-of-range `wrap_ttl`: `VerbError`, mapped identifier and `RefusalClass` equal in both runs, never `RegistrarBusy` (`limiting_preserves_the_callers_answer_across_every_pre_decision_refusal`)
- [x] Bucket isolation: with `predecision_refusal` drained, an admitted mint is still performed and recorded (`a_pre_decision_flood_does_not_starve_an_admitted_mint`, `a_drained_refusal_bucket_does_not_starve_a_live_mint`)
- [x] Ordering: a limited invocation on either bucket appends nothing to the record store, and a limited `admission` invocation is asserted through a seam that fails if OpenBao is reached (`a_limited_invocation_on_either_bucket_appends_nothing`)
- [x] A flood on either bucket emits no `tracing` event at all: a counting subscriber installed for the flood sees zero events while the sink counts every limited invocation, and the test fails if a per-invocation log line comes back (`a_limited_invocation_emits_no_daemon_log_line`)
- [x] The pure checks are evaluated once per invocation, not once for the limiter and again for the verb (`each_invocation_charges_its_bucket_exactly_once`)
- [x] Admission throttle: the typed retryable variant with a retry value of at least 1, no OpenBao call, no record, one event carrying the `admission` bucket, and a codec test that it encodes as `RegistrarBusy { retry_after }` with the retryable class (`the_throttle_is_its_own_variant_rather_than_an_unavailability`, `the_retry_payload_tracks_the_configured_refill_interval`)
- [x] Bucket selection: a refusal raised only after the pure checks pass spends an `admission` token, not a `predecision_refusal` one (`a_derivation_refusal_spends_an_admission_token_and_the_next_mint_is_throttled`)
- [x] Token-bucket behaviour driven with `tokio::time` pause/advance rather than sleeping: monotonic refill, no double budget across a boundary, no tokens from a wall-clock jump, a fresh bucket starting full, a recreated limiter starting full again
- [x] Bucket-key stability: two sequential connections from one caller spend the same bucket
- [x] Bounded map: far more distinct identities than the cap leaves the entry count at or below it; a full entry at the cap is reclaimed and the new key admitted; a saturated drained map charges the overflow bucket for its `(verb, bucket)`, including where no entry carries that pair; no path removes an entry below its burst
- [x] Config: the four keys load with their defaults; zero for each, and a negative, fractional, string, boolean and past-`u32::MAX` value, each fail naming `registrar.<key>`; a table setting only the pre-existing keys loads with their values unchanged
- [x] Sink: the default sink's per-bucket counts each track a flood at their own bucket and neither moves for the other's; the limiter functions with a no-op sink; a forwarding wrapper still advances both counters
- [x] Legitimate burst: the shipped defaults absorb the reference bring-up wave without limiting
- [x] The documented `[registrar]` block loads verbatim and equals the shipped defaults, so `deny_unknown_fields` fails a rename the pages do not follow (`the_documented_registrar_table_loads_as_one_block`)
- [x] `cargo fmt -- --config group_imports=StdExternalCrate --check` and `cargo clippy --all-targets -- -D warnings` are clean, on macOS and in a Linux container (the endpoint module, and so the `map_refusal` arm, is `#[cfg(target_os = "linux")]` and macOS lints neither)
- [x] `cargo test` passes on macOS; `cargo test --lib registrar::endpoint` passes in a Linux container (124 tests)
- [x] `markdownlint-cli2` and `./scripts/check-docs.sh` pass, with `mkdocs.yml` unchanged; `cargo audit` reports only the pre-existing allowed `RUSTSEC-2025-0134` warning
- [x] `scripts/impl/run-registrar-verbs-e2e.sh` passes (17 tests, exit 0), as do the rest of `scripts/preflight/run-all.sh`'s host-runnable parts: `ci/check.sh` through all nine steps, the five `scripts/validate-*.sh`, `ci/deploy-no-build-smoke.sh`, and `extra/cli-scenarios.sh` end to end (`bootroot init`, `service add`, both `verify`s ok). Only `ci/e2e-matrix.sh` cannot run here — it hits documented macOS walls (no passwordless `sudo`, no `docker0` gateway) that reproduce identically at the merge base — and CI confirms it: all thirteen Docker E2E arms are green on this head, alongside Quality Check and Unit & CLI Smoke.


AcoPiper:
**[Reviewer Round 1]**

Request changes.

- **P1:** [src/registrar/verbs.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-786-review/src/registrar/verbs.rs:811) emits a `tracing::warn!` for every limited invocation. The limiter suppresses the unbounded audit records, but a caller can now flood the same path to produce unbounded daemon logs instead; the warning includes a fresh request ID, so each attempt is a distinct log line. The production agent initializes the default `tracing_subscriber::fmt` subscriber ([bootroot-agent.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-786-review/src/bin/bootroot-agent.rs:15)), and the registrar service leaves output at systemd’s journal default. This recreates a disk-pressure route precisely when the limiter should make the flood cheap and bounded. Remove the per-invocation warning (the required limited-invocation sink already records the event), or coalesce/rate-limit it independently.

Otherwise, the bucket selection, pre-intent ordering, overflow behavior, throttle mapping, configuration validation, documentation, and test coverage align well with #786. PR linkage and test-plan hygiene are also in order.

AcoPiper:

AcoPiper:
**[Author Round 1]**

All checks green on the new head (21/21). Here is the disposition of the one review item.

## P1 — per-invocation `tracing::warn!` on the limited path — **Fixed**

The finding is correct, and it defeats the issue's own premise: the limiter exists so a flood stops producing one unbounded write per request, and the warning turned each suppressed record pair back into a journal line — with a fresh request ID, so nothing collapses it. The production agent's default `tracing_subscriber::fmt` sends that to systemd's journal on the one host that must not be restarted. The issue is also explicit that emitting the sink event is this issue's whole share of the evidence, and that coalescing belongs to the sibling record issue, so removal (rather than a counter here) is what fits.

What changed:

- `src/registrar/verbs.rs:815` — `charge` no longer emits anything; the `Limited` arm just returns `retry_after`. Its doc comment now records *why* there is no log line, so it does not come back as an "observability" fix.
- `src/registrar/verbs/limiter.rs:220` — the `LimitedInvocationSink` trait doc now states the constraint the seam has to carry: an implementation must not turn one event into one unbounded write (log line, record, file append); count in memory or coalesce. That is the seam the sibling record work builds on, so the rule belongs on the trait rather than only at the call site.
- `src/registrar/verbs/tests.rs:2745` — new regression test `a_limited_invocation_emits_no_daemon_log_line`: both bursts are spent first, then a counting `tracing::Subscriber` is installed for a 16-invocation flood across both buckets and must see **zero** events, while the sink's per-bucket counts must total all 16. I verified it is a real guard by temporarily re-adding a bare `warn!` — it fails with `left: 16, right: 0`.
- `docs/en/operations.md` / `docs/ko/operations.md` — one paragraph each, beside the counters: nothing is written per limited invocation, the journal included, and why; so an operator does not read journal silence during a flood as a missing limiter. No `mkdocs.yml` change. No CHANGELOG entry: nothing serves registrar verbs in the last release, so no user could observe this.

Verification on this head: `cargo fmt --check` (with `group_imports=StdExternalCrate`), `cargo clippy --all-targets -- -D warnings`, full `cargo test` (0 failed), `scripts/preflight/ci/check.sh` exit 0 through all nine steps (`cargo audit` reporting only the pre-existing allowed `RUSTSEC-2025-0134`), and `scripts/impl/run-registrar-verbs-e2e.sh` exit 0 (17 tests). `ci/e2e-matrix.sh` is host-blocked on this Mac as before; CI covers it — all thirteen Docker E2E arms, Quality Check and Unit & CLI Smoke are green on `1df46d2`.

The PR description was updated: a summary paragraph stating that a limited invocation writes nothing per invocation anywhere, and a test-plan line for the new test. `Closes #786` is unchanged.

AcoPiper:
**[Reviewer Round 2]**

**[Reviewer Round 2]**

Approved.

P1 is resolved: limited charges now publish only to the in-memory sink; no per-invocation tracing event remains in the limiter path ([verbs.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-786-review/src/registrar/verbs.rs:816)). The new regression test exercises floods across both buckets and asserts zero tracing events while verifying sink counts ([tests.rs](/Users/aco/.agentcoop/worktrees/aicers/bootroot/issue-786-review/src/registrar/verbs/tests.rs:2750)).

The updated implementation continues to satisfy #786’s bucket selection, admission-only throttling, bounded entry map with overflow buckets, configuration validation, and test-plan requirements. No unresolved or new findings.

AcoPiper:

AcoPiper:
## Suggested squash commit

**Title**

```text
Bound both registrar verbs with two token buckets
```

**Body**

```text
Every mint and deregister writes an intent record at entry and an
outcome record when it finishes, refusals included, so a caller could
grow the record store for free by flooding the refusals the verbs' pure
checks settle before any OpenBao work. A full filesystem stops every
mint, and because the OpenBao file audit device is mandatory and
OpenBao fails requests it cannot audit, it can stop OpenBao serving on
the one host that must not be restarted.

Both verbs now run their pure checks first and exactly once, charge a
token bucket keyed on (client identity, verb, bucket) over the
two-value predecision_refusal/admission enum, and only then write the
intent line. Moving those checks ahead of that write is the only
reordering: they are pure, so running them early costs nothing and
changes no answer.

Limiting suppresses the record, never the answer. A pre-decision
refusal keeps exactly the refusal its own input earned, while at
admission the daemon has determined nothing, so the new
VerbError::Throttled adds an answer rather than replacing one and
routes to the existing RegistrarBusy identifier. A limited invocation
writes nothing per invocation anywhere, the daemon journal included:
one log line per flooded request is the unbounded write the buckets
exist to remove.

The map is capped at 256 entries, reclaiming only entries refilled to
their full burst and falling back to four per-(verb, bucket) overflow
buckets so the miss path is total. Four rate_limit_* keys join the
[registrar] table with a hand-written Deserialize, so a negative,
fractional, string, boolean or past-u32::MAX value is a load error
naming the key rather than a silent coercion.

Closes #786
```

PR #920, opened 2026-08-26
Bound the OpenBao audit device's growth by rotating it in place (#917)

AcoPiper:
## Summary

`bootroot-agent` now rotates OpenBao's file audit device in place, on a host whose registrar endpoint is enabled. The device was the one audit artifact on the bootroot host with no bound on its size — OpenBao's file device does not rotate itself, and OpenBao fails requests it cannot audit, so an audit filesystem that fills stops OpenBao serving.

A dedicated task spawned from `run_daemon` runs a pass every 60 seconds (a file-top constant, injectable for tests, with its `JoinHandle` pushed onto the same `handles` vector the other daemon tasks use — deliberately not the per-profile `check_interval`, which is per-profile, jittered and operator-configurable). Each pass copies the active log's stable prefix — trimmed back to its last record boundary — into a staging file, fsyncs it, links it into place as `audit-<YYYYMMDDTHHMMSSZ>-<NNNNNN>.log` beside the active log, confirms that name reached the copy it just staged and unlinks the staging name, fsyncs the directory, and only then truncates `audit.log` in place and fsyncs it. OpenBao is never restarted, resealed or signalled; the active log is never unlinked, renamed or replaced; `openbao/openbao.hcl`'s audit stanza is untouched and the device's registration in `sys/audit` is unchanged.

Every pass then evaluates the retained set against a hard byte budget and trims oldest-first, whether or not a rotation happened. Three new `[registrar]` keys carry the bound — `openbao_audit_max_file_bytes` (default `67108864`, floor 1 MiB), `openbao_audit_max_retained_files` (default `7`) and `openbao_audit_min_retain_days` (default `90`). `S × N` is the hard ceiling on the generations bootroot owns; the retention age is soft and yields to it. Validation rejects a configuration whose `S × (N + 1)` does not fit in `u64`, so the runtime products can be total `saturating_mul` with no error path.

The device's own total is not bounded, and nothing here calls any figure other than `S × N` a bound: bootroot is not the writer, and the only way to stop OpenBao writing is to stop it serving. Instead a pass that leaves either obligation unmet increments a counter, `warn`s per increment, and `error`s at five consecutive failures and each further multiple, naming the run length, which obligation is unmet, the active log's size and the retained set's total. A rotation failure never fails an OpenBao request, stops the daemon or takes the endpoint down; the next tick simply retries. A pass owing nothing — a quiet host, an unprovisioned host, an active log below `S` — neither increments nor escalates. Only an *absent* device directory is that unprovisioned host: one present but unusable leaves the retained set unassertable and counts, and an `audit.log` that is not the regular file OpenBao writes is refused before its size is read, so a planted link cannot pass as a quiet host by pointing somewhere small. The directory itself is established with `symlink_metadata` and a pass acts only through a real one — a link there is the first component of every path a pass would build, which no `O_NOFOLLOW` open can refuse, so the pass stops at the path rather than copying, publishing, truncating and deleting inside a directory bootroot never provisioned; a link resolving to nothing is refused on the same terms rather than read as the absent directory.

That size check settles a name, not a file, and the device directory is writable by the container's audit user — so the copy and the truncate each re-establish it on the descriptor they actually hold. Both open `O_NONBLOCK` as well as `O_NOFOLLOW` and then `fstat`: `O_NOFOLLOW` refuses a symbolic link and permits a FIFO, and a blocking open of one waits for the other end, inside `spawn_blocking`, which would leave the rotation task parked forever — no retry, no counter, no escalation, no notice of shutdown, and so the one unmet obligation this design cannot report. Refused, a replacement is an ordinary failed pass.

The publish is settled the same way. `rename` resolves the staging name at the instant of the call, and that name sits in the same writable directory — so a file left there between the copy's flush and the publish would be published in the copy's place, and the truncate after it would destroy the records the pass believed it had captured. A pass therefore `link`s the name into place and establishes, by device and inode number against the descriptor it still holds open, that the published name reached the copy it staged. `link` also refuses a name already present rather than replacing it, so a collision can never overwrite a retained generation, and that refusal never unlinks the file it collided with. Both are ordinary failed passes, before the commit point, that leave the active log whole.

Both facts are established once more immediately before the truncate, which is the commit point: past it, getting either wrong costs records rather than a pass. The generation's name is re-read against the staged inode, because the directory flush between the publish and the truncate is a disk round trip and the name can stop leading to the copy inside it — and by then the staging name is gone, so the copy is reachable only through the descriptor the pass is about to drop. The truncate's own descriptor is compared against the one the copy was read through, because a *second regular file* planted at `audit.log` answers every question its type can be asked exactly as the first does: emptying it would rotate a file the pass never read while OpenBao went on appending to the original through the descriptor it still holds — the log unbounded, the pass reporting a completed rotation, and the counter never moving. Both descriptors are held open for the rest of the pass, so neither identity can be freed and its number recycled onto a replacement. Neither check closes its window outright, since a check and the operation it guards cannot be one operation; both put every interval the pass itself introduces behind them.

Recovery is scoped to before the commit point: a failed truncate or a failed directory flush unlinks the staging copy, but nothing after a successful truncate is ever rolled back, because past that point the generation is the only copy of those records. A partial trim is likewise never rolled back — deletions are oldest-first, so each moves the set toward the bound and the next pass carries on. Those deletions count as durable only once the directory holding them is flushed, so a failed flush is carried into the pass's outcome rather than warned about and dropped: the bound stays unmet, and every later pass re-attempts the flush, including one with nothing to delete. That debt does not survive the process holding it and the deletions it covers do, so a rotation task starts owing a directory flush rather than assuming one happened: one `fsync` on its first pass, whether or not it has anything to delete, in place of a restarted daemon reporting a ceiling that only the page cache establishes.

Generations are created at mode `0600` with the device directory's uid and gid, both before the rename, and live in the `<audit_store_dir>/openbao` the install path already provisions — reached through the existing `openbao_dir` accessor, with no new directory created. Names carry an always-present zero-padded six-digit sequence set to one past the greatest present under a stamp floored at the newest existing generation's, so a clock stepped backwards cannot reorder the set and a trimmed gap is never reused.

With `[registrar_endpoint] enabled` false or absent, no task is spawned, the three keys have no effect and the device stays on the `openbao-audit` named volume.

Both operations pages lose the bullet telling operators to rotate the device with an external tool and a `SIGHUP`; the mechanism, its loss window, its crash residual, the retained-set layout, the budget arithmetic and the failure escalation are documented in `docs/en/` and `docs/ko/`, with no `mkdocs.yml` nav change.

Closes #917

## Test plan

**Footprint**

- [x] `hard_ceiling_holds_after_every_completed_evaluation` — many rotations, including oversized active logs
- [x] `conditional_envelope_under_the_stated_model` — names its model, truncate suspended for the in-pass peak
- [x] `a_long_run_of_failed_passes_leaves_the_total_unbounded`
- [x] `a_rotation_that_could_not_trim_counts_as_unmet` — publish, empty, then force the trim to fail
- [x] `a_lowered_bound_is_repaired_without_a_rotation`
- [x] `an_active_log_past_the_bound_is_rotated`

**Rotation mechanics**

- [x] `a_trailing_partial_record_is_destroyed_rather_than_carried_in`
- [x] `the_publish_is_ordered_and_each_pre_commit_failure_unlinks_the_copy`
- [x] `a_failure_past_the_commit_point_keeps_the_generation_and_still_trims`
- [x] `a_partial_trim_is_not_rolled_back`
- [x] `a_prefix_with_no_record_boundary_skips_the_rotation_but_still_trims`
- [x] `the_trim_enforces_bytes_and_not_only_the_count`
- [x] `a_generation_larger_than_the_whole_budget_is_dropped_whole`
- [x] `generations_are_named_and_ordered_by_an_always_present_sequence`
- [x] `a_clock_stepped_backwards_cannot_reorder_the_set`
- [x] `staging_copies_and_generations_carry_the_directory_owner_at_mode_0600`
- [x] `an_active_log_replaced_by_a_symlink_is_refused_rather_than_followed` — a target at `S`, one below it, and a link pointing at nothing
- [x] `an_active_log_replaced_after_the_size_check_is_refused_at_each_open` — a FIFO planted in each window, on a thread with a deadline so a parked open fails rather than hangs
- [x] `a_staging_path_replaced_before_the_publish_is_refused` — the staging name replaced in the window before the publish, active log intact
- [x] `a_publish_never_overwrites_an_existing_generation` — the chosen name occupied after it was chosen; the colliding file survives untouched
- [x] `a_generation_replaced_after_the_publish_is_refused_before_the_truncate` — the verified name replaced past the directory flush; the active log keeps every record and the impostor is left alone
- [x] `an_active_log_swapped_for_another_regular_file_is_refused_at_the_truncate` — the original held open to model OpenBao's writer; the replacement is never emptied
- [x] `a_pass_creates_no_directory_and_publishes_beside_the_active_log`

**Failure handling**

- [x] `the_task_retries_a_failed_pass_on_the_next_tick` — through an injected interval
- [x] `the_budget_arithmetic_is_total_at_runtime_and_checked_at_validation`
- [x] `consecutive_failures_warn_per_pass_and_escalate_at_each_multiple`
- [x] `a_pass_with_nothing_owed_never_increments_or_escalates`
- [x] `a_missing_active_log_is_a_no_op_and_an_absent_directory_evaluates_nothing`
- [x] `a_device_directory_that_is_not_a_directory_is_a_failed_pass` — only `NotFound` is the unprovisioned host
- [x] `a_symlinked_device_directory_is_refused_rather_than_followed` — one resolving to a populated directory, one resolving to nothing
- [x] `a_trim_whose_flush_failed_stays_unmet_until_a_later_flush_succeeds` — the retry lands on a pass that deleted nothing
- [x] `the_trim_flush_debt_is_owed_again_by_a_new_rotation_instance` — the debt is re-owed across a restart, not inherited from memory

**Configuration and gating**

- [x] `the_registrar_table_reads_the_openbao_audit_rotation_keys` — the three defaults, each rejection naming its key, pre-existing `[registrar]` keys kept
- [x] `validation_rejects_an_openbao_audit_budget_that_cannot_be_computed`
- [x] `the_ceiling_wins_over_the_retention_age`
- [x] `the_openbao_audit_rotation_task_is_spawned_only_behind_the_endpoint_gate`

**Against a running deployment** (`scripts/impl/run-registrar-internal-init-e2e.sh`)

- [x] E2E no-restart — `StartedAt`, `RestartCount`, `Pid` unchanged and `sys/seal-status` reporting `sealed=false` / `initialized=true` throughout
- [x] E2E append semantics — the new active log begins at offset 0 with a well-formed record and no leading NUL bytes
- [x] E2E device registration — `verify_audit_file` still succeeds and `sys/audit` is unchanged
- [x] E2E shared assertion — the unmodified `assert_openbao_audit_log` passes on the rotated deployment, with a fresh AppRole login and KV read driven after the rotation

**Documentation and hygiene**

- [x] Neither operations page still instructs the operator to rotate the device with an external tool or a signal
- [x] `./scripts/check-docs.sh` passes
- [ ] `scripts/preflight/ci/e2e-matrix.sh` passes
- [x] `cargo clippy --all-targets -- -D warnings` is clean and `cargo fmt -- --check --config group_imports=StdExternalCrate` passes

### Verified locally

**Unit tests — all 36 named above pass, on macOS and on Linux.** `cargo test --lib registrar::openbao_audit` gives 33 passed / 1 ignored, and the three config/daemon tests pass alongside them; `cargo test --locked --lib --no-fail-fast` is 914 passed / 0 failed. The same run was repeated inside a `rust:latest` container so the Linux target is covered too. Each of the two new commit-point tests was confirmed to fail with its check removed and to pass with it in place. `cargo fmt -- --check --config group_imports=StdExternalCrate` and `cargo clippy --locked --all-targets -- -D warnings` are clean on both platforms, and `./scripts/check-docs.sh` exits 0 through all its steps.

**The four E2E items were verified against a live OpenBao, outside the harness.** `scripts/impl/run-registrar-internal-init-e2e.sh` still cannot run on this machine — it aborts at `[startup] FAIL passwordless sudo is required` — so its assertions were reproduced directly instead, on a rig that has the property the mechanism depends on: OpenBao 2.5.5 running `openbao/openbao.hcl` unmodified (the same declarative `audit { type = "file" }` stanza, at the same `/openbao/audit/audit.log`), with the device's directory on a **named volume** shared with a Linux container that ran the shipped rotation code through the ignored `a_live_openbao_audit_device_rotates_in_place` test — one kernel, one filesystem, which is what the deployment target is and what a macOS host bind mount is not.

- **No restart** — `StartedAt`, `RestartCount` and `Pid` were byte-identical before and after (`2026-08-26T15:16:10.3086903Z 0 2185227`), across two consecutive rotations, and `sys/seal-status` reported `sealed=false` / `initialized=true` throughout.
- **Append semantics** — after the rotation the active log's first byte is `0x7B` (`{`), its first line parses as a JSON audit record (`type=request`), and the file contains no NUL byte anywhere (13011 bytes in, 13011 out through `tr -d '\0'`).
- **Device registration** — `verify_audit_file` succeeded before and after, and `sys/audit` was unchanged: `{"file/": {"options": {"file_path": "/openbao/audit/audit.log"}, "path": "file/", "type": "file", ...}}` on both sides.
- **Shared assertion** — an AppRole login and a `secret/data/...` read were driven *after* the rotation, then the unmodified `assert_openbao_audit_log` from `scripts/impl/lib/audit-log.sh` was sourced and run against the rotated deployment: it passed.

The published generations came out as `audit-20260826T151702Z-000000.log` and `audit-20260826T151916Z-000000.log`, mode `0600`, uid/gid `100:1000` — the device directory's owner, established by the rotation running as root against a directory it does not own, which is the fixed-owner publish path.

**`scripts/preflight/ci/e2e-matrix.sh` is the one item left unchecked**, and it is owed to CI. It gets three arms in: `run-local-lifecycle.sh` passes all 24 phases through `assert-openbao-audit-log` and `cleanup`, and `run-remote-lifecycle.sh` likewise — both endpoint-disabled hosts, so both confirm this change leaves the unprovisioned path alone. It then dies in `run-rotation-recovery.sh` at `scenario-a` with `Leaf certificate at certs/daemon-c1.crt does not chain to CA bundle ...`, a pre-existing failure on this host that reproduces byte-identically at the merge base and has nothing to do with this branch. The `registrar-internal-init` arm sits past that wall and needs passwordless sudo besides.


AcoPiper:

AcoPiper:
**[Author Round 4]**

Round 4 had one review item. Pushed as `881f12f`.

## [P1] Root token exposed through process arguments — **Fixed**

The finding was valid on both lines, and the reach was wider than either suggests: `-H "X-Vault-Token: …"` sat in `openbao_api`, so it published the token on *every* authenticated call in the arm, not once; and `sudo env NAME=<token>` published it for the whole of the rotation test. Both contradicted the handling immediately above them, where `init`'s summary is written root-only and deleted the moment it has been read so the token isn't left readable.

The token no longer passes through a shell variable at all. `run_init` stages it out of the summary on a pipe — `jq` runs unelevated on `cat`'s output, and the elevated shell that writes each file reads the value from stdin, so only paths are ever arguments — into two root-owned `0600` files under `RUN_ROOT`:

- `openbao-root-token` — the token alone (`printf %s "$(cat)"` drops `jq -r`'s newline; `umask 077` sets the mode at creation, not after).
- `openbao-curl.conf` — `header = "X-Vault-Token: <token>"`.

Then each consumer names a path: `openbao_api` uses `curl -K "$OPENBAO_CURL_CONFIG"`, and the ignored rotation test takes `BOOTROOT_OPENBAO_AUDIT_E2E_TOKEN_FILE` instead of `…_TOKEN`, read and trimmed by a new `required_secret_file`. The empty-token guard moved with it: `[ -n "$OPENBAO_ROOT_TOKEN" ]` became `sudo -n test -s`, same failure message. Teardown is unchanged — both files are under `RUN_ROOT`.

**One thing beyond the two lines named:** the post-rotation AppRole login was passing a freshly minted `secret_id` as `curl -d '{…}'` — same argv channel, same visibility. `openbao_api_unauthed` now reads its body from stdin via `-d @-`, written by the `printf` builtin, which forks no process at all. `--stepca-password` / `--http-hmac` on the `init` line are the same shape but are `main`'s and untouched by this branch, so I left them.

## Verification

The harness itself can't run here — it aborts at `[startup] FAIL passwordless sudo is required`, so I verified each mechanism directly, unelevated, since `sudo` only adds elevation:

- `curl -K` against a local listener really sends `X-Vault-Token`, and `-d @-` sends the stdin body with the same `Content-Length`/`Content-Type` the old `-d "$data"` produced — nothing about the request changes for OpenBao.
- the staging pipeline produces `-rw-------`, exact token bytes with no trailing newline, and a zero-byte file that `test -s` rejects when `root_token` is absent.
- `bash -n` and `shellcheck -x` clean (only pre-existing SC1091 source-follow infos); `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check` and `cargo test --lib --no-run` clean.

The `registrar-internal-init` CI arm exercises the elevated path end to end. Checks are queued now (CodeQL pending, the rest starting).

No docs or CHANGELOG entry is owed — this is harness-internal credential plumbing with no user-visible behaviour, and `docs/{en,ko}/e2e-ci.md` don't describe how the arm carries the token. The PR description doesn't name these environment variables, so it stays accurate as written.

AcoPiper:

AcoPiper:
**[Author Round 6]**

CI is still queued (runner backlog); I'm polling it in the background. Here's the disposition of Round 6's single item.

## [P1] Publishing from the mutable staging pathname — **Fixed**

The finding is correct, and it is the one failure this module's ordering exists to prevent. `rename` resolves `.audit-rotate.staging` at the instant of the call. That name lives in a directory the container's audit user can write, so between the copy's flush and the publish that user can unlink it and leave a file of their own there. `rename` would publish the replacement, the directory flush would make it durable, and the truncate at `src/registrar/openbao_audit.rs:910` would then destroy the records the pass believed it had captured. Every other failure in this module leaves the active log whole; this one loses it.

**The publish is now a `link` and an identity check** (`publish`, `src/registrar/openbao_audit.rs:1113`), and each half does distinct work:

- `std::fs::hard_link` refuses an existing target with `EEXIST` where `rename` replaces it silently — the second half of the review item. A name collision means the directory changed under the sequence the pass derived, and overwriting is exactly how an already-retained generation's records would be destroyed.
- The published name's `(device, inode)` is compared against the staged descriptor's. That is what catches the replacement — after the link, but still before the commit point, so the active log is intact when the pass fails.

`stage_prefix` now returns the staging `File` rather than the staged length, and the caller holds it open across the publish. That is load-bearing: an open descriptor keeps the inode from being freed, so its number cannot be recycled onto the impostor and pass the comparison. It also settles the one thing `std::fs::hard_link` leaves to the platform — whether a symbolic link at the source is followed — since either answer yields an inode that is not the staged one.

One thing the review did not name, which the fix had to get right: **unlinking the target belongs inside `publish`, not in the caller's recovery.** A caller that ran `recover(&target)` on any publish error would, in the `EEXIST` case, delete the very generation the refusal just protected. Only the code that made the link knows whether there is one to undo, so `publish` owns that and the caller recovers the staging name alone.

### Regression tests

- `a_staging_path_replaced_before_the_publish_is_refused` — uses the existing post-stage hook, exactly as suggested: unlinks the staged copy and leaves an empty regular file at the name. Asserts the pass fails, the active log is byte-for-byte what it was, no generation was published, and the staging name is cleared.
- `a_publish_never_overwrites_an_existing_generation` — occupies the name the pass has already chosen, after it chose it. Asserts the colliding file survives with its original contents and the active log is untouched.

Verified both actually catch the regression: with `publish` reverted to `std::fs::rename`, both fail and the rest of the suite still passes.

### Verification

`cargo test --locked --lib --no-fail-fast` — 912 passed / 0 failed. `cargo clippy --locked --all-targets -- -D warnings`, `cargo fmt --check --config group_imports=StdExternalCrate` and `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --document-private-items` clean. The module tests were re-run inside `rust:latest` (31 passed / 1 ignored) because `link` semantics are platform-specific, and I confirmed there that with `protected_hardlinks=1` root can hard-link a mode-`0600` file owned by another uid — the deployment's publish path, where the staging copy has just been chowned to the device directory's owner. `./scripts/check-docs.sh` exits 0; markdownlint clean.

### Docs

`docs/en/operations.md` and `docs/ko/operations.md`: the mechanism paragraph now says the copy is linked into place and the name confirmed rather than renamed, with a new paragraph on why; the envelope's doubling term is re-explained (a hard link cannot cross a filesystem, so the staging copy still must live in the device's own directory); the crash-residual section gains the new residual — a crash inside the publish can leave `.audit-rotate.staging` as a second name for the new generation, costing no space, counted by no bound or trim, and removed by the next pass. The module docs carry the same in the step list and the `publish` rustdoc. No CHANGELOG entry: the feature is unreleased and announced once under `### Added`, so no user of the last release can observe this hardening.

Pushed as `c988901`; PR description updated (summary, two new test-plan items, corrected counts, `Closes #917` intact).

AcoPiper:
**[Reviewer Round 7]**

Request changes.

Round 6’s staging-path replacement test is fixed, but its underlying P1 remains open:

- [P1] The published generation can still be replaced after `publish` verifies it and before the commit point. [`publish`](src/registrar/openbao_audit.rs:1165) checks `target` immediately after `hard_link`, then [`rotate`](src/registrar/openbao_audit.rs:926) syncs the directory and later truncates the active log at [`:946`](src/registrar/openbao_audit.rs:946). The container audit user can unlink the verified target and create an empty regular file during that interval. The staging name has already been removed, so the original staged inode is held only by `staged`; after the successful pass returns it is gone, while the real active audit log has been truncated. The new regression test covers replacement before the link, not this remaining post-verification window. Add a test through the existing `on_publish` hook that replaces `target` and asserts the active log is not truncated, then make the commit path re-establish that the published name still identifies the staged inode before treating the generation as durable.

New finding:

- [P1] A regular-file swap of `audit.log` between staging and truncation is reported as a successful rotation of the wrong inode. [`stage_prefix`](src/registrar/openbao_audit.rs:1044) copies one opened active-log descriptor, but [`truncate_active`](src/registrar/openbao_audit.rs:1213) opens the pathname again and only verifies that its new descriptor is regular. Replacing `audit.log` with another regular file in the existing `before_truncate` window therefore passes `open_active`, truncates the replacement, and returns `Completed`; OpenBao can continue writing through its still-open original, now-unlinked inode. This defeats rotation without incrementing the failure counter. Preserve and compare the active inode identity across copy and truncation (or truncate the verified source descriptor), and add a regression test that holds the original file open to model OpenBao’s writer.

AcoPiper:
**[Author Round 7]**

Round 7's two items are both fixed, pushed as `a96204a`, and the disposition is posted to the PR. CI is running (I'm watching it in the background).

## [P1] Published generation replaceable after `publish` verifies it — **Fixed**

Valid, and the window is a disk round trip wide: `sync_dir` runs between the publish's identity check and the truncate, and by then the staging name is gone, so the copy is reachable only through the descriptor `rotate` is about to drop.

`rotate` now calls `Self::confirm_published(&copy.staged, &target)` as the last thing before `truncate_active` (`src/registrar/openbao_audit.rs:966`), after the `before_truncate` hook. It re-stats `target` and compares `(device, inode)` against the still-open staged descriptor, so the staged inode cannot have been freed and its number recycled onto an impostor. A mismatch is an ordinary failed pass before the commit point.

One deliberate deviation: a name the check does not recognise is **not** unlinked — this pass no longer holds a link to it, so unlinking would delete a file that is not its own, the same reasoning `publish` already applies to an `EEXIST` collision.

Test `a_generation_replaced_after_the_publish_is_refused_before_the_truncate` drives that window through `on_publish`.

## [P1] Regular-file swap of `audit.log` between staging and truncation — **Fixed**

Valid, and the worse of the two: `open_active` asks what the descriptor *is*, and a second regular file answers as the first does, so the pass emptied a file it never read while OpenBao kept appending to the original — reported as `Completed`, counter unmoved.

`stage_prefix` now returns `StagedCopy { staged, source }`, holding the active log's read descriptor open for the rest of the pass; `truncate_active(&copy.source)` compares what it opened against that identity and refuses with `InvalidInput` before `set_len`. I took the compare-and-refuse half of the suggestion rather than truncating the source descriptor directly: refusing reports the tampering through the failure counter, where truncating the old inode would leave the pathname growing unbounded and say nothing.

Test `an_active_log_swapped_for_another_regular_file_is_refused_at_the_truncate` holds the original open to model OpenBao's writer and asserts the replacement is never emptied.

Both new tests were confirmed to fail with their check removed and pass with it in place; the other 31 are unaffected either way. Docs: a new paragraph in `docs/en/operations.md` and `docs/ko/operations.md` covering both checks and stating plainly that neither closes its window outright — what they close is every interval the pass itself introduces. No CHANGELOG entry (unreleased feature, already announced once under `### Added`). Clippy, fmt, `cargo test --locked --lib` (914 passed), `check-docs.sh` and markdownlint are clean locally; the PR description was updated for the new mechanism and test-plan items.

AcoPiper:
Rebased onto `main` at 34f8dc9 ("Self-issue the registrar surface's certificates").

One conflict, in `CHANGELOG.md`: both that commit and this branch appended a new bullet at the same position under `## [Unreleased]` → `### Added`. Resolved by keeping both entries, main's certificate self-issuance bullet first and this branch's audit-device rotation bullet after it — neither was reworded or dropped.

Everything else merged automatically. The overlapping files (`src/config.rs`, `src/config/validation.rs`, `src/daemon.rs`, `src/registrar.rs`, `agent.toml.example`, and the four `docs/` pages) took additions from both sides in disjoint regions; the rebased commit's diffstat is identical to its pre-rebase form (14 files, 4783 insertions, 9 deletions), so nothing was lost in the auto-merge.

Verified locally after the rebase: `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `cargo doc` with `-D warnings`, `markdownlint-cli2`, and `./scripts/check-docs.sh` all clean, and the full `cargo test --no-fail-fast` suite green across 39 targets.

Issue #767, opened 2026-08-04
[Umbrella] Self-issue the registrar surface's two certificates from the daemon


sehkone:
# Self-issue the registrar surface's two certificates from the daemon

## Summary

An endpoint-enabled bootroot host needs two certificates that no path in this repository can currently produce. The registrar authenticates to the host-local endpoint with a **client certificate** and holds no OpenBao credential of its own; the endpoint presents a **server certificate** the registrar pins. Both names live in the reserved `bootroot-` namespace and are unmintable through `service add`, so today both are supplied entirely out of band and an endpoint-enabled host cannot obtain either for itself.

This effort makes the daemon mint both leaves itself, under the bootroot-internal privileged credential, and writes them where the endpoint and the registrar read them — and then reconciles the project's accepted design document with the recovery timing that choice implies.

## Children
- [x] #913
- [x] #914
- [x] #921
- [x] #922
- [x] #930

The two are strictly ordered, and the order is not a convenience. The implementation child changes **when** an already-expired registrar leaf is repaired: at daemon start, before the endpoint's TLS material loads, rather than at the daemon's first renewal tick. The documentation child amends the accepted RFC to describe that. Amending the RFC first would leave it describing a daemon that does not exist, so the RFC child runs after the implementation child rather than beside it.

The critical path is entirely the implementation child. It touches configuration deserialization, the daemon's startup ordering, and an issuance path that reads its ACME inputs from OpenBao, and its test surface is the eight unusable-material states it must answer by issuing rather than by refusing. The documentation child is short, touches one RFC section, and carries no code, no configuration and no test — it is schedule risk only in the sense that skipping it leaves the accepted design and the shipped daemon stating different recovery timings, with nothing recording which is current.

## Shared background

**The two reserved names, and why they are unmintable.** `<instance>.bootroot-registrar.<host>.<domain>` is the registrar's client leaf and `<instance>.bootroot-registrar-endpoint.<host>.<domain>` is the endpoint's server leaf. Both fall inside the reserved `bootroot-` prefix, and the reserved-prefix guard makes both unproducible through `service add`. That is by construction rather than omission: it is what keeps an operator-authored service identity from ever occupying either name. The consequence is that the only producer these names can have is bootroot's own daemon.

[...]

PR #946, opened 2026-08-28
Reconcile RFC 0001 lapse recovery timing (#914)

### LINKED ISSUE #914: Reconcile RFC 0001 section 4's lapse-recovery timing with start-time certificate repair
# Reconcile RFC 0001 section 4's lapse-recovery timing with start-time certificate repair

## Context

`docs/rfcs/0001-registrar-role-and-non-self-propagation.md` is the accepted design document for bootroot's registrar role. Its section 4, "The guarantee and the mechanism", fixes who repairs a lapsed registrar credential and when. Two passages in that section state a recovery timing the daemon no longer follows.

The daemon repairs an already-expired registrar leaf **at start**, before the registrar endpoint's TLS material is loaded, rather than at its first renewal tick. That was a deliberate choice. Deferring to the first renewal tick would leave the endpoint enabled and unable to serve for the length of a renewal lead time, on precisely the restart an operator performed in order to recover, and during that window every enrollment in the deployment fails while the machine holding the fix idles. Refusing to start instead would brick the daemon in a state whose only repair requires the daemon to be running — the same argument section 4 itself makes for not routing recovery through re-provisioning the host.

Keeping the two leaves valid over time — the renewal schedule, the reload contract, the health field and the lapse surface — is separate work that has not landed when this amendment does. So the amendment records the repair that exists, and states nothing about when or how a leaf that expires *while* bootrootd is running is repaired.

What section 4 decides is *who* repairs a lapse — bootrootd, never bootler, never re-provisioning the host — and that is unchanged. What moved is *when within bootrootd's own lifecycle* the repair happens. Leaving the section as written would have the accepted design document and the shipped daemon state different recovery timings, with nothing recording which is current, so a later reader consulting the RFC would be misled about how a lapse is recovered.

This is a documentation-only change. It amends one RFC section and touches no code, no configuration and no test.

## Scope

Amend `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` section 4 so that it describes the behaviour the daemon implements. Two passages are affected, both inside the section 4 bullet that begins "**Lapse recovery is bootrootd's, NOT bootler's — and the renewal must be handed off to the registrar.**"

### Passage one — the recovery-timing sentence

It currently reads:

> Because the daemon issues and renews the certificate under its own credential, it needs nothing from the registrar to do so: a lapsed registrar is repaired by **bootrootd running and reaching its next renewal tick**, not by re-provisioning the host.

Amend it so that it states that a leaf **already expired when bootrootd starts** is repaired at that start, before the registrar endpoint loads its TLS material. Scope the claim to the state found at start, and stop there: the amended sentence must **not** assert that a leaf expiring while bootrootd is already running is repaired by the renewal mechanism, or by anything else. Renewal of these two leaves is later work that has not landed when this amendment does, and an RFC sentence asserting a repair nothing performs yet is the same silent disagreement between document and daemon that this issue exists to end. The RFC's existing statements about renewal stand exactly as written and are neither extended nor qualified — the opening clause that the daemon issues and renews the certificate under its own credential, and the reload-contract paragraph that follows. Keep the sentence's existing point intact and explicit: recovery is bootrootd's, and never re-provisioning the host. Leave the surrounding statements as they are, in particular that bootler provisions the **initial** certificate and nothing more, and the parenthetical correcting the earlier pre-self-issue draft.

### Passage two — the ordering trap

It currently reads:

> Ordering trap worth stating: if bootrootd is down **through** the renewal deadline, the registrar first reports `EndpointUnreachable` and only after the operator starts it does it report `CredentialInvalid` — one cause, two remediations, the second historically wrong. Both must point at bootrootd.

Amend it to record three things:

1. **The server-side symptom sequence has collapsed.** Starting bootrootd now re-issues the endpoint's server leaf before the endpoint accepts a connection, so where bootrootd comes up the registrar's next dial succeeds rather than producing a second, different failure report. State it with that qualifier rather than as an unconditional success — where bootrootd does not come up, item 3 is what the operator sees.
2. **The caller's own client leaf is still observable before the repair can run.** The registrar reads its configured client material directly and refuses before dialing, so a bootrootd that was down through *that* leaf's expiry still yields a credential-lapse report until an operator starts it. What start-time repair changes there is the sequence the operator sees — one credential-lapse report instead of *unreachable* followed by *credential invalid* — not whether the lapse is observable at all.
3. **A new sub-case replaces the collapsed one.** If bootrootd cannot bring the endpoint up at start, it refuses to come up at all, so the registrar goes on reporting the endpoint unreachable, and the remedy is the failure named in bootrootd's startup diagnostic rather than anything about credentials. Write the sub-case around *bootrootd not coming up*, not around issuance failing specifically: issuance failing — an unreachable CA or a failed privileged-credential read — is the likely cause, but a successful issuance whose replacement the endpoint's own TLS loader then rejects reaches the same place, and the registrar's report is identical. A host clock far enough out of step with the CA's is how the second arises, in either direction.

Keep the property the trap exists to assert: **both symptoms still point at bootrootd**, so an operator is never sent toward a second, wrong remediation.

### Where the boundary runs, and what the amended passage must not say

The sub-case is written around an **observable** — the registrar goes on seeing the endpoint unreachable — and that observable is reached by more than one refusal. The amended text must keep the observable and the refusal apart, because the start-time repair work narrows the refusals without collapsing them into one:

- **Start-time certificate repair adds exactly one post-validation startup refusal: issuance itself having failed**, with a diagnostic naming the material paths and the failure. That is the refusal the daemon gains, and the amended passage must not attribute any other to it.
- **A successful issuance the endpoint's TLS loader then rejects is not that refusal.** It is the loader's own pre-existing refusal, applied to material issuance had already written, and it is in the sub-case because the *symptom* is identical — not because the cause is. Do not write it as an issuance failure, and do not present it as something start-time repair introduced.
- **So the remedy stays on the diagnostic, never on a presumed cause.** The two paths emit different diagnostics: the issuance one names the material paths and the failing read, request or write; the loader's names what it refused about the material it read. "The failure bootrootd's startup diagnostic names" is accurate for both, and is the wording to keep. Do not write the remedy as "re-run issuance", "check the CA", "check the clock" or anything else that presumes which path fired.
- **The sub-case concerns a start that already passed configuration validation.** An endpoint-enabled start is refused earlier — before any issuance, any OpenBao read and any endpoint TLS load — when a required `[registrar_endpoint]` certificate or key path is unset, when the host is not Linux, or when `[registrar] state_file` is absent. Those are a misconfigured host, not the trap's scenario, which is a correctly configured daemon that was down through a deadline and is then started. Do not enumerate them in the amended passage and do not let its wording read as though they were the sub-case.

Amend only what those two passages state. Leave section 4's other decisions untouched — the explicit client-identity requirement, the reload contract and the preference for roxyd re-reading the certificate on every dial, and the remaining-lifetime surface.

## Acceptance criteria

- [x] `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` section 4 no longer presents "bootrootd running and reaching its next renewal tick" as the recovery for an already-expired leaf, and instead states that a leaf already expired when bootrootd starts is repaired at that start, before the registrar endpoint loads its TLS material.
- [x] The amended sentence makes no claim about a leaf that expires while bootrootd is already running — it does not say such a leaf is repaired by the renewal mechanism, and does not substitute another repair for it.
- [x] Section 4's pre-existing renewal statements — the clause that the daemon issues and renews the certificate under its own credential, and the reload-contract paragraph — are present and unaltered.
- [x] The amended text explicitly preserves that recovery is bootrootd's and never re-provisioning the host, and leaves intact the statement that bootler provisions only the initial certificate.
- [x] The ordering-trap passage records that the server-side second symptom has collapsed where bootrootd comes up, that the caller's own client-leaf lapse remains observable before the repair can run, and that **bootrootd not coming up at start** is the sub-case replacing the collapsed symptom — written around that outcome rather than around issuance failing alone — with the remedy being the failure bootrootd's startup diagnostic names.
- [x] The amended passage does not describe the endpoint TLS loader's rejection of just-issued material as an issuance failure, and does not attribute any refusal other than issuance having failed to start-time certificate repair.
- [x] The amended passage's remedy wording names the startup diagnostic rather than a presumed cause: it does not tell an operator to re-run issuance, check the CA or check the clock.
- [x] The amended passage does not enumerate the configuration-validation refusals — an unset `[registrar_endpoint]` path, a non-Linux host, an absent `[registrar] state_file` — and does not read as though they were the sub-case.
- [x] The amended ordering-trap passage still asserts that both symptoms point at bootrootd.
- [x] No other section 4 decision is altered, and no section outside section 4 is touched.
- [x] No file outside `docs/rfcs/` is changed: no source file, no configuration file, no test, no `mkdocs.yml`.
- [x] After this issue, no sentence in the RFC and no behaviour in the daemon state different recovery timings for an already-expired registrar leaf.

## Constraints

- This is a documentation change. Do not modify source, configuration, tests, or `mkdocs.yml`.
- Do not rewrite section 4 wholesale. Amend the two passages named in Scope and leave the section's other decisions, and its surrounding argument, as they are.
- Do not remove the ordering trap. It is amended, not deleted: the property that both symptoms point at bootrootd is the reason the passage exists.
- Do not collapse the sub-case's two paths into one cause. The sub-case groups them by the symptom the registrar reports; the amended text must not state or imply that the loader's rejection of just-issued material is an issuance failure, that issuance failure is the only way the daemon fails to come up, or that either is the refusal an operator should assume.
- Do not describe the renewal schedule, a reload contract, a health-response member, or a lapse-reporting surface beyond what the two passages already touch. The RFC's existing statements about those stand.
- Do not have the amended text assert any repair for a leaf that expires while bootrootd is running. No renewal mechanism for these two leaves exists at the point this amendment lands, so such a sentence would document behaviour no code performs — replacing one disagreement between the RFC and the daemon with another. The amendment's subject is the state found at start.
- Do not add a `CHANGELOG.md` entry. This amends a design document and changes nothing a user of the last release could observe.
- Match the file's existing line-wrapping style in the amended prose rather than introducing a different one.
- Configure Markdown lint, if the amended prose needs it at all, at the narrowest scope that works — the line, then the file — and never by adding a rule to the repository-root `.markdownlint-cli2.yaml`.

## Out of scope

- Any change to daemon, configuration or test code.
- Any RFC section other than section 4, and any other file under `docs/rfcs/`.
- Documenting the same departure under `docs/en/` or `docs/ko/`. The operational documentation already carries it.
- The renewal schedule, the reload contract, the health field and the lapse surface, all of which are separate work — including saying in the RFC what repairs a leaf that expires while bootrootd is running. That belongs to whichever issue builds the renewal, alongside the mechanism itself.

## Test plan

- [x] `./scripts/check-docs.sh` passes, which is what CI runs for a change under `docs/`.
- [x] Markdown lint passes on `docs/rfcs/0001-registrar-role-and-non-self-propagation.md`, with no rule configuration added to the repository-root `.markdownlint-cli2.yaml`.
- [x] `git diff --name-only` against the base branch lists exactly `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` and nothing else.
- [x] Reading section 4 end to end, the recovery-timing sentence and the ordering-trap passage describe start-time repair, name bootrootd as the repairer, exclude re-provisioning the host, and carry the sub-case written around bootrootd not coming up at start — not around issuance failing alone.
- [x] `git diff` of the amended sentence adds no claim about expiry reached while bootrootd is running, and leaves section 4's pre-existing renewal clause and reload-contract paragraph byte-identical.
- [x] Reading the amended sub-case as an operator: the remedy it gives is "read bootrootd's startup diagnostic", it does not name issuance as the assumed cause, and it does not describe the loader's rejection of just-issued material as an issuance failure.
- [x] Reading the amended sub-case against the implementing issue's refusal list: the only post-validation refusal it attributes to start-time certificate repair is issuance having failed, and it names no configuration-validation refusal.

## Dependencies

Depends on #913, the issue that makes the daemon self-issue and repair the registrar surface's two certificates at start. It is closed and its behaviour is on `main`: `ensure_registrar_surface_certificates` in `src/registrar_certs.rs`, called from `src/bin/bootroot-agent.rs` before the daemon runs, with `src/tls.rs` recording that the path is deliberately restricted to start-time registrar repair — verified at `f52806e7d75e59d69b26aced8d6c3df759b3d416`. This issue documents behaviour that has already landed, so amending the RFC first would describe a daemon that does not yet exist. Part of #767, whose wave 1 — #913, #921, #922 and #930 — is complete and whose wave 2 is this issue alone.

## Pointers

- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md`, section 4 "The guarantee and the mechanism" — the bullet beginning "**Lapse recovery is bootrootd's, NOT bootler's**" opens at line 336 and holds both passages. The recovery-timing sentence runs from line 338 through the parenthetical ending at line 345; the reload-contract paragraph follows; the ordering-trap paragraph is lines 354–357. Line numbers are a convenience for locating the bullet, not an anchor — find the passages by their quoted text.
- `src/registrar/endpoint/tls.rs` `build_server_config`, reached from `RegistrarEndpoint::activate` in `src/bin/bootroot-agent.rs` — the endpoint's pre-existing TLS loader, whose `?` at the activation call site is why its rejection of just-issued material stops the daemon coming up. Background for writing the sub-case accurately; this issue changes no code.
- `.markdownlint-cli2.yaml` at the repository root — carries only `globs`, `ignores` and `MD024: siblings_only`; no rule configuration belongs there
- `scripts/check-docs.sh` — the documentation build CI runs for any change under `docs/`

sehkone:
## Summary

- Updates RFC 0001 section 4 to describe repair of a registrar leaf already expired at bootrootd start, before endpoint TLS material is loaded.
- Documents the collapsed server-side symptom sequence, the still-observable caller client-leaf lapse, and the startup-failure sub-case without conflating issuance and TLS-loader refusals.
- Preserves bootrootd as the sole recovery path and leaves the RFC's existing renewal and reload-contract decisions unchanged.

Closes #914

Part of #767

## Test plan

- [x] Run `./scripts/check-docs.sh`.
- [x] Run `markdownlint-cli2 "docs/rfcs/0001-registrar-role-and-non-self-propagation.md"`.
- [x] Confirm `git diff --name-only origin/main` contains only `docs/rfcs/0001-registrar-role-and-non-self-propagation.md`.
- [x] Review section 4 to verify it scopes repair to a leaf expired at startup, retains the renewal and reload-contract text, keeps both observed symptoms pointed at bootrootd, and uses the startup diagnostic as the remedy when bootrootd does not come up.

### DIFF
--- a/docs/rfcs/0001-registrar-role-and-non-self-propagation.md
+++ b/docs/rfcs/0001-registrar-role-and-non-self-propagation.md
@@ -336,12 +336,13 @@ Two consequences worth stating:
 - **Lapse recovery is bootrootd's, NOT bootler's — and the renewal must be
   handed off to the registrar.** Because the daemon issues and renews the
   certificate under its own credential, it needs nothing from the registrar to
-  do so: a lapsed registrar is repaired by **bootrootd running and reaching
-  its next renewal tick**, not by re-provisioning the host. (An earlier
-  draft said the sole recovery was re-running bootler; that was the
-  pre-self-issue model and is wrong under this decision. bootler provisions
-  the **initial** certificate and nothing more. RFC-E §9's `CredentialInvalid`
-  remedy names bootrootd for the same reason.)
+  do so: a registrar leaf already expired when bootrootd starts is repaired by
+  bootrootd at that start, before the registrar endpoint loads its TLS
+  material, and never by re-provisioning the host. (An earlier draft said the
+  sole recovery was re-running bootler; that was the pre-self-issue model and
+  is wrong under this decision. bootler provisions the **initial** certificate
+  and nothing more. RFC-E §9's `CredentialInvalid` remedy names bootrootd for
+  the same reason.)
   **The renewal needs a reload contract or it fails exactly as this deployment
   has already failed once:** the daemon writes a renewed certificate to disk
   and, absent a handoff, **nothing tells roxyd to reload it**, so roxyd keeps
@@ -351,10 +352,22 @@ Two consequences worth stating:
   once already. So either the renewal notifies the registrar, or **roxyd
   re-reads the certificate from disk on every dial**; the latter is preferred
   because it needs no cross-process signal.
-  Ordering trap worth stating: if bootrootd is down **through** the renewal
-  deadline, the registrar first reports `EndpointUnreachable` and only after
-  the operator starts it does it report `CredentialInvalid` — one cause, two
-  remediations, the second historically wrong. Both must point at bootrootd.
+  Ordering trap worth stating: when bootrootd comes up after being down
+  **through** the renewal deadline, it re-issues the endpoint's server leaf
+  before the endpoint accepts a connection, so the registrar's next dial
+  succeeds rather than producing a second, different failure report. The
+  registrar still reads its configured client material directly and refuses
+  before dialing: a client leaf that expired while bootrootd was down yields
+  `CredentialInvalid` until an operator starts it. The operator therefore sees
+  one credential-lapse report rather than `EndpointUnreachable` followed by
+  `CredentialInvalid`. If bootrootd does not come up at start, the registrar
+  goes on reporting `EndpointUnreachable`; the remedy is the failure named in
+  bootrootd's startup diagnostic. Start-time repair adds exactly one
+  post-validation startup refusal: issuance itself failing, whose diagnostic
+  names the material paths and the failed read, request, or write. If issuance
+  succeeds but the endpoint's own TLS loader rejects the replacement material,
+  its pre-existing refusal reaches the same sub-case. Both symptoms still
+  point at bootrootd.
   The daemon surfaces the certificate's remaining lifetime so a failing
   renewal is reported rather than discovered when the next install fails.
 


sehkone:
**[Reviewer Round 1]**

No findings. The RFC amendment accurately scopes repair to expired material found at startup and preserves the no-claim boundary for in-process expiry.

The ordering-trap rewrite correctly distinguishes the repaired server-side sequence, the still-observable client-leaf lapse, and the broader “bootrootd does not come up” sub-case without mislabeling TLS-loader rejection as issuance failure. See [RFC section 4](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-914-review/docs/rfcs/0001-registrar-role-and-non-self-propagation.md:336).

PR hygiene is also in order: it closes #914, carries the permitted `Part of #767`, includes a test plan, and has no review-thread issue creation.

sehkone:

PR #948, opened 2026-08-28
Refuse unmounted audit store bring-up (#924)

### LINKED ISSUE #924: Refuse the bring-up and the boot path on an unmounted audit store
# Refuse the bring-up and the boot path on an unmounted audit store

## Context

The registrar's two audit artifacts — the daemon's verb records under `<audit_store_dir>/records` and the OpenBao file audit device under `<audit_store_dir>/openbao` — share one directory on the bootroot host, sized by `[registrar] audit_store_reserve_bytes` and placed by `audit_store_dir`, with `audit_store_enforcement` selecting `filesystem` (the default) or `directory`. On a host whose registrar endpoint is enabled, `bootroot init` creates that layout and renders a Compose override that replaces the `openbao` service's `volumes:` list so the container's unchanged `/openbao/audit` is a bind mount of `<audit_store_dir>/openbao`.

By the time this issue starts, the provisioning half of the reserve exists and is the issue this one depends on. In `filesystem` mode the installer derives a loopback image at `<audit_store_dir>.img`, sized to exactly `audit_store_reserve_bytes`, fully allocated, uid 0, mode `0600`, carrying an ext4 filesystem made with `mkfs.ext4 -m 0 -E nodiscard`; it renders three artifacts into `<compose dir>/audit-store/` — a generated `<escaped audit_store_dir>.mount` unit plus ordering drop-ins for `docker.service` and `bootroot-registrar.service` — and it verifies, with metadata reads only, that the image meets its whole contract, that all three artifacts are installed under `/etc/systemd/system` and byte-identical to what was rendered, that `audit_store_dir` is mounted `ext4` from a loop device whose backing file is the derived image, and that `records/` and `openbao/` are present on the mounted filesystem at mode `0700`. Those three steps are named phase 1 (derive, evaluate, preflight, render), phase 2 (every step that changes the host, which the operator performs and bootroot only renders) and phase 3 (verify and report). A run reports exactly one outcome: **enforced**, **provisioned, not activated**, **failed**, or — for an explicitly configured `directory` deployment — **`unenforced (directory)`**, which is a success. In `filesystem` mode bootroot creates `audit_store_dir` alone and neither subdirectory beneath it, since a subdirectory made before the mount is one the mount hides.

**All of that lives on the fresh-host path only.** The phases run inside `bootroot init` and its phase-1 refusals are repeated by `bootroot reinit` ahead of its wipe, and both are commands an operator can only use before the deployment is up: `bootroot init` bails on an already-initialised host, because `diagnose_partial_init` refuses when OpenBao reports itself initialised and no `--root-token` was supplied. So a deployment that is already running has no command that renders the artifacts and no command that re-verifies the mount, and `bootroot infra up` — the one command it can run — brings the Compose stack up against whatever `audit_store_dir` happens to be.

**And the boot path has no refusal at all.** The ordering drop-ins are `Wants=` plus `After=` and nothing stronger, because a hard relation on `docker.service` would take the whole Compose stack down over a fault local to one directory. A mount job that fails at boot therefore does not stop Docker, Compose's `restart: always` brings the OpenBao container back, Docker creates a missing bind source by default, and the mandatory audit device is written to the root filesystem under an empty mount point — healthy-looking, protecting nothing, on the one path where neither an operator nor a bootroot command is watching. The provisioning work documents that as a residual precisely because closing it is this issue's.

**How far that can be closed is bounded by what Compose's guard actually does, and the bound is stated here rather than discovered later.** `create_host_path: false` stops Docker from *creating* a bind source that is missing; where the path already exists, Compose binds it and the container starts. So the guard covers every state a `filesystem`-mode store can be in **once bootroot has stopped creating the subdirectories beneath an unmounted store** — a fresh host, where the provisioning work creates `audit_store_dir` alone; an activated host whose mount fails at a later boot, whose underlying directory was proven empty before activation and cannot have been written to beneath the mount; and the maintenance window during which an existing store's contents have been moved aside. It does **not** cover the one state that predates all of them: a host provisioned before this work, whose `<audit_store_dir>/openbao` exists on the underlying filesystem because the store's creation routine makes all three directories in one unconditional pass. Every endpoint-enabled deployment in existence is in that state today, and for each of them the guard is inert until those contents are relocated onto the reserve — a separate documented procedure. That is a **retained residual**, not a closed one, and this issue documents it as such.

This issue closes the first outright and the second within that bound. It makes `bootroot infra up` carry the whole bootroot-side pass and refuse the bring-up short of **enforced**, and it declares the audit bind so that Docker refuses to manufacture an absent bind source — the same fail-closed verdict, delivered by the only component that runs on the boot path, for every store whose subdirectories have not been left behind on the underlying filesystem.

**"The bootroot-side pass" means phases 1 and 3 — everything bootroot does and nothing the operator must.** The run is root on both surfaces, so the division is not privilege: it is that bootroot renders, reads and reports, while every step that changes the host is a command the operator runs. This issue moves no step across that line.

**What that means for a deployment that is already running, stated plainly because it is a real consequence and not a side effect.** An existing endpoint-enabled `filesystem`-mode host has an unmounted store the moment this lands, so it cannot bring its stack up until enforcement is activated. Activating it means relocating the records that store already holds, and that machinery — a tracked holding directory, a capacity check against the mounted filesystem, a verified copy, a resume and a rollback — is the issue that depends on this one. Two things bound the exposure, and both are documented here rather than left for an operator to discover: the entire surface is gated on `[registrar_endpoint] enabled`, which no deployment has on unless it asked for it; and `audit_store_enforcement = "directory"` is an explicit, supported configuration that turns every refusal in this issue off. Do not soften the refusal to close that window, and do not render a partial relocation sequence to paper over it — a half-specified migration over the only copy of a deployment's audit records is worse than a refusal with a documented opt-out.

### The live-deployment surface is `bootroot infra up`

`bootroot infra up` is the one command an operator on a running deployment can use against the audit store: it already resolves the rendered Compose override, recovers `audit_store_dir` from it and re-checks that directory against the store's ownership contract on hosts whose recorded registrar predicate is enabled. This issue makes it carry **the whole bootroot-side pass** — phase 1 and phase 3, the same outcome line the fresh-host path prints, and a refusal of the bring-up **before the OpenBao container starts** on anything short of **enforced**, performing no step that changes the host.

**It has to render, not merely verify.** A verify-only `infra up` would leave an operator with no way to obtain the three artifacts or the outstanding privileged commands, because the command that renders them cannot run on their host. An implementation that only verifies fails this issue. The artifacts land where the fresh-host path already puts them, `<compose dir>/audit-store/`, and are not written a second time under a different name.

**The rendered "run this command again" step names `bootroot infra up`, not `bootroot init`.** That step is already parameterised on the command the run was invoked as, so this issue supplies its own — `bootroot infra up --agent-config <path>`, the invocation the operator can actually re-run — and adds no second copy of the step and no second catalogue string. A rendered instruction to re-run `bootroot init` on a live deployment names a command that refuses on that host, which is worse than no instruction at all.

**The mode gate is unchanged here.** A `directory` host reaches none of the phases: `infra up` reports **`unenforced (directory)`**, refuses nothing and brings the stack up, exactly as `bootroot init` reports it. Phases 1 and 3 are `filesystem`-mode work on both surfaces.

**`bootroot init` and `bootroot reinit` are unchanged by this issue.** They already carry the phases; nothing here moves, duplicates or re-tunes them, and the fresh-host sequence documented against `bootroot init` stays exactly as it is. `bootroot infra install` is not part of this surface and is unchanged.

### `--agent-config` on `infra up`

Phase 3 needs two values `infra up` cannot reach today: `audit_store_enforcement`, to know whether to run at all, and `audit_store_reserve_bytes`, to check the image's exact size. Neither is recorded in `state.json`, beside the rendered artifacts or in the Compose override, and neither may be recorded there by this issue — a ceiling read back from a file bootroot wrote is a ceiling that agrees with itself. So **`InfraUpArgs` gains `--agent-config`**, and the two values are read from that file alone.

- **The flag is mandatory whenever `state.json` records the registrar predicate as enabled **or** an audit override is rendered** — the same rule `plan_audit_store` already applies on the `init` side, so one rule covers both commands. Do not widen it.
- That does not weaken the guarantee for a deployment which has not enabled the endpoint, because such a deployment provably has no rendered override: `apply_audit_store` writes one only on its `Provision` arm, reached only when the predicate is on. The state where the second clause bites is a host that *had* the endpoint enabled and is turning it off, whose override `init` has not yet unwound.
- **Where the flag is required and absent, the bring-up hard-refuses and starts no container.** A gate an operator can omit is one they omit on the run that needed it.
- The same **enablement cross-check** `bootroot init` applies runs here: a configuration whose `[registrar_endpoint] enabled` disagrees with `state.json` fails, with the same message shape.
- `error_audit_store_agent_config_required` names `bootroot init` in its text, so it takes the command as a parameter or gains a sibling message, in `en` **and** `ko`.
- **With the predicate off and no override rendered** — the state of every deployment that has not enabled the endpoint — `bootroot infra up` behaves, prints and exits exactly as it does on the default branch, for **either** value of `audit_store_enforcement`. It requires no flag, opens no configuration file, derives no image path, `stat`s nothing and prints no outcome line. Add no read, no `stat` and no diagnostic to that path.
- Every in-tree invocation of `bootroot infra up` that runs against an endpoint-enabled host must pass the flag, including the one in `scripts/impl/run-registrar-internal-init-e2e.sh:762`.

### The refusal `infra up` raises (`filesystem` mode)

Phases 1 and 3 run where the audit Compose override is resolved, gated on the endpoint predicate exactly as the fresh-host path is. Anything short of **enforced** refuses the bring-up under that outcome, naming the rendered artifact paths and the exact remaining commands — the same refusal shape the store's ownership faults already use — and starts no container. The refusal happens before any Docker call that could bring OpenBao up against an unmounted store.

**Do not reimplement the phases.** Call the existing phase-1 and phase-3 entry points and the existing outcome reporting. A second copy of the image contract, the mount-identity proof or the outcome vocabulary is two definitions of **enforced** that will disagree, on the surface where the disagreement is invisible until a store is already full.

**With the endpoint disabled nothing here runs**, and nothing is refused: there is no audit store this surface is responsible for, and the deployment produces no verb records to protect.

### The boot-path refusal: Docker will not manufacture the bind source

The refusal above covers the path an operator drives. It does not cover the one nobody watches. A mount job that fails at boot does not stop `docker.service`, `restart: always` brings the OpenBao container back, Docker creates a missing bind source by default, and the mandatory audit device is written to the root filesystem under an empty mount point — exactly the failure this effort exists to end.

**The mechanism is `create_host_path: false` on the audit bind in the rendered Compose override.** With it, Docker refuses to create the missing host path and the container fails to start. That is the same fail-closed verdict `infra up` reaches for the bring-up, delivered by the only component that runs on the boot path — for every store whose bind source is absent, which is what the boundary below fixes precisely.

**This is a deliberate departure from the enforcement work's original wording, recorded here so it is not read as drift.** That wording forbade rendering or modifying the Compose override at all, on the ground that the file belonged to the layout work landing ahead of it. The layout work has landed and the override is on the default branch, and what the wording did not account for is that the boot path has no other refusal available: no bootroot command and no operator runs there. The prohibition therefore holds everywhere it was aimed — the provisioning issue leaves the file byte-identical, and nothing here changes what is mounted, the container path, the file's location or mode, or the command that writes it — and this issue changes one entry's syntax to add one key.

- **It forces the long mount syntax for that one entry.** `create_host_path` lives under `bind:` in the long form, so the audit entry stops being the short `'<source>:/openbao/audit'` scalar and becomes a mapping with `type: bind`, `source:`, `target: /openbao/audit` and `bind: { create_host_path: false }`. **Only that entry changes**: the other two `volumes:` members keep their short form, the `!override` tag stays, and the container path, the file's location and mode, when it is written and which command writes it are all unchanged.
- **`source:` keeps the renderer's existing quoting rules** — a YAML single-quoted scalar with `'` doubled and `$` doubled against Compose interpolation — because Compose still interpolates every value it reads.
- **The refusal of a store path holding `:` or a control character stays, and the reason it stays is recorded.** It exists today because Compose splits a short-form bind on `:`, and in the long form `source:` is its own key so a colon no longer moves the boundary. Keep it anyway: the reader, the systemd escaping and the rendered shell commands all still have to survive that path, and widening an input rule is not this issue's business. Note that in the code, so a later reader does not delete it as vestigial.
- **The reader is changed in the same commit, and accepts both forms.** `read_audit_override_store_dir` recovers the store path by stripping `:/openbao/audit` off a short-form entry; against the long form it finds nothing and every `bootroot infra up` fails to resolve the override. It must read `source:` out of the audit entry's mapping — **and must still read the short form**, because every host provisioned before this change carries one. Renderer and reader are one contract and are tested as one: what the renderer emits round-trips, and so does the short form the previous renderer emitted.
- **`bootroot infra up` upgrades a short-form audit bind in place, once**, which is the only way the guard reaches a deployment that is already running: `bootroot init` cannot be re-run there. The rewrite **preserves the bind source byte-for-byte as it was read** and never re-derives it from `--agent-config` — re-deriving would silently relocate the audit device wherever the two disagreed, which is a far larger change than a syntax upgrade. A host already carrying the long form is left byte-identical. The write is atomic, like the renderer's own. Where the configuration and the override disagree about the store, that is the existing cross-check's business, not this rewrite's.
- **Where it bites.** With the mount up, `<audit_store_dir>/openbao` exists on the mounted filesystem and the container starts exactly as before. With the mount absent and the underlying store empty — a fresh `filesystem`-mode host, where bootroot creates `audit_store_dir` alone; an activated host whose mount fails at a later boot; the window in which an existing store's contents have been moved aside — the bind source does not exist, so the container fails to start instead of writing to the root filesystem.
- **Where it does not bite, stated as a retained residual and never as a closed hole.** Compose's key governs *creation* only: an existing bind source is bound as it stands. A host provisioned before this work carries `<audit_store_dir>/openbao` on the underlying filesystem, the store's creation routine having made all three directories in one unconditional pass, and on such a host a boot with a failed mount still binds that directory and still writes the audit device to the root filesystem. The guard is inert there until the contents are relocated onto the reserve, which is a separate documented procedure. **No message, comment, test name or manual passage may claim this issue closes the boot path for such a host.** bootroot does not close it by deleting or moving those contents — never touching an operator's audit records is the rule this whole surface is built on — and no other actor runs on the boot path, so documenting the bound and naming the relocation as what retires it is the whole of the answer here.
- In `directory` mode `create_layout` still creates all three directories, so the bind source always exists and nothing changes. On a host that has never enabled the endpoint no override is rendered at all, so nothing changes there either.
- **The trade is deliberate and is documented.** Where the guard bites, a boot with a failed mount leaves the OpenBao container failing to start rather than quietly writing its audit device elsewhere. A visible outage of the audited service is the intended outcome, and the manual says so beside the recovery: fix the mount, then start the stack.

### A store that already holds records is refused and left alone

Phase 1 reports **provisioned, not activated** for a `filesystem`-mode store that still holds content on the underlying filesystem, and fails outright when that content exceeds `audit_store_reserve_bytes`. Running that phase from `infra up` puts the refusal in front of a live deployment for the first time, and the refusal is all this issue adds to it. Its rendering rule is the one the provisioning work already fixes and this surface inherits unchanged: the three artifacts are still written, **no phase-2 command at all** is rendered, and the outcome names the store path, says that it holds content, lists the rendered artifact paths and states the two supported ways forward — relocating those records, a separate documented procedure, or configuring `audit_store_enforcement = "directory"`.

**That state is also where the bind-source guard is inert**, its `openbao/` being present on the underlying filesystem, so the outcome for such a store says both things at once: the bring-up is refused here, and an unattended boot still binds that directory until the contents are relocated.

**It renders no part of a relocation.** No move-aside, no holding directory, no `cp`, no manifest, no comparison, no release, no `rm`, no rename — and it deletes, moves, unmounts and mounts over nothing. Those steps only carry an audit trail safely when they come with the checks the dependent issue specifies: a holding directory that is a state bootroot recognises ahead of every other verdict, a capacity check against the mounted filesystem's own available bytes rather than the configured reserve, refusal of a mount point or a special file inside the set being copied, NUL-safe manifest comparison gating the close, closure by rename rather than a recursive delete, and a rendered resume and rollback. Rendering the bare commands without them produces a sequence an operator can half-finish with no way for bootroot to notice, over the only copy of a deployment's audit records. That whole sequence — the rendering **and** the documentation — belongs to the issue that depends on this one, and none of it is written here.

### Documentation

Document, in **both** `docs/en/` and `docs/ko/`, on the existing audit-store and configuration pages with no `mkdocs.yml` nav change. The provisioning work already wrote those sections — the two modes, the phases, the four outcomes, the fresh-host sequence, the removal sequence, the residuals, and the rule that a changed reserve is replace-and-migrate rather than a resize. Extend those passages; do not open a second audit-store section that states any of it again:

- **The `infra up` refusal**: that a `filesystem`-mode deployment whose store is not verified as the mounted reserve has `bootroot infra up` refused before the stack comes up, with the store path, the outstanding commands and the rendered artifact paths named; that `infra up` renders as well as verifies, so it is the command that produces the artifacts on a host where `bootroot init` cannot run; and that none of it runs on a host whose registrar endpoint is not enabled.
- **The bind-source guard and its exact boundary**: that the audit bind is declared so Docker will not create the host path, so a boot whose mount job failed leaves the OpenBao container failing to start rather than writing its audit device to the root filesystem; that this is the intended trade and its recovery is to fix the mount and start the stack; that the entry's syntax changed for that reason while nothing about what is mounted did; and that `bootroot infra up` upgrades an override written before this change, in place and without moving the store. **Say in the same passage that the guard governs creation only**, so a store that still holds `openbao/` on the underlying filesystem — which is every deployment provisioned before this work — is bound as it stands and keeps writing its device to the root filesystem on a boot whose mount failed, until those contents are relocated onto the reserve by the separate documented procedure. **Rewrite the boot-path residual the provisioning work recorded to that narrower form** rather than deleting it or leaving two statements side by side, and name relocation as what retires it. Leave the other residuals it recorded alone — including the one about a daemon that still creates `records/` beneath an unmounted store, which is separate work and whose passage this issue does not touch.
- **`--agent-config` on `bootroot infra up`**: when it is required, what it is read for, and that a deployment which has never enabled the endpoint needs it no more than it did before.
- **What an operator of an existing deployment should do**, stated without a relocation procedure and added to the existing passage on a store that already holds records: that store is refused rather than mounted over on this surface too, moving those records onto the reserve is a separate documented procedure, and `audit_store_enforcement = "directory"` is the supported configuration for a deployment that is not ready to activate, turning the refusal off and leaving the reserve as the budget it is today.
- **The size-mismatch refusal a changed `audit_store_reserve_bytes` produces**, written into the passage that already states the replace-and-migrate rule and is where that error already points. `bootroot infra up` fails naming both sizes; a new value below the phase-1 minimum — the larger of the fixed ext4 floor and `audit_max_file_bytes` × (`audit_max_retained_files` + 1) — is refused by that check first, so nothing starts against a size that cannot be formatted. The rule does not change: the image is never resized, truncated or reformatted in place, in either direction, and the replacement is a documented operator procedure run in a maintenance window with both writers stopped.
  **The numbered steps that carry the records from the old image to the new one are deliberately not written here.** They copy a mounted store's whole contents onto another filesystem, which needs the pinned copy, the three verification comparisons, the resume and the rollback that the relocation issue defines; a second, thinner copy procedure sitting beside the refusal is exactly the hazard that issue exists to remove. It adds those steps to this same passage. So the passage this issue writes must read completely on its own without them — refusal, rule, and the statement that the replacement is a separate documented procedure.

## Acceptance criteria

- [x] `bootroot infra up` takes **`--agent-config`**, mandatory whenever `state.json` records the registrar predicate as enabled **or** an audit override is rendered — the rule `plan_audit_store` already applies — refused with the same shape and the same enablement cross-check, and it reads `audit_store_enforcement` and `audit_store_reserve_bytes` from that file alone; neither value is recorded anywhere else.
- [x] In `filesystem` mode `bootroot infra up` runs **phase 1 and phase 3** while resolving the audit Compose override — rendering the three artifacts into `<compose dir>/audit-store/` and reporting the outstanding privileged commands exactly as the fresh-host path does — and refuses the bring-up before the OpenBao container starts on anything short of **enforced**, performing no privileged step. An implementation that only verifies fails this criterion, leaving a live deployment with no command that can render the artifacts.
- [x] The phases are **called, not reimplemented**: no second image-contract check, no second mount-identity proof and no second outcome vocabulary appears on this surface.
- [x] The rendered re-run step names the command actually invoked: `bootroot infra up --agent-config <path>` on this surface, `bootroot init` on the fresh-host one, from one parameterised message rather than two. No message rendered by `infra up` tells an operator to re-run `bootroot init`.
- [x] In `directory` mode `bootroot infra up` runs neither phase, reports **`unenforced (directory)`**, refuses nothing and brings the stack up, deriving no image path and `stat`ing no image.
- [x] With the recorded predicate **off and no audit override rendered** — the state of every deployment that has not enabled the endpoint — `bootroot infra up` behaves, prints and exits exactly as it does on the default branch, for **either** value of `audit_store_enforcement`, requiring and reading no `--agent-config` and opening no configuration file.
- [x] **The rendered audit bind carries `create_host_path: false`**, in the long mount syntax, with `type: bind`, the quoted `source:`, `target: /openbao/audit` and nothing else changed: the other two `volumes:` entries keep their short form, `!override` stays, the container path, the file's path and mode and the command that writes it are unchanged, and no second override is added. With the bind source absent, a Compose bring-up fails to create it and the container does not start.
- [x] **The guard's boundary is stated wherever its effect is claimed.** With the bind source **present** on the underlying filesystem — the state of every host provisioned before this work — the container still starts and still writes there, and that is asserted rather than assumed. The outcome text, the code comment beside the rendered key and both manuals say that the guard governs creation only, that such a host keeps the boot-path exposure until its contents are relocated onto the reserve, and that relocation is what retires it. No string added by this issue claims the boot path is closed for that host, and the provisioning work's boot-path residual is rewritten to that narrower form rather than removed.
- [x] **`read_audit_override_store_dir` round-trips both forms**: the long form this issue renders, and the short form every host provisioned before it carries. The store path it returns is identical for the two, and the existing refusal of a store path containing `:` or a control character still fires in the renderer, with a comment recording why it outlives the syntax that first required it.
- [x] **`bootroot infra up` upgrades a short-form audit bind in place, exactly once**, preserving the bind source byte-for-byte as read rather than re-deriving it from `--agent-config`; a host already on the long form is left byte-identical; the write is atomic. An implementation that re-derives the source from the configuration fails this criterion.
- [x] With `[registrar_endpoint] enabled` false, nothing this issue adds runs for **either** value of `audit_store_enforcement`: `bootroot infra up` derives no image path, `stat`s nothing, prints no outcome line and adds no warning.
- [x] `bootroot init` and `bootroot reinit` are behaviourally unchanged by this issue — same outcomes, same rendered commands, same exit statuses — and `bootroot infra install` is untouched.
- [x] On a `filesystem`-mode host with a non-empty underlying store, `bootroot infra up` reports **provisioned, not activated**, names the store and the rendered artifact paths, and **renders no relocation step whatever**: no move-aside, no holding-directory path, no `cp`, no manifest, no comparison, no release, no `rm` and no rename. Nothing is deleted, moved, unmounted or mounted over. A rendered command matching any of those shapes fails this criterion.
- [x] Bootroot performs no step that changes the host on this surface, whatever uid the run has: no `mkfs`, no `install` under `/etc/systemd/system`, no `systemctl`, no `chown`/`chmod`, no `mkdir` beneath a `filesystem`-mode store, no mount, no rename of a store and no copy.
- [x] Every in-tree `bootroot infra up` invocation that runs against an endpoint-enabled host passes `--agent-config`, and the E2E suites pass unchanged otherwise.
- [x] Every operator-facing string this issue adds or changes exists in **both** `en` and `ko`, including the reworked `error_audit_store_agent_config_required` that no longer hard-codes `bootroot init`.
- [x] Everything listed under Documentation is documented in **both** `docs/en/` and `docs/ko/` — the `infra up` refusal, the bind-source guard with its trade and its recovery, the `--agent-config` rule, the `directory` opt-out and the size-mismatch refusal with the never-resize-in-place rule — with no `mkdocs.yml` nav change, and `./scripts/check-docs.sh` passes. The boot-path residual the provisioning work recorded is rewritten to the guard's narrower boundary rather than deleted or left standing unchanged beside it, and no relocation or activation procedure is documented here.
- [x] `cargo clippy --all-targets` is warning-free under `-D warnings` and `cargo fmt -- --config group_imports=StdExternalCrate --check` passes.

## Constraints

- Everything this issue adds is gated on `[registrar_endpoint] enabled`. Do not change behaviour, output or exit status for a deployment that has not enabled the endpoint and has no override rendered, and do not add a warning for one whose configuration selects `filesystem` mode.
- `audit_store_enforcement` is the only input that selects a mode. Never infer one, never fall back from `filesystem` to `directory`, never write the key back, and never apply a `filesystem`-mode refusal to a `directory` deployment.
- Do not record `audit_store_enforcement` or `audit_store_reserve_bytes` anywhere — not in `state.json`, not beside the rendered artifacts, not in the Compose override. `infra up` reads them through `--agent-config` and does not proceed without it wherever that flag is required. Do not widen the requirement past the rule `plan_audit_store` already applies, and add no read, `stat` or diagnostic to the path where it is not required.
- Do not reimplement the phases, and do not define a second outcome vocabulary.
- Do not soften the refusal, add an override flag that skips it, or downgrade it to a warning, to spare a deployment that has not yet activated. The supported answer for such a deployment is explicit `audit_store_enforcement = "directory"` configuration, which is documented.
- **Render no relocation of an existing store, and document none.** No move-aside, no `<audit_store_dir>.pre-mount` path, no `cp`, no manifest or tree comparison, no release step, no recursive delete and no rename of a store appears in a rendered command, in a message, or in either manual. That sequence is only safe with the tracking, capacity check, refusals, verification gate, resume and rollback the dependent issue specifies, and shipping a bare version of it first is shipping a way to lose audit records.
- **Change nothing about the Compose override except the audit entry's syntax and its `create_host_path: false`.** Not what is mounted, not the container path, not the other two `volumes:` entries, not `!override`, not the file's path or mode, not when it is written. Do not relax the renderer's refusal of a `:` or a control character in the store path. Do not re-derive the bind source when upgrading an existing override, and do not drop short-form support from the reader — every host provisioned before this change carries one. Do not add a second Compose override.
- Do not change the store's layout contract or ownership rules, the derivation of `audit_record_dir`, the derived image path or unit name, or the rendered systemd artifacts' content. Do not add an `audit_store_*` key.
- Do not have bootroot perform any privileged step, whatever uid the run has, and do not start any container before the refusal decision has been made.
- Do not change `bootroot init`, `bootroot reinit` or `bootroot infra install`.

## Out of scope

- **The provisioning half itself.** The image contract, the free-space preflight, the artifact renderers, the phase-3 checks, the four outcomes and the store-creation rules are the issue this one depends on. Call them; do not re-derive them.
- **The daemon's own refusal of the registrar verbs against an unmounted store**, and the residual that a started daemon creates `records/` beneath an unmounted store because opening the record store creates its directories. That is separate work with its own issue; this one adds no check to `src/daemon.rs`, touches neither `AuditRecordStore::open` nor `prepare()`, and leaves that residual's manual passage as the provisioning work wrote it.
- **Every other edit to the rendered Compose override or its reader.** This issue changes the audit entry's syntax and adds `create_host_path: false` to it, and nothing else in that file.
- **Moving records onto the reserve, in every part — at first activation and on a reserve change alike.** The holding directory as a tracked state, the migration outcome, the capacity check against the mounted filesystem, the mount-point and special-file refusals, the manifest comparison, the close-by-rename, the resume and the rollback, the numbered steps that replace the image when the reserve changes, every command that renders any of them, and the operator procedures in both manuals are the dependent issue's. This issue neither renders nor documents any of it: it owns the size-mismatch refusal and the never-resize-in-place rule, and stops there.
- **Measuring the store.** The `statvfs`-backed capacity probe, the mode-dependent usage measurement, the headroom arithmetic, the low-water alarm and any `registrar_health` member for it are separate work; this issue computes none of it.
- **Re-verifying the empty-mount-point invariant after activation, and watching for a mount lost while the container runs.** Once mounted, the directory beneath cannot be named by any read available here, and a running container keeps the mount it started with. Both are documented residuals of the provisioning issue.
- **Bounding the OpenBao audit device's own growth** — rotating, pruning or capping it. Implementing a project quota or repartitioning a host for a dedicated audit partition. The registrar's rate limiter, in every part.

## Test plan

- [x] `infra up` flag test: with the recorded predicate on, **or** with it off and an audit override present, an invocation without `--agent-config` fails naming the flag and starts no container; with the predicate off and no override — the never-enabled state — it succeeds untouched and opens no configuration file; a configuration whose `[registrar_endpoint] enabled` disagrees with `state.json` fails the same cross-check `bootroot init` applies.
- [x] `infra up` rendering test: with the flag, on an endpoint-enabled `filesystem` host with no artifacts present, it **writes the three artifacts** into `<compose dir>/audit-store/`, reports **provisioned, not activated** with the exact remaining commands, starts no container and performs no privileged step. The regression a verify-only implementation fails. The rendered remaining commands close with `bootroot infra up --agent-config`, and no string in that output tells the operator to re-run `bootroot init`; the same phase run from `bootroot init` still closes with `bootroot init`.
- [x] `infra up` outcome test: on the same host with everything in place it reports **enforced** and brings the stack up; on a `directory` host it reports **`unenforced (directory)`**, exits 0 and brings the stack up, deriving no image path and running no verification. Drive phase 3's reads through injected metadata so the test needs no root.
- [x] Refusal-ordering test: on a `filesystem`-mode host short of **enforced**, the refusal is raised before any Docker call, and no container is started; the outcome names the rendered artifact paths and the remaining commands.
- [x] **Override render-and-read test**: the rendered audit entry is the long form carrying `type: bind`, the quoted `source:`, `target: /openbao/audit` and `bind: { create_host_path: false }`, while the other two `volumes:` entries and the `!override` tag are byte-identical to the default branch's output. `read_audit_override_store_dir` returns the same store path from that output and from a pinned short-form fixture written by the previous renderer — the regression a reader rewritten only for the long form fails, by breaking every host provisioned before this change. A store path holding a `$` and one holding a `'` each round-trip; one holding a `:` is still refused by the renderer.
- [x] **Override upgrade test**: `bootroot infra up` against a pinned short-form override rewrites only the audit entry, leaves the bind source byte-identical to what was read even when `--agent-config` names a different `audit_store_dir` — the regression a re-deriving implementation fails, by relocating the audit device — and leaves a long-form override untouched. A second run changes nothing.
- [x] **E2E bind-guard assertion, over both sides of the boundary**, where a real loop device exists. (1) *Absent source*: with `filesystem` mode configured, the mount unit not enabled and the underlying store empty, a Compose bring-up leaves the OpenBao container failed to start and creates no `<audit_store_dir>/openbao` on the root filesystem — the assertion the provisioning issue deliberately left unwritten. (2) *Pre-existing source*: the same host with `<audit_store_dir>/openbao` present on the underlying filesystem, as every deployment provisioned before this work has it, brings the container **up**, which is the documented residual rather than a bug; the test asserts that outcome explicitly and carries a comment naming relocation as what retires it, so nobody later reads the guard as covering this case. (3) With the mount up the container starts normally and `assert-openbao-audit-log` still passes.
- [x] Endpoint-disabled test: with `[registrar_endpoint] enabled` false and `audit_store_enforcement` set to each of `filesystem` and `directory`, `bootroot infra up` output and exit status are unchanged from the pre-change behaviour and no configuration file is opened.
- [x] **No-relocation test**: on an unmounted `filesystem`-mode store holding `records/`, `openbao/`, and both, `bootroot infra up` reports **provisioned, not activated**, leaves every byte in place, and the full set of rendered commands contains no `mv` of the store, no `cp`, no `find`, no `cmp`, no `diff`, no `rm` and no reference to a `.pre-mount` path. Assert over the rendered command list rather than by eye, so an implementation that reintroduces the bare sequence fails.
- [x] Unchanged-surface test: `bootroot init` and `bootroot reinit` produce the same outcomes, rendered commands and exit statuses as before this issue on each of a fresh, a provisioned-not-activated and an enforced host; `src/daemon.rs`, `AuditRecordStore::open` and `prepare()` are byte-unchanged.
- [x] Privilege-boundary test asserting no code path added here issues `mkfs`, `install` under `/etc/systemd/system`, `systemctl`, a `chown`/`chmod`, a `mkdir` under a `filesystem`-mode store, a mount, a rename of a store or a copy — whatever uid it runs under.
- [x] Message-catalogue test asserting every string this issue adds or changes resolves in both `en` and `ko`, and that the reworked agent-config refusal names the command it was invoked as rather than a hard-coded `bootroot init`.
- [x] E2E: every in-tree `bootroot infra up` invocation against an endpoint-enabled host passes `--agent-config` — `scripts/impl/run-registrar-internal-init-e2e.sh:762` included — the lifecycle suites still reach **enforced**, and an `infra up` run against a deliberately unmounted store refuses before the OpenBao container starts and leaves no container running.
- [x] Documentation check: `./scripts/check-docs.sh` passes, and a review checklist confirms the en/ko pair covers the `infra up` refusal, the bind-source guard with its trade and its recovery, the `--agent-config` rule, the `directory` opt-out and the size-mismatch refusal with the never-resize-in-place rule — that these extend the existing audit-store section rather than opening a second one, that the boot-path residual the provisioning work recorded is **rewritten to the guard's narrower boundary** — creation only, with a pre-existing `openbao/` still bound and relocation named as what retires it — rather than deleted or left standing unchanged, while the other residuals it recorded are untouched, and that neither page carries a relocation, an activation or an image-replacement sequence.
- [x] Tests use `tempfile::tempdir()` and never a fixed path; no test binds a fixed port or reaches the network.

## Dependencies

Depends on **#923**, the issue that provisions and verifies the reserve — the loopback image, the generated mount unit and drop-ins, the three phases and the four outcomes. This issue calls those and adds no second definition of them, so it lands after it.

**#926**, the issue that relocates records onto the mounted reserve, depends on this one: its holding directory, its migration outcome, its capacity check against the mounted filesystem and its rendered resume and rollback all attach to the live-deployment surface this issue creates, and the absent bind source this issue makes fatal is what keeps the OpenBao container out of a half-populated store during that window. The numbered steps this issue's size-mismatch refusal points at are added there.

The daemon's own refusal of the registrar verbs against an unmounted store, **#925**, is a sibling of this issue rather than a predecessor or a successor: it depends on the same provisioning work, shares no code with this surface, and may land before or after. Part of #773, which is part of #775.

## Pointers

- `src/commands/infra.rs:176` (`resolve_audit_override`, inside `run_infra_up`) — where `bootroot infra up` re-checks the store today and where phases 1 and 3 join it; `src/commands/infra.rs:370` (`run_infra_install`) — not this surface
- `src/cli/args.rs:817` (`InfraUpArgs`) — where `--agent-config` goes, mirroring the rule in `plan_audit_store`
- `src/commands/audit_store.rs` — `plan_audit_store` at `:460` for the conditionally-mandatory `--agent-config` rule, the enablement cross-check and the endpoint gate; `load_agent_config`, `audit_override_path`, and the phase entry points this issue calls. `render_audit_override` at `:209` — the short-form bind this issue converts, and the `:`/control-character refusal that stays; `compose_quote` at `:227` and `compose_unquote` — the scalar rules `source:` keeps; `read_audit_override_store_dir` at `:259` and `collect_openbao_volume_entries` — the suffix-stripping reader that must learn the long form without forgetting the short one; `OPENBAO_AUDIT_CONTAINER_PATH`, `OPENBAO_DATA_MOUNT` and `OPENBAO_CONFIG_MOUNT` — the two entries that do not change
- `docker-compose.yml:11` — the `openbao-audit:/openbao/audit` entry the override replaces; `docker-compose.yml:5` — `restart: always`, why the boot path needs a guard of its own
- `src/i18n/en.rs:591` — `error_audit_store_agent_config_required`, whose text hard-codes `bootroot init`; `src/i18n/ko.rs` and `src/i18n/audit_store.rs` for its counterpart and for every new refusal
- `src/commands/init/steps/orchestrator.rs:338` — where `apply_audit_store` is called on the fresh-host path; `diagnose_partial_init` at `:455`, which is why a live host cannot re-run `bootroot init`
- `src/registrar/audit_store.rs` — `layout_of`, `check_store_directory` and the `RECORDS_SUBDIR` / `OPENBAO_SUBDIR` / `STORE_DIR_MODE` constants the store path resolution uses
- `scripts/impl/run-registrar-internal-init-e2e.sh:762` — the `infra up` invocation that must pass the new flag; `scripts/impl/run-local-lifecycle.sh`, `scripts/impl/run-remote-lifecycle.sh` — the lifecycle suites that must still reach **enforced**
- `docs/en/operations.md:252` / `docs/ko/operations.md:247` — the "shared audit store" sections; `docs/en/configuration.md:479` / `docs/ko/configuration.md:455`; `docs/en/operations.md:625` — the operator-installed unit pattern
- `docs/rfcs/0001-registrar-role-and-non-self-propagation.md` §5.6, §6

sehkone:
## Summary

- Make `bootroot infra up` load the required agent configuration, render and verify the audit-store phases, and refuse the stack before Docker starts when filesystem enforcement is not active.
- Render the audit bind in long form with `create_host_path: false`, retain short-form reading, and atomically upgrade existing overrides without changing their source path.
- Update translated operator guidance, release notes, and the registrar E2E invocation for the guarded bring-up flow.

Closes #924. Part of #773. Part of #775.

## Test plan

- [x] `scripts/preflight/ci/check.sh` (formatting, Clippy, rustdoc, Python/Biome/Markdown checks, docs build, security audit, and Compose validation).
- [x] `scripts/preflight/ci/test-core.sh`.
- [x] `bash -n scripts/impl/run-registrar-internal-init-e2e.sh`.
- [ ] `scripts/preflight/ci/e2e-matrix.sh --skip-hosts` was attempted: its local and remote lifecycle runs passed, then reinit recovery stopped because `172.17.0.1` is not assigned to this Docker-host environment; the dedicated registrar-internal-init arm separately stopped at its required passwordless `sudo -n` prerequisite. CI's green Docker E2E jobs gate both privileged coverage paths.


sehkone:
**[Reviewer Round 1]**

Changes requested — two findings:

- P2: The rendered recovery command leaves `--agent-config` unquoted in [audit_store.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/src/commands/audit_store.rs:894). A valid config path containing spaces will not rerun; shell metacharacters also turn an operator-copyable root command into shell input. Quote the path with the existing POSIX-shell quoting approach and test a path with spaces/metacharacters.

- P2: The new filesystem-mode `infra up` must write/replace staged artifacts ([audit_store.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/src/commands/audit_store.rs:923)), which requires root once `init` has created the staging directory. The E2E was updated accordingly, but the manual presents the new command without that prerequisite and still calls `infra up` unprivileged ([operations.md](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/docs/en/operations.md:840), [operations.md](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/docs/en/operations.md:855)). Document the root-run requirement (and mirror it in Korean), or provide a clear early privilege refusal.

sehkone:

sehkone:
**[Author Round 1]**

Review dispositions:

- Fixed: quoted the rendered `--agent-config` recovery path with the existing POSIX-shell path quoter. Added coverage for spaces, `;`, `$()`, quotes, and `$HOME`.
- Fixed: documented the root requirement for filesystem-mode `infra up` in English and Korean operations guides, and corrected the mounted-reserve E2E documentation.

Pushed commit `091b0f5a` to [PR #948](https://github.com/aicers/bootroot/pull/948).

Validation: focused regression test, formatting, `check.sh`, `test-core.sh`, and docs build passed. The E2E matrix’s local/remote no-hosts lifecycles passed; it stopped at the known reinit host-address limitation (`172.17.0.1` unavailable).

sehkone:
**[Reviewer Round 2]**

Round 1 findings are resolved: the recovery path is safely shell-quoted, and the filesystem-mode root requirement is documented in both manuals.

Changes requested:

- P2: The unchanged localized error for a colon in `audit_store_dir` now gives the wrong reason. The new long-form Compose bind has a separate `source:` field, so Compose no longer splits that path on `:`; the code comment correctly explains that the restriction remains for the reader, systemd escaping, and rendered commands. Update both message catalogues to reflect that rationale while retaining the refusal. [en.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/src/i18n/en.rs:606) and [ko.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/src/i18n/ko.rs:606).

Verdict: not approved pending this documentation/message correction.

sehkone:

sehkone:
**[Author Round 2]**

Fixed — updated English and Korean audit-store path-refusal messages to explain why colons remain unsupported under the long Compose bind form. Updated adjacent test documentation accordingly.

Pushed commit `ac2b4853` to PR #948.

Validation: formatting, Clippy, rustdoc, docs build, full `cargo test` passed. Full preflight reached E2E coverage; its privileged matrix remains environment-gated as already documented in the PR.

sehkone:
**[Reviewer Round 3]**

**[Reviewer Round 3]**

Round 2’s localized colon-path rationale is correctly updated in both catalogues; no prior findings remain open.

Changes requested:

- P2: The required end-to-end `infra up` refusal against an unmounted filesystem store is still not tested. [run-registrar-internal-init-e2e.sh:741](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/scripts/impl/run-registrar-internal-init-e2e.sh:741) tests `create_host_path: false` by invoking Compose directly, while the Rust test only exercises `prepare_audit_store_for_infra_up` without a Docker call. Neither would fail if the audit-store preparation regressed to run after Compose had started OpenBao. Add an E2E case that invokes `bootroot infra up --agent-config ...` on the deliberately unmounted store and asserts its nonzero `provisioned, not activated` refusal, with no OpenBao container running.

Verdict: **NOT_APPROVED** pending that coverage.

sehkone:

sehkone:
**[Author Round 3]**

Fixed — added a registrar E2E case that runs `bootroot infra up --agent-config` against the deliberately unmounted filesystem store. It requires the `provisioned, not activated` refusal and verifies neither an OpenBao container nor `<store>/openbao` was created.

Pushed `a5aab33f` to PR #948.

Validation: syntax check and preflight quality checks passed. The focused E2E is locally blocked by unavailable Docker and passwordless `sudo`; CI will run it.

sehkone:
**[Reviewer Round 4]**

**[Reviewer Round 4]**

Approved.

The prior P2 is resolved: the new E2E invokes `bootroot infra up --agent-config` against the deliberately unmounted filesystem store, requires the `provisioned, not activated` refusal, and verifies that neither the OpenBao container nor the absent bind source was created ([run-registrar-internal-init-e2e.sh](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/scripts/impl/run-registrar-internal-init-e2e.sh:738)).

The updated implementation continues to gate before Compose pull/up ([infra.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-924-review/src/commands/infra.rs:169)), retains the documented legacy-bind residual, and the PR linkage/test plan are appropriate. No unresolved or new findings.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Refuse unmounted audit store bring-up
```

**Body**

```text
Prevent `infra up` from starting OpenBao before the filesystem-backed
audit reserve has been verified, and make failed mounts fail visibly at
boot instead of allowing Docker to create an unprotected audit path.

Closes #924
Part of #773
Part of #775
```

sehkone:
Rebased onto main commit 529ef04e (Refuse registrar without audit mount). Resolved the English and Korean operations-guide conflicts by retaining the branch’s Compose bind behavior and main’s daemon-side unmounted-store refusal.

Issue #952, opened 2026-08-28
Clarify retained audit-store exposure


sehkone:
# Clarify retained audit-store exposure

## Context

In `filesystem` enforcement, a store that holds content on the underlying filesystem is intentionally reported as `provisioned, not activated`; bootroot renders the three reserve artifacts but renders no phase-2 command and does not move, delete, mount over, or otherwise alter the records. `bootroot infra up` now raises that shared outcome before starting Compose.

The generated OpenBao audit bind uses `create_host_path: false`, which prevents Docker from creating an absent source. It does not protect an underlying `<audit_store_dir>/openbao` that already exists: on a failed mount boot Docker binds that directory and the audit device still writes to the root filesystem until the records are relocated to the reserve.

The current non-empty-store finding names the refusal and the two ways forward, but does not state this retained boot-path boundary. The existing-records passages in both manuals also introduce the state through `bootroot init` and do not explicitly say that `bootroot infra up` refuses it too. This issue corrects those operator-facing omissions without changing enforcement behavior or adding a relocation procedure.

## Scope

Update the existing non-empty-store outcome finding in both message catalogues and the existing-records passages in `docs/en/operations.md` and `docs/ko/operations.md`. The shared outcome must make clear that, on the live `bootroot infra up` surface, the store is refused before the stack starts; a pre-existing underlying `openbao/` remains a bind source because Compose's guard governs creation only; and a failed-mount boot therefore remains exposed until records are relocated onto the reserve. It must continue to state that bootroot does not move, delete, or mount over the records and that `audit_store_enforcement = "directory"` is the supported opt-out.

Keep using the existing `ReserveReport::NotActivated` outcome and `audit_reserve_finding_store_not_empty` finding. Add focused coverage for the rendered outcome in English and Korean, and for the two manual passages, so later wording changes cannot silently remove the boundary.

## Acceptance criteria

- [x] For an underlying non-empty filesystem-mode store, the outcome emitted through `bootroot infra up` explicitly says that bring-up is refused and no container starts.
- [x] That outcome explicitly says `create_host_path: false` governs creation only: an existing underlying `<audit_store_dir>/openbao` is still bound on a failed-mount boot and keeps the audit device on the root filesystem until relocation onto the reserve retires the exposure.
- [x] The English and Korean outcome strings retain the existing no-move/no-delete/no-mount-over guarantee, the separate-relocation statement, and the `audit_store_enforcement = "directory"` opt-out.
[...]

PR #955, opened 2026-08-28
Strengthen relocation-procedure regression coverage (#954)

### LINKED ISSUE #954: Strengthen relocation-procedure regression coverage
# Strengthen relocation-procedure regression coverage

## Context

The existing-records passages in `docs/en/operations.md` and `docs/ko/operations.md` correctly say that a non-empty underlying audit store is refused and that relocation is separate work. The regression test `manual_existing_records_passages_name_the_live_refusal_without_a_procedure` (`src/commands/audit_store/reserve.rs:3748`) scopes its inspection to those passages, requires the live `bootroot infra up` and retained-bind wording, rejects a fenced block, and rejects six backticked command prefixes: `` `mv ``, `` `cp ``, `` `rsync ``, `` `rm ``, `` `mount `` and `` `umount ``.

That list is short of the prohibition it guards. #924 names the forbidden shapes explicitly — "no `mv` of the store, no `cp`, no `find`, no `cmp`, no `diff`, no `rm` and no reference to a `.pre-mount` path" — and says none of them may appear "in a rendered command, in a message, or in either manual". Four of those are absent from the test's list: `` `find ``, `` `cmp ``, `` `diff `` and `.pre-mount`. A passage that added a tree comparison or named the move-aside path would pass a test written to stop exactly that.

**What this issue deliberately does not attempt.** An earlier framing of this gap asked the test to reject inline rename or replacement instructions written as prose, including the semantically equivalent Korean. That is not implementable here and is not attempted. The test is a `contains` check over a passage, so a prose ban is a list of forbidden substrings: it is bypassed by any paraphrase not on the list, and it collides with the assertions this same test requires — the passage **must** say "until the records are relocated onto the reserve" and "레코드를 예약량으로 옮길 때까지 이 노출은 남습니다", so relocation vocabulary is mandatory in the very text a prose ban would scan. Distinguishing "relocation is separate work" from "relocate it yourself" is a judgement, not a substring match.

The machine-checkable surface for that judgement already exists and is elsewhere: #924 requires the no-relocation property to be asserted "over the rendered command list rather than by eye", and those assertions are in place. The manual guard is the second belt, and its job is to catch the concrete and the obvious — a fenced procedure, a backticked command, a named path. This issue completes that list and nothing more.

## Scope

Add the four missing literals to the existing test's rejected set: the backticked command prefixes `` `find ``, `` `cmp `` and `` `diff ``, and the bare token `.pre-mount`, which is a path fragment rather than a command and is rejected wherever it appears in the passage.

Keep the test bounded to the two existing-records passages, keep its fenced-block rejection, keep the six prefixes it already rejects, and keep every positive assertion it already makes for both locales: the live `bootroot infra up` refusal, the creation-only bind boundary, the root-filesystem exposure until relocation, and the statement that relocation is separate work.

## Acceptance criteria

- [x] The test rejects `` `find ``, `` `cmp `` and `` `diff `` in either existing-records passage, on the same terms as the six prefixes it already rejects.
- [x] The test rejects `.pre-mount` anywhere in either passage, backticked or not.
- [x] The six existing prefixes, the fenced-block rejection, and the passage scoping are unchanged.
- [x] Every positive assertion the test already makes for both locales is unchanged, and both manual passages still pass without edits.
- [x] The test makes no attempt to match prose intent, and no forbidden substring is added that occurs in the wording the test itself requires.
- [x] No operator-facing manual wording changes.

## Constraints

- Do not add a relocation procedure, command, configuration key, outcome, error class, or enforcement mode.
- Do not change `ReserveReport::NotActivated`, audit-store enforcement, the Compose bind or override reader/upgrader, Docker behavior, registrar behavior, or the manual section structure.
- Do not widen the test's scope beyond the two existing-records passages.
- Do not add a natural-language pattern, in either locale, for rename or replacement intent.
- Keep English and Korean coverage semantically aligned.

## Out of scope

- Moving, copying, renaming, deleting, mounting, unmounting, or replacing audit-store records or images in production. Migration and replacement procedures remain separate work in #926.
- Any change to the live `infra up` outcome text or to the two manual passages' current operator guidance.
- The rendered-command assertions, which already carry the no-relocation property and are not this issue's.

## Test plan

- [x] Run the focused Rust test containing `manual_existing_records_passages_name_the_live_refusal_without_a_procedure` and confirm it passes against the manuals as they stand.
- [x] For each of the four added literals, mutate a copy of each locale's passage to contain it and confirm the test fails, so the additions are proven to bite rather than assumed to.
- [x] Confirm none of the four occurs in the strings the test requires, so no addition can make a correct passage fail.
- [x] Run `cargo fmt -- --config group_imports=StdExternalCrate --check` and `cargo clippy --all-targets -- -D warnings`.
- [x] Run the repository CI checks required for the changed test source.

## Dependencies

Depends on no unmerged issue. Part of #952. Part of #773.

## Pointers

- `src/commands/audit_store/reserve.rs:3748` — the existing manual-passage regression test.
- `src/commands/audit_store/reserve.rs:3795` — its fenced-block check.
- `src/commands/audit_store/reserve.rs:3801` — the six-prefix list this issue extends.
- `docs/en/operations.md:1081` and `docs/ko/operations.md:1021` — the two existing-records passages the test must continue to bound.

sehkone:
## Summary

- Extend the existing existing-records manual guard with `find`, `cmp`, and `diff` command prefixes.
- Reject the `.pre-mount` path fragment in either localized passage while retaining all existing positive checks and scope boundaries.

Closes #954. Part of #952. Part of #773.

## Test plan

- [x] Run `cargo test manual_existing_records_passages_name_the_live_refusal_without_a_procedure`.
- [x] Run `cargo fmt -- --config group_imports=StdExternalCrate --check` and `cargo clippy --all-targets -- -D warnings`.
- [x] Run `scripts/preflight/ci/check.sh`.
- [x] Run `scripts/preflight/ci/test-core.sh`.
- [x] Mutate a copy of each locale passage with every new literal and confirm the focused regression fails.
- [x] Confirm the current English and Korean passages contain none of the four newly forbidden literals.

### DIFF
--- a/src/commands/audit_store/reserve.rs
+++ b/src/commands/audit_store/reserve.rs
@@ -3796,12 +3796,19 @@ mod tests {
                 !passage.contains("\n```"),
                 "{manual}: the existing-records passage must not add a relocation procedure"
             );
-            for command in ["`mv ", "`cp ", "`rsync ", "`rm ", "`mount ", "`umount "] {
+            for command in [
+                "`mv ", "`cp ", "`rsync ", "`rm ", "`mount ", "`umount ", "`find ", "`cmp ",
+                "`diff ",
+            ] {
                 assert!(
                     !passage.contains(command),
                     "{manual}: the existing-records passage must not add a relocation command: {command}"
                 );
             }
+            assert!(
+                !passage.contains(".pre-mount"),
+                "{manual}: the existing-records passage must not name the relocation path .pre-mount"
+            );
         }
     }
 


sehkone:
**[Reviewer Round 1]**

Approved — no findings.

The PR precisely extends the existing bounded manual guard in [reserve.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-954-review/src/commands/audit_store/reserve.rs:3799) with the three required inline command prefixes and rejects `.pre-mount` anywhere in either scoped passage. It leaves the positive locale assertions, heading-based passage boundaries, existing six checks, and fenced-block prohibition intact.

The PR body correctly uses `Closes #954`, retains only the permitted inherited `Part of #952` and `Part of #773` references, and includes a completed test plan.

sehkone:

Issue #926, opened 2026-08-26
Move audit records onto the reserve, at activation and when the reserve changes


AcoPiper:
# Move audit records onto the reserve, at activation and when the reserve changes

## Context

The registrar's two audit artifacts share one directory on the bootroot host. `[registrar] audit_store_dir` (default `/var/lib/bootroot/audit-store`) is the only definition of where that store is; the daemon's verb records go to `<audit_store_dir>/records` and the OpenBao file audit device writes `<audit_store_dir>/openbao/audit.log`, bound into the container through a rendered Compose override.

Separate work makes `audit_store_reserve_bytes` a kernel-enforced ceiling in the default `filesystem` mode: a fully allocated loopback image at `<audit_store_dir>.img`, an ext4 filesystem made with `mkfs.ext4 -m 0 -E nodiscard,lazy_itable_init=0`, a generated systemd mount unit that mounts it at `audit_store_dir` and is restored on boot, and an audit bind declared `create_host_path: false` so Docker refuses to manufacture the bind source when the mount is not there. That work runs in three phases — bootroot derives, evaluates, preflights and renders in phase 1; the operator performs every step that changes the host in phase 2; bootroot verifies with metadata reads and reports in phase 3 — and reports exactly one outcome per run: **enforced**, **provisioned, not activated**, **failed**, or **`unenforced (directory)`** for an explicitly configured `directory` deployment. Two properties of it are the reason this issue exists:

- **The mount hides whatever is underneath it.** A store that already holds records keeps them — intact, but invisible and outside the reserve — the moment a filesystem is mounted over it. So the underlying directory must be empty before activation.
- **One writer is held out by the deployment; the other is held out only by being stopped.** The audit bind is declared `create_host_path: false`, so from the aside-rename onward — when `<audit_store_dir>/openbao` no longer exists — the OpenBao container fails to start rather than writing into whatever the path resolves to. The daemon's verb records are the other half, and nothing enforces them once a mount exists: a daemon-side gate refuses the verbs only while `audit_store_dir` is not a mount point, which stops holding the moment the mount is established part-way through. So the migration runs with **both** writers down throughout, and for the records that is the only mechanism — procedural, not a property of the mount.

[...]

PR #961, opened 2026-08-29
Verify a source root lost+found on both sides (#959)

### LINKED ISSUE #959: Verify root lost+found metadata during migration
# Verify root lost+found metadata during migration

## Context

The audit-store migration copies a holding directory into the root of a newly formatted ext4 reserve. ext4 creates `lost+found` at that destination root, while the holding directory normally has no corresponding entry. The current renderer excludes a directory named `./lost+found` from both metadata manifests so that the normal migration can complete. That symmetric exception also omits type, mode, uid, gid, and link-count verification when the holding directory itself legitimately contains a root `lost+found` directory. Its contents are still compared, but a metadata mismatch on that directory can pass the verification gate.

## Scope

Change the migration metadata verification so that a destination-only ext4-created root `lost+found` does not prevent a faithful migration from closing, while a directory with that name at the holding root is verified against its copied destination counterpart. Preserve the existing complete verification contract for all other paths and keep regular-file content and size checks covering descendants.

Update the English and Korean operator documentation only as needed to describe the exact rendered verification behavior without creating a second copy or verification contract.

## Acceptance criteria

- [x] A newly formatted ext4 destination whose holding source lacks a root `lost+found` verifies successfully without treating the destination's seeded directory as audit data.
- [x] When the holding source contains a root directory named `lost+found`, a mismatch of its type, mode, numeric uid, numeric gid, or link count makes the metadata comparison fail and prevents the closing rename.
- [x] Contents beneath a source root `lost+found` remain included in the existing metadata, regular-file-size, and content comparisons.
- [x] A non-directory entry named `lost+found` remains covered by the ordinary manifest rules.
- [x] The rendered commands remain NUL-delimited, staged, pipeline-free, and byte-compared with `cmp`; no recursive delete is introduced.
- [x] English and Korean audit-store documentation remain semantically aligned with the final behavior.

## Constraints

- Do not simply exempt `./lost+found` symmetrically from metadata verification.
- Do not weaken verification for any other directory or regular file, and do not compare directory sizes or modification times.
- Do not follow symlinks, change the migration-set refusal rules, or alter the type guards.
- Do not change the copy form, reserve capacity gate, holding-directory lifecycle, or image-replacement procedure.
- bootroot must continue to render and verify only; it must not copy, rename, delete, mount, or unmount host data.
- Keep the procedure free of pipelines and recursive deletion.

## Out of scope

- Rendering additional recovery or rollback steps for collision states.
- Changing ext4 provisioning, the reserve image format, or the audit-store directory layout.
- Adding a separate verification procedure for reserve replacement.

## Test plan

- [x] Extend migration command tests to cover a source without `lost+found`, a source-root `lost+found` directory with matching metadata, and each compared metadata mismatch.
- [x] Confirm descendants beneath that directory still participate in all applicable comparisons and a same-named regular file is not exempt.
- [x] Run the focused migration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --config group_imports=StdExternalCrate --check`, and `./scripts/check-docs.sh`.

## Dependencies

Part of #926. This follow-up has no follow-up issue dependency and is filed under umbrella #773.

## Pointers

- `src/commands/audit_store/migration.rs:95` — the current root-directory exclusion constant.
- `src/commands/audit_store/migration.rs:598` and `src/commands/audit_store/migration.rs:612` — the source and destination metadata manifest rendering.
- `src/commands/audit_store/migration.rs:990` and `src/commands/audit_store/migration.rs:1009` — existing coverage for descendants and a non-directory with this name.
- `docs/en/operations.md:1450` and `docs/ko/operations.md:1370` — current operator documentation for the exception.

sehkone:
## What this changes

The migration metadata manifests dropped `./lost+found` from **both** sides. That exemption exists because `mke2fs` seeds a `lost+found` directory in the root of every filesystem it makes, and that root is the *destination* of the relocation copy — without it, the destination carries one record the holding directory never had and no correct migration could ever verify.

Applied symmetrically, it also silenced the source side. A holding directory that legitimately holds a root `lost+found` — the whole source tree of the replacement procedure, whose source is itself a mounted reserve — had that directory's type, mode, numeric uid, numeric gid and link count compared against nothing at all, and a mismatch on it passed the gate that guards the closing rename.

The exemption now belongs to the destination side alone, and even there it is conditional. The rendered shell asks, at the moment it runs, whether the source root carries a `lost+found` directory:

```sh
if [ -d '<holding>/lost+found' ] && [ ! -L '<holding>/lost+found' ]; then find . -xdev -mindepth 1 -printf '%y %m %U %G %n %P\0' > "$SCRATCH"/meta.destination.raw; else find . -xdev -mindepth 1 ! \( -path './lost+found' -type d \) -printf '%y %m %U %G %n %P\0' > "$SCRATCH"/meta.destination.raw; fi
```

The source walk carries no filter at all — a record the copy did not put there cannot arise on the side the copy came *from*.

The predicate is `-d` **and not** `-L` rather than `-d` alone: `test -d` answers for a symbolic link's target, and nothing in this procedure follows one. `find -type d` does not either, so the two spellings agree on every entry.

Everything else stands: the walks are NUL-delimited, each stage is its own command over its own staged file, the three manifests are byte-compared with `cmp`, and no recursive delete is introduced. The branch is a plain `if`, not a pipeline, so a failed walk inside it still stops the sequence — `set -e` fires on the branch body, and there is a test for exactly that, driven down *both* branches.

Also here: a CodeQL `rust/cleartext-logging` alert. The ownership case labelled its assertion with the uid and gid it had just set, and the query reads any value bound to `uid` reaching a message as a leak — a red required check over two loop constants in a Linux-only test module. The message names the manifest field the case moves instead, which is what a failure needed to distinguish anyway.

## Acceptance criteria

- **A newly formatted ext4 destination whose holding source lacks a root `lost+found` verifies successfully** — the `else` branch, covered by `the_reserves_own_lost_and_found_does_not_fail_the_verification` and by the `registrar-internal-init` E2E scenario, which now also asserts the conditional is rendered and that the branch taken really was the exempting one.
- **A source-root `lost+found` mismatch fails the comparison** — `a_mismatched_source_root_lost_and_found_fails_the_verification` breaks type, mode and link count in turn over a copy that had already passed; `a_reowned_source_root_lost_and_found_fails_the_verification` does uid and gid where the run has the privilege to `chown`, and the rendered `%y %m %U %G %n %P` form is asserted unconditionally beside it.
- **Contents beneath a source root `lost+found` stay in every comparison** — `everything_under_lost_and_found_is_still_compared` now breaks the metadata, size and content passes separately, each over a copy that had already passed, rather than only the same-sized content case.
- **A non-directory named `lost+found` stays under the ordinary rules** — `a_regular_file_merely_bearing_the_name_is_not_exempt` at the source root, and `a_non_directory_bearing_the_name_at_the_destination_root_is_not_exempt` at the destination root, which is where the exemption is actually spelled and so where `-type d` has to be the thing doing the work. `a_symlink_bearing_the_name_at_the_source_root_is_refused_by_the_guard` covers the third shape the name can take: a link to a real directory, which `test -d` alone would read as a directory, and which the type guard refuses before the question is ever asked.
- **The rendered commands stay NUL-delimited, staged, pipeline-free and `cmp`-compared** — `the_rendered_migration_commands_are_pinned_exactly` pins the whole sequence verbatim and re-checks the pipe, `diff -r`, `rm -r` and trailing-field bans.
- **Both manuals stay aligned** — `both_manuals_pin_the_asymmetric_lost_and_found_exemption` asserts each page's relocation block carries the exclusion exactly *once* and carries the conditional, so a symmetric form cannot come back in either language.

## Test plan

The `rendered_sequence` module is Linux-only (GNU `find -printf`, `sort -z`, `sha256sum --zero`), so it was run in a `rust:latest` container; everything else was run on macOS as well.

- [x] A source with no root `lost+found` verifies, and the destination's seeded directory is not treated as audit data
- [x] A source-root `lost+found` whose type, mode, numeric uid, numeric gid or link count differs fails the metadata comparison and stops before the closing rename
- [x] Descendants beneath a source root `lost+found` still participate in the metadata, regular-file-size and content comparisons — each broken separately
- [x] A regular file and a symlink bearing the name are not exempt, at the source root and at the destination root
- [x] `cargo test --bin bootroot audit_store::migration` — 27 passed, in the container both as **root** (so the uid/gid case ran) and as an **unprivileged user** (so the fail-closed-inside-the-conditional case ran, down both of its branches)
- [x] `cargo test` — green on macOS, and green in the Linux container (lib 1269, bin 1303, plus the other bins and integration suites). The 8 `tests/bootroot_rotate.rs` cases that fail in that container fail on a missing `docker` CLI and pass locally on macOS where Docker is present: `46 passed; 0 failed`. The container's own `umask` matters: at the image default of `002` the `registrar::audit::scan` and `registrar::openbao_audit` suites refuse their own fixtures as group-writable, which is the safe-path check working rather than a failure of this branch
- [x] `cargo clippy --all-targets -- -D warnings` — clean on macOS and Linux
- [x] `cargo fmt -- --config group_imports=StdExternalCrate --check` — clean on macOS and Linux
- [x] `./scripts/check-docs.sh` — `mkdocs build --strict` passes and the theme verification passes
- [x] `markdownlint-cli2` over the repository — 0 issues
- [x] `shellcheck -x scripts/impl/run-registrar-internal-init-e2e.sh` — clean (only `SC1091` info for sourced libs, resolved when run from the script's directory)
- [x] All CI checks green on the pushed branch, including `Docker E2E (registrar-internal-init)`

## Local run notes

- **`scripts/preflight/ci/e2e-matrix.sh` did not run here.** Its step 13 — `run-registrar-internal-init-e2e.sh`, the one arm that exercises this rendering end to end — requires passwordless `sudo`, and `sudo -n true` on this machine reports "a password is required"; the matrix's own usage text states `--skip-hosts` does not stand in for it. Port 8200 is also held by an unrelated live process from another session, which the earlier steps' port preflight refuses to start over, and tearing that down is not mine to do. CI's `Docker E2E` jobs gate that arm, and `Docker E2E (registrar-internal-init)` is green. Nothing was weakened, skipped, marked `continue-on-error` or deleted to get a local run to pass. What did run locally is the `rendered_sequence` suite, which executes the rendered shell for real against fixture trees on Linux, including every `lost+found` shape on both sides.
- **No `CHANGELOG.md` entry.** The audit-store reserve, its relocation procedure and its verification are all announced under `[Unreleased]`; no user of the last release ever saw the symmetric exemption, so this is a rework of unreleased work rather than an observable change.
- A directory's `%n` is two plus its subdirectory count, so no fixture can move that field alone without also changing the entries the manifest lists underneath. The link-count case therefore faults on both, which still demonstrates the field is compared rather than skipped; it is not an isolated one-field test and the code comment says so.

Closes #959

Part of #926

### DIFF
--- a/docs/en/operations.md
+++ b/docs/en/operations.md
@@ -1336,8 +1336,8 @@ records:
      ! -type d ! -type f -exec false {} +
    find <audit_store_dir> -xdev -mindepth 1 ! -type d ! -type f -exec false {} +
    cd <audit_store_dir>.pre-mount
-   find . -xdev -mindepth 1 ! \( -path './lost+found' -type d \) \
-     -printf '%y %m %U %G %n %P\0' > "$SCRATCH"/meta.source.raw
+   find . -xdev -mindepth 1 -printf '%y %m %U %G %n %P\0' \
+     > "$SCRATCH"/meta.source.raw
    sort -z -o "$SCRATCH"/meta.source "$SCRATCH"/meta.source.raw
    find . -xdev -mindepth 1 -type f -printf '%s %P\0' > "$SCRATCH"/size.source.raw
    sort -z -o "$SCRATCH"/size.source "$SCRATCH"/size.source.raw
@@ -1345,8 +1345,14 @@ records:
    sort -z -o "$SCRATCH"/files.source "$SCRATCH"/files.source.raw
    xargs -0 -r sha256sum --binary --zero < "$SCRATCH"/files.source > "$SCRATCH"/content.source
    cd <audit_store_dir>
-   find . -xdev -mindepth 1 ! \( -path './lost+found' -type d \) \
-     -printf '%y %m %U %G %n %P\0' > "$SCRATCH"/meta.destination.raw
+   if [ -d <audit_store_dir>.pre-mount/lost+found ] \
+     && [ ! -L <audit_store_dir>.pre-mount/lost+found ]; then
+     find . -xdev -mindepth 1 -printf '%y %m %U %G %n %P\0' \
+       > "$SCRATCH"/meta.destination.raw
+   else
+     find . -xdev -mindepth 1 ! \( -path './lost+found' -type d \) \
+       -printf '%y %m %U %G %n %P\0' > "$SCRATCH"/meta.destination.raw
+   fi
    sort -z -o "$SCRATCH"/meta.destination "$SCRATCH"/meta.destination.raw
    find . -xdev -mindepth 1 -type f -printf '%s %P\0' > "$SCRATCH"/size.destination.raw
    sort -z -o "$SCRATCH"/size.destination "$SCRATCH"/size.destination.raw
@@ -1447,13 +1453,25 @@ afterwards. The metadata pass carries the **hard-link count**, so a copy
 that expanded one source inode into two destination files fails even
 though every path matches on type, mode, ownership, size and content.
 Directory sizes and modification times are excluded, both differing
-legitimately between two filesystems. So is `lost+found`, and only where
-it is a directory at a tree's own root: `mkfs.ext4` creates it in every
-reserve it makes, so it stands on a side the copy did not put it on and
-would fault every correct migration. Only that one entry is exempt —
-anything underneath it is still compared under its `lost+found/…` path,
-so nothing is smuggled past the verification by being put there. A
-mismatch is decided by exit status, never by reading output.
+legitimately between two filesystems. `lost+found` is dropped from the
+**destination** manifest alone, and only where it is a directory at that
+tree's own root **and** the source root holds no directory of that name:
+`mkfs.ext4` creates it in every reserve it makes, so a source without one
+would otherwise be faulted for a record the copy did not put there. Which
+of the two destination walks runs is decided by the shell at the moment
+it runs them, from the source it is about to be compared against — so a
+source that *does* carry one, as the live store of
+[the replacement procedure](#changing-the-reserve) does, being itself a
+mounted reserve, has that directory's type, mode, numeric uid and gid and
+link count compared like any other directory's, and a mismatch on it
+stops the sequence before the closing rename. The test is spelled `-d`
+**and not** `-L` because `test -d` answers for a symbolic link's target
+and nothing in this procedure follows one, which is what `find -type d`
+also does. Only ever that one entry, and only on that one side — anything
+underneath it is still compared under its `lost+found/…` path, so nothing
+is smuggled past the verification by being put there, and a regular file
+merely bearing the name is compared in full. A mismatch is decided by exit
+status, never by reading output.
 
 **Resume and rollback.** Both are rendered, both are idempotent, and
 bootroot performs neither.
--- a/docs/ko/operations.md
+++ b/docs/ko/operations.md
@@ -1264,8 +1264,8 @@ reinit의 사전 점검은 이 표면의 1단계 거부를 모두 순수 읽기�
      ! -type d ! -type f -exec false {} +
    find <audit_store_dir> -xdev -mindepth 1 ! -type d ! -type f -exec false {} +
    cd <audit_store_dir>.pre-mount
-   find . -xdev -mindepth 1 ! \( -path './lost+found' -type d \) \
-     -printf '%y %m %U %G %n %P\0' > "$SCRATCH"/meta.source.raw
+   find . -xdev -mindepth 1 -printf '%y %m %U %G %n %P\0' \
+     > "$SCRATCH"/meta.source.raw
    sort -z -o "$SCRATCH"/meta.source "$SCRATCH"/meta.source.raw
    find . -xdev -mindepth 1 -type f -printf '%s %P\0' > "$SCRATCH"/size.source.raw
    sort -z -o "$SCRATCH"/size.source "$SCRATCH"/size.source.raw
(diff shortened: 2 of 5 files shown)


sehkone:
**[Reviewer Round 1]**

Review: no findings. I would approve PR #961.

The destination-only conditional correctly preserves ext4’s seeded `lost+found` exception while making a source-root directory participate in the full metadata manifest. The rendered sequence remains staged, NUL-delimited, pipeline-free, and fail-closed.

The tests meaningfully cover the new branches: source/destination metadata mismatches, descendants, same-named non-directories, symlink refusal, and both conditional paths. English and Korean documentation match the rendered behavior. PR linkage and test-plan hygiene are also correct.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Verify a source root lost+found on both sides
```

**Body**

```text
The migration metadata manifests dropped `./lost+found` from both
sides. That exemption exists because `mke2fs` seeds a `lost+found`
directory in the root of every filesystem it makes, and that root is
the destination of the relocation copy — without it the destination
carries one record the holding directory never had, and no correct
migration could ever verify.

Applied symmetrically, it also silenced the source side. A holding
directory that legitimately holds a root `lost+found` — the whole
source tree of the replacement procedure, whose source is itself a
mounted reserve — had that directory's type, mode, numeric uid,
numeric gid and link count compared against nothing at all, and a
mismatch on it passed the gate that guards the closing rename.

The exemption now belongs to the destination side alone, and even
there it is conditional: the rendered shell asks, at the moment it
runs, whether the source root carries a `lost+found` directory, and
records its counterpart when it does. The source walk carries no
filter at all — a record the copy did not put there cannot arise on
the side the copy came from.

The predicate is `-d` and not `-L` rather than `-d` alone, because
`test -d` answers for a symbolic link's target and nothing in this
procedure follows one. `find -type d` does not either, so the two
spellings agree on every entry.

Everything else stands: the walks are NUL-delimited, each stage is
its own command over its own staged file, the three manifests are
byte-compared with `cmp`, and no recursive delete is introduced. The
branch is a plain `if`, not a pipeline, so a failed walk inside it
still stops the sequence.

Closes #959

Part of #926
```

Issue #770, opened 2026-08-04
Renew the registrar surface's certificates, with a reload contract and a lapse surface


sehkone:
# Renew the registrar surface's certificates, with a reload contract and a lapse surface

## Summary

Minting the registrar surface's two leaves once is not enough. Both expire; a lapsed leaf stops **every** enrollment in the deployment, and the registrar cannot re-mint itself. This group keeps both leaves valid over time and makes a *failing* renewal visible before it becomes an outage. **It is a grouping issue and carries no work of its own.**

**The seam between the children is doing versus telling, and it is drawn where the wire is.** The first child renews: it adds both leaves to the daemon's existing certificate-renewal mechanism, preserves the name invariants a fresh key pair leaves behind, and delivers both halves of the reload contract — the server-side certificate swap and the caller-side torn-pair-safe per-dial load. It puts **nothing** on the wire and touches no serialization fixture. The second child reports and classifies: it populates the endpoint health container's `certificates` member, fixes a precedence between simultaneous conditions, turns an actual lapse into a typed non-retryable error on both leaves, classifies every post-connect failure by layer, and documents the remediation. Every wire-visible consequence in this group is the second child's, and that line is checkable in review rather than a matter of taste.

**Only one of those two reporting shapes is on the wire, and conflating them would send this group looking for an error identifier it must not add.** The health member rides ordinary responses, so it is serialized and fixture-pinned. The lapse *error* is not: a lapsed client leaf is caught before dialing and a lapsed server leaf fails the handshake, and neither produces a response to carry an identifier — so it is a **local typed error in this repository's client**, asserted by variant, and the translation into the ecosystem's permanent unavailability reason happens in the co-located registrar's repository out of what it observes. Neither child adds a wire identifier or a reason.

They are strictly ordered — reporting depends on renewal — and the whole interface between them is one in-process accessor, described below.

## Children

- [x] Renew both registrar leaves on the daemon's existing loop and honour the reload contract — Surface
- [ ] Report the registrar certificates' lifetime and turn a lapse into a typed non-retryable error — Surface

## Shared background

[...]

PR #962, opened 2026-08-29
Renew registrar leaves and honour the reload contract (#768)

### LINKED ISSUE #768: Renew registrar leaves and honour the reload contract
# Renew registrar leaves and honour the reload contract

## Context

The registrar surface uses two expiring certificate/key pairs: a registrar client leaf that the caller presents and an endpoint server leaf that callers pin. The daemon already performs start-time issuance before endpoint activation, has a profile-renewal loop with an immediate first pass, configured cadence and shutdown ownership, and exposes server-swap and per-dial client-load seams. It does not yet put these leaves on that lifecycle.

This issue adds one daemon-owned renewal adapter for both leaves, their in-process renewal state, safe server/client reloads, and the caller reload documentation. The open reporting sibling owns wire-visible health reporting and typed lapse classification; this issue supplies its in-process input but puts nothing on the wire.

## Scope

Create the registrar-certificate renewal adapter in the daemon lifecycle. It is created only when the endpoint is enabled, owns both leaves, receives the daemon shutdown signal, and its retained task handle joins with the daemon's other tasks. It runs immediately, then uses the rendered internal credential agent configuration's sole profile: its daemon check interval and jitter set cadence, `renew_before` is lead time, and its retry configuration governs issuance retries.

Initialize the shared per-leaf renewal-state accessor after the existing start-time issuance has ensured usable material and before endpoint serving begins. Read each enabled leaf's valid certificate and record notAfter, `never_attempted`, and no attempt time. This is not an attempt and does not contact OpenBao or the CA.

On each tick, renew a leaf inside lead time or whose chain no longer verifies against `trust.ca_bundle_path`; lead time is checked first. The shared eligibility predicate retains its no-bundle rule: with `ca_bundle_path` unconfigured, chain drift does not trigger renewal. That state cannot activate an enabled endpoint: `build_server_config` requires both `trust.ca_bundle_path` and non-empty `[trust].trusted_ca_sha256`, with no system-root or pin-file fallback for incoming mTLS. Preserve and test the predicate below activation; do not create a verifier fallback. Use the existing daemon eligibility and issuance-retry policy.

Refactor registrar issuance for renewal so it returns the chain, candidate certificate, and fresh key without writing live output—not the configured pair and not `trust.ca_bundle_path`. Candidate files use the finished files' required permissions; a private rollback snapshot is written no wider than the key mode whatever the file it copies, and records that file's own mode so a restore puts it back. All of them are removed on every exit after creation, including issuance, validation, publication, rollback, and TLS-swap failure. Validate every candidate before a live write: its key matches the leaf; a client candidate has the reserved registrar SAN and recognises to the same instance/host/domain; a server candidate has the exact endpoint SAN and chains to an anchor in the existing endpoint pin file. Start-time issuance may use the new helper, but renewal must not call the present direct-to-live `issue_surface_pair` path.

The endpoint pin file and the incoming client verifier have distinct roles. The pin file decides only whether a replacement server leaf is safe for pinned callers; it never supplies incoming mTLS roots. Rebuild `WebPkiClientVerifier` from the would-be post-merge bytes of `trust.ca_bundle_path`, retaining exactly PEM certificates whose DER SHA-256 is in `[trust].trusted_ca_sha256`. Do not use the returned chain alone, the endpoint pin file, all bundle entries, or system roots. Before live publication, build a complete next `ServerConfig` from staged inputs: the candidate server pair when renewing that leaf, the current live server pair when renewing only the client leaf, and the staged merged bundle in either case. This build must finish before any live path changes.

Publish a validated renewal as a recoverable transaction. Before the first live write, capture restore artifacts for the affected live bundle and pair, including bytes, final modes, and ownership. Atomically replace the merged bundle, write the live certificate then key through the established two-rename contract, and only then atomically exchange the already-built active `TlsAcceptor`/configuration. The accept loop loads the active acceptor immediately before each handshake and holds no lock across an await. Replacing `EndpointCertResolver` alone is insufficient because the old acceptor retains the old client verifier. Existing handshakes and connections keep their old configuration and are not dropped; the socket is not rebound.

The active configuration changes only after every live file write succeeds; its exchange is infallible because the full replacement was built first. If a bundle, certificate, or key write fails, do not exchange it; restore every live path whose publication may have started from the saved artifacts, leave the old active configuration installed, record a failed attempt with the publication and, if applicable, rollback error, and retry on the next tick. A successful rollback restores prior bytes, modes, and ownership. If rollback itself fails, do not conceal the mixed disk state or claim it was restored: keep the old active configuration, record both errors, clean all private artifacts, and let a later tick retry. The per-dial loader remains responsible for never presenting a mismatched pair. An invalid or unpinned server candidate is a refusal: discard it, record the pin-file reason, and change no live or active state.

Each successful issuance uses a fresh key. Never rewrite the endpoint pin file or add a leaf fingerprint. During CA rotation the tick chooses no ordering and does not repair the bootroot-internal credential. The rotation's mandatory Phase-4 tail repairs that credential under additive transitional trust before Phase 5 reissues services. If the tick reaches a registrar leaf while the credential cannot authenticate, record an ordinary failed issuance and retry next tick. On chain drift the client leaf is reissued; reissue the server leaf only when its candidate anchor is already in the unchanged pin file.

Update the accessor only when initialization or an attempt supplies state. A success replaces notAfter and records `succeeded` with attempt time. An issuance, publication, rollback, authentication, or pin-file-refusal failure retains notAfter and records `failed` with attempt time and reason. A no-op tick changes nothing. Enabled endpoints have entries for both leaves.

Honour the client reload contract through the completed in-repository per-dial loader. Each dial rereads its pair and verifies key-to-leaf matching. The writer uses two renames, so a reader in the interval retries rather than presenting a mismatch: five reads total, waits of 1 ms, 2 ms, 4 ms, and 8 ms after the first four mismatches, then a typed mismatch error. This fixed local policy is not the daemon issuance-retry policy. Document this contract in the existing English and Korean endpoint documentation, naming the in-repository client as reference behaviour.

## Acceptance criteria

- [x] With the endpoint enabled, one adapter registers both leaves, initializes both accessor entries from valid on-disk certificates before serving, and joins on shutdown. With it disabled, no adapter, state entries, renewal work, OpenBao request, or CA request exists.
- [x] The first pass is immediate; later passes use the internal agent configuration's sole profile cadence, jitter, lead time, and issuance retry settings. Tests can drive a pass without wall-clock waiting.
- [x] Initialization records observed notAfter, `never_attempted`, and no timestamp without an issuance request; one driven pass cannot read role_id/secret_id or use AppRole.
- [x] A due leaf is issued off-live with a fresh key. Candidate key/SAN/pin validation, staged CA-bundle construction, next TLS configuration construction, and restore snapshots precede every live write; candidates and snapshots are cleaned on every exit.
- [x] A client candidate preserves its recognised instance, host, and domain. A server candidate has the exact endpoint SAN and chains to an existing endpoint-pin-file anchor.
- [x] With a configured bundle, an in-date leaf outside lead time whose chain no longer reaches it is eligible through the shared drift check; without it, drift does not trigger renewal and the enabled endpoint refuses activation rather than falling back to other roots.
- [x] For an unpinned server candidate, the live bundle, pair, active TLS configuration, and pin file remain byte-for-byte unchanged; the accessor records failed with the pin-file reason and attempt time.
- [x] The rebuilt incoming verifier uses only staged post-merge `trust.ca_bundle_path` certificates whose DER fingerprints occur in `trusted_ca_sha256`; pin-file anchors, unpinned bundle certificates, a returned chain alone, and system roots cannot substitute.
- [x] Successful publication replaces the merged bundle, writes certificate then key, and only then swaps the prebuilt active TLS configuration. The next handshake presents the renewed server leaf and accepts the renewed client leaf without restart, signal, socket rebind, socket-path replacement, or dropped in-flight connections.
- [x] An injected bundle, certificate, or key publication failure retains the old active configuration, restores every changed live path from saved bytes/mode/ownership, records failure, and cleans private artifacts. An injected rollback failure retains the old active configuration, records both errors, cleans artifacts, and does not claim the live files were restored.
- [x] A successful attempt updates notAfter and `succeeded`; every ordinary issuance, publication, rollback, authentication, or pin-file refusal failure retains notAfter and records `failed`; a no-op pass is unchanged.
- [x] A reader forced between the two renames retries until it finds a matching pair and never presents a mismatch. A permanent mismatch performs exactly five reads, follows the 1/2/4/8 ms waits, and returns the typed mismatch error.
- [x] No response field, serialization fixture, wire-visible renewal state, or typed lapse classification is added. English and Korean endpoint documentation state the per-dial reload contract.
- [x] Changed Rust code passes the repository-required formatting and warning-free clippy invocations.

## Constraints

- Do not create a second daemon, supervisor, scheduler, lead-time policy, or daemon issuance-retry policy.
- Do not use AppRole, role_id, secret_id, or the per-service ACME renewal path.
- Do not publish renewal output before candidate key/SAN/server-anchor, staged-bundle, and next-TLS-configuration validation completes.
- Do not make the pin file, returned chain alone, all bundle entries, or system roots an incoming client-verifier source.
- Do not rewrite the endpoint pin file, add a leaf pin, swap active TLS after a publication failure, or describe a failed rollback as restored.
- Do not restart the daemon, rebind or unlink its socket, require an external signal, or drop in-flight connections for a TLS replacement.
- Do not repair certificates at start, repair the internal credential, add request-path recovery, or impose rotation ordering from this tick.
- Keep tests deterministic: inject/control ticks, publication failures, rollback failures, and loader delays.

## Out of scope

- Initial issuance, start-up repair, certificate naming, and CSR construction.
- The endpoint pin-file format and its provisioning tool.
- Renewal or recovery of the bootroot-internal privileged credential.
- Health-response schema, lifetime reporting, lapse precedence, and caller-visible lapse errors.

## Test plan

- [x] Test enabled/disabled adapter creation, initialization, shutdown, and absence of disabled-path OpenBao/CA activity.
- [x] Test immediate and scheduled ticks with controllable time and shutdown; assert no-AppRole access.
- [x] Drive lead-time renewal and assert candidate validation, staged verifier build, and snapshot creation before publication; assert fresh key, preserved client identity, valid server SAN, unchanged pins, and accessor success.
- [x] Drive drift: verify client replacement, server publication with a pinned incoming anchor, and unpinned-server refusal with unchanged live/active state and retained notAfter. Verify no-bundle drift opt-out and endpoint activation refusal.
- [x] Test the incoming verifier's exact source set, including rejection of pin-file-only, unpinned-bundle, returned-chain-only, and no-bundle substitutes.
- [x] Test successful active configuration replacement and mTLS after trust-anchor rotation; assert unchanged socket inode and undropped in-flight connection.
- [x] Inject each live publication failure and assert rollback, old acceptor, failed state, and cleanup. Inject rollback failure and assert old acceptor, both errors, and cleanup.
- [x] Test an ordinary issuance/internal-credential authentication failure and a no-op pass.
- [x] Test next-dial client renewal without caller recreation and the deterministic torn-pair success/exhaustion paths.
- [x] Run documentation build and required formatting, lint, unit, and applicable Docker E2E preflight checks.

## Dependencies

Part of #770. The callable registrar issuance unit, endpoint certificate-swap seam, and in-repository per-dial client loader are already present in origin/main.

## Pointers

- `src/bin/bootroot-agent.rs` for start-time issuance and activation order.
- `src/daemon.rs` for task ownership, cadence, shutdown, and eligibility.
- `src/registrar_certs.rs` (`issue_surface_pair`) for the publication boundary.
- `src/acme/flow.rs` (`issue_certificate_with_bootstrap`) and `src/fs_util.rs` (`write_cert_and_key`) for current bundle/pair publication.
- `src/registrar/internal/agent_config.rs` for renewal settings.
- `src/cert_chain.rs` for chain-drift verification.
- `src/registrar/endpoint.rs`, `src/registrar/endpoint/serve.rs`, and `src/registrar/endpoint/tls.rs` for active acceptor replacement and verifier construction; `src/registrar/endpoint/client.rs` for the loader.
- `src/registrar/endpoint_pin.rs` for server-anchor validation.
- `docs/en/operations.md` and `docs/ko/operations.md` for caller documentation.

sehkone:
Closes #768

Part of #770

## Summary

Both leaves the registrar surface runs on expire, and nothing renewed them: `ensure_registrar_surface_certificates` mints them once at start and deliberately leaves a usable pair alone, so a lapse took every enrollment in the deployment down and the registrar could not re-mint itself. This adds one daemon-owned renewal adapter that keeps both valid, and delivers both halves of the reload contract. Nothing goes on the wire: no response field, serialization fixture, wire-visible renewal state or lapse classification is added.

### The adapter

Created only where `[registrar_endpoint] enabled = true`, from the activated endpoint the daemon already holds, and armed **before** anything at all is spawned — beside the accept task's other start-up dependencies — so the per-leaf state accessor is initialized from the certificates already on disk before the endpoint serves anything. Its handle joins with the daemon's other tasks and it takes the same shutdown watch. With the endpoint disabled it is never built: no adapter, no state entry, no pass, no OpenBao request and no CA request.

Arming it is part of starting, not something logged over a running endpoint. An enabled endpoint owes both leaves an adapter and both accessor entries, and a failed attempt has no lifetime to make a missing entry with, so a preparation that cannot be completed — a rendered internal `agent.toml` that is missing or no longer loads, a deployment state file that cannot be resolved, a leaf certificate on disk that no longer parses — fails the invocation with a named error, exactly as an absent internal credential or an unloadable provisioning config already does. Serving two certificates nothing will renew is the outage this whole change exists to prevent.

Cadence, lead time and the issuance-retry budget all come from the rendered internal agent configuration's sole profile. No second scheduler, no lead-time constant of its own, no second retry policy: eligibility is the daemon's existing predicate, split out as `daemon::should_renew_certificate` so the two registrar leaves — which have no `[[profiles]]` entry — are judged by the same rule rather than by a copy of it, no-bundle opt-out and all. The retry budget is selected the same way, through `daemon::select_retry_backoff`: that profile may carry a `[profiles.retry]` of its own, and where it does it wins over the config's top-level `[retry]`, exactly as it does for a service profile.

### Publication as a transaction

Nothing touches a live path until everything that can fail has already succeeded:

1. **Issue off-live.** `acme::flow` is split at exactly the point publication used to begin, so `issue_certificate_material` returns the chain, the candidate certificate and a fresh key instead of writing them. `issue_certificate_with_bootstrap` is now that call plus `publish_issued_material`, so the two paths cannot drift.
2. **Validate the candidate.** Key-to-leaf; the reserved SAN, with a client candidate recognising to the same instance, host and domain, and a server candidate carrying the exact endpoint SAN and chaining to an anchor the endpoint pin file already names. The pin file is read and never written.
3. **Stage the merged bundle**, computed but not written.
4. **Build the whole next `ServerConfig`** — from the candidate server pair when that is the leaf being renewed and the live one when it is not, and from the staged bundle in either case. The incoming `WebPkiClientVerifier` is rebuilt from the post-merge bundle's pinned subset and from nothing else.
5. **Snapshot, then publish**: the bundle, then the certificate and key through the existing two-rename contract.
6. **Exchange the active configuration**, infallibly, because the replacement was built in step 4.

A failed write restores every path the publication reached — bytes, mode and ownership — leaves the old configuration installed and records a failed attempt. A rollback that itself fails carries both errors and explicitly does not describe the files as restored. Candidate files and snapshots live in a `0700` working directory removed on every exit.

### The server-side swap

`ActivatedEndpoint` now holds its acceptor and resolver together behind an `RwLock<Arc<ActiveTls>>`, and the accept loop loads it immediately before each handshake, holding no lock across an await. Renewal exchanges the **whole** configuration: the old acceptor retains the old client verifier, so replacing `EndpointCertResolver` alone would leave a trust-anchor rotation undecided until a restart. No restart, no signal, no socket rebind, no dropped in-flight connection.

### The caller-side reload

`load_matching_client_material` rereads the pair on every dial and checks the key is the leaf's. A reader that lands between the writer's two renames re-reads instead of presenting a mismatch: five reads total, with 1/2/4/8 ms waits after the first four, then a typed `ClientMaterialError::KeyMismatch`. That is a fixed local policy of the reader's and deliberately not the daemon's issuance-retry backoff. It is documented as the caller contract in `docs/en/operations.md` and `docs/ko/operations.md`, naming the in-repository client as the reference behaviour.

## Test plan

- [x] Enabled/disabled adapter creation, accessor initialization before serving, shutdown join, and the absence of any disabled-path OpenBao or CA activity
- [x] An enabled endpoint whose renewal cannot be armed fails `run_daemon` with a named error before anything is spawned, so the accept task is never reached
- [x] Immediate first pass and scheduled later passes under `tokio::time::pause`, with a driven pass reading no role_id/secret_id and using no AppRole
- [x] Initialization records observed notAfter, `never_attempted` and no timestamp without issuing anything, and a leaf it cannot observe refuses the adapter instead of leaving the endpoint one entry short
- [x] The cadence's retry budget: the internal profile's own `[profiles.retry]` wins, and the rendered `[retry]` table is the fallback
- [x] Lead-time renewal: candidate key/SAN/pin validation, staged bundle, next TLS configuration and snapshots all precede the first live write; fresh key, preserved client instance/host/domain, exact server SAN, unchanged pin file, accessor success
- [x] Chain drift: client replacement, server publication behind a pinned anchor, unpinned-server refusal leaving live bundle, pair, active TLS and pin file byte-for-byte unchanged with notAfter retained, and the no-bundle opt-out plus the activation refusal it cannot stand behind
- [x] The incoming verifier's exact source set, rejecting pin-file-only, unpinned-bundle, returned-chain-only and no-bundle substitutes
- [x] Whole-configuration exchange over a real handshake: renewed leaf presented and renewed client anchors accepted, with unchanged socket inode and an undropped in-flight connection
- [x] Injected bundle, certificate and key publication failures — rollback, old acceptor retained, failed state, artifacts cleaned — and an injected rollback failure recording both errors without claiming restoration
- [x] An ordinary internal-credential authentication failure recorded as a failed attempt, and an unchanged no-op pass
- [x] Next-dial client renewal without recreating the caller, and the deterministic torn-pair success and five-read exhaustion paths with their 1/2/4/8 ms waits
- [x] `cargo test --lib` (1299 passed) plus the other integration test binaries; clippy `-D warnings` and the suite were also run on Linux in a `rust:latest` container as a non-root user, because `src/registrar/endpoint/*` is Linux-gated and never compiles on macOS
- [x] `scripts/preflight/ci/check.sh`: fmt, clippy, rustdoc, ruff, biome, markdownlint, `mkdocs build --strict`, `cargo audit` (only the two pre-existing allowed advisories)
- [ ] `scripts/preflight/ci/e2e-matrix.sh` — **did not run on this machine.** The matrix needs passwordless `sudo` for its `bootroot init` step, which this host does not supply, and port 8200 is held by an unrelated local process. `tests/bootroot_rotate.rs` likewise fails inside the Linux container for want of a `docker` binary; those tests never reach any code this branch changes. CI's `Docker E2E` and `test-core` jobs gate the arm that could not run here.


sehkone:
**[Reviewer Round 1]**

Found two issues:

- **[P1] Do not serve an enabled endpoint after renewal setup fails.** [`src/daemon.rs`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/daemon.rs:889) logs a `RegistrarCertRenewal::prepare` error and continues to spawn the accept loop. For example, a missing or malformed rendered internal `agent.toml` leaves an enabled endpoint running permanently without the required adapter, state entries, or renewal work—contrary to the issue’s enabled-endpoint contract. Similarly, [`assemble`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal.rs:783) tolerates certificate-observation failures, while [`record_failure`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal.rs:224) will not create the missing entry. Fail preparation before serving, or otherwise guarantee both state entries and the adapter.

- **[P2] Honor the internal profile’s retry override.** [`RenewalCadence::from_internal_config`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal.rs:718) always copies `internal.retry.backoff_secs`. The daemon’s established policy uses a profile-level override when present ([`select_retry_backoff`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/daemon.rs:1270)). The internal-config loader permits such an override, so registrar issuance can use a different retry budget from its sole profile. Select `profile.retry` first and add a coverage case for it.

sehkone:

sehkone:
**[Author Round 1]**

CI is running on the pushed commit. Here is the disposition of each review item.

## [P1] Do not serve an enabled endpoint after renewal setup fails — **Fixed**

Accepted in full; the issue's contract is explicit that an enabled endpoint has an adapter and both accessor entries, and the old code could satisfy neither while still serving.

- `run_daemon` now arms the adapter with `prepare_registrar_cert_renewal` immediately after `resolve_registrar_service`, **before anything is spawned**, and propagates the error with `?`. This is the rule the accept task's other dependencies already follow (`resolve_registrar_service`'s own doc: a dependency it cannot resolve fails the invocation rather than leaving the adopted socket with no accept task). `spawn_registrar_cert_renewal` is now infallible and just spawns the already-armed loop; the `error!`-and-continue path is gone.
- `RegistrarCertRenewal::assemble` returns `Result` and fails when a leaf's certificate cannot be read or parsed, so `record_failure`'s inability to create a missing entry is no longer reachable — the two entries and the adapter are guaranteed together or the daemon does not start. Docs on `record_success`/`record_failure` updated accordingly.
- Practical risk is small: `build_registrar_handler` → `InternalCredential::load` already calls `load_internal_config`, so a missing or malformed rendered `agent.toml` was *already* a hard startup failure. The genuinely new refusal is an unparsable leaf certificate — a file start-time issuance has just made usable.
- Tests: `an_unarmable_registrar_renewal_stops_the_daemon` (clobbered client leaf → `run_daemon` returns an error naming both the arming step and the path), `a_leaf_that_cannot_be_observed_refuses_the_adapter`, and `the_adapter_is_armed_before_anything_is_spawned` (source-order assertion mirroring the existing `the_registrar_service_is_resolved_before_anything_is_spawned`).
- Two existing daemon tests called `run_daemon` with `enabled = true` and *no* registrar-endpoint certificate paths at all — a state configuration validation refuses in production, and one the test seam allowed only because the old code shrugged. They now get a fixture, `configure_registrar_surface_material`, that provisions the four paths the way start-time issuance would.

## [P2] Honour the internal profile's retry override — **Fixed**

Valid: `load_internal_config` permits a `[profiles.retry]` on the sole profile, and the daemon's established policy prefers it. `RenewalCadence::from_internal_config` now calls `daemon::select_retry_backoff(&internal, profile)` (widened to `pub(crate)`, with a doc comment recording why it is shared) instead of copying `internal.retry.backoff_secs`. Covered by `the_cadence_takes_the_profiles_own_retry_budget` and `the_cadence_falls_back_to_the_top_level_retry_budget`, which render a real internal agent config and read the cadence back off it.

## Docs

`docs/en/operations.md` and `docs/ko/operations.md` gained a paragraph stating that arming the loop is part of starting, and the `CHANGELOG.md` entry now says a host that enables the endpoint but cannot arm renewal fails to start instead of serving certificates nothing would renew. PR body updated to match (both `Closes #768` and the inherited `Part of #770` prose retained).

## Verification

Linux container (`rust:1-bookworm`, non-root, since `src/registrar/endpoint/*` never compiles on macOS): `cargo fmt --check` (with `group_imports=StdExternalCrate`), `cargo clippy --all-targets -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc`, `cargo test --lib` (1306 passed, 0 failed), `cargo test --bin bootroot` (1295 passed, 0 failed). On macOS: `markdownlint-cli2` and `./scripts/check-docs.sh` (mkdocs `--strict`) both clean. `e2e-matrix.sh` still cannot run on this host — unchanged from the PR body's existing note.

sehkone:
**[Reviewer Round 2]**

**[Reviewer Round 2: NOT_APPROVED]**

- **[P1] Serialize the shared CA-bundle transaction with other in-process writers.** [`registrar_renewal.rs`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal.rs:1122) stages and snapshots `trust.ca_bundle_path`, then publishes or rolls it back at [1166–1181](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal.rs:1166). Meanwhile the normal internal-profile renewal independently merges and writes that same bundle in [`acme/flow.rs`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/acme/flow.rs:609). Both loops are spawned independently in [`daemon.rs`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/daemon.rs:243) and [268](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/daemon.rs:268), and the existing locks are per profile, not per bundle.

  Consequently, an internal-profile write can land after the registrar snapshot; if the registrar pair write then fails, its rollback restores the stale snapshot and discards the other renewal’s bundle update. A successful registrar write can similarly overwrite an intervening merge. Put all in-process bundle read/merge, snapshot, publication, and rollback paths behind one shared bundle lock (including the normal profile writer), and add a deterministic interleaving test.

Round 1’s two findings are resolved: preparation now fails before spawning the endpoint, and cadence correctly uses the internal profile’s retry override. PR linkage and test-plan hygiene look correct.

sehkone:

sehkone:
**[Author Round 2]**

CI is running on `baf29b2d`. Here is the disposition of Round 2's single item.

## [P1] Serialize the shared CA-bundle transaction with other in-process writers — **Fixed**

The finding is correct, and it is worse than the two writers it names — there are three, and one of the races predates this branch.

**What was wrong.** `[trust] ca_bundle_path` has no owner. `acme::flow::write_merged_ca_bundle` reads it, merges the issued chain in and writes it back — and up to three profiles issue concurrently by default (`max_concurrent_issuances` = 3). `fast_poll::apply_trust_to_disk` replaces the same file wholesale on a trust update. `registrar_renewal::publish_candidate` stages a merge, snapshots it, publishes it and restores it when the pair write that follows fails. `ProfileLocks` serializes none of them: its keys are profile labels, and this file has no profile.

**The fix.** A new `src/ca_bundle_lock.rs` — a path-keyed registry (`LazyLock<BundleLocks>`, the shape `registrar::verbs`' `ID_LOCKS` already uses) handing out one `tokio::Mutex` per bundle path. All three writers hold it across their whole read-merge-write span: the registrar's from before `stage_bundle`'s read through the last restore `roll_back` can perform, the profile writer's inside `write_merged_ca_bundle`, the fast-poll apply's scoped to its replacement (not its `agent.toml` rewrite).

A `static` rather than a plumbed-through registry, deliberately: the per-profile publication is reached from the CLI as well as the daemon loop, several frames below anything that knows a daemon exists, and an argument every future call site must remember to pass is a lock whose correctness compiles away when forgotten. It is also deliberately *not* inside `fs_util::write_ca_bundle`, which the holding transactions call — that would deadlock them against themselves; both that function's docs and the module's say so.

**Deterministic interleaving test**, plus three more:
- `an_interleaved_profile_merge_survives_a_registrar_rollback` — a `LivePaths` double gates the pair write open, a real `write_merged_ca_bundle` is spawned against the same bundle, and the assertion is the invariant the lock creates: whichever way the two are scheduled, the second anchor is in the bundle afterwards. The wait is condition-based (`merge.is_finished()`, capped), not a clock. I measured it: with the `hold` removed from `publish_candidate` it fails **10 runs out of 10**; with the lock it passes 10 out of 10. (The fixed-yield version I wrote first caught it only 1 in 5 — that is why the wait is a condition.)
- `the_registrar_publication_waits_for_the_shared_bundle_lock` and `the_profile_bundle_writer_waits_for_the_shared_bundle_lock` — with the guard held externally, neither writes a byte to any live path, and both proceed on release.
- `apply_trust_to_disk_waits_for_the_shared_bundle_lock` in `fast_poll`, and `one_path_is_one_lock_and_two_paths_are_two` in the new module.

**Scope note:** including `fast_poll` goes one writer beyond what the review named. It replaces the same file from an independently spawned loop and would land inside either transaction's span in exactly the same way, so leaving it out would have left the reported bug half-fixed.

**Docs.** `CHANGELOG.md` gains a `### Fixed` entry — the lost update between two concurrent profile merges is observable to a user of the last release, unlike the registrar half of it — and `docs/en/concepts.md` / `docs/ko/concepts.md` note beside the existing merge explanation that the bundle is exclusively held per write and that readers are never blocked. PR body updated with an "One writer at a time on the shared CA bundle" section and a test-plan bullet; `Closes #768` and the inherited `Part of #770` prose both retained.

**Verification** (Linux container, `rust:1-bookworm`, non-root, since `src/registrar/endpoint/*` never compiles on macOS): `cargo fmt --check` with `group_imports=StdExternalCrate`, `cargo clippy --all-targets -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc` — all clean; `cargo test --lib` 1311 passed, `cargo test --bin bootroot` 1295 passed. One `testbin` run failed on `preflight_compose_published_ports_checks_openbao_localhost_during_install` — the known port-8200 flake on this host, not reproducible on rerun. On macOS: `markdownlint-cli2` and `./scripts/check-docs.sh` (mkdocs `--strict`) clean. `e2e-matrix.sh` still cannot run here, unchanged from the PR body's note.

sehkone:
**[Reviewer Round 3]**

**[Reviewer Round 3: NOT_APPROVED]**

- **[P2] Exercise the reload contract through the renewal path.** [`the_next_handshake_uses_the_renewed_material_without_a_restart`](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal/tests.rs:1510) never runs the accept loop or completes a TLS handshake. It only opens an unserved raw Unix socket and checks that the certificate file changed. Consequently, it would still pass if `publish_candidate` stopped calling `swap_active_tls`, leaving live files renewed but the active acceptor on the old server certificate/client verifier. Add an integration test that serves the harness endpoint, publishes through `RegistrarCertRenewal`, and verifies a subsequent real mTLS handshake presents the renewed server leaf and accepts a client under the rebuilt trust anchor.

Round 2’s P1 is resolved: the path-keyed bundle lock now covers registrar staging/snapshot/rollback, profile merge publication, and fast-poll trust replacement, with a deterministic rollback interleaving test.

sehkone:

sehkone:
**[Author Round 3]**

CI is running on `b748619e`. Disposition of Round 3's single item:

## [P2] Exercise the reload contract through the renewal path — **Fixed**

The finding is correct. `the_next_handshake_uses_the_renewed_material_without_a_restart` opened a raw, unserved `UnixStream` and then compared PEM text on disk. Nothing in it touched the acceptor, so deleting `swap_active_tls` from `publish_candidate` would have left it green while the endpoint served renewed files behind the old server certificate and the old client verifier.

It is replaced by two tests that run the production accept loop (`serve::run`) over the harness endpoint and read every conclusion off the wire:

- **`the_next_handshake_presents_the_renewed_server_leaf`** — dials and completes a real mTLS handshake before the swap, holds that connection open with its request unsent, publishes a server candidate through `RegistrarCertRenewal::publish_candidate`, then asserts (a) the socket inode is unchanged, (b) the in-flight connection still completes its round trip under the configuration it handshook with, and (c) a fresh handshake is *presented* the renewed leaf — the fingerprint comes from `peer_certificates()`, not from the file.
- **`the_next_handshake_accepts_a_caller_under_the_rebuilt_anchor`** — pins a rotated CA in `trusted_ca_sha256` but leaves it out of the live bundle, dials with a caller under it and asserts the live verifier refuses it with `UnknownCA`, publishes exactly that material as the client-leaf renewal, then dials again with the now-live pair and is admitted. The only thing that changed is the verifier the publication rebuilt from the staged bundle.

Acceptance is asserted through a round trip rather than through `connect`, because in TLS 1.3 the client finishes its handshake before the server has validated the client certificate. The handler behind the endpoint answers with the transport-authenticated caller identity, so a returned answer proves both the `WebPkiClientVerifier` accepted the chain and the accept loop recognized the SAN. The caller's TLS configuration is composed by `endpoint::client::build_client_config` — the one place in the tree that builds it — over material read from the live paths each dial, so a dial after a renewal carries no caller state across.

**Negative control, measured:** with `self.endpoint.swap_active_tls(...)` replaced by a no-op, both tests fail — the first on the presented leaf, the second on `AlertReceived(UnknownCA)`. The production line was restored before the run below.

## Docs

None. This round is test-only; no behaviour, interface, or user-visible change, so `CHANGELOG.md` and the manuals are untouched. The PR body's test-plan bullet was rewritten to say what is now actually exercised (production accept loop, wire-read assertions, the negative control), with `Closes #768` and the inherited `Part of #770` prose both retained.

## Verification

Linux container (`rust:1-bookworm`, non-root, since `src/registrar/endpoint/*` never compiles on macOS): `cargo fmt --check` with `group_imports=StdExternalCrate`, `cargo clippy --all-targets -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc --no-deps` — all clean; `cargo test --lib` 1312 passed / 0 failed, `cargo test --bin bootroot` 1295 passed / 0 failed. `e2e-matrix.sh` still cannot run on this host, unchanged from the PR body's note — and this round changes no code any E2E path reaches.

sehkone:
**[Reviewer Round 4]**

**[Reviewer Round 4: APPROVED]**

Round 3’s P2 is resolved. The new tests use the production accept loop and real mTLS handshakes to prove both server-leaf replacement and client-verifier replacement after publication ([tests](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal/tests.rs:1675), [tests](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal/tests.rs:1761)). The accept loop loads the active acceptor immediately before the handshake, and publication swaps it only after all live writes succeed ([serve.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar/endpoint/serve.rs:382), [registrar_renewal.rs](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-768-review/src/registrar_renewal.rs:1177)).

All prior findings are resolved. I found no new issues. PR linkage and test-plan hygiene are correct.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Renew the registrar surface's two leaves
```

**Body**

```text
Both leaves the registrar surface runs on expire, and nothing renewed
them: the surface mints them once at start and deliberately leaves a
usable pair alone, so a lapse took every enrollment in the deployment
down and left the registrar unable to re-mint itself.

One daemon-owned adapter now keeps both valid. It is built only where
the endpoint is enabled, armed before anything is spawned so the
per-leaf state is initialized from the certificates already on disk
before the endpoint serves, and joined with the daemon's other tasks
under the same shutdown watch. Cadence, lead time and the issuance
retry budget all come from the rendered internal agent configuration's
sole profile, and eligibility is the daemon's own predicate rather
than a copy of it.

A pass touches no live path until everything that can fail has already
succeeded: issue off-live, validate the candidate's key, its SAN and a
server candidate's anchor against the pin file, stage the merged
bundle, build the whole next server configuration from it, snapshot,
publish, and only then exchange the active configuration. A failed
write restores every path the publication reached and leaves the old
configuration installed; a rollback that itself fails carries both
errors and does not claim the files were restored. The pin file is
read and never written.

The endpoint holds its acceptor and resolver together and the accept
loop loads them immediately before each handshake, so a renewal needs
no restart, no signal and no socket rebind, and drops no connection in
flight. Replacing the certificate resolver alone would not do: the old
acceptor retains the old client verifier, leaving a trust-anchor
rotation undecided until a restart.

The shared CA bundle gained a third writer here, and nothing had ever
serialized the two it already had — the per-profile publication's
read-merge-write and the fast-poll trust apply's wholesale replacement
— so a merge computed from bytes another writer had already replaced
could drop an anchor the host was told to trust, with nothing
recording that it went missing. All three now take one lock per bundle
path across their whole span.

Callers reload through the per-dial loader, which rereads the pair on
every dial and re-reads rather than present a mismatch when it lands
between the writer's two renames. That contract is documented for
callers in English and Korean, naming the in-repository client as the
reference behaviour.

Nothing here goes on the wire: no response field, no serialization
fixture, no wire-visible renewal state and no lapse classification.

Closes #768

Part of #770
```

PR #963, opened 2026-08-29
Prove the audit bind guard mid-migration (#960)

### LINKED ISSUE #960: Prove the audit bind refuses to create its source during the migration window
# Prove the audit bind refuses to create its source during the migration window

## Context

The audit bind is declared `create_host_path: false` so Docker refuses to manufacture the bind source when the reserve mount is not there. The migration window is where that guarantee matters most: after the aside rename the underlying store is empty, the reserve may be mounted but still carries no `openbao/`, and a Compose bring-up in that state must fail rather than create the directory on the mounted reserve. A created `openbao/` there would be a bind source the migration did not put there, standing on the destination side of a copy that has not run.

Rust coverage proves `bootroot infra up` refuses bring-up under **migration incomplete** (`src/commands/audit_store.rs:3240`). That is a different assertion: it proves bootroot declines to start containers, not that Docker itself refuses the bind when something else brings the stack up. The migration E2E scenario never attempts a Compose bring-up inside the window, so the `create_host_path: false` guarantee is asserted nowhere against the state the window actually produces.

**What was withdrawn from this issue, and why it is not to be re-added.** An earlier revision also asked that every mount-absent open-migration report render the outstanding activation commands, including under the migration-set refusals. That reading came from an acceptance criterion in #926 that had not been narrowed the way its capacity-figures bullet was; #926 has since been amended, and the withholding is deliberate. A rollback is rendered for the mount state its run observed, so one rendered with the mount absent carries no `umount`; rendering an activation beside it would let an operator run the activation, take the rollback, and meet `rmdir` against a live mount point, stopping under `sh -e` before the restoring rename. Nothing in this issue changes what is rendered.

## Scope

Add an E2E scenario that exercises the migration window and proves the bind guard holds inside it. The scenario reaches an activated but still empty reserve — `.pre-mount` present, the mount active, `<audit_store_dir>/openbao` absent — attempts the normal Compose bring-up, and proves the OpenBao container fails to start with no entry created beneath the mounted store.

The scenario runs through the same `registrar-internal-init` lifecycle harness the existing migration coverage uses, and starts from an existing audit store so the aside rename and the full window are exercised rather than simulated. It leaves both writers stopped until the normal closing path.

## Acceptance criteria

- [x] The E2E suite attempts `docker compose up` while `.pre-mount` exists, the reserve mount is active, and `<audit_store_dir>/openbao` is absent.
- [x] It proves the OpenBao container fails to start and that no entry is created beneath the mounted store.
- [x] The fixture starts with an existing audit store, so the aside rename and the full migration window are exercised, and it leaves both writers stopped until the normal closing path.
- [x] The scenario runs through the existing `registrar-internal-init` lifecycle harness rather than a new one.
- [x] The existing `infra up` refusal coverage is kept and unchanged.
- [x] No change weakens `create_host_path: false`, changes the audit-store layout, or lets bootroot perform a host-changing migration step.
- [x] No rendering behaviour changes: the commands any outcome renders, and the outcomes themselves, are exactly what they are today.

## Constraints

- Keep the test operator-faithful: it must use the rendered or normal Compose surface, and must not simulate Docker's result by creating or removing paths directly.
- Do not create, copy, rename, delete, mount, or unmount a store from bootroot.
- Do not change `withholds_activation`, the migration-set refusals, holding-directory precedence, or any rendered command list.
- Do not add a second activation renderer or a new outcome vocabulary.
- Do not use a recursive delete or a pipeline in anything the scenario renders or runs.

## Out of scope

- **Rendering the activation for mount-absent reports under a migration-set refusal.** Withdrawn deliberately, for the reason in Context; #926's acceptance criteria now state the withholding.
- Choosing a rollback procedure for collision states where the reverse rename target already exists or a foreign filesystem occupies the store.
- Changing the migration copy, capacity, manifest, or reserve-replacement contracts.
- Altering daemon-side registrar-verb behavior.

## Test plan

- [x] Extend `assert_the_rendered_steps_migrate_an_existing_store`, or add a focused sibling, in `scripts/impl/run-registrar-internal-init-e2e.sh` with the mid-window Compose attempt and the assertion that no bind source appears on the mounted reserve.
- [x] Assert the OpenBao container's failure by its own exit rather than by the absence of a side effect alone, so a bring-up that silently does nothing cannot pass.
- [x] Run the registrar-internal-init E2E job, plus `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --config group_imports=StdExternalCrate --check`, and `./scripts/check-docs.sh`.
- [x] Assert the existing `infra up` refusal test still passes unchanged.

## Dependencies

Part of #926. This follow-up has no follow-up issue dependency and is filed under umbrella #773.

## Pointers

- `src/commands/audit_store.rs:3240` — existing `infra up` refusal coverage, which this issue keeps and does not duplicate.
- `scripts/impl/run-registrar-internal-init-e2e.sh:773` — existing bind-guard boundary test.
- `scripts/impl/run-registrar-internal-init-e2e.sh:1223` — the migration E2E function that lacks the mid-window Compose attempt.
- `src/commands/audit_store/migration.rs:264` — `withholds_activation`, whose behaviour this issue leaves untouched; its doc comment records why the mount-absent refusal states withhold.

sehkone:
## Summary

The audit bind is declared `create_host_path: false` so Docker refuses to manufacture the bind source when the reserve mount is not there. The migration window is where that guarantee matters most — the store is aside under `.pre-mount`, the reserve is mounted but still carries no `openbao/`, and a bring-up that created that directory would leave a bind source on the destination side of a copy that has not run.

`bootroot infra up` already refuses that state (`src/commands/audit_store.rs`), but that is a different assertion: it proves bootroot declines to start containers, not that Docker itself refuses the bind when something else brings the stack up. The migration E2E scenario never attempted a Compose bring-up inside the window, so the guarantee was asserted nowhere against the state the window actually produces.

Closes #960

Part of #926

## Changes

All of it is in `scripts/impl/run-registrar-internal-init-e2e.sh`, inside the existing `assert_the_rendered_steps_migrate_an_existing_store` lifecycle path. No Rust changed.

- **`assert_the_audit_bind_refuses_to_create_its_source_mid_migration`** — a focused sibling called between pass 2 (the activation) and pass 3 (the copy), which is exactly the window. It restates its preconditions rather than assuming them (`.pre-mount` present, the mount unit `active`, `<audit_store_dir>/openbao` absent), checks that the override the refusal pass rendered still names that store and still carries `create_host_path: false`, then runs the ordinary `docker compose up -d --no-deps openbao` over it.
- **The failure is read from the bring-up itself** — its non-zero exit, the error naming the bind source, and the container not running — so a bring-up that returned success having quietly done nothing cannot pass on the absent source alone. Only then does it assert the destination was not manufactured: `openbao/` still absent, and the mounted store's entry list identical to the one taken before the attempt.
- **Both writers are stopped for the window.** Rendered step 1 stops them, and `rendered_migration_commands` drops those two lines — one would stop this scenario's whole Compose stack, the other names a unit no CI host has installed. The OpenBao writer is now stopped by service through the same Compose surface just before the aside rename, stays down across every pass, and is brought back only on the closing path at the end of the section. The registrar writer is asserted not running rather than stopped: stopping a unit this scenario did not install would be a host-changing step it has no way to undo.

The fixture starts from an existing audit store, so the aside rename and the full window are exercised rather than simulated. Nothing is created, removed, mounted or unmounted to stand in for Docker's result, and no rendered command list, outcome, or `withholds_activation` behaviour is touched.

## Test plan

- [x] `docker compose up` is attempted while `.pre-mount` exists, the reserve mount is active, and `<audit_store_dir>/openbao` is absent.
- [x] The OpenBao container's failure is asserted by its own non-zero exit and by the error naming the bind source, not by the absence of a side effect alone, and no entry is created beneath the mounted store.
- [x] The scenario runs through the existing `registrar-internal-init` lifecycle harness rather than a new one, starting from an existing audit store and leaving both writers stopped until the normal closing path.
- [x] The existing `infra up` refusal coverage is kept and unchanged — `cargo test --bin bootroot commands::audit_store::` passes 136 tests, including `infra_up_refuses_the_bring_up_while_a_migration_is_open`.
- [x] `cargo clippy --all-targets -- -D warnings` passes.
- [x] `cargo fmt -- --config group_imports=StdExternalCrate --check` passes.
- [x] `./scripts/check-docs.sh` passes.
- [x] `shellcheck -x` and `bash -n` on the changed script are clean.
- [x] `./scripts/validate-e2e-run-scope.sh` and `./scripts/validate-e2e-leftover-check.sh` pass.
- [x] The Linux `Docker E2E` `registrar-internal-init` job passes — it ran green on this commit ([job 99074231742](https://github.com/aicers/bootroot/actions/runs/33242493022/job/99074231742)), and its log shows the new assertions actually executing inside the `reserve_activation_is_possible` guard rather than being skipped: `PASS both writers are stopped before the store moves aside`, `PASS the reserve is mounted for the mid-window bring-up`, `the mounted store holds [lost+found] before the mid-migration bring-up`, `PASS the mid-migration bring-up fails on the guarded audit bind source`, `PASS the refused bring-up added no entry beneath the mounted store`, and `PASS the writers come back only once the migration has closed`. It could not run on the development host — macOS with no passwordless sudo, no systemd, no loop device and no `mkfs.ext4` — so that CI job is the only arm that exercises it.

### DIFF
--- a/scripts/impl/run-registrar-internal-init-e2e.sh
+++ b/scripts/impl/run-registrar-internal-init-e2e.sh
@@ -1273,6 +1273,27 @@ assert_the_rendered_steps_migrate_an_existing_store() {
   RESERVE_IMAGE="$image"
   RESERVE_UNIT_NAME="$unit"
 
+  # The rendered step 1 is the maintenance window itself: stop both
+  # writers before the store moves aside.  `rendered_migration_commands`
+  # drops those two lines -- one would stop this scenario's whole
+  # Compose stack rather than the writer, and the other names a unit no
+  # CI host has installed -- so the OpenBao writer is stopped here, by
+  # service, through the same Compose surface.  It stays down for the
+  # whole window and comes back only on the closing path at the end of
+  # this section, which is the shape of the window an operator opens.
+  instance_compose stop openbao >>"$RUN_LOG" 2>&1 ||
+    fail "could not stop the OpenBao writer for the migration window"
+  instance_compose rm -f openbao >>"$RUN_LOG" 2>&1 ||
+    fail "could not remove the OpenBao writer for the migration window"
+  # The other writer is `bootroot-registrar.service`, and this is an
+  # assertion rather than a stop: no CI host has that unit installed,
+  # and stopping one this scenario did not install would be a
+  # host-changing step it has no way to undo.
+  if systemctl is-active --quiet bootroot-registrar.service 2>/dev/null; then
+    fail "the registrar writer is running; the migration window needs both writers stopped"
+  fi
+  pass "both writers are stopped before the store moves aside"
+
   run_rendered_migration_commands "$first" "$store"
   sudo -n test -d "${store}.pre-mount" ||
     fail "the aside rename did not create ${store}.pre-mount"
@@ -1308,6 +1329,8 @@ assert_the_rendered_steps_migrate_an_existing_store() {
     fail "records/ was created on the mounted reserve before the copy"
   fi
   pass "the reserve is mounted and the store on it is still empty"
+
+  assert_the_audit_bind_refuses_to_create_its_source_mid_migration "$store" "$unit"
 
   # Pass 3: migration incomplete, now with the capacity verdict, the
   # copy, the verification and the closing rename.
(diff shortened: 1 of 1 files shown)


sehkone:
**[Reviewer Round 1]**

Approved — no findings.

The new E2E helper correctly verifies the actual migration window: holding directory exists, reserve is active, and `openbao/` is absent before Compose is invoked. It then requires Compose itself to fail on that source and confirms neither the bind source nor any top-level reserve entry was created. See [run-registrar-internal-init-e2e.sh](/Users/sehkone/.agentcoop/worktrees/aicers/bootroot/issue-960-review/scripts/impl/run-registrar-internal-init-e2e.sh:1476).

The surrounding lifecycle change also keeps OpenBao stopped from before the aside rename until after the migration closes, avoiding a writer race in the newly exercised window. PR linkage and test-plan hygiene are correct (`Closes #960`, permitted `Part of #926`); there are no review-thread issue creations.

sehkone:

